PR #7, opened 2025-06-23
generate.go: remove

jba:
Remove the code generator. It has served its purpose. From now on, we will modify protocol.go directly.


jba:
The diffs in protocol.go are because (apparently—news to me) `gofmt` will not format a file with the special "generated" comment, and its formatting differs slightly from the `go/format` package that the generator used.

PR #8, opened 2025-06-24
mcp: move to 2025-06-18 schema

jba:
Add types and fields for the latest MCP schema.

The main challenge here was dealing with the presence of _meta fields on many types, like Tool, Prompt and Resource.

- Distinguish between the _meta field of a request's params, which includes a progress token, from the _meta field on other types, which is simply a map.

- Redefine GetMeta to return a pointer to the map. We wanted to avoid that, but I don't see a clean way to do so. An alternative would be to have GetMeta return a map, and SetMeta to set it.

- Provide GetProgressToken for requests only (not notifications or other types).

- Still to do: deal with the meta and annotation fields of ContentBlock, which are currently duplicated for Resource and ResourceContents.

### DIFF
--- a/mcp/protocol.go
+++ b/mcp/protocol.go
@@ -40,29 +40,27 @@ type Annotations struct {
 type CallToolParams struct {
 	// This property is reserved by the protocol to allow clients and servers to
 	// attach additional metadata to their responses.
-	Meta      ParamsMeta `json:"_meta,omitempty"`
-	Arguments any        `json:"arguments,omitempty"`
-	Name      string     `json:"name"`
+	Meta      `json:"_meta,omitempty"`
+	Arguments any    `json:"arguments,omitempty"`
+	Name      string `json:"name"`
 }
 
-func (x *CallToolParams) GetMeta() *map[string]any { return &x.Meta.Data }
-func (x *CallToolParams) GetProgressToken() *any   { return &x.Meta.ProgressToken }
+func (x *CallToolParams) GetProgressToken() any  { return getProgressToken(x) }
+func (x *CallToolParams) SetProgressToken(t any) { setProgressToken(x, t) }
 
 type CallToolParamsFor[In any] struct {
 	// This property is reserved by the protocol to allow clients and servers to
 	// attach additional metadata to their responses.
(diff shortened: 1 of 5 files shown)


mcp/protocol.go
```diff
@@ -443,68 +491,86 @@ type ProgressNotificationParams struct {
...
 
 type PromptListChangedParams struct {
 	// This property is reserved by the protocol to allow clients and servers to
 	// attach additional metadata to their responses.
-	Meta Meta `json:"_meta,omitempty"`
+	Meta ParamsMeta `json:"_meta,omitempty"`
 }
 
-func (x *PromptListChangedParams) GetMeta() *Meta { return &x.Meta }
+func (x *PromptListChangedParams) GetMeta() *map[string]any { return &x.Meta.Data }
```
findleyr: What do you think about the following embedded type?
We can always export it later.

```
type paramsMeta struct {
  Meta map[string]any `json:"_meta,omitempty"`
}

func (paramsMeta) GetMeta() *map[string]any
func (paramsMeta) GetProgressToken() any
func (paramsMeta) SetProgressToken(any)
```

mcp/content.go
```diff
@@ -10,58 +10,67 @@ import (
...
-	Resource    *ResourceContents `json:"resource,omitempty"`
-	Annotations *Annotations      `json:"annotations,omitempty"`
+// The Type field must be one of "text", "image", "audio", "resource_link" or "resource".
+// The constructors above populate this field appropriately.
+// Although at most one of Text, Data, ResourceLink and Resource should be non-zero,
+// consumers of ContentBlock use the Type field to determine which value to use;
+// values in the other fields are ignored.
+// TODO(jba,rfindley): rethink this type. Each kind (text, image, etc.) should have its own
+// meta and annotations, otherwise they're duplicated for Resource and ResourceContents.
+type ContentBlock struct {
```
findleyr, line 25: Where did this name come from? Why not 'Content'?
Is the point that 'Content' is a collection of content blocks?

jba: From the spec.

mcp/protocol.go
jba: As discussed, did something very similar.

PR #36, opened 2025-06-25
design: add design for rate limiting

### LINKED ISSUE #22: Rate limiting support / Best practices
Add utilities / hooks / documentation to support rate limiting requests.
(e.g. search for 'rate limiting' at
https://modelcontextprotocol.io/specification/2025-03-26/client/sampling)

samthanawalla:
I added an example for how rate limiting should be implemented using middleware.

Fixes #22 

## Motivation and Context

Rate limiting can be enabled using middleware. This adds an example on how to do that.

## How Has This Been Tested?

N/A

## Breaking Changes

No

## Types of changes

- [x] Documentation update

## Checklist

### DIFF
--- a/examples/rate-limiting/go.mod
+++ b/examples/rate-limiting/go.mod
@@ -1,0 +1,8 @@
+module github.com/modelcontextprotocol/go-sdk/examples/rate-limiting
+
+go 1.25
+
+require (
+	github.com/modelcontextprotocol/go-sdk v0.0.0-20250625185707-09181c2c2e89
+	golang.org/x/time v0.12.0
+)
--- a/examples/rate-limiting/main.go
+++ b/examples/rate-limiting/main.go
@@ -1,0 +1,54 @@
+// Copyright 2025 The Go MCP SDK Authors. All rights reserved.
+// Use of this source code is governed by an MIT-style
+// license that can be found in the LICENSE file.
+
+package main
+
+import (
+	"context"
+	"errors"
+	"time"
+
+	"github.com/modelcontextprotocol/go-sdk/mcp"
+	"golang.org/x/time/rate"
+)
+
+// GlobalRateLimiterMiddleware creates a middleware that applies a global rate limit.
+// Every request attempting to pass through will try to acquire a token.
+// If a token cannot be acquired immediately, the request will be rejected.
+func GlobalRateLimiterMiddleware[S mcp.Session](limiter *rate.Limiter) mcp.Middleware[S] {
+	return func(next mcp.MethodHandler[S]) mcp.MethodHandler[S] {
+		return func(ctx context.Context, session S, method string, params mcp.Params) (mcp.Result, error) {
+			if !limiter.Allow() {
+				return nil, errors.New("JSON RPC overloaded")
+			}
+			return next(ctx, session, method, params)
+		}
+	}
+}
+
+// PerMethodRateLimiterMiddleware creates a middleware that applies rate limiting
+// on a per-method basis.
+// Methods not specified in limiters will not be rate limited by this middleware.
+func PerMethodRateLimiterMiddleware[S mcp.Session](limiters map[string]*rate.Limiter) mcp.Middleware[S] {
+	return func(next mcp.MethodHandler[S]) mcp.MethodHandler[S] {
+		return func(ctx context.Context, session S, method string, params mcp.Params) (mcp.Result, error) {
+			if limiter, ok := limiters[method]; ok {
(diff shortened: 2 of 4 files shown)


findleyr:
I think the section on rate limiting should be brief, suggesting that it be implemented with middleware. Can you instead promote your example to a fully fledged example in the `examples/` directory? Then it will be discoverable.

samthanawalla:
Sounds good. I added examples/rate-limiting.
For the per session rate limiting, we will need a way to expose a session ID but we can discuss later.

examples/rate-limiting/go.mod
```diff
@@ -0,0 +1,8 @@
+module github.com/modelcontextprotocol/go-sdk/examples/rate-limiting
```
jba, line 1: I guess we want a separate module because of x/time, right? Doc that in the commit message.
(I personally don't think it's such a big deal to include it, but if you feel it's best I'm okay with that.)

examples/rate-limiting/main.go
```diff
@@ -0,0 +1,68 @@
...
+	"time"
+
+	"github.com/modelcontextprotocol/go-sdk/mcp"
+	"golang.org/x/time/rate"
+)
+
+// GlobalRateLimiterMiddleware creates a middleware that applies a global rate limit.
+// Every request attempting to pass through will try to acquire a token.
+// If a token cannot be acquired immediately, the request will be rejected.
+func GlobalRateLimiterMiddleware[S mcp.Session](limit rate.Limit, burst int) mcp.Middleware[S] {
```
jba: I wonder if these should just accept a *rate.Limiter. Then users can query it, use it for other things (for example, a combined rate limit across multiple clients or servers), and so on.

examples/rate-limiting/main.go
```diff
@@ -0,0 +1,68 @@
...
+			return next(ctx, session, method, params)
+		}
+	}
+}
+
+// PerMethodRateLimiterMiddleware creates a middleware that applies rate limiting
+// on a per-method basis.
+// `methodLimits` maps method names to their specific rate limit configurations.
+// Methods not specified in `methodLimits` will not be rate limited by this middleware.
+func PerMethodRateLimiterMiddleware[S mcp.Session](methodLimits map[string]struct {
```
jba: ...and here, using a rate.Limiter would avoid the awkward struct type literal.

jba (changes requested):

examples/rate-limiting/main.go
samthanawalla: Done

examples/rate-limiting/main.go
samthanawalla: Good idea. Done

examples/rate-limiting/go.mod
samthanawalla: Yes, documented in the commit message.

examples/rate-limiting/main.go
```diff
@@ -0,0 +1,54 @@
...
+			return next(ctx, session, method, params)
+		}
+	}
+}
+
+func main() {
+	server := mcp.NewServer("greeter1", "v0.0.1", nil)
+	server.AddReceivingMiddleware(GlobalRateLimiterMiddleware[*mcp.ServerSession](rate.NewLimiter(rate.Every(time.Second/5), 10)))
+	server.AddReceivingMiddleware(PerMethodRateLimiterMiddleware[*mcp.ServerSession](map[string]*rate.Limiter{
+		"callTool":  rate.NewLimiter(rate.Every(time.Second), 5),  // 5 callTool requests/sec
```
jba: not quite right: it's 1 a second, with bursts of up to 5. 
For example, with Every(5*time.Second) and a burst of 5, output would be:
```
Starting rate limiter demo...                                                                                                  
Message 1 sent at 15:22:56                                                                                                     
Message 2 sent at 15:22:56                                                                                                     
Message 3 sent at 15:22:56                                                                                                     
Message 4 sent at 15:22:56                                                                                                     
Message 5 sent at 15:22:56                                                                                                     
Message 6 sent at 15:23:01                                                                                                     
Message 7 sent at 15:23:06 
...
```
In other words, you can do up to 5 all at once, but then you're back to one every 5 seconds.
You don't have to explain all that, just say "once a second with a burst of 5".

jba (approved):

examples/rate-limiting/main.go
samthanawalla: Thanks for the clarification. Updated

jba (approved):

Issue #24, opened 2025-06-24
Keepalive support


findleyr:
Add support for automatic keepalive 'ping' requests.

Design: https://github.com/modelcontextprotocol/go-sdk/blob/main/design/design.md#ping--keepalive

rwjblue-glean:
👋 sorry for the late ping, I realize that I was supposed to confirm that it was ok for new contributors first (I didn't notice that part in the CONTRIBUTING.md at first 🤦 ). 

I sent #39 to implement this.

findleyr:
No worries, and thanks for reading CONTRIBUTING.md!

No harm done: we hadn't started working on this and your contribution is very welcome!

PR #39, opened 2025-06-25
mcp: Implement KeepAlive for client and server

rwjblue-glean:
## Motivation and Context

Implement `KeepAlive` features from the design doc

Closes https://github.com/modelcontextprotocol/go-sdk/issues/24

## How Has This Been Tested?

Tested locally (unit tests).

## Breaking Changes

N/A

## Types of changes

- [x] New feature (non-breaking change which adds functionality)
- [ ] Documentation update

## Checklist

## Additional context

I can send a follow-up PR adding an example with a StreamableHTTP server that uses the keepalive if that would be helpful.

### DIFF
--- a/mcp/client.go
+++ b/mcp/client.go
@@ -9,6 +9,7 @@ import (
 	"iter"
 	"slices"
 	"sync"
+	"time"
 
 	"github.com/modelcontextprotocol/go-sdk/internal/jsonrpc2"
 )
@@ -56,6 +57,10 @@ type ClientOptions struct {
 	ResourceListChangedHandler  func(context.Context, *ClientSession, *ResourceListChangedParams)
 	LoggingMessageHandler       func(context.Context, *ClientSession, *LoggingMessageParams)
 	ProgressNotificationHandler func(context.Context, *ClientSession, *ProgressNotificationParams)
+	// If non-zero, defines an interval for regular "ping" requests.
+	// If the peer fails to respond to pings originating from the keepalive check,
+	// the session is automatically closed.
+	KeepAlive time.Duration
 }
 
 // bind implements the binder[*ClientSession] interface, so that Clients can
@@ -118,6 +123,11 @@ func (c *Client) Connect(ctx context.Context, t Transport) (cs *ClientSession, e
 		_ = cs.Close()
 		return nil, err
 	}
+
+	if c.opts.KeepAlive > 0 {
+		cs.startKeepalive(c.opts.KeepAlive)
+	}
+
 	return cs, nil
 }
 
@@ -131,12 +141,21 @@ type ClientSession struct {
 	conn             *jsonrpc2.Connection
 	client           *Client
 	initializeResult *InitializeResult
+	keepaliveCancel  context.CancelFunc
 }
 
 // Close performs a graceful close of the connection, preventing new requests
 // from being handled, and waiting for ongoing requests to return. Close then
 // terminates the connection.
 func (cs *ClientSession) Close() error {
+	// Note: keepaliveCancel access is safe without a mutex because:
+	// 1. keepaliveCancel is only written once during startKeepalive (happens-before all Close calls)
+	// 2. context.CancelFunc is safe to call multiple times and from multiple goroutines
+	// 3. The keepalive goroutine calls Close on ping failure, but this is safe since
+	//    Close is idempotent and conn.Close() handles concurrent calls correctly
+	if cs.keepaliveCancel != nil {
+		cs.keepaliveCancel()
+	}
 	return cs.conn.Close()
 }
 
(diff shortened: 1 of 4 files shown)


mcp/client.go
```diff
@@ -131,12 +141,16 @@ type ClientSession struct {
...
 	client           *Client
 	initializeResult *InitializeResult
+	keepaliveCancel  context.CancelFunc
 }
 
 // Close performs a graceful close of the connection, preventing new requests
 // from being handled, and waiting for ongoing requests to return. Close then
 // terminates the connection.
 func (cs *ClientSession) Close() error {
+	if cs.keepaliveCancel != nil {
```
jba, line 156: It looks like a user Close and the Close from the startKeepalive goroutine could race.
I don't think they can, but it's worth an explanation, in terms of happens-before relations, why no mutex is needed for keepaliveCancel. (If you don't understand what I'm talking about, I can explain further.)

mcp/mcp_test.go
```diff
@@ -838,3 +838,98 @@ func falseSchema() *jsonschema.Schema { return &jsonschema.Schema{Not: &jsonsche
 func nopHandler(context.Context, *ServerSession, *CallToolParamsFor[map[string]any]) (*CallToolResult, error) {
 	return nil, nil
 }
+
+func TestKeepAlive(t *testing.T) {
```
jba, line 842: Add a TODO to try to use the new synctest package for this test.

mcp/server.go
```diff
@@ -688,6 +700,33 @@ func (ss *ServerSession) Wait() error {
 	return ss.conn.Wait()
 }
 
+// startKeepalive starts the keepalive mechanism for this server session.
+func (ss *ServerSession) startKeepalive(interval time.Duration) {
```
jba, line 709: Can you factor this out so there is only one copy of the code?

jba (changes requested):

mcp/client.go
rwjblue-glean: Ya, I agree, I'm pretty sure that we do not need a mutex here, I've added the following comment explaining the flow and why it's safe:

```go
	// Note: keepaliveCancel access is safe without a mutex because:
	// 1. keepaliveCancel is only written once during startKeepalive (happens-before user Close calls)
	// 2. context.CancelFunc is safe to call multiple times and from multiple goroutines
	// 3. The keepalive goroutine calls Close on ping failure, but this is safe since
	//    Close is idempotent and conn.Close() handles concurrent calls correctly

```

Please double check my reasoning though!

mcp/server.go
rwjblue-glean: Great call, updated!

mcp/mcp_test.go
rwjblue-glean: I added the TODO (explaining that when we bump to 1.24+ we should experiment with it), and I also played around with it locally (it's totally nicer for these kinds of tests!).

mcp/client.go
```diff
@@ -131,12 +141,21 @@ type ClientSession struct {
...
 	initializeResult *InitializeResult
+	keepaliveCancel  context.CancelFunc
 }
 
 // Close performs a graceful close of the connection, preventing new requests
 // from being handled, and waiting for ongoing requests to return. Close then
 // terminates the connection.
 func (cs *ClientSession) Close() error {
+	// Note: keepaliveCancel access is safe without a mutex because:
+	// 1. keepaliveCancel is only written once during startKeepalive (happens-before user Close calls)
```
jba: s/user/all/

mcp/shared.go
```diff
@@ -308,3 +308,38 @@ type listResult[T any] interface {
...
+	Ping(ctx context.Context, params *PingParams) error
+	Close() error
+}
+
+// startKeepalive starts the keepalive mechanism for a session.
+// It assigns the cancel function to the provided cancelPtr and starts a goroutine
+// that sends ping messages at the specified interval.
+func startKeepalive(session keepaliveSession, interval time.Duration, cancelPtr *context.CancelFunc) {
+	ctx, cancel := context.WithCancel(context.Background())
+	*cancelPtr = cancel
```
jba, line 326: Add a comment saying that we must assign this before starting the goroutine, so we cannot return it. 

(I first thought "why not just return the cancel func instead of passing in a pointer?" But it would be racy.)

jba (changes requested):

mcp/shared.go
rwjblue-glean: Good suggestion, thank you! Updated with this comment:

```go
	// Assign cancel function before starting goroutine to avoid race condition.
	// We cannot return it because the caller may need to cancel during the
	// window between goroutine scheduling and function return.
```

jba (approved):

Issue #33, opened 2025-06-25
Tracking issue: adding examples


findleyr:
We should add many more examples to the `examples` directory. There are any number of sources for this:
- Other SDKs
- https://modelcontextprotocol.io/examples
- Some Go specific examples, such as integration with `go list`.

MegaGrindStone:
I'm interested to contribute on implementing [memory](https://github.com/modelcontextprotocol/servers/tree/main/src/memory) server example.

findleyr:
Thank you! That would be very helpful.

nsega:
I am also interested to contribute on implementing [sequentialthinking](https://github.com/modelcontextprotocol/servers/tree/main/src/sequentialthinking) server example, one of examples on [Example Servers](https://modelcontextprotocol.io/examples).

sbshah97:
I can contribute to the https://github.com/modelcontextprotocol/servers/tree/main/src/time time server example?

gk16:
I am very insterested to contribute on a [PostgreSQL MCP server](https://github.com/modelcontextprotocol/servers-archived/tree/main/src/postgres) example. I have create a PR https://github.com/modelcontextprotocol/go-sdk/pull/70 for it.

jba:
Update: according to https://github.com/modelcontextprotocol/servers/issues/2220#issuecomment-3029130120, we should keep the examples here.

There is no need to implement every example in modelcontextprotocol/servers. In fact, some of them might do too much.

@MegaGrindStone, you can reopen your PR or start a new one.

jba:
I've edited the top post and removed the help-wanted badge.

I think with the four or five examples we have, we're in a good place. Each example is both a review burden and an ongoing maintenance burden, so let's stop there for now.

nicholasf:
Just catching up with all of the above. Keen to understand the state of where the SSE implementation is and to see if I can help. I'll keep reading.

jba:
I believe our SSE implementation is complete, or at least as complete as it's going to get. SSE is not in the most recent spec.

@findleyr can say more about that.

findleyr:
@nicholasf yes, our SSE implementation should be complete. Please let us know if you find deficiencies.

nicholasf:
Thanks, will do. I've dropped back to just developing functionality using stdio for now.

findleyr:
I think we have a good number of examples at this point. Closing as complete.

Issue #20, opened 2025-06-24
Tool output schemas


findleyr:
Add support for output schemas, as described in the [spec](https://modelcontextprotocol.io/specification/2025-06-18/server/tools#output-schema):

If an output schema is provided, servers MUST provide structured results that conform to this schema.

jba:
There is one wrinkle here: If you use NewServerTool[In, Out], then you'll have an output schema, so you must provide a StructuredContent. But what if you want an input schema but not an output schema? Maybe you just want to return a simple string. A StructuredContent must be an object.

The problem is really with NewServerTool. One answer is: you just can't use it in this case. Maybe that's sufficient.

/cc @findleyr @samthanawalla

findleyr:
@jba see discussion regarding tool binding. Let's see if we can modify our tool binding APIs so that there's exactly one way to do this.

jba:
My recommended solution to the problem I raised above is to avoid generating an output schema if Out = any.

jba:
This is supported by the v0.2.0 API.

PR #50, opened 2025-06-25
mcp: memory example

MegaGrindStone:
## Motivation and Context
Part of #33 . This PR adds an example of implementing mcp on [memory server](https://github.com/modelcontextprotocol/servers/tree/main/src/memory).

## Types of changes

- [ ] Documentation update

## Checklist

- [ ] I have added appropriate error handling

### DIFF
--- a/examples/memory/kb.go
+++ b/examples/memory/kb.go
@@ -1,0 +1,676 @@
+// Copyright 2025 The Go MCP SDK Authors. All rights reserved.
+// Use of this source code is governed by an MIT-style
+// license that can be found in the LICENSE file.
+
+package main
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"os"
+	"slices"
+	"strings"
+
+	"github.com/modelcontextprotocol/go-sdk/mcp"
+)
+
+// store is an interface for knowledge base persistence.
+type store interface {
+	Read() ([]byte, error)
+	Write(data []byte) error
+}
+
+// memoryStore implements the store interface for in-memory persistence.
+type memoryStore struct {
+	data []byte
+}
+
+// Read reads data from the memory. If the data is empty, it returns an empty slice.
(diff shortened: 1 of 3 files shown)


MegaGrindStone:
One thing I just realized when implementing this is that the `StructuredContent` isn't marshalled into `CallToolResult`, which surprised me. Then, when I looks around, I think it's related to #20, isn't it? Or, do I miss something? If that is the case, then for now, I marshall the result into json string to be added as text content of `CallToolResult`.

jba:
That should be fixed by https://github.com/modelcontextprotocol/go-sdk/pull/46. Can you confirm?

jba:
I think this should live in the servers repo. See https://github.com/modelcontextprotocol/servers/issues/2220.

jba:
We can keep working on it here until we figure out the story with that.

MegaGrindStone:
> That should be fixed by https://github.com/modelcontextprotocol/go-sdk/pull/46. Can you confirm?

I'm sorry if I wasn't being clear, but the example is already working by marshalling the result into JSON string, and the external client could read the result as expected.

But, what was bothering me is, like being said in #20, if we pass the result struct directly (without marshalling into JSON) into StructuredContent, the result is not delivered to the client.

MegaGrindStone:
> I think this should live in the servers repo. See https://github.com/modelcontextprotocol/servers/issues/2220.

That would be great! And I think, this example (and the sequential thinking server #51) is more suited on that servers folder/repo than in the examples, as both doesn't show the usage of the mcp client implementation. Then, when we know where to put the servers repo, I could rewrite the examples to utilize the memory server, as a server and client implementation.

MegaGrindStone:
Based on the new version of #33, I think this pull request is no longer relevant. So I think it's best to close this.

jba:
We're still working on getting permission to review and approve on the servers repo. Once we have that we'll resume.

PR #51, opened 2025-06-26
mcp: add sequential thinking server example

nsega:
## Motivation and Context

This is part of #33.  This PR adds example of implementing mcp on [sequentialthinking server](https://github.com/modelcontextprotocol/servers/tree/main/src/sequentialthinking), from [Example Servers](https://modelcontextprotocol.io/examples).

The PR includes dynamic and reflective problem-solving through structured thinking processes. 
The server provides tools for:
- starting thinking sessions
- adding sequential thoughts with step tracking
- revising previous thoughts
- creating alternative reasoning branches
-  reviewing complete processes.

Also, features include thread-safe session management, adaptive planning that adjusts step counts dynamically.

## How Has This Been Tested?

Testing with 8 test functions covering all functionality with test code.

## Breaking Changes

N/A

## Types of changes

- [ ] Documentation update

## Checklist

- [ ] I have added appropriate error handling


examples/sequentialthinking/README.md
nsega, line 1: Note: Though we might not need to prepare for README.md for each example, I created this file, just in case.

jba:
After this lands, it would be great to also add it to the servers repo. See #52.

examples/sequentialthinking/README.md
jba: It's definitely helpful.

examples/sequentialthinking/README.md
```diff
@@ -0,0 +1,211 @@
...
+
+Each thinking session maintains:
+
+- **Session metadata**: ID, problem statement, creation time, current status
+- **Thought sequence**: Ordered list of thoughts with timestamps and revision history  
+- **Progress tracking**: Current step and estimated total steps
+- **Branch relationships**: Links to alternative reasoning paths
+- **Status management**: Active, completed, or paused sessions
+
+## Testing
```
jba: You can remove this section. It is already understood.

examples/sequentialthinking/main.go
```diff
@@ -0,0 +1,394 @@
...
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/modelcontextprotocol/go-sdk/mcp"
+)
+
+var httpAddr = flag.String("http", "", "if set, use streamable HTTP at this address, instead of stdin/stdout")
+
+// Thought represents a single step in the thinking process
```
jba: nit: period at end of sentence

examples/sequentialthinking/main.go
```diff
@@ -0,0 +1,394 @@
...
+	sessions map[string]*ThinkingSession
+}
+
+func NewSessionStore() *SessionStore {
+	return &SessionStore{
+		sessions: make(map[string]*ThinkingSession),
+	}
+}
+
+func (s *SessionStore) GetSession(id string) (*ThinkingSession, bool) {
```
jba: Remove the Get (idiomatic Go)

examples/sequentialthinking/main.go
jba: "A Thought is..."

examples/sequentialthinking/main.go
```diff
@@ -0,0 +1,394 @@
...
+// Thought represents a single step in the thinking process
+type Thought struct {
+	ID       int       `json:"id"`
+	Content  string    `json:"content"`
+	Created  time.Time `json:"created"`
+	Revised  bool      `json:"revised"`
+	ParentID *int      `json:"parentId,omitempty"` // For branching
+}
+
+// ThinkingSession represents an active thinking session
```
jba: ditto: "A ThinkingSession is"

examples/sequentialthinking/main.go
```diff
@@ -0,0 +1,394 @@
...
+	"time"
+
+	"github.com/modelcontextprotocol/go-sdk/mcp"
+)
+
+var httpAddr = flag.String("http", "", "if set, use streamable HTTP at this address, instead of stdin/stdout")
+
+// Thought represents a single step in the thinking process
+type Thought struct {
+	ID       int       `json:"id"`
```
jba: We'll need brief doc for each of these. For example:
```
   // Globally unique ID of the thought.
  ID int

  ...
   // ID of parent, or nil if this is a root.
   ParentID *int
```

Also, are the IDs globally unique, or just on this server, or per session? Note in the doc.

examples/sequentialthinking/main.go
```diff
@@ -0,0 +1,394 @@
...
+	Created  time.Time `json:"created"`
+	Revised  bool      `json:"revised"`
+	ParentID *int      `json:"parentId,omitempty"` // For branching
+}
+
+// ThinkingSession represents an active thinking session
+type ThinkingSession struct {
+	ID             string    `json:"id"`
+	Problem        string    `json:"problem"`
+	Thoughts       []Thought `json:"thoughts"`
```
jba: A Thought has mutable state, so *Thought is probably better.

examples/sequentialthinking/main.go
```diff
@@ -0,0 +1,394 @@
...
+	Status         string    `json:"status"` // "active", "completed", "paused"
+	Created        time.Time `json:"created"`
+	LastActivity   time.Time `json:"lastActivity"`
+	Branches       []string  `json:"branches,omitempty"` // Alternative thought paths
+}
+
+// Global session store (in a real implementation, this might be a database)
+type SessionStore struct {
+	mu       sync.RWMutex
+	sessions map[string]*ThinkingSession
```
jba: I usually write something like
```
sessions map[string]*ThinkingSession // key is session ID
```

examples/sequentialthinking/main.go
```diff
@@ -0,0 +1,394 @@
...
+func (s *SessionStore) SetSession(session *ThinkingSession) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	s.sessions[session.ID] = session
+}
+
+func (s *SessionStore) ListSessions() []*ThinkingSession {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+	sessions := make([]*ThinkingSession, 0, len(s.sessions))
```
jba: slices.Clone

examples/sequentialthinking/main.go
```diff
@@ -0,0 +1,394 @@
...
+	return session, exists
+}
+
+func (s *SessionStore) SetSession(session *ThinkingSession) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	s.sessions[session.ID] = session
+}
+
+func (s *SessionStore) ListSessions() []*ThinkingSession {
```
jba: Sessions

examples/sequentialthinking/main.go
```diff
@@ -0,0 +1,394 @@
...
+	sessions := make([]*ThinkingSession, 0, len(s.sessions))
+	for _, session := range s.sessions {
+		sessions = append(sessions, session)
+	}
+	return sessions
+}
+
+var store = NewSessionStore()
+
+// Tool argument structures
```
jba: Separate doc for each.

examples/sequentialthinking/main.go
```diff
@@ -0,0 +1,394 @@
...
+	SessionID      string `json:"sessionId,omitempty"`
+	EstimatedSteps int    `json:"estimatedSteps,omitempty"`
+}
+
+type ContinueThinkingArgs struct {
+	SessionID      string `json:"sessionId"`
+	Thought        string `json:"thought"`
+	NextNeeded     *bool  `json:"nextNeeded,omitempty"`
+	ReviseStep     *int   `json:"reviseStep,omitempty"`
+	CreateBranch   *bool  `json:"createBranch,omitempty"`
```
jba: Is this and the one above a *bool because the default is true?

examples/sequentialthinking/main.go
```diff
@@ -0,0 +1,394 @@
...
+	SessionID string `json:"sessionId"`
+}
+
+// Tool implementations
+func StartThinking(ctx context.Context, ss *mcp.ServerSession, params *mcp.CallToolParamsFor[StartThinkingArgs]) (*mcp.CallToolResultFor[any], error) {
+	args := params.Arguments
+
+	sessionID := args.SessionID
+	if sessionID == "" {
+		sessionID = fmt.Sprintf("session_%d", time.Now().Unix())
```
jba: Insufficiently unique: this Unix returns times to the second. I wouldn't even use UnixNano. Just generate random text. See our randText function in util.go.

examples/sequentialthinking/main.go
```diff
@@ -0,0 +1,394 @@
...
+
+	estimatedSteps := args.EstimatedSteps
+	if estimatedSteps == 0 {
+		estimatedSteps = 5 // Default estimate
+	}
+
+	session := &ThinkingSession{
+		ID:             sessionID,
+		Problem:        args.Problem,
+		Thoughts:       []Thought{},
```
jba: Omit, nil is fine.

examples/sequentialthinking/main.go
```diff
@@ -0,0 +1,394 @@
...
+		CurrentThought: 0,
```
jba: omit, zero is the default.

examples/sequentialthinking/main.go
```diff
@@ -0,0 +1,394 @@
...
+		EstimatedTotal: estimatedSteps,
+		Status:         "active",
+		Created:        time.Now(),
+		LastActivity:   time.Now(),
+		Branches:       []string{},
```
jba: omit

examples/sequentialthinking/main.go
```diff
@@ -0,0 +1,394 @@
...
+				&mcp.TextContent{
+					Text: fmt.Sprintf("Revised step %d in session '%s':\n%s",
+						*args.ReviseStep, args.SessionID, args.Thought),
+				},
+			},
+		}, nil
+	}
+
+	// Handle branching
+	if args.CreateBranch != nil && *args.CreateBranch {
```
jba: It looks like the default for CreateBranch is false. So it doesn't need to be a pointer.

examples/sequentialthinking/main.go
```diff
@@ -0,0 +1,394 @@
...
+	session.LastActivity = time.Now()
+
+	// Handle revision of existing thought
+	if args.ReviseStep != nil {
+		stepIndex := *args.ReviseStep - 1
+		if stepIndex < 0 || stepIndex >= len(session.Thoughts) {
+			return nil, fmt.Errorf("invalid step number: %d", *args.ReviseStep)
+		}
+
+		session.Thoughts[stepIndex].Content = args.Thought
```
jba: Why doesn't session need locking? Can you be sure that two requests for the same session don't arrive concurrently?

examples/sequentialthinking/main.go
```diff
@@ -0,0 +1,394 @@
...
+	// Handle branching
+	if args.CreateBranch != nil && *args.CreateBranch {
+		branchID := fmt.Sprintf("%s_branch_%d", args.SessionID, len(session.Branches)+1)
+		session.Branches = append(session.Branches, branchID)
+
+		// Create a new session for the branch
+		branchSession := &ThinkingSession{
+			ID:             branchID,
+			Problem:        session.Problem + " (Alternative branch)",
+			Thoughts:       append([]Thought{}, session.Thoughts...), // Copy existing thoughts
```
jba: Use slices.Clone.

examples/sequentialthinking/main.go
```diff
@@ -0,0 +1,394 @@
...
+	if args.ReviseStep != nil {
+		stepIndex := *args.ReviseStep - 1
+		if stepIndex < 0 || stepIndex >= len(session.Thoughts) {
+			return nil, fmt.Errorf("invalid step number: %d", *args.ReviseStep)
+		}
+
+		session.Thoughts[stepIndex].Content = args.Thought
+		session.Thoughts[stepIndex].Revised = true
+
+		store.SetSession(session)
```
jba: I don't think you need this: you are modifying a *ThinkingSession. (I _am_ concerned about locking; see above.)

examples/sequentialthinking/main.go
```diff
@@ -0,0 +1,394 @@
...
+			EstimatedTotal: session.EstimatedTotal,
+			Status:         "active",
+			Created:        time.Now(),
+			LastActivity:   time.Now(),
+			Branches:       []string{},
+		}
+
+		store.SetSession(branchSession)
+		session.LastActivity = time.Now()
+		store.SetSession(session)
```
jba: Unnecessary.

examples/sequentialthinking/main.go
```diff
@@ -0,0 +1,394 @@
...
+				&mcp.TextContent{
+					Text: fmt.Sprintf("Created branch '%s' from session '%s'. You can now continue thinking in either session.",
+						branchID, args.SessionID),
+				},
+			},
+		}, nil
+	}
+
+	// Add new thought
+	thoughtID := len(session.Thoughts) + 1
```
jba: OK, so it's unique per session. It's also just the index into Thoughts, so maybe Index would be a better field name.

examples/sequentialthinking/main.go
```diff
@@ -0,0 +1,394 @@
...
+		Content: args.Thought,
+		Created: time.Now(),
+		Revised: false,
+	}
+
+	session.Thoughts = append(session.Thoughts, thought)
+	session.CurrentThought = thoughtID
+
+	// Update estimated total if provided
+	if args.EstimatedTotal != nil {
```
jba: Does it make sense to have EstimatedTotal == 0? I don't think so. In that case, you can use 0 as "not present" and drop the pointer.

examples/sequentialthinking/main.go
```diff
@@ -0,0 +1,394 @@
...
+		session.EstimatedTotal = *args.EstimatedTotal
+	}
+
+	// Check if thinking is complete
+	if args.NextNeeded != nil && !*args.NextNeeded {
+		session.Status = "completed"
+	}
+
+	store.SetSession(session)
```
jba: ditto

examples/sequentialthinking/main.go
```diff
@@ -0,0 +1,394 @@
...
+func ReviewThinking(ctx context.Context, ss *mcp.ServerSession, params *mcp.CallToolParamsFor[ReviewThinkingArgs]) (*mcp.CallToolResultFor[any], error) {
+	args := params.Arguments
+
+	session, exists := store.GetSession(args.SessionID)
+	if !exists {
+		return nil, fmt.Errorf("session %s not found", args.SessionID)
+	}
+
+	var review strings.Builder
+	review.WriteString(fmt.Sprintf("=== Thinking Review: %s ===\n", session.ID))
```
jba: fmt.Fprintf(&review, "...",...)

examples/sequentialthinking/main.go
```diff
@@ -0,0 +1,394 @@
...
+		Content: []mcp.Content{
+			&mcp.TextContent{
+				Text: review.String(),
+			},
+		},
+	}, nil
+}
+
+// Resource handler for thinking sessions
+func GetThinkingHistory(ctx context.Context, ss *mcp.ServerSession, params *mcp.ReadResourceParams) (*mcp.ReadResourceResult, error) {
```
jba: Remove `Get`

examples/sequentialthinking/main.go
```diff
@@ -0,0 +1,394 @@
...
+	// Extract session ID from URI (e.g., "thinking://session_123")
+	parts := strings.Split(params.URI, "://")
```
jba: use url.Parse

examples/sequentialthinking/main.go
```diff
@@ -0,0 +1,394 @@
...
+}
+
+func main() {
+	flag.Parse()
+
+	server := mcp.NewServer("sequential-thinking", "v0.0.1", nil)
+
+	// Add thinking tools
+	server.AddTools(
+		mcp.NewServerTool(
```
jba: Unfortunately, this will soon be a bug.
According to the spec, tools with an output schema must provide a StructuredOutput field in the result.
We don't check for that today, but I'm working on the PR.
NewServerTool creates both an input and output schema.

Can you try creating these tools without NewServerTool, using the part of NewServerTool that does input schemas, and leaving out the outputschema? It would be useful to see how complicated that gets.

examples/sequentialthinking/main.go
nsega: The defaults I expected are actually:
  - `NextNeeded` : default behavior is `true` for continue thinking.
  - `CreateBranch` : default behavior is `false` for don't create  a branch.

Since only CreateBranch doesn't need a pointer in this case, I updated the field from *bool to bool.

examples/sequentialthinking/main.go
nsega: Thank you for your suggestion. Since I didn't know the randText function, I am glad to have found a way to generate sufficient uniqueness. I copied randText() from util.go.

examples/sequentialthinking/main.go
nsega: You're totally right. Just omitted the field.

examples/sequentialthinking/main.go
nsega: This is also you're right. Just omitted the field.

examples/sequentialthinking/main.go
nsega: Just omitted the field, too.

examples/sequentialthinking/main.go
nsega: Your point is very true. Thanks to your feedback, I noticed that there is a case where two requests for the same session arrive concurrently.

To address this, I added Version filed to ThinkingSession for optimistic concurrency control and updated the ContinueThinking function to use the new CompareAndSwap function for all mutation, such as thought revision, branch creation, and adding new thoughts. How does it sound to you?

examples/sequentialthinking/main.go
nsega: Understood. By creating CompareAndSwap function, I removed this part from here.

examples/sequentialthinking/main.go
nsega: Since I addressed the locking of session, this part has been changed using a deep copy.

examples/sequentialthinking/main.go
nsega: Your suggestion totally makes sense to me. Just update the field name `ID` of Thoughts to the field name `Index`.

examples/sequentialthinking/main.go
nsega: Glad to hear the feedback. Your point completely makes sense to me. So, I updated *int of the field EstimatedTotal to int.

examples/sequentialthinking/main.go
nsega: Addressed It.

examples/sequentialthinking/main.go
nsega: Understood. Updated these parts to using fmt.Fprint in ReviewThinking function.

examples/sequentialthinking/main.go
nsega: Addressed it.

examples/sequentialthinking/main.go
nsega: Updated it to use url.Parse.

examples/sequentialthinking/main.go
nsega: It is my careless. Just updated it.

nsega: Updated the comment by following the format you suggested.

examples/sequentialthinking/main.go
nsega: Since this ID is unique per session, I updated the file name `ID` to `Index` to make its meaning easier to understand, as you pointed out.

examples/sequentialthinking/main.go
nsega: Updated the part by following the format.

examples/sequentialthinking/main.go
nsega: You're totally right. Updated the the filed to *Thought.

examples/sequentialthinking/main.go
nsega: Thanks for your input. Glad to know that. Added the comment by following your suggestion.

examples/sequentialthinking/main.go
nsega: Removed the Get by following idiomatic Go.

examples/sequentialthinking/main.go
nsega: Updated the function name to Sessions.

examples/sequentialthinking/main.go
nsega: Updated the part to use slices.Clone.

examples/sequentialthinking/main.go
nsega: Added the doc for each struct respectively.

examples/sequentialthinking/main.go
nsega: Thank you for your explanation. I understand the current status of the PR you're working on. 

Yes, I can. I tried to implement these tools without NewServerTool. As I may have missed something, I would appreciate it if you could provide me with further feedback on the implementation.

nsega (commented):
Thank you so much for your feedback. I addressed the feedback. Could you please take a look at it again at your convenience?

examples/sequentialthinking/README.md
nsega: Understood. I appreciate your input.

examples/sequentialthinking/main.go
```diff
@@ -0,0 +1,615 @@
...
+		var branchID string
+		var branchSession *ThinkingSession
+
+		err := store.CompareAndSwap(args.SessionID, func(session *ThinkingSession) (*ThinkingSession, error) {
+			branchID = fmt.Sprintf("%s_branch_%d", args.SessionID, len(session.Branches)+1)
+			session.Branches = append(session.Branches, branchID)
+			session.LastActivity = time.Now()
+
+			// Create a new session for the branch (deep copy thoughts)
+			thoughtsCopy := make([]*Thought, len(session.Thoughts))
```
jba: You deep-copy the thoughts in two places, so factor it out to a separate function.

examples/sequentialthinking/main.go
```diff
@@ -0,0 +1,615 @@
...
+	session, exists := store.Session(args.SessionID)
+	if !exists {
+		return nil, fmt.Errorf("session %s not found", args.SessionID)
+	}
+
+	var review strings.Builder
+	fmt.Fprintf(&review, "=== Thinking Review: %s ===\n", session.ID)
+	fmt.Fprintf(&review, "Problem: %s\n", session.Problem)
+	fmt.Fprintf(&review, "Status: %s\n", session.Status)
+	fmt.Fprintf(&review, "Steps: %d of ~%d\n", len(session.Thoughts), session.EstimatedTotal)
```
jba: You take a read lock for Thoughts in CompareAndSwap. Why not here?

examples/sequentialthinking/main.go
```diff
@@ -0,0 +1,615 @@
...
+	}
+	if u.Scheme != "thinking" {
+		return nil, fmt.Errorf("invalid thinking resource URI scheme: %s", u.Scheme)
+	}
+
+	sessionID := u.Host
+	if sessionID == "sessions" {
+		// List all sessions
+		sessions := store.Sessions()
+		data, err := json.MarshalIndent(sessions, "", "  ")
```
jba, line 451: This is going to read all sessions with no lock.

examples/sequentialthinking/main.go
```diff
@@ -0,0 +1,615 @@
...
+		}, nil
+	}
+
+	// Get specific session
+	session, exists := store.Session(sessionID)
+	if !exists {
+		return nil, fmt.Errorf("session %s not found", sessionID)
+	}
+
+	data, err := json.MarshalIndent(session, "", "  ")
```
jba, line 473: Ditto on reading.

jba (changes requested):
Let's think through the locking. Is it possible to concurrently read and write fields of a ThinkingSession? I don't think so, but check all field references to see if there are updates to existing (as opposed to fresh) copies.

If there are none, then I don't think you need the read lock on a session in CompareAndSwap.

Either way, doc the situation somewhere.

examples/sequentialthinking/main.go
nsega: Thank you for your input. I factored it out to the separate function `deepCopyThoughts` .

examples/sequentialthinking/main.go
nsega: You're right. I missed taking  a read lock here. I added the read lock by creating a snapshot of the data we need while holding the lock before reading it.

examples/sequentialthinking/main.go
nsega: You're right about that, too. I've fixed this part by using store.SessionsSnapshot(), which returns a deep copy of all sessions with read lock.

examples/sequentialthinking/main.go
nsega: I've fixed the same one here, too. as with https://github.com/modelcontextprotocol/go-sdk/pull/51#discussion_r2179770944

nsega:
@jba I appreciate your additional feedback. I addressed it. Could you please take a look at it again?

In my understanding, the Go runtime includes best-effort detection for concurrent map misuse that may result in panics. However, I would say that a read lock is necessary for SessionThink, as there's best-effort detection that may or may not catch the race. 
Also, just in case, I read through [map.go](https://cs.opensource.google/go/go/+/refs/tags/go1.23.0:src/runtime/map.go) and verified with `go test -race` to make sure no race conditions remain, just to be safe. 

If I've misunderstood anything, I'd really appreciate it if you could let me know. Thank you.

examples/sequentialthinking/main.go
```diff
@@ -146,6 +164,39 @@ func (s *SessionStore) Sessions() []*ThinkingSession {
...
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+
+	session, exists := s.sessions[id]
+	if !exists {
+		return nil, false
+	}
+
+	// Create a deep copy
+	sessionCopy := *session
```
jba: This code occurs twice. Factor it out into a `ThinkingSession.clone` method.

jba (changes requested):

examples/sequentialthinking/main.go
nsega: @jba Thank you for pointing that out. I factored it out to a `ThinkingSession.clone` method. Can you please take a look at it again when you have chance?

examples/sequentialthinking/main.go
```diff
@@ -0,0 +1,669 @@
...
+	return &sessionCopy
+}
+
+// A SessionStore is a global session store (in a real implementation, this might be a database).
+//
+// Locking Strategy:
+// The SessionStore uses a RWMutex to protect the sessions map from concurrent access.
+// All ThinkingSession modifications happen on deep copies, never on shared instances.
+// This means:
+// - Read locks protect map access (reading from a Go map during writes causes panics)
```
jba: Remove the part in parentheses. It's not always true, and it gives the wrong impression that we lock to avoid panics. We lock to avoid data races; panics are just (sometimes) a consequence of data races.

jba: End these sentences with periods.

jba: One more thing, about code review etiquette. When you finish these small fixes, you can just reply "done" and resolve the comment. I know that may seem rude to you, but that's what everyone does. You are being very polite—too polite!

examples/sequentialthinking/main.go
```diff
@@ -0,0 +1,669 @@
...
+	defer s.mu.RUnlock()
+	return slices.Collect(maps.Values(s.sessions))
+}
+
+// SessionsSnapshot returns a deep copy of all sessions for safe concurrent access.
+func (s *SessionStore) SessionsSnapshot() []*ThinkingSession {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+
+	sessions := make([]*ThinkingSession, 0, len(s.sessions))
```
jba: `var sessions []*ThinkingSession`
You don't need to bother with allocating the slice. Even though it will be faster, it is a premature optimization.

Or here is another way to do this, which I think is very readable:
```
sessions := slices.Clone(s.sessions)
for i, s := range sessions {
    sessions[i] = s.clone()
}
```

examples/sequentialthinking/main.go
```diff
@@ -0,0 +1,669 @@
...
+	for _, session := range s.sessions {
+		// Create a deep copy of each session
```
jba: remove comment

examples/sequentialthinking/main.go
```diff
@@ -0,0 +1,669 @@
...
+func (s *SessionStore) SessionSnapshot(id string) (*ThinkingSession, bool) {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+
+	session, exists := s.sessions[id]
+	if !exists {
+		return nil, false
+	}
+
+	// Create a deep copy
```
jba: remove comment

examples/sequentialthinking/main.go
```diff
@@ -0,0 +1,669 @@
...
+	sessions := make([]*ThinkingSession, 0, len(s.sessions))
+	for _, session := range s.sessions {
+		// Create a deep copy of each session
+		sessions = append(sessions, session.clone())
+	}
+	return sessions
+}
+
+// SessionSnapshot returns a deep copy of a session for safe concurrent access.
+func (s *SessionStore) SessionSnapshot(id string) (*ThinkingSession, bool) {
```
jba, line 182: "The second return value reports whether a session with the given id exists."

examples/sequentialthinking/main.go
```diff
@@ -0,0 +1,669 @@
...
+// ThinkingHistoryArgs are the arguments for retrieving thinking history.
+type ThinkingHistoryArgs struct {
+	SessionID string `json:"sessionId"`
+}
+
+// deepCopyThoughts creates a deep copy of a slice of thoughts.
+func deepCopyThoughts(thoughts []*Thought) []*Thought {
+	thoughtsCopy := make([]*Thought, len(thoughts))
+	for i, t := range thoughts {
+		thoughtCopy := *t
```
jba: Use a short name because the scope is small. `c` is fine (short for "copy"), or `t2`, or `tt`. Whatever you prefer.

examples/sequentialthinking/main.go
```diff
@@ -0,0 +1,669 @@
...
+	continueThinkingSchema, err := jsonschema.For[ContinueThinkingArgs]()
+	if err != nil {
+		log.Fatalf("Failed to create continue_thinking schema: %v", err)
+	}
+	reviewThinkingSchema, err := jsonschema.For[ReviewThinkingArgs]()
+	if err != nil {
+		log.Fatalf("Failed to create review_thinking schema: %v", err)
+	}
+
+	server.AddTools(
```
jba: Unfortunately this won't merge because of the new API. Can you modify it to use AddTool?
Try the generic `mcp.AddTool` function first. If that doesn't work for you, let me know why.

examples/sequentialthinking/main.go
nsega: Done.

examples/sequentialthinking/main.go
nsega: Glad to know that. You're right. It's not always true. Just removed it.

nsega: DONE.

nsega: I really appreciate you pointing that out. I wasn’t aware of that review etiquette, so it was eye-opening in a good way! Thanks for helping me learn something new.

examples/sequentialthinking/main.go
nsega: I opted for the first one, though I also preferred the second suggestion. I found the second one doesn't work because `s.sessions` is a `map[string]*ThinkingSession`, not a slice. As you know, `slices.Clone`  only works on slices, not maps.

examples/sequentialthinking/main.go
nsega: done.

examples/sequentialthinking/main.go
nsega: done.

examples/sequentialthinking/main.go
nsega: done.

examples/sequentialthinking/main.go
nsega: Used `t2`.  done.

examples/sequentialthinking/main.go
nsega: Yes, I could update the target implementations of the main func to use `mcp.AddTool` instead of `server.AddTools` with manual schemas.

nsega:
@jba Thanks so much for your feedback. I addressed them. Could you please take a look at it again?

jba (approved):
Thanks for all your hard work on this!

nsega:
@jba Thank you very much for your kind review so far! I was really excited to contribute to go-sdk for the first time.

PR #55, opened 2025-06-26
mcp: implement completion

samthanawalla:
This CL implements a completion handler for the server and a complete method for the client.

## Motivation and Context

This adds support for completion.

## How Has This Been Tested?

Added tests and an example.

## Breaking Changes

N/A

## Types of changes

- [ x] New feature (non-breaking change which adds functionality)
- [ ] Documentation update

## Checklist

- [ x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [ x] My code follows the repository's style guidelines
- [ x] New and existing tests pass locally
- [ x] I have added appropriate error handling
- [ x] I have added or updated documentation as needed

### DIFF
--- a/examples/completion/main.go
+++ b/examples/completion/main.go
@@ -1,0 +1,49 @@
+// Copyright 2025 The Go MCP SDK Authors. All rights reserved.
+// Use of this source code is governed by an MIT-style
+// license that can be found in the LICENSE file.
+
+package main
+
+import (
+	"context"
+	"fmt"
+	"log"
+
+	"github.com/modelcontextprotocol/go-sdk/mcp"
+)
+
+// This example demonstrates the minimal code to declare and assign
+// a CompletionHandler to an MCP Server's options.
+func main() {
+	// Define your custom CompletionHandler logic
+	myCompletionHandler := func(_ context.Context, _ *mcp.ServerSession, params *mcp.CompleteParams) (*mcp.CompleteResult, error) {
+		// In a real application, you'd implement actual completion logic here
+		// For this example, we return a fixed set of suggestions.
+		var suggestions []string
+		switch params.Ref.Type {
+		case "ref/prompt":
+			suggestions = []string{"suggestion1", "suggestion2", "suggestion3"}
+		case "ref/resource":
+			suggestions = []string{"suggestion4", "suggestion5", "suggestion6"}
+		default:
+			return nil, fmt.Errorf("unrecognized content type %s", params.Ref.Type)
+		}
+
+		return &mcp.CompleteResult{
+			Completion: mcp.CompletionResultDetails{
+				HasMore: false,
+				Total:   int64(len(suggestions)),
+				Values:  suggestions,
+			},
+		}, nil
+	}
+
+	// Create the MCP Server instance and assign the handler
+	// No server running, just showing the configuration.
+	_ = mcp.NewServer("myServer", "v1.0.0", &mcp.ServerOptions{
+		CompletionHandler: myCompletionHandler,
+	})
+
+	log.Println("MCP Server instance created with a CompletionHandler assigned (but not running).")
+	log.Println("This example demonstrates configuration, not live interaction.")
+}
--- a/mcp/completion.go
+++ b/mcp/completion.go
@@ -1,0 +1,35 @@
+// Copyright 2025 The Go MCP SDK Authors. All rights reserved.
+// Use of this source code is governed by an MIT-style
+// license that can be found in the LICENSE file.
+
+package mcp
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+)
+
+type Reference struct {
+	Type string `json:"type"`
+	Name string `json:"name,omitempty"`
+	URI  string ` json:"uri,omitempty"`
+}
+
+func (r *Reference) UnmarshalJSON(data []byte) error {
+	type wireReference Reference // for naive unmarshaling
+	var r2 wireReference
+	if err := json.Unmarshal(data, &r2); err != nil {
+		return err
+	}
+	switch r2.Type {
+	case "ref/prompt", "ref/resource":
+	default:
+		return fmt.Errorf("unrecognized content type %s", r2.Type)
+	}
+	*r = Reference(r2)
+	return nil
+}
+
+// A CompletionHandler handles completion.
+type CompletionHandler func(context.Context, *ServerSession, *CompleteParams) (*CompleteResult, error)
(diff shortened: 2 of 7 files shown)


examples/completion/main.go
```diff
@@ -0,0 +1,49 @@
...
+	"fmt"
+	"log"
+
+	"github.com/modelcontextprotocol/go-sdk/mcp"
+)
+
+// This example demonstrates the minimal code to declare and assign
+// a CompletionHandler to an MCP Server's options.
+func main() {
+	// Define your custom CompletionHandler logic
```
jba: nit: end with period

mcp/completion.go
```diff
@@ -0,0 +1,35 @@
...
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+)
+
+type Reference struct {
+	Type string `json:"type"`
+	Name string `json:"name,omitempty"`
+	URI  string ` json:"uri,omitempty"`
```
jba: Remove space between ` and j

mcp/completion.go
```diff
@@ -0,0 +1,35 @@
...
+
+package mcp
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+)
+
+type Reference struct {
```
jba: How about CompleteReference so it's not such a general name?
(The spec doesn't have this type, but I think we should.)

mcp/completion.go
```diff
@@ -0,0 +1,35 @@
...
+	Type string `json:"type"`
+	Name string `json:"name,omitempty"`
```
jba: doc this is only valid for ref/prompt
ditto URI for ref/resource

mcp/completion.go
```diff
@@ -0,0 +1,35 @@
...
+
+func (r *Reference) UnmarshalJSON(data []byte) error {
+	type wireReference Reference // for naive unmarshaling
+	var r2 wireReference
+	if err := json.Unmarshal(data, &r2); err != nil {
+		return err
+	}
+	switch r2.Type {
+	case "ref/prompt", "ref/resource":
+	default:
```
jba: Might as well also check that names go with prompts and uris with resources.

mcp/completion.go
jba: should live in protocol.go

mcp/streamable_test.go
```diff
@@ -143,10 +143,11 @@ func TestStreamableServerTransport(t *testing.T) {
 	initReq := req(1, "initialize", &InitializeParams{})
 	initResp := resp(1, &InitializeResult{
 		Capabilities: &serverCapabilities{
-			Logging:   &loggingCapabilities{},
-			Prompts:   &promptCapabilities{ListChanged: true},
-			Resources: &resourceCapabilities{ListChanged: true},
-			Tools:     &toolCapabilities{ListChanged: true},
+			Completions: completionCapabilities{},
```
jba: why not a pointer?

mcp/protocol.go
```diff
@@ -163,6 +163,36 @@ type ClientCapabilities struct {
...
+	// attach additional metadata to their responses.
+	Meta `json:"_meta,omitempty"`
+	// The argument's information
+	Argument CompleteParamsArgument `json:"argument"`
+	Ref      *Reference             `json:"ref"`
+}
+
+type CompletionResultDetails struct {
+	HasMore bool     `json:"hasMore,omitempty"`
+	Total   int64    `json:"total,omitempty"`
```
jba: Why int64? I think int would be easier to use.

mcp/completion.go
```diff
@@ -0,0 +1,35 @@
...
+	"encoding/json"
+	"fmt"
+)
+
+type Reference struct {
+	Type string `json:"type"`
+	Name string `json:"name,omitempty"`
+	URI  string ` json:"uri,omitempty"`
+}
+
```
jba: Add a marshal too, with the same checks.

mcp/protocol.go
```diff
@@ -163,6 +163,36 @@ type ClientCapabilities struct {
 	Elicitation *ElicitationCapabilities `json:"elicitation,omitempty"`
 }
 
+type CompleteParamsArgument struct {
+	// The name of the argument
+	Name string `json:"name"`
+	// The value of the argument to use for completion matching.
+	Value string `json:"value"`
```
SamMorrowDrums, line 190: > For prompts or URI templates with multiple arguments, clients should include previous completions in the context.arguments object to provide context for subsequent requests.

https://modelcontextprotocol.io/specification/2025-06-18/server/utilities/completion#requesting-completions

This is in the new specification, but we worked hard to get it included and I assume the Go SDK will support the new spec soon, so it would be good to plan for the additional fields even if commented out.

An example use case is GitHub Repository URIs where each subsquent path completion for a Template Resource is dependent on the previous, so they must all be passed back for completions to even be useful in that context.

Image

examples/completion/main.go
samthanawalla: Done

mcp/completion.go
samthanawalla: Done

samthanawalla: Done

mcp/completion.go
samthanawalla: Done

mcp/completion.go
samthanawalla: Done

mcp/completion.go
samthanawalla: Done

mcp/completion.go
samthanawalla: Done

mcp/protocol.go
samthanawalla: Done

mcp/streamable_test.go

mcp/protocol.go
samthanawalla: Thanks for the heads up. The spec update makes a lot of sense. I included them in this CL.

jba (approved):

mcp/protocol.go
```diff
@@ -182,6 +183,93 @@ type ClientCapabilities struct {
...
+}
+
+// CompleteContext represents additional, optional context for completions.
+type CompleteContext struct {
+	// Previously-resolved variables in a URI template or prompt.
+	Arguments []string `json:"arguments,omitempty"`
```
SamMorrowDrums: This should be a
`map[string]string` I believe 

```json
"context": {
      "arguments": {
        "language": "python"
      }
    }

```

SamMorrowDrums (changes requested):

mcp/protocol.go
samthanawalla: Thank you so much for catching that! 🙏 

Updated

jba (approved):

PR #57, opened 2025-06-26
[WIP] Use interfaces for Tool definitions

SamMorrowDrums:
## Motivation and Context

This PR is an experiment to see how using interfaces could look, so that the `in` and `out` types are in control of how their schema is generated, how their params are parsed and how their results are returned.

Related discussion and issues:
- https://github.com/modelcontextprotocol/go-sdk/discussions/37
- https://github.com/modelcontextprotocol/go-sdk/issues/20

The primary change is to update `NewServerTool` and `ToolHandlerFor` to work with interfaces rather than pure generics:

```golang
// A ToolInput is a tool definition defining the input of a tool.
type ToolInput interface {
	// Schema returns the JSON schema for the tool's input.
	Schema() (*jsonschema.Schema, error)
	// SetParams sets the parameters for the tool's input from raw JSON.
	SetParams(json.RawMessage) error
}

// A ToolOutput is an interface defining the output of a tool.
type ToolOutput interface {
	// Schema returns the JSON schema for the tool's output.
	// Schema() (*jsonschema.Schema, error) // TODO implement output schema in future
	// Result returns the result of the tool's execution.
	Result() (*CallToolResult, error)
}

```

**Note: this change is not fully fleshed out yet, as of opening this PR it is the minimum change to get everything building, and doesn't finish fleshing out all the ideas that become possible with this change, nor does it fully sell the benefits.**

## How Has This Been Tested?

All test cases have been updated

## Breaking Changes

Yes, this is an extremely breaking change.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

### DIFF
--- a/examples/hello/main.go
+++ b/examples/hello/main.go
@@ -6,6 +6,7 @@ package main
 
 import (
 	"context"
+	"encoding/json"
 	"flag"
 	"fmt"
 	"log"
@@ -13,6 +14,7 @@ import (
 	"net/url"
 	"os"
 
+	"github.com/modelcontextprotocol/go-sdk/jsonschema"
 	"github.com/modelcontextprotocol/go-sdk/mcp"
 )
 
@@ -22,11 +24,36 @@ type HiArgs struct {
 	Name string `json:"name"`
 }
 
-func SayHi(ctx context.Context, ss *mcp.ServerSession, params *mcp.CallToolParamsFor[HiArgs]) (*mcp.CallToolResultFor[struct{}], error) {
-	return &mcp.CallToolResultFor[struct{}]{
+func (h *HiArgs) Schema() (*jsonschema.Schema, error) {
+	return jsonschema.For[HiArgs]()
+}
+
+func (h *HiArgs) SetParams(raw json.RawMessage) error {
+	return json.Unmarshal(raw, h)
+}
+
+type HiResult struct {
+	Message string
+}
+
+func (h *HiResult) Result() (*mcp.CallToolResult, error) {
+	return &mcp.CallToolResult{
 		Content: []mcp.Content{
-			&mcp.TextContent{Text: "Hi " + params.Arguments.Name},
+			&mcp.TextContent{Text: h.Message},
 		},
+	}, nil
+}
+
+func SayHi(ctx context.Context, ss *mcp.ServerSession, params *mcp.CallToolParamsFor[json.RawMessage]) (*HiResult, error) {
+	var args HiArgs
+	if params.Arguments != nil {
+		if err := args.SetParams(params.Arguments); err != nil {
+			return nil, err
+		}
+	}
+
+	return &HiResult{
+		Message: "Hi " + args.Name,
 	}, nil
 }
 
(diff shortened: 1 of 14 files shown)


jba:
Thinking about this.
It seems like a lot of work to define methods for each ToolInput and ToolOutput. Maybe this could be some sort of fallback if inference or explicitly writing the schema doesn't work out. Let's explore other options and come back to this.

examples/hello/main.go
```diff
@@ -22,14 +24,39 @@ type HiArgs struct {
 	Name string `json:"name"`
 }
 
-func SayHi(ctx context.Context, ss *mcp.ServerSession, params *mcp.CallToolParamsFor[HiArgs]) (*mcp.CallToolResultFor[struct{}], error) {
-	return &mcp.CallToolResultFor[struct{}]{
+func (h *HiArgs) Schema() (*jsonschema.Schema, error) {
+	return jsonschema.For[HiArgs]()
+}
+
+func (h *HiArgs) SetParams(raw json.RawMessage) error {
```
jba, line 31: When would this be any different? In other words, can we drop the interface method and just do this automatically?

mcp/tool.go
```diff
@@ -46,22 +62,25 @@ type ServerTool struct {
 // schema may be customized using the [Input] option.
 //
 // TODO(jba): check that structured content is set in response.
-func NewServerTool[In, Out any](name, description string, handler ToolHandlerFor[In, Out], opts ...ToolOption) *ServerTool {
+func NewServerTool[In ToolInput, Out ToolOutput](name, description string, handler ToolHandlerFor[In, Out], opts ...ToolOption) *ServerTool {
```
jba, line 65: My main concern is that this forces all the input and output types to implement the methods.
This would only be worth it if the type is used for more than one tool.
But how likely is that?

examples/hello/main.go
SamMorrowDrums: I was thinking that some users might want to control the parsing more specifically, so the pure generic version was not flexible enough, it made the assumption that you would always parse the args into the type in this way - and that may not always be the case.

mcp/tool.go
SamMorrowDrums: > This would only be worth it if the type is used for more than one tool.

It's worth it for being able to define how the parsing and schema generation occurs. An example in the discussion is a request was to be able to order the schema in a specific way to encourage the LLM to provide the arguments in the same order. That kind of flexibility is not possible with the totally generic approach.

Also, for example I could have list GitHub Issues, and Search GitHub Issues, and could provide the same out type for both tools?

Pretty much the entire dance here was about inverting control back to the author to be able to change how the tool is represented directly.

SamMorrowDrums:
> It seems like a lot of work to define methods for each ToolInput and ToolOutput. Maybe this could be some sort of fallback if inference or explicitly writing the schema doesn't work out. Let's explore other options and come back to this.

@jba I think supporting both would be totally reasonable, this was my attempt to express the idea and show it concretely and see how we felt. I hope I did a good enough job of imagining how it could be. Not strongly tied to it - just wanted to have something to throw a stick at.

My personal goal is that I would like to be able to keep a concrete type in the return value of a handler, so handlers can wrap other handlers and not lose data in between. Too many SDKs assume that tools will always return the value, instead of a type that can serialize the value, but it makes composing new server tools out of existing server tools more difficult.

jba:
I think we went a different way with this. I'd like to close this PR, if that's all right with you @SamMorrowDrums. If there are use cases that our current API doesn't work for but this approach does, let's discuss them.

SamMorrowDrums:
Thanks @jba will close. It was more of a hypothetical to see what it would look and served its purpose.

PR #58, opened 2025-06-26
mcp: Add example of using Middleware for logging purposes

rwjblue-glean:
## Motivation and Context
I needed to implement some logging for our experimental server that we are building out using the SDK, and figured I could propose an example that might make it easier for folks in the future.

For #33 

## How Has This Been Tested?

Implemented a small number of tests to ensure that the example continues to function over time.

## Breaking Changes

N/A

## Types of changes

- [x] Documentation update

## Checklist


examples/logging-middleware/main.go
```diff
@@ -0,0 +1,200 @@
...
+			method string,
+			params mcp.Params,
+		) (mcp.Result, error) {
+			start := time.Now()
+
+			// Note: Currently there's no way to access the HTTP Mcp-Session-Id header
+			// from middleware. In a real implementation, you'd want the actual session ID
+			// to correlate with HTTP access logs. For now, we use the session pointer
+			// as a unique identifier within this process.
+			sessionID := fmt.Sprintf("%p", session)
```
rwjblue-glean, line 40: I struggled to find a way to properly gain access to the `Mcp-Session-Id`, which seems kinda important. I would love to update the example to fix that if I'm just missing something super obvious. 

I kinda expected to be able to do something like `session.SessionId()` or somesuch, though that obviously doesn't work (also, not all sessions are even going to have a session id conceptually -- e.g. stdio and in memory transports have no logical session id).

examples/logging-middleware/main.go
```diff
@@ -0,0 +1,200 @@
...
+			method string,
+			params mcp.Params,
+		) (mcp.Result, error) {
+			start := time.Now()
+
+			// Note: The actual HTTP session ID (from Mcp-Session-Id header) is not
+			// accessible from middleware with the current SDK design. This would require
+			// SDK changes to pass the session ID through context. For now, we use the
+			// session pointer as a unique identifier within this process.
+			sessionID := fmt.Sprintf("%p", session)
```
rwjblue-glean: I struggled to find a way to properly gain access to the `Mcp-Session-Id`, which seems kinda important. I would love to update the example to fix that if I'm just missing something super obvious. 

I kinda expected to be able to do something like `session.SessionId()` or somesuch, though that obviously doesn't work (also, not all sessions are even going to have a session id conceptually -- e.g. stdio and in memory transports have no logical session id).

---

As it stands now, this "works" but really doesn't help folks that want to implement a Streamable HTTP server and have easy correlation between incoming request logs (e.g. that might have the `Mcp-Session-Id` header) and the middleware logging output.

examples/logging-middleware/main.go
```diff
@@ -0,0 +1,200 @@
...
+	"net/http"
+	"os"
+	"time"
+
+	"github.com/modelcontextprotocol/go-sdk/mcp"
+)
+
+var httpAddr = flag.String("http", "", "if set, use streamable HTTP at this address, instead of stdin/stdout")
+
+// LoggingMiddleware provides comprehensive MCP-level logging
```
jba: nit: end with a period

examples/logging-middleware/main.go
jba: We clearly need to fix this. I'll add something.

examples/logging-middleware/main.go
jba: Also, rephrase to avoid "marketing" adjectives like "comprehensive". Keep it neutral. And clarify that this is server-side logging, to distinguish it from the MCP logging feature, which goes back to the client.

examples/logging-middleware/main.go
```diff
@@ -0,0 +1,200 @@
...
+			result, err := next(ctx, session, method, params)
+
+			duration := time.Since(start)
+
+			if err != nil {
+				logger.Error("MCP method failed",
+					"method", method,
+					"session_id", sessionID,
+					"duration_ms", duration.Milliseconds(),
+					"error", err.Error(),
```
jba: Use "err". We didn't make this an exported constant in log/slog, but we want it to be the convention.

jba: Also, I think you can pass in `err` and its Error method will get called. Confirm that.

rwjblue-glean:
Thank you @jba for taking a look! I will work on updating this evening.

examples/logging-middleware/main.go
jba: See #67.

jba: #67 has landed. Please try again.

rwjblue-glean: Awesome, thank you!

rwjblue-glean:
Sorry for the delay here @jba (I was out on vacation, and this slipped between the cracks); I've updated as you requested please take a look when you have time.

mcp/example_middleware_test.go
```diff
@@ -0,0 +1,143 @@
...
+	"github.com/modelcontextprotocol/go-sdk/mcp"
+)
+
+// This example demonstrates server side logging using the mcp.Middleware system.
+func Example_loggingMiddleware() {
+	// Create a logger for demonstration purposes.
+	logger := slog.New(slog.NewTextHandler(os.Stdout, &slog.HandlerOptions{
+		Level: slog.LevelInfo,
+		ReplaceAttr: func(groups []string, a slog.Attr) slog.Attr {
+			// Simplify timestamp format for consistent output
```
jba: end sentence with period

mcp/example_middleware_test.go
```diff
@@ -0,0 +1,143 @@
...
+	"github.com/modelcontextprotocol/go-sdk/jsonschema"
+	"github.com/modelcontextprotocol/go-sdk/mcp"
+)
+
+// This example demonstrates server side logging using the mcp.Middleware system.
+func Example_loggingMiddleware() {
+	// Create a logger for demonstration purposes.
+	logger := slog.New(slog.NewTextHandler(os.Stdout, &slog.HandlerOptions{
+		Level: slog.LevelInfo,
+		ReplaceAttr: func(groups []string, a slog.Attr) slog.Attr {
```
jba: s/groups/_/

mcp/example_middleware_test.go
```diff
@@ -0,0 +1,143 @@
...
+
+	// Create logging middleware
+	loggingMiddleware := func(next mcp.MethodHandler[*mcp.ServerSession]) mcp.MethodHandler[*mcp.ServerSession] {
+		return func(
+			ctx context.Context,
+			session *mcp.ServerSession,
+			method string,
+			params mcp.Params,
+		) (mcp.Result, error) {
+			start := time.Now()
```
jba: Move until after the logging.

jba (approved):
Thanks a lot!

rwjblue-glean:
Thank you for all of your help @jba!

Issue #65, opened 2025-06-27
add a  Session.ID method


jba:
**Is your feature request related to a problem? Please describe.**
Sometimes it's useful to have the ID of a ClientSession or ServerSession. For example, per-session rate limiting.

**Describe the solution you'd like**
A way to get the session ID.

**Describe alternatives you've considered**
Using the session pointer. That requires guaranteeing that the pointer doesn't change, which we may not want to do.

jba:
Also, the spec talks about an Mcp-Session-Id header, and that is the value we should expose.

findleyr:
The Mcp-Session-Id header only applies to HTTP transports. I didn't expose the session ID from the session initially because it wasn't a documented part of the spec outside of the transport.

But I think it is completely reasonable to expose an intrinsic ID for the session, even if it is not directly prescribed by the spec.

PR #67, opened 2025-06-28
mcp: add Session.ID

jba:
Support retrieving the session ID from client and server sessions.

For #65.

### DIFF
--- a/mcp/transport.go
+++ b/mcp/transport.go
@@ -50,7 +50,7 @@ type Connection interface {
 	Read(context.Context) (JSONRPCMessage, error)
 	Write(context.Context, JSONRPCMessage) error
 	Close() error // may be called concurrently by both peers
-	sessionID() string
+	SessionID() string
 }
 
 // A StdioTransport is a [Transport] that communicates over stdin/stdout using
@@ -95,7 +95,7 @@ type binder[T handler] interface {
 
 type handler interface {
 	handle(ctx context.Context, req *JSONRPCRequest) (any, error)
-	setSessionIDFunc(func() string) // so Sessions can get the session ID
+	setConn(Connection)
 }
 
 func connect[H handler](ctx context.Context, t Transport, b binder[H]) (H, error) {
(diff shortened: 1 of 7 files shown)


jba:
@findleyr Please take a careful look. I'm not familiar with the machinery underlying sessions, and there may be a much simpler way. I'm also not sure if I've gotten all the locking right.

rwjblue-glean (commented):
Thank you!!

findleyr (commented):
Why don't we just set the session ID when creating each session? Then no locking is required. What am I missing?

samthanawalla:
> Why don't we just set the session ID when creating each session? Then no locking is required. What am I missing?

IIUC, the session ID is not known when the session is created but only when a successful POST happens.

I think we could change the sessionID to an atomic.Value to avoid locking.

jba:
> Why don't we just set the session ID when creating each session? Then no locking is required. What am I missing?

The ID comes from the server, so clients can't create it.

streamableClientConn locks because you already had locking around its sessionID. I frankly didn't look deeply into why.

There is no other locking in this PR.

@samthanawalla We're not concerned with locking overhead here, because it's dwarfed by the cost of the networking.

@findleyr I was more concerned with the twists and turns I needed to surface these IDs, but as I look over the diffs, there's really not much code.

So I'm satisfied with this PR if you are.

mcp/transport.go
```diff
@@ -50,6 +50,7 @@ type Connection interface {
 	Read(context.Context) (JSONRPCMessage, error)
 	Write(context.Context, JSONRPCMessage) error
 	Close() error // may be called concurrently by both peers
+	sessionID() string
```
findleyr: This must be exported, as users must be able to define custom transports.

mcp/transport.go
```diff
@@ -124,6 +126,7 @@ func connect[H handler](ctx context.Context, t Transport, b binder[H]) (H, error
 		},
 	})
 	assert(preempter.conn != nil, "unbound preempter")
+	h.setSessionIDFunc(conn.sessionID)
```
findleyr: Since this is an internal detail, I think we could just set the store the Connection on the session, rather than indirecting through this func value. WDYT?

findleyr (dismissed):
LGTM, with a couple actionable comments.

mcp/transport.go
jba: Done.

mcp/transport.go
jba: I did what I think you said. PTAL.

findleyr (approved):
Thanks, LGTM.

Issue #69, opened 2025-06-28
jsonschema: support additional keywords


jba:
The JSON Schema spec allows additional keywords, beyond those in the various specified vocabularies.

The `Schema` type should support that. The obvious way is to add a map[string]any field. We then have to ensure that the map keys are JSON-marshaled at top level, along with the other keywords. That is a bit tricky, but we have [un]marshalStructWithMap in mcp/util.go.

albertsundjaja:
hello! I'm interested in contributing to accelerate the release :)

I have taken a look at the source code and my plan for tackling this is to

1. add a `map[string]any` field in `Schema` type defined in `jsonschema/schema.go`
2. extend the `[UN]MarshalJSON` functions to use the `[un]marshalStructWithMap`
- the `Extra` fields will hold the extra fields that are not defined in `Schema` but during marshal, Extra will be flattened to top level key-value pairs
3. add / refactor necessary unit tests for the above (unknown fields are no longer dropped)

Let me know if I can pick this up!

jba:
Thanks! I assigned you.
You'll have to copy the [un]marshalStructWithMap functions into jsonschema.

jba:
@qiaodev if Schema had an Extra field, and you could do:
```
schema.Extra["propertyOrdering"] = []string{"a", "b", "c"}
```
would that work for genai?

qiaodev:
> would that work for genai?

Thanks for the effort. It works for genai.

PR #72, opened 2025-06-30
jsonchema: support additional keywords (#69)

albertsundjaja:
Support additional (extra) keywords beyond those specified in the `Schema` type. The current behavior is additional keywords are dropped. 

## Motivation and Context

https://github.com/modelcontextprotocol/go-sdk/issues/69

Mainly to adhere with the JSON Schema specs because additional keywords are allowed

## How Has This Been Tested?

Unit tests are added to test the behavior

## Breaking Changes

No

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation update

## Checklist

## Additional context

- There are similar util functions in `mcp/util.go` which don't seem to be used. Maybe we should move it to `internal/util` instead of `jsonchema/util.go` package?
- The method `MarshalJSON` is only on `pointer receiver`, is this intended? (that means it will only work if we [un]marshal pointer to the `Schema` instance)
- The fix is using the [un]marshalStructWithMap from `mcp/util.go` and there are requirements for the function to work properly but these are not strictly checked e.g. need `json:"-"` tag for the Extra field. Just double checking that this will not trip people up

Happy to make required changes! :)

### DIFF
--- a/jsonschema/schema.go
+++ b/jsonschema/schema.go
@@ -127,6 +127,9 @@ type Schema struct {
 	// https://json-schema.org/draft/2020-12/draft-bhutton-json-schema-validation-00#rfc.section.7
 	Format string `json:"format,omitempty"`
 
+	// Extra allows for additional keywords beyond those specified
+	Extra map[string]any `json:"-"`
+
 	// computed fields
 
 	// This schema's base schema.
@@ -236,7 +239,7 @@ func (s *Schema) MarshalJSON() ([]byte, error) {
 		Type:                 typ,
 		schemaWithoutMethods: (*schemaWithoutMethods)(s),
 	}
-	return json.Marshal(ms)
+	return marshalStructWithMap(&ms, "Extra")
 }
 
 func (s *Schema) UnmarshalJSON(data []byte) error {
@@ -269,7 +272,7 @@ func (s *Schema) UnmarshalJSON(data []byte) error {
 	}{
 		schemaWithoutMethods: (*schemaWithoutMethods)(s),
 	}
-	if err := json.Unmarshal(data, &ms); err != nil {
+	if err := unmarshalStructWithMap(data, &ms, "Extra"); err != nil {
 		return err
 	}
 	// Unmarshal "type" as either Type or Types.
--- a/jsonschema/schema_test.go
+++ b/jsonschema/schema_test.go
@@ -26,6 +26,7 @@ func TestGoRoundTrip(t *testing.T) {
 		{Const: Ptr(any(map[string]any{}))},
 		{Default: mustMarshal(1)},
 		{Default: mustMarshal(nil)},
+		{Extra: map[string]any{"test": "value"}},
 	} {
 		data, err := json.Marshal(s)
 		if err != nil {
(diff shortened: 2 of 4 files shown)


jsonschema/schema.go
```diff
@@ -127,6 +127,9 @@ type Schema struct {
 	// https://json-schema.org/draft/2020-12/draft-bhutton-json-schema-validation-00#rfc.section.7
 	Format string `json:"format,omitempty"`
 
+	// Extra allows for additional keywords beyond those specified
```
jba: Nit: end with period.

jsonschema/util.go
```diff
@@ -282,3 +285,129 @@ func assert(cond bool, msg string) {
...
+	}
+	return nil
+}
+
+var jsonNamesMap sync.Map // from reflect.Type to map[string]bool
+
+// jsonNames returns the set of JSON object keys that t will marshal into,
+// including fields from embedded structs in t.
+// t must be a struct type.
+// note that there is a similar function in mcp/util.go, but they are not the same
```
jba: What's different?

jsonschema/util.go
```diff
@@ -282,3 +285,129 @@ func assert(cond bool, msg string) {
...
+	}
+	// "{X}" + "{Y}" => "{X,Y}"
+	res := append(structBytes[:len(structBytes)-1], ',')
+	res = append(res, mapBytes[1:]...)
+	return res, nil
+}
+
+// unmarshalStructWithMap is the inverse of marshalStructWithMap.
+// T has the same restrictions as in that function.
+// note that there is a similar function in mcp/util.go, but they are not the same
```
jba: What's different?

jsonschema/util.go
```diff
@@ -282,3 +285,129 @@ func assert(cond bool, msg string) {
...
+//
+// the call marshalJSONWithMap(s, "Extra") would return
+//
+//	{"A": 1, "B": 2}
+//
+// It is an error if the map contains the same key as another struct field's
+// JSON name.
+//
+// marshalStructWithMap calls json.Marshal on a value of type T, so T must not
+// have a MarshalJSON method that calls this function, on pain of infinite regress.
```
jba, line 313: I would have thought that we'd have a problem, because Schema has a MarshalJSON method. What's going on here? Why doesn't the json.Marshal call on L338 cause a stack overflow?

jba (changes requested):

jsonschema/util.go
albertsundjaja: oh that's because of this clever implementation in the `MarshalJSON` method

```
func (s *Schema) MarshalJSON() ([]byte, error) {
        ...
	ms := struct {
		Type any `json:"type,omitempty"`
		*schemaWithoutMethods
	}{
		Type:                 typ,
		schemaWithoutMethods: (*schemaWithoutMethods)(s),
	}
	return marshalStructWithMap(&ms, "Extra")
```

so `ms` doesn't actually have `MarshalJSON`

jba: Ah, right! I was going to compliment you on the cleverness but then I realized it was me! (It's actually a standard trick.)

jsonschema/schema_test.go
```diff
@@ -64,11 +65,13 @@ func TestJSONRoundTrip(t *testing.T) {
 			`{"$vocabulary":{"b":true, "a":false}}`,
 			`{"$vocabulary":{"a":false,"b":true}}`,
 		},
-		{`{"unk":0}`, `{}`}, // unknown fields are dropped, unfortunately
+		{`{"unk":0}`, `{"unk":0}`},     // unknown fields are not dropped
```
jba: Add a test combining standard fields like "type" with extra fields like "unk".

jba:
We're close! One more small test case, please.

albertsundjaja:
fantastic! didn't realize the plan is to move this package to a separate repo hence the need to copy the util functions

PR #85, opened 2025-07-02
jsonschema: make schema state independent of resolve arguments

### LINKED ISSUE #84: jsonschema: remove Schema resolution fields
A Schema holds information related to its resolution. This makes the schema dependent on the call to Resolve, but it should be effectively stateless.

jba:
Move schema state that depends on resolution into the Resolved struct.
The remaining unexported schema fields are dependent only on the schema itself and its sub-schemas.

Now a schema can be resolved multiple times, as itself or as part of other schemas.

Fixes #84.

### DIFF
--- a/jsonschema/resolve_test.go
+++ b/jsonschema/resolve_test.go
@@ -91,13 +91,13 @@ func TestPaths(t *testing.T) {
 		{root.PrefixItems[1].Items, "/prefixItems/1/items"},
 	}
 	rs := newResolved(root)
-	if err := root.checkStructure(rs.resolvedInfo); err != nil {
+	if err := root.checkStructure(rs.resolvedInfos); err != nil {
 		t.Fatal(err)
 	}
 
 	var got []item
 	for s := range root.all() {
-		got = append(got, item{s, rs.resolvedInfo[s].path})
+		got = append(got, item{s, rs.resolvedInfos[s].path})
 	}
 	if !slices.Equal(got, want) {
 		t.Errorf("\ngot  %v\nwant %v", got, want)
@@ -133,7 +133,7 @@ func TestResolveURIs(t *testing.T) {
 			}
 
 			rs := newResolved(root)
-			if err := root.check(rs.resolvedInfo); err != nil {
+			if err := root.check(rs.resolvedInfos); err != nil {
 				t.Fatal(err)
 			}
 			if err := resolveURIs(rs, base); err != nil {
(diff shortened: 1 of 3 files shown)


jba:
@findleyr: @adonovan should review. Having trouble getting his username to show up.

jsonschema/resolve_test.go
```diff
@@ -193,13 +195,14 @@ func TestRefCycle(t *testing.T) {
 	}
 
 	rs, err := schemas["root"].Resolve(&ResolveOptions{Loader: loader})
+	t.Logf("%#v", rs.resolvedInfo)
```
adonovan: delete

jsonschema/schema.go
```diff
@@ -130,37 +129,13 @@ type Schema struct {
...
-	// fragmentless URI.
-	base *Schema
-
-	// The URI for the schema, if it is the root or has an ID.
-	// Otherwise nil.
-	// Invariants:
-	//   s.base.uri != nil.
-	//   s.base == s <=> s.uri != nil
-	uri *url.URL
+	// These fields are independent of arguments to Schema.Resolved,
```
adonovan: I don't understand how there can be any computed fields in a Schema if it is to be stateless, and a valid operand of many independent calls to Resolve.

jsonschema/schema.go
```diff
@@ -186,27 +161,22 @@ type anchorInfo struct {
...
 }
 
-// ResolvedRef returns the Schema to which this schema's $ref keyword
-// refers, or nil if it doesn't have a $ref.
-// It returns nil if this schema has not been resolved, meaning that
-// [Schema.Resolve] was called on it or one of its ancestors.
-func (s *Schema) ResolvedRef() *Schema {
-	return s.resolvedRef
-}
+// // ResolvedRef returns the Schema to which this schema's $ref keyword
```
adonovan: delete

jsonschema/resolve_test.go
jba: done.

jsonschema/schema.go
jba: done

jsonschema/schema.go
jba: The unexported fields are the outputs of functions on the other fields. They are cached computations.
For example, `pattern` is always the result of compiling `Pattern`.
I am assuming a Schema is not changed after Resolve, as documented.
(How to thread-protect the setting of these cached fields will be the topic of another PR.)
However, thinking about this, `path` is not in fact a function of the schema, but its ancestors as well.

jba: I moved everything into resolvedInfo.

I am trying to take advantage of the fact that a GitHub PR consists of multiple commits, by structuring the work like a stack of CLs. Of the two new commits, the first makes this change, and the second is just a rename. You can view them separately via the "Changes from..." dropdown on the "Files changed" tab.

Issue #10, opened 2025-06-24
Streamable HTTP support


findleyr:
This issue tracks work on the Streamable HTTP transport:

- [x] Add a `StreamableServerTransport` type.
- [x] Add a `StreamableClientTransport` type.
- [x] Support 'stateless' mode for server transport (not documented, but implicit: see https://github.com/modelcontextprotocol/modelcontextprotocol/issues/282#issuecomment-2998128187).
- [x] Add configurable event store for the streamable server transport (see TypeScript's 'EventStore').
- [x] Support retryability/replayability for the streamable client transport.
- [x] Support a single JSON response (application/json)

samthanawalla:
I can take a look at "Support retryability/replayability for the streamable client transport."

findleyr:
@samthanawalla sure, that would be great. I started thinking about persistence in https://go.dev/cl/682555, but haven't done anything toward retryability. I think there is some good prior art in other SDKs. For example:
https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/client/streamableHttp.ts#L51

I will work on server persistence APIs.

samthanawalla:
We also need to support application/json #129

jpmcb:
> Support 'stateless' mode for server transport

I'd be happy to take a look at this since it's important for alot of servers I've been building.

Adding an option to `ServerOptions` to not expect session management seems like an ideal solution. This is how alot of other SDKs implement this (like in the Python SDK):

```python
from mcp.server.fastmcp import FastMCP

# Stateless server (no session persistence)
mcp = FastMCP("StatelessServer", stateless_http=True)
```

findleyr:
@jpmcb thank you, and sorry for the missing support. I've been away for a few weeks, but @samthanawalla, @jba and I have a meeting tomorrow to talk about outstanding items, of which streamable transport is the most important.  I think it's likely that we'll prioritize this above essentially everything else, and with Sam working on configurable session storage, it makes sense for us to do stateless mode as well.

whiletrue111:
So does it support streamable HTTP?

jba:
> So does it support streamable HTTP?

Yes, at main.

findleyr:
Stateless support is next after #257.

findleyr:
This will be also be completed by #307.

PR #86, opened 2025-07-02
mcp: add retry and replay to the Streamable HTTP implementation

samthanawalla:
Adds exponential backoff and jitter to the client-side POST. Implements replay using Last-Event-ID for resumability.

Since these concepts are intertwined- I have added this in a single CL.

For #10

### DIFF
--- a/mcp/streamable.go
+++ b/mcp/streamable.go
@@ -7,13 +7,17 @@ package mcp
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"io"
+	"math/rand"
+	"net"
 	"net/http"
 	"strconv"
 	"strings"
 	"sync"
 	"sync/atomic"
+	"time"
 
 	"github.com/modelcontextprotocol/go-sdk/internal/jsonrpc2"
 )
@@ -90,7 +94,7 @@ func (h *StreamableHTTPHandler) ServeHTTP(w http.ResponseWriter, req *http.Reque
 	var session *StreamableServerTransport
 	if id := req.Header.Get("Mcp-Session-Id"); id != "" {
 		h.sessionsMu.Lock()
-		session, _ = h.sessions[id]
+		session = h.sessions[id]
 		h.sessionsMu.Unlock()
 		if session == nil {
 			http.Error(w, "session not found", http.StatusNotFound)
@@ -594,6 +598,15 @@ type StreamableClientTransportOptions struct {
 	// HTTPClient is the client to use for making HTTP requests. If nil,
 	// http.DefaultClient is used.
 	HTTPClient *http.Client
+	// MaxRetries specifies the maximum number of retries for sending a message
+	// or re-establishing a hanging GET connection. If 0, no retries are performed
+	// beyond the initial attempt.
+	MaxRetries int
+
+	// InitialBackoff is the initial duration to wait before the first retry
+	// attempt. Subsequent retries use exponential backoff. If 0, a default
+	// of 1 second is used.
+	InitialBackoff time.Duration
 }
 
 // NewStreamableClientTransport returns a new client transport that connects to
@@ -602,6 +615,13 @@ func NewStreamableClientTransport(url string, opts *StreamableClientTransportOpt
 	t := &StreamableClientTransport{url: url}
 	if opts != nil {
 		t.opts = *opts
+	} else {
+		t.opts = StreamableClientTransportOptions{}
+	}
+
+	// Set default initial backoff if not specified.
+	if t.opts.InitialBackoff == 0 {
+		t.opts.InitialBackoff = time.Second
 	}
 	return t
 }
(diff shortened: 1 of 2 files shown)


mcp/streamable.go
```diff
@@ -7,13 +7,17 @@ package mcp
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"io"
+	"math/rand"
```
findleyr, line 13: Use math/rand/v2
(did an LLM help write this? We need to teach them about math/rand/v2!)

mcp/streamable.go
```diff
@@ -602,6 +615,13 @@ func NewStreamableClientTransport(url string, opts *StreamableClientTransportOpt
 	t := &StreamableClientTransport{url: url}
 	if opts != nil {
 		t.opts = *opts
+	} else {
```
findleyr, line 618: No need for this: opts are already the zero value.

mcp/streamable.go
```diff
@@ -654,120 +701,370 @@ func (s *streamableClientConn) Read(ctx context.Context) (JSONRPCMessage, error)
 	case <-ctx.Done():
 		return nil, ctx.Err()
 	case <-s.done:
+		s.mu.Lock()
+		defer s.mu.Unlock()
+		if s.err != nil {
+			return nil, s.err // Return explicit error if connection closed due to error
```
findleyr, line 707: This looks like a logically distinct change. Given the size and complexity of this PR, it would be good to separate out those changes that are unrelated to replay support into one or more separate CLs, with tests.

mcp/streamable.go
```diff
@@ -654,120 +701,370 @@ func (s *streamableClientConn) Read(ctx context.Context) (JSONRPCMessage, error)
...
+						// Continue
+					}
+
+					gotSessionID, sendErr := s.postMessage(ctx, currentSessionID, msgToSend)
+					if sendErr == nil {
+						// If sessionID was not set and we got one, update it.
+						if currentSessionID == "" && gotSessionID != "" {
+							s.sessionID.Store(gotSessionID)
+						}
+						// Undefined behavior when currentSessionID != gotSessionID
```
findleyr, line 769: What is the rationale for undefined behavior here? Why isn't this an error?

mcp/streamable.go
```diff
@@ -654,120 +701,370 @@ func (s *streamableClientConn) Read(ctx context.Context) (JSONRPCMessage, error)
...
+					select {
+					case <-s.done:
+						return
+					case <-ctx.Done(): // Check if the individual send context was cancelled
+						return
+					default:
+						// Continue
+					}
+
+					gotSessionID, sendErr := s.postMessage(ctx, currentSessionID, msgToSend)
```
findleyr, line 763: I think there's a fundamental problem here (fixable, but fundamental): the spec says we can _resume_ an sse stream by issuing a subsequent GET requests, but does not say that we should retry post requests.

Imagine that the server-side is a stateful server (rare but possible): we don't want to perform duplicate actions. Instead, we should only resume hanging requests once we've received the response header as well as a nonempty event ID. Then we can issue a GET with that last-event-id to resume the streaming of responses.

mcp/streamable.go
```diff
@@ -654,120 +701,370 @@ func (s *streamableClientConn) Read(ctx context.Context) (JSONRPCMessage, error)
...
-		s._sessionID = gotSessionID
+// startMessageWriter continuously sends messages from the pendingMessages channel,
+// applying retry logic for transient errors.
+func (s *streamableClientConn) startMessageWriter() {
+	for {
+		select {
+		case <-s.done:
+			return // Connection is closed
+		case msg := <-s.pendingMessages:
+			// Use a new context for each send attempt to allow individual retries to be cancelled
```
jba, line 742: I don't understand this sentence. If the overall connection is cancelled, don't we want all the retries to be cancelled too? Why retain the ability to cancel each one separately?

mcp/streamable.go
```diff
@@ -654,120 +701,370 @@ func (s *streamableClientConn) Read(ctx context.Context) (JSONRPCMessage, error)
...
+					}
+
+					lastErr = sendErr // Store the latest error
+					if !isRetryable(sendErr) || i == s.maxRetries {
+						break // Not a retryable error or max retries reached
+					}
+
+					// Apply exponential backoff with jitter
+					backoffDuration := s.initialBackoff * time.Duration(1<<uint(i))
+					jitter := time.Duration(s.randSource.Int63n(int64(backoffDuration / 2))) // Jitter up to half of backoff
```
jba, line 780: I've also seen jitter implemented as picking the delay randomly between zero and backoffDuration. I wonder what the tradeoffs are of each way. It probably doesn't matter, but the other one is slightly easier to implement.

mcp/streamable.go
```diff
@@ -654,120 +701,370 @@ func (s *streamableClientConn) Read(ctx context.Context) (JSONRPCMessage, error)
...
-		for evt, err := range scanEvents(resp.Body) {
-			if err != nil {
-				// TODO: surface this error; possibly break the stream
-				return
+	for {
+		select {
+		case <-s.done:
+			return // Connection is closed.
+		default:
+			// Continue
```
jba, line 864: omit comment

mcp/streamable.go
```diff
@@ -654,120 +701,370 @@ func (s *streamableClientConn) Read(ctx context.Context) (JSONRPCMessage, error)
...
+		}
+
+		sessionID := s.sessionID.Load().(string)
+		if sessionID == "" {
+			// Session ID not yet established (first POST hasn't completed).
+			// Wait and retry.
+			time.Sleep(100 * time.Millisecond) // Avoid busy-waiting
```
jba, line 871: I think you could avoid this if you kept sessionID in a chan string with capacity 1. This code would receive from the channel, blocking until it was ready. Other code that didn't want to block could do a select. Everyone who reads it would immediately put it back in the channel.

jba: Not sure if it's worth it.

mcp/streamable.go
```diff
@@ -654,120 +701,370 @@ func (s *streamableClientConn) Read(ctx context.Context) (JSONRPCMessage, error)
...
+
+		// Apply exponential backoff with jitter
+		delay := backoffDuration + time.Duration(s.randSource.Int63n(int64(backoffDuration/2)))
+		select {
+		case <-s.done:
+			return // Connection closed during backoff
+		case <-time.After(delay):
+			retries++
+			backoffDuration *= 2                  // Exponential increase
+			if backoffDuration > 30*time.Second { // Cap backoff duration
```
jba, line 916: Should't you cap it on the other side too?

jba, line 916: Why are this backoff delay computations so different between the client and server sides? Can they be pulled out into something more general?

mcp/streamable.go
```diff
@@ -654,120 +701,370 @@ func (s *streamableClientConn) Read(ctx context.Context) (JSONRPCMessage, error)
...
+			s.mu.Lock()
+			s.lastEventID = evt.id
+			s.mu.Unlock()
+		}
+		select {
+		case s.incoming <- evt.data:
+			// Message successfully sent to incoming channel
+		case <-s.done:
+			// Connection closed while trying to send incoming message
+			return io.EOF
```
jba, line 977: Should this be io.ErrUnexpectedEOF?

mcp/streamable.go
```diff
@@ -654,120 +701,370 @@ func (s *streamableClientConn) Read(ctx context.Context) (JSONRPCMessage, error)
...
+func isRetryable(err error) bool {
+	if err == nil {
+		return false
+	}
+
+	// Check if the error is an httpStatusError and if its status code is retryable.
+	var httpErr *httpStatusError
+	if errors.As(err, &httpErr) {
+		switch httpErr.StatusCode {
+		case http.StatusRequestTimeout, // 408
```
jba, line 994: OOC, where did this list come from?

mcp/streamable_test.go
```diff
@@ -97,6 +98,292 @@ func TestStreamableTransports(t *testing.T) {
...
+	// This will happen concurrently with the client's GET retries.
+	go func() {
+		_, callErr := session.CallTool(ctx, &CallToolParams{Name: "noop"})
+		if callErr != nil && !strings.Contains(callErr.Error(), "context canceled") {
+			t.Errorf("CallTool returned unexpected error: %v", callErr)
+		}
+	}()
+
+	// Wait for all messages to be received, or timeout
+	allMessages := []string{}
```
jba, line 289: var allMessages []string

samthanawalla:
My approach was wrong, oops.

As Rob pointed out, we shouldn't retry POST, I will try again and send a separate PR.

PR #87, opened 2025-07-02
examples/rate-limiting: add example for session based rate limiting

### LINKED ISSUE #22: Rate limiting support / Best practices

samthanawalla:
Added an example for session based rate limiting after session.ID() was exposed.

Fixes #22

### PR Tips

the [rules for Go commit messages](https://go.dev/wiki/CommitMessage), with the following
changes and additions:

- Markdown is allowed.


examples/rate-limiting/main.go
```diff
@@ -43,12 +45,36 @@ func PerMethodRateLimiterMiddleware[S mcp.Session](limiters map[string]*rate.Lim
 	}
 }
 
+// PerSessionRateLimiterMiddleware creates a middleware that applies rate limiting
+// on a per-session basis for receiving requests.
+func PerSessionRateLimiterMiddleware[S mcp.Session](limit rate.Limit, burst int) mcp.Middleware[S] {
+	// A map to store limiters, keyed by the session ID.
+	// Use sync.Map for concurrent access as sessions can connect/disconnect concurrently.
```
jba: sync.Map isn't wrong, but it has some special use cases (see third paragraph of https://pkg.go.dev/sync#Map). Let's use a regular map with a mutex.

examples/rate-limiting/main.go
```diff
@@ -43,12 +45,36 @@ func PerMethodRateLimiterMiddleware[S mcp.Session](limiters map[string]*rate.Lim
...
+	var sessionLimiters sync.Map // map[string]*rate.Limiter
+
+	return func(next mcp.MethodHandler[S]) mcp.MethodHandler[S] {
+		return func(ctx context.Context, session S, method string, params mcp.Params) (mcp.Result, error) {
+			// Load or CREATE a new limiter for this session if it doesn't exist
+			actualLimiter, _ := sessionLimiters.LoadOrStore(session.ID(), rate.NewLimiter(limit, burst))
```
jba: One thing I'm not clear about is whether the session ID is set at this point. For example, the client doesn't see it until after the initialize message, I think. Can you look over the code and document the reason it is set here, or if you can't prove that it _is_ set, then check for emptiness.

examples/rate-limiting/main.go
samthanawalla: Done

examples/rate-limiting/main.go
samthanawalla: Done

Issue #56, opened 2025-06-26
InitializeResult returns Prompts and Resources in Capabilities, even when no Prompts or Resources are added to the MCP Server


netroy:
**Is your feature request related to a problem? Please describe.**
When one creates a new `mcp.Server`, there are no Prompts, Resources, or Tools defined on it. At this point if a client connects to this server, the `capabilities` property on the `initialize` response should not return any capabilities where the featureSet is empty, because this makes the client think that there might be these features available.  

**Describe the solution you'd like**
`InitializeResult.Capabilities` should only include capabilities that have a non-empty featureset.

**Additional context**
I can send a PR to address this. I just wanted to check if this makes sense to you folks.

jba:
What if the server adds tools or other features after the client connects?
It looks like the typescript API requires you to signal that you might have tools by calling server.setToolRequestHandlers. We could do something like that.

What are the consequences of saying you support a feature when you don't?

jba:
If this matters, I would say: let's add Server.WillAddTools(). Then at the time Server.Run is called, you must have either added a tool, or called WillAddTools to set the capability.

jba:
We could also add `HasTools`, etc. to ServerOptions. Less clutter that way.

martinemde:
Here's a few extra considerations related to this:

- A server should be able to declare that it doesn't support prompts, resources, and/or tools.
- A server should be able to inform the client that it will not send `list changed` notifications even if it does support one of the capabilities.

My expectation would be that `ServerOptions` support some form of `hasPrompts: false` for each type.

Here's the log from Claude desktop talking to a go-sdk MCP server I implemented that has 1 tool. Note that the client seems to re-request prompts, resources, and tools regularly (`id`s are increasing, so this is within the same session)

```
2025-07-03T00:36:59.563Z [say] [info] Message from server: {"jsonrpc":"2.0","id":0,"result":{"capabilities":{"completions":{},"logging":{},"prompts":{"listChanged":true},"resources":{"listChanged":true},"tools":{"listChanged":true}},"protocolVersion":"2024-11-05","serverInfo":{"name":"studio-mcp","version":"dev"}}} { metadata: undefined }
2025-07-03T00:36:59.564Z [say] [info] Message from client: {"method":"notifications/initialized","jsonrpc":"2.0"} { metadata: undefined }
[...]

PR #93, opened 2025-07-03
mcp: advertise a capability if and only if added

jba:
If tools are added to a Server when Run is called, it will send toolCapabilities to the client. Otherwise it won't. Ditto for prompts and resources.

For #56.


jba:
Added test, rebased onto main.

samthanawalla (approved):

PR #94, opened 2025-07-04
mcp: support "mcp" struct tags 

### LINKED ISSUE #47: Support setting field descriptions using struct tags
In https://github.com/modelcontextprotocol/go-sdk/discussions/37, we're discussing the awkwardness of having to set property descriptions using tool options.

Ideally, the inferred schema should be sufficient for simple use cases, and options would only be needed for more sophisticated APIs.

I think we should implement the `mcp` struct tag, to allow setting the property description inline with the type definition. Then we could at least see if this is sufficient for most use cases.

```go
type SayHiParams struct {
  Name string `json:"name" mcp:"the name to greet"
}
```

CC @qiaodev @jba

jba:
Add the SchemaFor[T] function, which infers the schema for a struct and also uses the values of "mcp" struct tags to set descriptions.  

Fixes #47.


jba:
wrong base; closing.

PR #95, opened 2025-07-04
mcp: don't omit required fields in ImageContent and AudioContent

martinemde:
This is a follow-up related to #91 (though not requiring it).

Updates `ImageContent` and `AudioContent` to use an inline wire format to ensure that required fields are not omitted during JSON marshaling, matching TypeScript schema requirements. This follows the TextContent approach.

- `ImageContent` requires `data` and `mimeType` fields.
- `AudioContent` requires `data` and `mimeType` fields.

### DIFF
--- a/mcp/content.go
+++ b/mcp/content.go
@@ -58,13 +58,25 @@ type ImageContent struct {
 }
 
 func (c *ImageContent) MarshalJSON() ([]byte, error) {
-	return json.Marshal(&wireContent{
+	// Custom wire format to ensure required fields are always included, even when empty.
+	data := c.Data
+	if data == nil {
+		data = []byte{}
+	}
+	wire := struct {
+		Type        string       `json:"type"`
+		MIMEType    string       `json:"mimeType"`
+		Data        []byte       `json:"data"`
+		Meta        Meta         `json:"_meta,omitempty"`
+		Annotations *Annotations `json:"annotations,omitempty"`
+	}{
 		Type:        "image",
 		MIMEType:    c.MIMEType,
-		Data:        c.Data,
+		Data:        data,
 		Meta:        c.Meta,
 		Annotations: c.Annotations,
-	})
+	}
+	return json.Marshal(wire)
 }
 
 func (c *ImageContent) fromWire(wire *wireContent) {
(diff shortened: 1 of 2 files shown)


findleyr:
@martinemde I think your suggestion to remove the 'omitempty' attrs is preferable: the missing 'mimeType' or 'data' field is really an error in business logic, not marshalling.

For data, we should be careful to replace nil with an empty byte slice, as JSON null is not valid.

Thanks for picking this up!

mcp/content.go
```diff
@@ -48,6 +48,12 @@ type ImageContent struct {
 }
 
 func (c *ImageContent) MarshalJSON() ([]byte, error) {
+	if len(c.Data) == 0 {
+		return nil, errors.New("ImageContent missing data")
+	}
+	if c.MIMEType == "" {
+		return nil, errors.New("ImageContent missing mimeType")
+	}
 	return json.Marshal(&wireContent{
```
findleyr: As noted, I think we should use an inline type rather than 'wireContent' here, so that we can make Data and MimeType not "omitempty". We should be sure to replace nil data with the empty slice though.

jba: This is a requested change.

martinemde: Thanks, will get to it asap.

martinemde:
This is updated. Thanks for your patience while I got around to this.

mcp/content.go
```diff
@@ -58,13 +58,25 @@ type ImageContent struct {
 }
 
 func (c *ImageContent) MarshalJSON() ([]byte, error) {
-	return json.Marshal(&wireContent{
+	// Custom wire format to ensure required fields are always included, even when empty.
```
jba, line 61: It looks like these two functions are almost identical. Factor out into a single function. It looks to me that that function's args are the type (image or audio) and the data.

jba (changes requested):

mcp/content.go
martinemde: We can if you think it's important.

Two copies of something doesn't feel like a strong deduplication signal to me given that it will split the currently very clear schema mapping into a function that abstracts away half of the type.

If you think this is a blocker, I'll refactor, since I'd prefer the sdk is correct regardless of the code inside.

jba (approved):

PR #96, opened 2025-07-04
mcp: ensure meta is not skipped on blob ResourceContents

martinemde:
I noticed this while working on #95. Seems Meta was accidentally skipped in the custom wire format struct.


findleyr (approved):
Thanks!

Issue #77, opened 2025-07-01
jsonschema: memory leak in schema generator


jbpratt:
**Describe the bug**
It seems there is a leak when trying to generate schemas from certain structs!

**To Reproduce**
```go
package main

import (
        "fmt"
        "runtime"
        "time"

        "github.com/modelcontextprotocol/go-sdk/jsonschema"
        v1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1"
)

func main() {
        ticker := time.NewTicker(5 * time.Second)
        go func() {
                for range ticker.C {
                        var m runtime.MemStats
                        runtime.ReadMemStats(&m)
                        fmt.Printf("Alloc: %d KB, Sys: %d KB, NumGC: %d\n",
                                m.Alloc/1024, m.Sys/1024, m.NumGC)
                }
        }()
        _, err := jsonschema.For[v1.Pipeline]()
        if err != nil {
                panic(err)
        }
}
```

**Expected behavior**
Generating the schema correctly!

**Logs**
```
❯ go run main.go
Alloc: 3126601 KB, Sys: 3275482 KB, NumGC: 13
Alloc: 6046591 KB, Sys: 6332210 KB, NumGC: 14
Alloc: 8992110 KB, Sys: 9524295 KB, NumGC: 15
Alloc: 12696764 KB, Sys: 13284884 KB, NumGC: 15
Alloc: 14358771 KB, Sys: 15591427 KB, NumGC: 16
signal: killed
```

**Additional context**
OS: Linux
Go version: 1.24.3
Deps:
```
github.com/modelcontextprotocol/go-sdk v0.0.0-20250630184440-2facfc6ffe0b
github.com/tektoncd/pipeline v1.1.0
```

![Image](https://github.com/user-attachments/assets/550af1d4-78c7-486d-95e0-77d2c834eeaf)

albertsundjaja:
hello!

I think the issue is because of a cycle in your type. The `jsonschema.For` have this comment (https://github.com/modelcontextprotocol/go-sdk/blob/main/jsonschema/infer.go#L39)

`// The types must not have cycles.`

It looks to me that `v1.Pipeline` have a cycle

```
type Pipeline struct {
        ...
	Spec PipelineSpec `json:"spec"`
}

type PipelineSpec struct {
	...
	Finally []PipelineTask `json:"finally,omitempty"`
}

type PipelineTask struct {
        ...
[...]

PR #97, opened 2025-07-05
jsonchema: add memoization and cycle detection (#77)

albertsundjaja:
This PR adds:
1. Cycle detection: return error instead of waiting for stack overflow*
2. Memoization: performance improvement

*This is a fix for https://github.com/modelcontextprotocol/go-sdk/issues/77
The root cause of the issue was the type being processed is deeply nested and has cycle. Hence, the heap memory growth is faster than the stack growth, it went OOM before stack overflow panic can kick in

### DIFF
--- a/jsonschema/infer.go
+++ b/jsonschema/infer.go
@@ -9,6 +9,7 @@ package jsonschema
 import (
 	"fmt"
 	"reflect"
+	"sync"
 
 	"github.com/modelcontextprotocol/go-sdk/internal/util"
 )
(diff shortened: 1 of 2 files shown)


jba:
Thanks. I agree, let's drop the caching and keep the cycle detection.

jsonschema/infer.go
```diff
@@ -39,15 +40,18 @@ import (
 // The types must not have cycles.
```
jba, line 39: Let's also change this to "It also returns an error if there is a cycle in the types."

jsonschema/infer_test.go
```diff
@@ -91,3 +91,96 @@ func TestForType(t *testing.T) {
 		})
 	}
 }
+
+func TestForWithMutation(t *testing.T) {
```
albertsundjaja, line 95: context: I'm keeping this unit test, I think they are still good to have

jba (approved):

jsonschema/infer_test.go
jba: Sure.

jba:
Thanks for doing this!

PR #99, opened 2025-07-07
mcp: add SSE session IDs

### LINKED ISSUE #98: SessionID is blank / empty when read from mcp.ServerSession
**Describe the bug**

First off: Thanks for your hard work. I'm aware this is early days and not production ready yet. I may just be holding things wrong, but I expected `mcp.ServerSession.ID()` to return some kind of unique ID associated with a session, and instead it returns an empty string.

I'm after a session ID so that I can store some data on behalf of the user and reference it in later calls (write some files to a user-specific workspace and read them later for example)

**To Reproduce**
1. Run any of the SSE server demos.
2. Log the value of mcp.ServerSession.ID() from a tool handler function
3. Connect to the server using some agent, claude for example `claude mcp add -t sse bbmcp http://localhost:6270/sse`, and trigger the tool

**Expected behavior**
Handler should print some unique session ID

**Logs**
N/A

**Additional context**
Thanks

jba:
ClientSessions and ServerSessions created by SSE have non-empty session IDs.

Fixes #98.

### DIFF
--- a/mcp/sse.go
+++ b/mcp/sse.go
@@ -110,8 +110,9 @@ func NewSSEHandler(getServer func(request *http.Request) *Server) *SSEHandler {
 //     [SSEServerTransport.ServeHTTP].
 //   - Close terminates the hanging GET.
 type SSEServerTransport struct {
-	endpoint string
-	incoming chan JSONRPCMessage // queue of incoming messages; never closed
+	endpoint  string
+	sessionID string
+	incoming  chan JSONRPCMessage // queue of incoming messages; never closed
 
 	// We must guard both pushes to the incoming queue and writes to the response
 	// writer, because incoming POST requests are arbitrarily concurrent and we
@@ -228,6 +229,7 @@ func (h *SSEHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 	}
 
 	transport := NewSSEServerTransport(endpoint.RequestURI(), w)
+	transport.sessionID = sessionID
 
 	// The session is terminated when the request exits.
 	h.mu.Lock()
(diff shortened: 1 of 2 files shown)


mcp/sse.go
```diff
@@ -393,6 +394,7 @@ func (c *SSEClientTransport) Connect(ctx context.Context) (Connection, error) {
 	s := &sseClientConn{
 		sseEndpoint: c.sseEndpoint,
 		msgEndpoint: msgEndpoint,
+		sessionID:   msgEndpoint.Query().Get("sessionid"),
```
findleyr, line 397: This is, unfortunately, not part of the spec.
Why do we want to have session IDs for the SSE connection? In theory, there is no way for the client to actually know its session id.

jba: It was requested in #98, and I obliged.
We could also say that SSE is unsupported, and drop this.

tylerwilliams: 👋 I filed this bug.

The part of the spec that led me down this session ID path was this: https://modelcontextprotocol.io/docs/concepts/transports#session-management.

Maybe there's another way to approach this (setting or grabbing the header manually) that I should look into?

tylerwilliams: I see SSE is deprecated https://modelcontextprotocol.io/docs/concepts/transports#server-sent-events-sse-deprecated so I think this is my bad and I should just be using `mcp.NewStreamableHTTPHandler` instead of `mcp.NewSSEHandler`. Sorry for the noise :(

jba: No problem! Closing this PR.

PR #101, opened 2025-07-07
jsonschema: support "jsonschema" struct tags

### LINKED ISSUE #47: Support setting field descriptions using struct tags

jba:
Struct fields may have the "jsonschema" struct tag, which is used as the description of the property.

Fixes #47.

### DIFF
--- a/jsonschema/infer.go
+++ b/jsonschema/infer.go
@@ -9,6 +9,7 @@ package jsonschema
 import (
 	"fmt"
 	"reflect"
+	"regexp"
 
 	"github.com/modelcontextprotocol/go-sdk/internal/util"
 )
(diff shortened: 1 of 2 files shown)


tiborvass:
For what it's worth I resolved the conflicts here: https://github.com/tiborvass/mcp-go-sdk/pull/new/js-struct-tag

jba:
@tiborvass Thanks. I wasn't sure how to use your commit so I just changed my local copy and pushed.

tiborvass:
Thank you! Just a priority feedback: this PR is what makes the SDK finally usable for me!

jba:
@tiborvass We fast-tracked this for you :)

tiborvass:
Thank you!!

PR #105, opened 2025-07-08
mcp: fix sseClientConn to use custom HTTP client

crspeller:
When attempting to create an OAuth authentication on top of the SSE transport I discovered the sseClientConn was using http.DefaultClient for requests instead of the custom client provided via SSEClientTransportOptions. 

This change stores the HTTP client in the sseClientConn struct and uses it for all requests. Also adds test to verify the custom client is properly used.

### DIFF
--- a/mcp/sse_test.go
+++ b/mcp/sse_test.go
@@ -37,12 +37,10 @@ func TestSSEServer(t *testing.T) {
 
 			var customClientUsed int64
 			customClient := &http.Client{
-				Transport: &roundTripperFunc{
-					fn: func(req *http.Request) (*http.Response, error) {
-						atomic.AddInt64(&customClientUsed, 1)
-						return http.DefaultTransport.RoundTrip(req)
-					},
-				},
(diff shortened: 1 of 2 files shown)


mcp/sse_test.go
```diff
@@ -162,3 +180,12 @@ func TestScanEvents(t *testing.T) {
 		})
 	}
 }
+
+// roundTripperFunc is a helper to create a custom RoundTripper
+type roundTripperFunc struct {
```
jba: `type roundTripperFunc func(...)`

You should be able to write a method on that.

jba:
LGTM. Rob will approve.

findleyr (commented):
Thanks for this fix! Once you've addressed Jonathan's comment, we can merge.

mcp/sse_test.go
crspeller: Thanks! TIL

jba (approved):

PR #111, opened 2025-07-09
mcp/server: make Run return  on context cancel

### LINKED ISSUE #107: A server connection using stdio transport cannot be closed gracefully
**Describe the bug**
When calling Connect on a server using `mcp.NewStdioTransport` and then subsequently calling `Close()` on the server session, the `Close()` call never returns.

**To Reproduce**

Run this mainprog:

```golang
func main() {
	server := mcp.NewServer("test-server", "1.0.0", nil)

	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt)
	defer stop()

	serverSession, err := server.Connect(ctx, mcp.NewStdioTransport())
	if err != nil {
		panic(err)
	}
	fmt.Println("started")

	<-ctx.Done()
	fmt.Println("stopping")
	if err := serverSession.Close(); err != nil {
		panic(err)
	}

	fmt.Println("stopped")
}
```

After observing the "started" message, stop the program with CTRL-C.

The "stopping" message will be logged, but "stopped" is not logged and the program does not exit.

**Expected behavior**
The program should exit.

**Logs**
NA

**Additional context**
I think this is happening because `jsonrpc2.Connection` expects the `reader.Read` to [return here](https://github.com/modelcontextprotocol/go-sdk/blob/main/internal/jsonrpc2/conn.go#L500) after `Close()` is called but it does not. See [this comment](https://github.com/modelcontextprotocol/go-sdk/blob/main/internal/jsonrpc2/conn.go#L140).

I think this is because the reader is a `json.Decoder` which will not return when the underlying reader is closed.

If this explanation makes sense, I can work on a fix, just let me know.

chriscasola:
Make `Server.Run` return when the provided context is canceled.

Add tests for when the `Run` context is cancelled and also for when the server process receives a signal.

Fixes #107


mcp/cmd_test.go
```diff
@@ -37,24 +39,89 @@ func runServer() {
...
+	onServerExit := make(chan error)
+	go func() {
+		onServerExit <- server.Run(ctx, serverTransport)
+	}()
+
+	// send a ping to the server to ensure it's running
+	client := mcp.NewClient("client", "v0.0.1", nil)
+	session, err := client.Connect(ctx, clientTransport)
+	if err != nil {
+		log.Fatal(err)
```
jba: This is a test. Use t.Fatal.

mcp/cmd_test.go
```diff
@@ -37,24 +39,89 @@ func runServer() {
...
+	if err := session.Ping(context.Background(), nil); err != nil {
+		log.Fatal(err)
+	}
+
+	// cancel the context to stop the server
+	cancel()
+
+	// wait for the server to exit
+	select {
+	case <-time.After(5 * time.Second):
```
jba, line 73: I think this could be shorter. Try 1 second.

And add a TODO to use `synctest` when we can use Go 1.24.

mcp/cmd_test.go
```diff
@@ -37,24 +39,89 @@ func runServer() {
...
 	defer cancel()
 
-	exe, err := os.Executable()
+	cmd := createServerCommand(t)
+
+	client := mcp.NewClient("client", "v0.0.1", nil)
+	session, err := client.Connect(ctx, mcp.NewCommandTransport(cmd))
 	if err != nil {
-		t.Fatal(err)
+		log.Fatal(err)
```
jba: I  must be missing something. Why switch to log?

mcp/server.go
```diff
@@ -404,13 +404,26 @@ func fileResourceHandler(dir string) ResourceHandler {
...
-	return ss.Wait()
+
+	ssClosed := make(chan error)
+	go func() {
+		ssClosed <- ss.Wait()
+	}()
+
+	select {
+	case <-ctx.Done():
+		ss.Close()
```
jba, line 422: This can potentially block indefinitely waiting for existing requests to finish. I think we want non-graceful shutdown in this case.

mcp/cmd_test.go
chriscasola: This test file was already using `log` -- I will replace it.

mcp/cmd_test.go
chriscasola: Reducing the wait time doesn't speed up the test (unless it is likely to fail because the server doesn't exit). Is it worth reducing the wait time at the risk of the test being flaky?

mcp/server.go
chriscasola: The behavior before this PR was also to block indefinitely. I'm not sure it makes sense to change now. Would it be better to handle a timeout on graceful shutdown within `ServerSession.Close`?

jba: OK, if the code was already doing it, leave it.

mcp/cmd_test.go
jba: OK, I see. Leave it.

mcp/cmd_test.go
jba: Never mind, I see the logs were there originally because this sometimes runs as a sub-process. You can leave them. Sorry for the noise.

chriscasola: No problem -- I think `runServer` is the only function called in the sub-process, and it is still using `log`, so should be set.

jba:
Thanks!

PR #112, opened 2025-07-09
mcp: negotiate protocol version

### LINKED ISSUE #103: implement version negotiation

The [MCP spec](https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#protocol-version-header) defines a client can send the `MCP-Protocol-Version` during the init phase

Some option on the `client.Connect` to change the spec version. Right now, [it is hardcoded](https://github.com/modelcontextprotocol/go-sdk/blob/main/mcp/client.go#L111) to the spec's default

jba:
Negotiate the protocol version properly, and set the header on HTTP transports.

We follow the logic of the Typescript SDK.

- On initialization, the client sends the latest version it can support, and accepts the version that the server returns if the client supports it. If not, the connection fails.

- The server accepts the client's version unless it doesn't support it, in which case it replies with its latest version.

Fixes #103.

### DIFF
--- a/mcp/shared.go
+++ b/mcp/shared.go
@@ -26,7 +26,11 @@ import (
 // It is the version that the client sends in the initialization request.
 const latestProtocolVersion = "2025-06-18"
 
-var supportedProtocolVersions = []string{latestProtocolVersion, "2025-03-26"}
+var supportedProtocolVersions = []string{
+	latestProtocolVersion,
+	"2025-03-26",
+	"2024-11-05",
+}
 
 // A MethodHandler handles MCP messages.
 // For methods, exactly one of the return values must be nil.


samthanawalla (dismissed):

mcp/shared.go
```diff
@@ -22,6 +22,12 @@ import (
 	"github.com/modelcontextprotocol/go-sdk/internal/jsonrpc2"
 )
 
+// latestProtocolVersion is the latest protocol version that this version of the SDK supports.
+// It is the version that the client sends in the initialization request.
+const latestProtocolVersion = "2025-06-18"
+
+var supportedProtocolVersions = []string{latestProtocolVersion, "2025-03-26"}
```
findleyr: Don't we also support the 2024-11 version?
If we don't say that, what will that mean for SSE usage?

findleyr (approved):

mcp/shared.go
jba: Added.

findleyr (approved):

PR #113, opened 2025-07-09
mcp: NewClient and NewServer take Implementation

### LINKED ISSUE #109: No way to provide client title in initialization
We have `NewClient(name, version string, opts *ClientOptions)`. The name and version are sent as part the initialization method. But in the latest protocol version there is also "title," which is a human-readable version of the name.

We haven't really painted ourselves into a corner here, since we could add it as a client option. But that's awkward—it is closely associated with the name.

Yet another reason to avoid spread arguments wherever the protocol is involved.

I suggest being annoyingly explicit about this, by exposing the `implementation` protocol type:
```
type implementation struct {
    Name string `json:"name"`
    Title   string `json:"title,omitempty"`
    Version string `json:"version"`
}
```
and having
```
NewClient(*Implementation, *ClientOptions)
```

jba:
Change the name and version arguments to NewClient and NewServer to a *Implementation, to future-proof against the spec.

Fixes #109.

### DIFF
--- a/design/design.md
+++ b/design/design.md
@@ -323,7 +323,7 @@ Sessions are created from either `Client` or `Server` using the `Connect` method
 
 ```go
 type Client struct { /* ... */ }
-func NewClient(name, version string, opts *ClientOptions) *Client
+func NewClient(impl *Implementation, opts *ClientOptions) *Client
 func (*Client) Connect(context.Context, Transport) (*ClientSession, error)
 func (*Client) Sessions() iter.Seq[*ClientSession]
 // Methods for adding/removing client features are described below.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -72,7 +72,7 @@ func main() {
 	ctx := context.Background()
 
 	// Create a new client, with no features.
-	client := mcp.NewClient("mcp-client", "v1.0.0", nil)
+	client := mcp.NewClient(nil, nil)
```
findleyr: I don't think nil should be a valid first argument. We should probably panic in that case (and perhaps also if implementation.Name or Version are missing)? WDYT?

jba: Maybe I don't understand how they're used. I don't see anything in the protocol that cares about the server name. There is no ListServers, just ListTools, etc. It is passed in the initialization exchange and never referred to again.

But we can allow nils later, if we want. Done.

PR #114, opened 2025-07-10
jsonrpc: expose encoding and decoding functions

### LINKED ISSUE #110: No way to implement custom Transport
MCP spec and design document both state that it should be possible to implement a custom transport protocol. This library tries to allow that by defining a `Transport` and `Connection` interfaces which rely on internal `jsonrpc2` types. Some of the types are re-exported, but unfortunately are impossible to construct without relying on internal `jsonrpc2` library.

Specifically `JSONRPCRequest` relies on `JSONRPCID`, which contains only private fields. Because `jsonrpc2.MakeID` is internal function, it is impossible to construct `JSONRPCID`. At minimum `jsonrpc2.MakeID` should also be re-exported. But it would be beneficial for custom transport implementations to also have access to `jsonrpc2.EncodeMessage` and `jsonrpc2.DecodeMessage` functions, and perhaps helpers in `jsonrpc2/wire.go` file.

I am happy to open a PR with the changes, but I am not sure how much of the `jsonrpc2` package should be re-exported.

viddrobnic:
Expose `jsonrpc2.MakeID`, `jsonrpc2.EncodeMessage` and `jsonrpc2.DecodeMessage` functions to allow implementing custom `mcp.Transport`. Add an example that demonstrates a custom transport implementation.

Fixes #110

### DIFF
--- a/examples/custom-transport/main.go
+++ b/examples/custom-transport/main.go
@@ -1,0 +1,109 @@
+// Copyright 2025 The Go MCP SDK Authors. All rights reserved.
+// Use of this source code is governed by an MIT-style
+// license that can be found in the LICENSE file.
+
+package main
+
+import (
+	"bufio"
+	"context"
+	"errors"
+	"io"
+	"log"
+	"os"
+
+	"github.com/modelcontextprotocol/go-sdk/jsonrpc"
+	"github.com/modelcontextprotocol/go-sdk/mcp"
+)
+
+// IOTransport is a simplified implementation of a transport that communicates using
+// newline-delimited JSON over an io.Reader and io.Writer. It is similar to ioTransport
+// in transport.go and serves as a demonstration of how to implement a custom transport.
+type IOTransport struct {
+	r *bufio.Reader
+	w io.Writer
+}
+
+// NewIOTransport creates a new IOTransport with the given io.Reader and io.Writer.
+func NewIOTransport(r io.Reader, w io.Writer) *IOTransport {
+	return &IOTransport{
+		r: bufio.NewReader(r),
+		w: w,
+	}
+}
+
+// ioConn is a connection that uses newlines to delimit messages. It implements [mcp.Connection].
+type ioConn struct {
+	r *bufio.Reader
+	w io.Writer
+}
+
+// Connect implements [mcp.Transport.Connect] by creating a new ioConn.
+func (t *IOTransport) Connect(ctx context.Context) (mcp.Connection, error) {
+	return &ioConn{
+		r: t.r,
+		w: t.w,
+	}, nil
+}
+
+// Read implements [mcp.Connection.Read], assuming messages are newline-delimited JSON.
+func (t *ioConn) Read(context.Context) (jsonrpc.Message, error) {
+	data, err := t.r.ReadBytes('\n')
+	if err != nil {
+		return nil, err
+	}
+
+	return jsonrpc.DecodeMessage(data[:len(data)-1])
+}
+
+// Write implements [mcp.Connection.Write], appending a newline delimiter after the message.
+func (t *ioConn) Write(_ context.Context, msg jsonrpc.Message) error {
+	data, err := jsonrpc.EncodeMessage(msg)
+	if err != nil {
+		return err
+	}
+
+	_, err1 := t.w.Write(data)
+	_, err2 := t.w.Write([]byte{'\n'})
+	return errors.Join(err1, err2)
+}
+
+// Close implements [mcp.Connection.Close]. Since this is a simplified example, it is a no-op.
+func (t *ioConn) Close() error {
+	return nil
+}
+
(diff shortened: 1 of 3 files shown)


design/design.md
```diff
@@ -65,6 +65,18 @@ type (
 	JSONRPCResponse = jsonrpc2.Response
 )
 
+func JOSNRPCMakeID(v any) (JSONRPCID, error) {
```
jba: JSON

design/design.md
```diff
@@ -65,6 +65,18 @@ type (
...
+	return jsonrpc2.MakeID(v)
```
jba: Omit the function definitions from the design doc.

design/design.md
```diff
@@ -639,6 +651,7 @@ mcp.AddTool(server, &mcp.Tool{Name: "subtract", Description: "subtract numbers"}
 
 The `AddTool` method requires an input schema, and optionally an output one. It will not modify them.
 The handler should accept a `CallToolParams` and return a `CallToolResult`.
+
```
jba: remove extra blank lines here and below.

examples/custom-transport/main.go
```diff
@@ -0,0 +1,78 @@
...
+	"bufio"
+	"context"
+	"errors"
+	"io"
+	"os"
+
+	"github.com/modelcontextprotocol/go-sdk/mcp"
+)
+
+// A very simplified implementation of transport.ioTransport that is then used instead of StdioTransport.
```
jba: I don't think I understand this sentence. What is transport.ioTransport? Is your IOTransport doing the same thing as StdioTransport, or something different?

examples/custom-transport/main.go
```diff
@@ -0,0 +1,78 @@
...
+// It servers as a demonstration of how a custom transport can be implemented.
```
jba: serves

examples/custom-transport/main.go
```diff
@@ -0,0 +1,78 @@
...
+type IOTransport struct {
+	r *bufio.Reader
+	w io.Writer
+}
+
+func (t *IOTransport) Connect(ctx context.Context) (mcp.Connection, error) {
+	return t, nil
```
jba: Shouldn't there be a separate connection object returned for each call?

examples/custom-transport/main.go
jba: I think you mean the ioTransport type in transport.go. Since that is not exported, you shouldn't refer to it this way in documentation. You can say something like "see the ioTransport type in transport.go" to make it clear that readers must look at the source code.

examples/custom-transport/main.go
```diff
@@ -0,0 +1,78 @@
...
+)
+
+// A very simplified implementation of transport.ioTransport that is then used instead of StdioTransport.
+// It servers as a demonstration of how a custom transport can be implemented.
+type IOTransport struct {
+	r *bufio.Reader
+	w io.Writer
+}
+
+func (t *IOTransport) Connect(ctx context.Context) (mcp.Connection, error) {
```
jba, line 42: If you're going to show an example, you should document each function so the user understands what it's supposed to do and how it's doing it.
For example:

Connect implements [mcp.Transport.Connect] by creating a ....

examples/custom-transport/main.go
```diff
@@ -0,0 +1,78 @@
...
+	return t, nil
+}
+
+func (t *IOTransport) Read(context.Context) (mcp.JSONRPCMessage, error) {
```
jba: This comment would say that it assumes the messages are newline-delimited JSON.

examples/custom-transport/main.go
```diff
@@ -0,0 +1,78 @@
...
+func main() {
+	server := mcp.NewServer("greeter", "v0.0.1", nil)
+	mcp.AddTool(server, &mcp.Tool{Name: "greet", Description: "say hi"}, SayHi)
+
+	// Run server with custom IOTransport using stdio.
+	transport := &IOTransport{
+		r: bufio.NewReader(os.Stdin),
+		w: os.Stdout,
+	}
+	server.Run(context.Background(), transport)
```
jba: Check the error returned by `Run`.

jba:
#117 moves the other JSONRPC symbols to their own package. When that lands, it would be great if you would rebase this onto it and make the corresponding changes.

jba:
#117 has landed.

jsonrpc/jsonrpc.go
```diff
@@ -18,3 +18,15 @@ type (
 	// Response is a JSON-RPC response.
 	Response = jsonrpc2.Response
 )
+
+func MakeID(v any) (ID, error) {
```
jba, line 26: Add doc for these. You can copy from jsonrpc2.

jba (approved):

jba:
Thanks!

PR #119, opened 2025-07-10
Add NewIOTransport function

joshcarp:
Adds new `NewIOTransport` function that returns a `Transport`. 
I needed this because I'm reading/writing to a custom file descriptor.


jba:
We'd probably also want to export the `IOTransport` type and return that directly, in case we want to add exported fields or methods to it.

This sounds reasonable, but we're going to discuss it internally a bit, as we would with most new API. That may take a couple of weeks due to vacations.

joshcarp:
Yeah i noticed this repo isn't returning interfaces, so as a convention it's probably a good idea to make IOTransport public too

findleyr:
See also https://github.com/modelcontextprotocol/go-sdk/pull/376. I think this is reasonable, but we need a proposal.

jba:
I believe this is obsoleted by #376.

findleyr:
Yes, this is obsolete with #376, which has a proposal.

PR #123, opened 2025-07-11
mcp/stream: fix prevent potential goroutine leak on blocked incoming channel

cryo-zd:
### Change:
Adds a regression test to ensure the `scanEvents` goroutine exits properly when the `done` channel is closed and the `incoming` channel is full.

Previously, there was a potential for goroutine leak if `scanEvents` was blocked on writing to `incoming` and the `done` channel was closed before it could exit. This issue has already been fixed in prior changes — this test is intended to capture that edge case and prevent regressions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,7 @@ import (
 )
 
 type HiParams struct {
-	Name string `json:"name", mcp:"the name of the person to greet"`
+	Name string `json:"name" jsonschema:"the name of the person to greet"`
 }
 
 func SayHi(ctx context.Context, cc *mcp.ServerSession, params *mcp.CallToolParamsFor[HiParams]) (*mcp.CallToolResultFor[any], error) {
--- a/examples/hello/main.go
+++ b/examples/hello/main.go
@@ -19,7 +19,7 @@ import (
 var httpAddr = flag.String("http", "", "if set, use streamable HTTP at this address, instead of stdin/stdout")
 
 type HiArgs struct {
-	Name string `json:"name" mcp:"the name to say hi to"`
+	Name string `json:"name" jsonschema:"the name to say hi to"`
 }
 
 func SayHi(ctx context.Context, ss *mcp.ServerSession, params *mcp.CallToolParamsFor[HiArgs]) (*mcp.CallToolResultFor[struct{}], error) {
--- a/internal/readme/server/server.go
+++ b/internal/readme/server/server.go
@@ -13,7 +13,7 @@ import (
 )
 
 type HiParams struct {
-	Name string `json:"name", mcp:"the name of the person to greet"`
+	Name string `json:"name" jsonschema:"the name of the person to greet"`
 }
 
 func SayHi(ctx context.Context, cc *mcp.ServerSession, params *mcp.CallToolParamsFor[HiParams]) (*mcp.CallToolResultFor[any], error) {
--- a/mcp/conformance_test.go
+++ b/mcp/conformance_test.go
@@ -23,6 +23,7 @@ import (
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/modelcontextprotocol/go-sdk/internal/jsonrpc2"
+	"github.com/modelcontextprotocol/go-sdk/jsonrpc"
 	"golang.org/x/tools/txtar"
 )
 
@@ -46,12 +47,12 @@ var update = flag.Bool("update", false, "if set, update conformance test data")
 // with -update to have the test runner update the expected output, which may
 // be client or server depending on the perspective of the test.
 type conformanceTest struct {
-	name                      string           // test name
-	path                      string           // path to test file
-	archive                   *txtar.Archive   // raw archive, for updating
-	tools, prompts, resources []string         // named features to include
-	client                    []JSONRPCMessage // client messages
-	server                    []JSONRPCMessage // server messages
+	name                      string            // test name
+	path                      string            // path to test file
+	archive                   *txtar.Archive    // raw archive, for updating
+	tools, prompts, resources []string          // named features to include
+	client                    []jsonrpc.Message // client messages
+	server                    []jsonrpc.Message // server messages
 }
 
 // TODO(rfindley): add client conformance tests.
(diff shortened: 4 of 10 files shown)


mcp/streamable_test.go
```diff
@@ -597,3 +599,49 @@ func TestEventID(t *testing.T) {
...
+	// Start the handleSSE goroutine manually
+	go conn.handleSSE(resp)
+
+	// Wait until incoming channel is filled
+	deadlineCtx, cancelFunc := context.WithTimeout(context.Background(), 30*time.Millisecond)
+	defer cancelFunc()
+	for len(conn.incoming) < cap(conn.incoming) { // Ensure enough events are written
+		select {
+		case <-deadlineCtx.Done():
+			t.Skipf("timeout when waiting for streamableClientConn.incoming to be full, test skipped: %v", deadlineCtx.Err())
```
jba: Make the timeout longer (100 milliseconds) and fail if this happens instead of skipping.

cryo-zd: Got it and updated.

jba (approved):

mcp/streamable_test.go
```diff
@@ -597,3 +599,49 @@ func TestEventID(t *testing.T) {
...
+		case <-deadlineCtx.Done():
+			t.Fatalf("timeout when waiting for streamableClientConn.incoming to be full: %v", deadlineCtx.Err())
+		default:
+			// Continue checking until the channel is full
+		}
+	}
+
+	// Now simulate calling Close() and blocking the goroutine
+	close(conn.done)
+	time.Sleep(50 * time.Millisecond) // Allow goroutine to exit
```
samthanawalla: Can we use a sync.WaitGroup instead of a sleep to remove flakiness?

mcp/streamable_test.go
```diff
@@ -597,3 +599,49 @@ func TestEventID(t *testing.T) {
...
+
+	// Wait until incoming channel is filled
+	deadlineCtx, cancelFunc := context.WithTimeout(context.Background(), 100*time.Millisecond)
+	defer cancelFunc()
+	for len(conn.incoming) < cap(conn.incoming) { // Ensure enough events are written
+		select {
+		case <-deadlineCtx.Done():
+			t.Fatalf("timeout when waiting for streamableClientConn.incoming to be full: %v", deadlineCtx.Err())
+		default:
+			// Continue checking until the channel is full
```
samthanawalla, line 635: this is a busy-wait loop.

Can we instead do something like:

select {
case <-conn.incoming:
...
case <-deadlineCtx.Done():
...
}

samthanawalla (changes requested):

mcp/streamable_test.go
cryo-zd: I considered replacing the loop with a `select`, but in this test case, our goal is to ensure that the `scanEvents` goroutine has not only started but has actually reached the point where it attempts to write into the `incoming` channel — and that it blocks there due to the channel being full.

It seems that using `select` to observe a write or read on `incoming` does not reliably indicate this state due to non-deterministic scheduling. There is a chance the goroutine hasn't been scheduled or hasn't reached the blocking point yet, leading to possible false negatives in leak detection.

To ensure accurate test behavior, I chose to wait explicitly until `incoming` is full.

mcp/streamable_test.go
cryo-zd: 🤔,   I agree that `sync.WaitGroup` is a great tool for waiting on goroutines, but in this case, since `scanEvents` is launched within `handleSSE` and doesn't expose a separate wait condition, `sync.WaitGroup` wouldn't directly help in waiting for `scanEvents` to exit.

samthanawalla: Ack, we could wrap the call to handleSSE:

```
go func() {
    conn.handleSSE(resp)
    wg.Done()
}
```

mcp/streamable_test.go
samthanawalla: Understood. Thanks for the clarification.

samthanawalla (changes requested):

mcp/streamable_test.go
cryo-zd: 👋, sorry for my delayed response.

I’ve implemented the proposed approach using a `sync.WaitGroup` around `handleSSE`, and replaced the `time.Sleep` with `wg.Wait()`. 

However, the test fails intermittently — although the leak has been correctly fixed. My understanding is that the `scanEvents` goroutine, which is launched inside `handleSSE`, might not observe the closed `done` channel and exit before `handleSSE` returns and `wg.Wait()` completes due to scheduling variance. So the test catches it as a false positive leak, even though the fix is correct.

Thanks again for the helpful suggestions and patience!
testFail

samthanawalla: Can you rebase with upstream and upload the test?

I added the waitgroup and I am not able to reproduce the failures.

samthanawalla: Could you also update the PR to mention this CL is to add a leak test.

It seems Rob beat you to the fix! https://github.com/modelcontextprotocol/go-sdk/pull/130

cryo-zd: Thanks for the quick response!

I actually noticed two days ago that the leak issue had been fixed — great to see that resolved.

I've rebased the branch and pushed the test update using `sync.WaitGroup` as suggested. On my local runs, the test sometimes passes, but occasionally fails due to a detected goroutine leak, which I believe is caused by scheduling variance.

This test currently relies on timing in two places to simulate the blocked goroutine scenario, and I personally feel this makes it a bit less elegant and deterministic. Given the potential for flakiness in CI, I'm also happy to close this PR if you'd prefer not to include this regression test.

test3

samthanawalla: I'm not sure why it's failing on your machine, it seems to work on mine after running it 10000 times 🤔  :

```
:~/w/go-sdk/mcp$ go test -count=10000 -race -run TestStreamableClientConnSSEGoroutineLeak
PASS
ok      github.com/modelcontextprotocol/go-sdk/mcp      12.092s
```

cryo-zd: 👋, sorry for the delay and thank you so much for your patience and continued feedback!

To help validate the issue further, I added a dedicated step to my GitHub Actions workflow in my [personal fork](https://github.com/cryo-zd/mcp-go-sdk/tree/test) using the following command:
```go
- name: Test
      run: go test -v ./...
- name: TestGoroutineLeak
      run: go test -count=1000 -v -race -run TestStreamableClientConnSSEGoroutineLeak ./mcp
```

This directly targets the `streamable_test.go` file in the `mcp` package, where the leak test is defined. I noticed a failure in this setup using the same environment as upstream CI.

Here's the link to the failing run: [link](https://github.com/cryo-zd/mcp-go-sdk/actions/runs/16440598591/job/46460842694)

When you have a moment, you could take a look and confirm whether it aligns with what you’d expect?

samthanawalla: Thanks for that!

I understand now that sleep is necessary to allow for scheduling variance. We can keep the sync.waitgroup and add back the small delay to allow for scanEvents to shut down. Then it should be good to submit. Thanks!

cryo-zd: Thanks again for all the helpful discussion and feedback.

While preparing to update the PR, I noticed that the latest version of the codebase has significantly changed the SSE handling logic( #133 ). The `handleSSE` function no longer starts a separate `scanEvents` goroutine. Instead, it now processes the stream directly through `processStream` in a blocking loop.

As a result, the potential `scanEvents` goroutine leak scenario we were testing for no longer exists in this implementation. The regression test now always passes, but it no longer provides meaningful protection or value.

To avoid unnecessary test complexity and maintenance cost, I'm going to go ahead and close this PR and the related issue.Please feel free to reopen if there's still interest in preserving a version of the test for historical reference.

Thanks again for your time and thoughtful review!

PR #125, opened 2025-07-11
internal/oauthex: OAuth extensions

jba:
Add a package for the extensions to OAuth 2.0 required by MCP.

This first PR adds Protected Resource Metadata.


internal/oauthex/resource_meta.go
RoyceLeonD, line 1: Hey guys, I'm tasked with moving a production MCP server from Python to Go, and just started looking at this repo, so forgive me if I'm missing something. But based on this PR, I  see that you:
- Mount discovery endpoints (/.well-known/oauth-protected-resource)
- Handle 401 responses with WWW-Authenticate headers
- Provide OAuth server discovery
- Use token binding to specific resources

This is obviously valuable since it is needed to protect the routes. And in my case (as in what I asume most people's case, the oauth server provides a vast majority of the inputs for this). But instead of 2.0, don't we need OAuth 2.1 (I have a feeling we kind of get this for free because this is just a discovery endpoint and the clinet will have to do PKCE with the end server), but I want to be clear that our testing revolves around 2.1 while also thinking about how an sdk end user will actually configure the ResourceMetadata. Like is it:
```
 server := mcp.NewServer(impl, caps)

  // Configure resource metadata (RFC 9728)
  server.SetResourceMetadata(&oauthex.ProtectedResourceMetadata
  {
      ResourceServers: []oauthex.ResourceServer{
          {
              Resource: "https://myapp.com/mcp",
              AuthorizationServers:
  []string{"https://myapp.com/oauth"},
              ScopesSupported: []string{"user:read",
  "user:write", "mcp:access"},
              // OAuth 2.1 specific requirements
              PKCERequired: true,
              TokenEndpointAuthMethodsSupported:
  []string{"client_secret_basic", "private_key_jwt"},
          },
      },
  })
```
Also i know this Pr is specifically for just defining the metadataserver, but is there a but is there a plan for how the data configured here will be used on the MCP tool calling side to   make the token accessible by the tool without needing to pull in custom HTTP headers?
  For example, would we eventually be able to do something like:
````
  server.RegisterTool(&mcp.ToolDefinition{
      Name: "search_users",
      RequiredScopes: []string{"user:read"}, // Links to
  ResourceMetadata.ScopesSupported
      Handler: func(ctx context.Context, session
  *mcp.ServerSession, params *mcp.CallToolParams)
  (*mcp.CallToolResult, error) {
          // Token and user context automatically available
  from OAuth flow
          user := session.GetAuthenticatedUser()
          // ... business logic
      },
  })
```

jba: @RoyceLeonD Can you file an issue for making the token attributes accessible to the server-side feature (tool, prompt, etc.)? Besides the user, what else is valuable?

jba: @RoyceLeonD I'd like you to take a look at #176 and tell me your thoughts.

internal/oauthex/resource_meta.go
```diff
@@ -0,0 +1,382 @@
...
+	if res.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("bad status %s", res.Status)
+	}
+	// Spec §3.2 requires application/json.
+	if ct := res.Header.Get("Content-Type"); ct != "application/json" {
+		return nil, fmt.Errorf("bad content type %q", ct)
+	}
+
+	var prm ProtectedResourceMetadata
+	dec := json.NewDecoder(res.Body)
```
rolandshoemaker, line 190: Do we have a general idea of how large the JSON body here is going to be? If so should we be using a LimitedReader to avoid malicious requests consuming arbitrary amounts of memory?

jba: Will do in a follow-up.

PR #126, opened 2025-07-11
jsonschema: add missing colon to link block

prattmic:
Without this, not only is the one link not valid, the entire section is considered invalid and none of the link references are used.


prattmic:
cc @jba @samthanawalla

findleyr:
Thanks! The CI indicates that this needs to be formatted with gofmt. Is that right?

prattmic:
Ah, kinda. One of the other links was misnamed (and thus unreferenced), so gofmt wanted to move it out of the link block. Fixed now.

jba (approved):

jba:
Thanks.

PR #133, opened 2025-07-14
mcp/streamable: add resumability for the Streamable transport

samthanawalla:
This CL implements a retry mechanism to resume SSE streams to recover from network failures.

For #10

I referenced 
* https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#resumability-and-redelivery 
* https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/client/streamableHttp.ts

### DIFF
--- a/mcp/streamable_test.go
+++ b/mcp/streamable_test.go
@@ -113,7 +113,7 @@ func TestStreamableTransports(t *testing.T) {
 // mid-stream network failure and receive replayed messages. It uses a proxy
 // that is killed and restarted to simulate a recoverable network outage.
 func TestClientReplay(t *testing.T) {
-	notifications := make(chan string, 10)
+	notifications := make(chan string)
 	// 1. Configure the real MCP server.
 	server := NewServer(testImpl, nil)
 


mcp/streamable.go
```diff
@@ -594,12 +597,38 @@ type StreamableClientTransport struct {
 	opts StreamableClientTransportOptions
 }
 
+// StreamableClientReconnectionOptions defines parameters for client reconnection attempts.
+type StreamableClientReconnectionOptions struct {
+	// InitialDelay is the base delay for the first reconnection attempt
```
jba: end with period

mcp/streamable.go
```diff
@@ -594,12 +597,38 @@ type StreamableClientTransport struct {
...
+	InitialDelay time.Duration // default: 1 second
```
jba: There's no need to doc the defaults since they're visible in the exported default value for this struct.

mcp/streamable.go
```diff
@@ -594,12 +597,38 @@ type StreamableClientTransport struct {
 	opts StreamableClientTransportOptions
 }
 
+// StreamableClientReconnectionOptions defines parameters for client reconnection attempts.
```
jba: How about s/Reconnection/Reconnect/?

mcp/streamable.go
```diff
@@ -800,3 +857,83 @@ func (s *streamableClientConn) Close() error {
...
+	}
+
+	s.handleSSE(resp, opts)
+}
+
+// scheduleReconnection schedules the next SSE reconnection attempt after a delay.
+func (s *streamableClientConn) scheduleReconnection(opts *startSSEOptions) {
+	reconnOpts := s.reconnectionOptions
+	if opts.attempt >= reconnOpts.MaxRetries {
+		return
```
jba: How does this manifest to the user? Is there a way to return an error?

mcp/streamable.go
```diff
@@ -704,11 +748,19 @@ func (s *streamableClientConn) Write(ctx context.Context, msg jsonrpc.Message) e
...
 		// locked
 		s._sessionID = gotSessionID
+		// With the session now established, launch the persistent background listener for server-pushed events.
+		go s.establishSSE(&startSSEOptions{})
 	}
 
 	return nil
 }
 
+// startSSEOptions holds parameters for initiating an SSE stream.
```
jba: nit: these aren't really options. This is state. `startSSEState`.

mcp/streamable.go
```diff
@@ -800,3 +857,83 @@ func (s *streamableClientConn) Close() error {
...
+	select {
+	case <-s.done:
+		return
+	case <-time.After(delay):
+		opts.attempt++
+		s.establishSSE(opts)
+	}
+}
+
+// calculateReconnectionDelay calculates a delay using exponential backoff with full jitter.
```
jba: "Reconnect" here too, I think. Save a few characters.

mcp/streamable.go
```diff
@@ -800,3 +857,83 @@ func (s *streamableClientConn) Close() error {
...
+func calculateReconnectionDelay(opts *StreamableClientReconnectionOptions, attempt int) time.Duration {
+	// Calculate the exponential backoff using the grow factor.
+	backoffDuration := time.Duration(float64(opts.InitialDelay) * math.Pow(opts.GrowFactor, float64(attempt)))
```
jba: Do you check GrowFactor anywhere? A negative value would behave weirdly.

jba (changes requested):
First preliminary review. With Rob out, I want to really understand this. That may take a day or two.

mcp/streamable.go
samthanawalla: SGTM

mcp/streamable.go

mcp/streamable.go

mcp/streamable.go

mcp/streamable.go

mcp/streamable.go
samthanawalla: Good point, added a check in Connect

mcp/streamable.go
samthanawalla: Good point, surfaced the error in s.err

mcp/streamable.go
```diff
@@ -594,12 +597,38 @@ type StreamableClientTransport struct {
 	opts StreamableClientTransportOptions
 }
 
+// StreamableClientReconnectionOptions defines parameters for client reconnection attempts.
+type StreamableClientReconnectionOptions struct {
```
jba: I think we should export only MaxRetries. For the others we can pick reasonable defaults that will probably work forever, or a long time. WCAAIL.

mcp/streamable.go
jba: I think we can also drop "Client", since there's no analogous server feature.

mcp/streamable.go
samthanawalla: And updated Read to surface this error

mcp/streamable.go
```diff
@@ -704,11 +760,19 @@ func (s *streamableClientConn) Write(ctx context.Context, msg jsonrpc.Message) e
 	if sessionID == "" {
 		// locked
 		s._sessionID = gotSessionID
+		// With the session now established, launch the persistent background listener for server-pushed events.
+		go s.establishSSE(&startSSEState{})
```
jba: This goroutine is not present in the current code. Why is it needed here?

More general question: is there a description of the design of resumability somewhere? Not looking for a design doc, just a short sketch.

mcp/streamable.go
```diff
@@ -754,17 +818,20 @@ func (s *streamableClientConn) postMessage(ctx context.Context, sessionID string
...
+func (s *streamableClientConn) handleSSE(resp *http.Response, opts *startSSEState) {
 	defer resp.Body.Close()
 
 	done := make(chan struct{})
 	go func() {
 		defer close(done)
 		for evt, err := range scanEvents(resp.Body) {
 			if err != nil {
-				// TODO: surface this error; possibly break the stream
+				s.scheduleReconnect(opts)
```
jba: Not every error is going to be worth reconnecting for, is it? There could be bugs or bad data. We'll probably need a way to distinguish these. For a later PR.

mcp/streamable.go
```diff
@@ -800,3 +869,87 @@ func (s *streamableClientConn) Close() error {
...
+
+	// Per the spec, a 405 response means the server doesn't support SSE streams at this endpoint.
+	if resp.StatusCode == http.StatusMethodNotAllowed {
+		resp.Body.Close()
+		return
+	}
+
+	if !strings.Contains(resp.Header.Get("Content-Type"), "text/event-stream") {
+		resp.Body.Close()
+		s.scheduleReconnect(opts)
```
jba: Why do we think reconnecting will help?

mcp/streamable.go
```diff
@@ -800,3 +869,87 @@ func (s *streamableClientConn) Close() error {
...
+		return
+	}
+
+	s.handleSSE(resp, opts)
```
jba: You started a goroutine with this in postMessage. Why are you calling it again here?

jba (changes requested):

mcp/streamable.go
samthanawalla: IIUC, the previous logic only covered [section 2.1](https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#sending-messages-to-the-server)

Which established an SSE stream if the server responds to a client POST with a text/event-stream.

However in [section 2.2](https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#listening-for-messages-from-the-server), it says: The client MAY issue an HTTP GET to the MCP endpoint. This can be used to open an SSE stream, allowing the server to communicate to the client, without the client first sending data via HTTP POST

So from my understanding, we need to proactively issue a client GET request after initialization to see if the server has any SSE streams waiting before a client POST.

Added a comment.

mcp/streamable.go
samthanawalla: Good point, I'll follow-up with another CL after this one.

mcp/streamable.go
samthanawalla: Ah I don't think it will, good point. Updated

mcp/streamable.go
samthanawalla: This is for the client GET initiated stream.

I think there are 2 flows to start an SSE stream: 
1. client GET request (here)
2. client POST request (postMessage)

mcp/streamable.go
samthanawalla: Done. Removed the growFactor check since it's unexported now

mcp/streamable.go

mcp/streamable.go
```diff
@@ -800,3 +869,86 @@ func (s *streamableClientConn) Close() error {
...
+	}
+
+	delay := calculateReconnectDelay(reconnOpts, opts.attempt)
+
+	select {
+	case <-s.done:
+		return
+	case <-time.After(delay):
+		opts.attempt++
+		s.establishSSE(opts)
```
jba: This is a recursive call to establishSSE, so in theory you could blow the stack.
Not sure what the right fix is.

samthanawalla: Changed to iterative instead of recursive

mcp/streamable.go
jba: Hmm...I don't really understand that part of the spec. But we can continue this discussion with Rob.

mcp/streamable.go
```diff
@@ -757,34 +809,118 @@ func (s *streamableClientConn) postMessage(ctx context.Context, sessionID string
...
+		eventID, clientClosed := s.processStream(resp)
+		lastEventID = eventID
+
+		// If the connection was closed by the client, we're done.
+		if clientClosed {
+			return
+		}
+
+		// The stream was interrupted or ended by the server. Attempt to reconnect.
+		newResp, reconnectErr := s.reconnect(lastEventID)
```
jba: `err` is fine.

mcp/streamable.go
```diff
@@ -757,34 +809,118 @@ func (s *streamableClientConn) postMessage(ctx context.Context, sessionID string
...
+		if reconnectErr != nil {
+			// All reconnection attempts failed. Set the final error, close the
+			// connection, and exit the goroutine.
+			s.mu.Lock()
+			s.err = reconnectErr
+			s.mu.Unlock()
+			s.Close()
```
jba, line 836: Could handleSSE return an error so we can propagate the error from Close?

mcp/streamable.go
```diff
@@ -757,34 +809,118 @@ func (s *streamableClientConn) postMessage(ctx context.Context, sessionID string
 	return sessionID, nil
 }
 
-func (s *streamableClientConn) handleSSE(resp *http.Response) {
+// handleSSE manages the entire lifecycle of an SSE connection. It processes
+// an incoming Server-Sent Events stream and automatically handles reconnection
+// logic if the stream breaks.
+func (s *streamableClientConn) handleSSE(initialResp *http.Response) {
```
jba, line 815: This looks very clean!

mcp/streamable.go
```diff
@@ -757,34 +809,118 @@ func (s *streamableClientConn) postMessage(ctx context.Context, sessionID string
...
+// an error if all retries are exhausted.
+func (s *streamableClientConn) reconnect(lastEventID string) (*http.Response, error) {
+	var finalErr error
+
+	for attempt := 0; attempt < s.ReconnectOptions.MaxRetries; attempt++ {
+		select {
+		case <-s.done:
+			return nil, fmt.Errorf("connection closed by client during reconnect")
+		case <-time.After(calculateReconnectDelay(s.ReconnectOptions, attempt)):
+			resp, reconnectErr := s.establishSSE(lastEventID)
```
jba: `err` again

mcp/streamable.go
```diff
@@ -757,34 +809,118 @@ func (s *streamableClientConn) postMessage(ctx context.Context, sessionID string
...
+			if reconnectErr != nil {
+				finalErr = reconnectErr // Store the error and try again.
+				continue
 			}
-			select {
-			case <-s.done:
-				return
-			case s.incoming <- evt.Data:
+
+			if !isResumable(resp) {
```
jba, line 890: I think we may also want to try resuming if the error from establishSSE is non-nil. For example, if the network is partitioned, that might manifest as a timeout error instead of an HTTP response. But we can leave that for a later PR.

mcp/streamable.go
```diff
@@ -757,34 +809,118 @@ func (s *streamableClientConn) postMessage(ctx context.Context, sessionID string
...
-	case <-s.done:
-	case <-done:
+// isResumable checks if an HTTP response indicates a valid SSE stream that can be processed.
+func isResumable(resp *http.Response) bool {
+	// Per the spec, a 405 response means the server doesn't support SSE streams at this endpoint.
+	if resp.StatusCode == http.StatusMethodNotAllowed {
+		return false
 	}
+
+	if !strings.Contains(resp.Header.Get("Content-Type"), "text/event-stream") {
```
jba: return strings.Contains...

mcp/streamable.go
```diff
@@ -803,3 +939,38 @@ func (s *streamableClientConn) Close() error {
...
+	if s._sessionID != "" {
+		req.Header.Set("Mcp-Session-Id", s._sessionID)
+	}
+	s.mu.Unlock()
+	if lastEventID != "" {
+		req.Header.Set("Last-Event-ID", lastEventID)
+	}
+	req.Header.Set("Accept", "text/event-stream")
+
+	resp, err := s.client.Do(req)
```
jba: return s.client.Do...

mcp/streamable_test.go
```diff
@@ -105,6 +108,156 @@ func TestStreamableTransports(t *testing.T) {
 	}
 }
 
+// TestClientReplayAfterProxyBreak verifies that the client can recover from a
+// mid-stream network failure and receive replayed messages. It uses a proxy
+// that is killed and restarted to simulate a recoverable network outage.
+func TestClientReplayAfterProxyBreak(t *testing.T) {
```
jba: nit: "proxy" is an implementation detail; remove it

mcp/streamable_test.go
```diff
@@ -105,6 +108,156 @@ func TestStreamableTransports(t *testing.T) {
...
+	// 1. Configure the real MCP server.
+	server := NewServer(testImpl, nil)
+
+	// Use a channel to synchronize the server's message sending with the test's
+	// proxy-killing action.
+	serverReadyToKillProxy := make(chan struct{})
+	var serverClosed sync.WaitGroup
+	AddTool(server, &Tool{Name: "multiMessageTool"}, func(ctx context.Context, ss *ServerSession, params *CallToolParamsFor[any]) (*CallToolResultFor[any], error) {
```
jba: CallTool{Params,Result} are aliases for CallTool{Params,Result}For[any]

mcp/streamable_test.go
```diff
@@ -105,6 +108,156 @@ func TestStreamableTransports(t *testing.T) {
...
+// mid-stream network failure and receive replayed messages. It uses a proxy
+// that is killed and restarted to simulate a recoverable network outage.
+func TestClientReplayAfterProxyBreak(t *testing.T) {
+	// 1. Configure the real MCP server.
+	server := NewServer(testImpl, nil)
+
+	// Use a channel to synchronize the server's message sending with the test's
+	// proxy-killing action.
+	serverReadyToKillProxy := make(chan struct{})
+	var serverClosed sync.WaitGroup
```
jba: Looks like this could be just another channel.

mcp/streamable_test.go
```diff
@@ -105,6 +108,156 @@ func TestStreamableTransports(t *testing.T) {
...
+			bgCtx := context.Background()
+			// Send the first two messages immediately.
+			_ = ss.NotifyProgress(bgCtx, &ProgressNotificationParams{Message: "msg1"})
+			_ = ss.NotifyProgress(bgCtx, &ProgressNotificationParams{Message: "msg2"})
+
+			// Signal the test that it can now kill the proxy.
+			serverClosed.Add(1)
+			close(serverReadyToKillProxy)
+			// Wait for the test to kill the proxy before sending the rest.
+			serverClosed.Wait()
```
jba: You'd wait for the channel to be closed here (`<-serverClosed`)

mcp/streamable_test.go
```diff
@@ -105,6 +108,156 @@ func TestStreamableTransports(t *testing.T) {
...
+		t.Fatalf("Failed to parse real server URL: %v", err)
+	}
+
+	// 2. Configure a proxy that sits between the client and the real server.
+	proxyHandler := httputil.NewSingleHostReverseProxy(realServerURL)
+	proxy := httptest.NewServer(proxyHandler)
+	proxyAddr := proxy.Listener.Addr().String() // Get the address to restart it later.
+
+	// 3. Configure the client to connect to the proxy with default options.
+	clientTransport := NewStreamableClientTransport(proxy.URL, &StreamableClientTransportOptions{
```
jba: I think `nil` should give the same behavior as this: retry should be the default, since it doesn't cost anything. Not sure what the default MaxRetries should be, but 5 seems OK.

mcp/streamable_test.go
```diff
@@ -105,6 +108,156 @@ func TestStreamableTransports(t *testing.T) {
...
+	defer cancel()
+
+	// 4. Connect, perform handshake, and trigger the tool.
+	conn, err := clientTransport.Connect(ctx)
+	if err != nil {
+		t.Fatalf("Connect() failed: %v", err)
+	}
+
+	// Perform handshake.
+	initReq := &jsonrpc.Request{ID: jsonrpc2.Int64ID(100), Method: "initialize", Params: mustMarshal(t, &InitializeParams{})}
```
jba: Why does this have to happen by hand?
I guess what I mean is, can you write this test with a `Client` instead of at the transport level? If you need to reach into the transport for something, maybe you could add test hooks to the transport code. That may ultimately be cleaner that re-writing the init protocol.

mcp/streamable_test.go
```diff
@@ -105,6 +108,156 @@ func TestStreamableTransports(t *testing.T) {
...
+		// Server has sent the first two messages and is paused.
+	case <-ctx.Done():
+		t.Fatalf("Context timed out before server was ready to kill proxy")
+	}
+
+	// 6. Simulate a total network failure by closing the proxy.
+	t.Log("--- Killing proxy to simulate network failure ---")
+	proxy.CloseClientConnections()
+	proxy.Close()
+	serverClosed.Done()
```
jba: Here you'd close the channel.

mcp/streamable_test.go
```diff
@@ -105,6 +108,156 @@ func TestStreamableTransports(t *testing.T) {
...
+// Helper to read a specific number of progress notifications.
+func readProgressNotifications(t *testing.T, ctx context.Context, conn Connection, count int) []jsonrpc.Message {
+	t.Helper()
+	var notifications []jsonrpc.Message
+	for len(notifications) < count && ctx.Err() == nil {
+		msg, err := conn.Read(ctx)
+		if err != nil {
+			t.Fatalf("Failed to read notification: %v", err)
+		}
+		if req, ok := msg.(*jsonrpc.Request); ok && req.Method == "notifications/progress" {
```
jba: Should it also be an error if you see a non-notification?

mcp/streamable.go

mcp/streamable.go
samthanawalla: 😎

mcp/streamable.go
samthanawalla: handleSSE can't return an error since it runs in its own goroutine

mcp/streamable.go

mcp/streamable.go
samthanawalla: With the way it's written- we will retry another attempt if the error is non-nil. Are you saying that we have to do something special in that case?

mcp/streamable.go

mcp/streamable.go

mcp/streamable_test.go

mcp/streamable_test.go
samthanawalla: Good point, done.

mcp/streamable_test.go

mcp/streamable_test.go
samthanawalla: Good point, done.

mcp/streamable_test.go

mcp/streamable_test.go
samthanawalla: Updated to only hook into progress notifications.

mcp/streamable_test.go
samthanawalla: Sorry I am not following, could you clarify what you mean?

mcp/streamable_test.go
samthanawalla: Good point. Updated to test with Client.

mcp/streamable_test.go
```diff
@@ -223,39 +197,37 @@ func TestClientReplayAfterProxyBreak(t *testing.T) {
 	go restartedProxy.Serve(listener)
```
jba, line 197: How is this goroutine stopped?

mcp/streamable_test.go
```diff
@@ -108,37 +109,37 @@ func TestStreamableTransports(t *testing.T) {
 	}
 }
 
-// TestClientReplayAfterProxyBreak verifies that the client can recover from a
+// TestClientReplay verifies that the client can recover from a
 // mid-stream network failure and receive replayed messages. It uses a proxy
 // that is killed and restarted to simulate a recoverable network outage.
-func TestClientReplayAfterProxyBreak(t *testing.T) {
+func TestClientReplay(t *testing.T) {
+	notifications := make(chan string, 10)
```
jba: Document how the channel buffer size affects correctness. (e.g. must be at least as large as the number of notifications sent in the tool handler on L125?)

mcp/streamable_test.go
samthanawalla: restartedProxy.Close() will terminate it.

"Serve always returns a non-nil error and closes l. After [Server.Shutdown] or [Server.Close], the returned error is [ErrServerClosed]."

mcp/streamable_test.go
samthanawalla: Looks like buffer size has no effect since we collect them all. Removed the size of 10.

jba (approved):

Issue #28, opened 2025-06-24
JSON schema: hosting, draft implementation, validation


findleyr:
This umbrella issue tracks various considerations for JSON schema support.

- [x] The current JSON schema package is useful. Perhaps it should be published outside of this repository.
- [x] There are many drafts of the spec. Clarify which draft MCP expects, and implement it strictly. See also https://github.com/modelcontextprotocol/modelcontextprotocol/pull/655
- [x] Support for schema validation with defaults.

jba:
I marked "schema validation with defaults" complete because we already have it: Resolved.ApplyDefaults applies the defaults, and then Resolved.Validates the result.

jba:
I think it should live in github.com/google/jsonschema-go or github.com/google/go-jsonschema, according to the current fashion.

There is go-cmp and go-cloud, but also generative-ai-go and adk-go.

mhpenta:
Agree it should live outside the mcp package. It has uses outside of the immediate mcp context, such as generating schemas to request json objects from LLMs in other contexts. 

I might personally contribute a fluent API wrapper for manual schema construction, similar to what I've done with github.com/mhpenta/jobj. Mine is just a subset of an older spec but the manual construction is useful in certain contexts, e.g., for very long descriptions. 

I like github.com/google/go-jsonschema

jba:
@mhpenta jobj looks very nice. It's also great just where it is. I'd prefer to keep the jsonschema API to a minimum. Perhaps our README could refer to your package to raise visibility. Anyway, we can defer this conversation for when the package has been moved and is up and running.

samthanawalla:
We are aiming to have the repo released by mid-august.

samthanawalla:
We moved the current state of jsonschema to https://github.com/google/jsonschema-go

We will need to:

- [x] move jsonschema issues/pr's to that repo
- [x] tag a version
- [x] update the go-sdk to depend on that module

andig:
Commented on naming in https://github.com/google/jsonschema-go/issues/5

PR #134, opened 2025-07-15
Checking extra tags in schema structs

S3B4SZ17:
Currently, we only support adding the `jsonschema` tag that will be mapped to the description of the schema. Adding support for more tags that will improve validation of the expected data from the client.

- Adding an option to support extra tags in the structs
  - Current list supports the ones that could be needed the most: `["default", "minimum", "maximum", "examples", "readOnly", "deprecated", "writeOnly"]`

Reference:
[jsonschema-go](https://github.com/swaggest/jsonschema-go?tab=readme-ov-file)

Related to #28

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ func main() {
 	// Call a tool on the server.
 	params := &mcp.CallToolParams{
 		Name:      "greet",
-		Arguments: map[string]any{"name": "you"},
+		Arguments: map[string]any{"name": "you", "age": 24, "funkos": []string{"Batman", "Superman"}},
 	}
 	res, err := session.CallTool(ctx, params)
 	if err != nil {
@@ -108,13 +108,15 @@ import (
 	"github.com/modelcontextprotocol/go-sdk/mcp"
 )
 
-type HiParams struct {
-	Name string `json:"name" jsonschema:"the name of the person to greet"`
+type HiArgs struct {
+	Name   string   `json:"name" jsonschema:"the name to say hi to"`
+	Age    int      `json:"age" jsonschema:"the age of the person" minimum:"0" maximum:"120" default:"20"`
+	Funkos []string `json:"funkos" jsonschema:"the funkos the person has" minimum:"1" examples:"[\"Batman\", \"Superman\"]" default:"[\"Batman\"]"`
 }
 
 func SayHi(ctx context.Context, cc *mcp.ServerSession, params *mcp.CallToolParamsFor[HiParams]) (*mcp.CallToolResultFor[any], error) {
 	return &mcp.CallToolResultFor[any]{
-		Content: []mcp.Content{&mcp.TextContent{Text: "Hi " + params.Arguments.Name}},
+		&mcp.TextContent{Text: fmt.Sprintf("Hi %s! You are %d years old and have %v funkos", params.Arguments.Name, params.Arguments.Age, params.Arguments.Funkos)},
 	}, nil
 }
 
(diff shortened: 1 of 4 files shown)


jba:
I am sorry you did all this work speculatively. I want to refer you to this section of [CONTRIBUTING.md](https://github.com/modelcontextprotocol/go-sdk/blob/main/CONTRIBUTING.md#contributing-code):

> For issues not labeled 'Help Wanted', it is recommended that you ask (and wait for confirmation) on the issue before contributing, to avoid duplication of effort or wasted work. For nontrivial changes that don't relate to an existing issue, please file an issue first.

If you had filed an issues, we could have told you there that this is not a feature we want to support. If we did anything like this, it would look like
```
Field Type `jsonschema:"minimum=1,the description"`
```
(That's why there's a restriction on descriptions not starting with `WORD=`.) But we aren't even sure about doing that at the moment.

If you need to set these fields on a schema, use `jsonschema.For` first, then set the schema fields directly.

mslacken:
@jba Is there any example how to correctly use `jsonschema.For`?
I have absolutely have no clue, how I would create a enum like `[]string`  for this, so that I can tell the client the valid parameters.
Btw. I am completely in favor in encoding the valid parameters in a `jsonschema` instead of using `ToolOption`, although this was a breaking change.

S3B4SZ17:
> @jba Is there any example how to correctly use `jsonschema.For`? I have absolutely have no clue, how I would create a enum like `[]string` for this, so that I can tell the client the valid parameters. Btw. I am completely in favor in encoding the valid parameters in a `jsonschema` instead of using `ToolOption`, although this was a breaking change.

The current solution is to define something like this:

```go
func GetExampleSchema() (*jsonschema.Schema, error) {
	schema, err := jsonschema.For[GetExample]()
	if err != nil {
		return nil, err
	}
	var minimum float64 = 1.0

	schema.Properties["scopeHours"].Description = "Number of hours to look back for events"
	schema.Properties["scopeHours"].Default = json.RawMessage(`24`)
	schema.Properties["scopeHours"].Minimum = &minimum
	schema.Properties["cursor"].Description = "Cursor for pagination"
        ...
	return schema, nil
}
...
func addTools() error {
...
   schema, err := eventsfeedsvc.GetExampleSchema()
    if err != nil {
	    return fmt.Errorf("failed to get example schema: %w", err)
    }
    mcp.AddTool(mcpServer, &mcp.Tool{Name: "get_example",
	    Description: "Description of get examples",
	    InputSchema: schema},
	    examplestools.GetExampleTool,
    )
    }
...
return nil
```
I would say it is cumbersome, and that is why I was thinking of implementing the helper handlers for the tags that can be inferred directly from the struct definition.

Another point is if you are using another SDK client to make your calls, the expected structs are already defined and could already have the needed tags instead of defining them like above.

@jba should I go ahead and open an issue for this proposal?

jba:
@S3B4SZ17 Yes, open an issue.

jba:
```
        var minimum float64 = 1.0
	schema.Properties["scopeHours"].Description = "Number of hours to look back for events"
	schema.Properties["scopeHours"].Default = json.RawMessage(`24`)
	schema.Properties["scopeHours"].Minimum = &minimum
	schema.Properties["cursor"].Description = "Cursor for pagination"
```
You could encode the two descriptions on the struct definition by using "jsonschema" tags.

You could write the minimum line as
``` 
schema.Properties["scopeHours"].Minimum = jsonschema.Ptr(1.0)
```
So not too cumbersome, I think. That has to be weighed against the complexity of more complicated struct tags.

S3B4SZ17:
Thanks for the suggestions @jba I will do it like that for now for our structs.

Created the issue to discuss more approaches.

Issue #23, opened 2025-06-24
Resource subscriptions support


findleyr:
Add support for Resource Subscriptions.
https://modelcontextprotocol.io/specification/2025-03-26/server/resources#subscriptions

Design:
https://github.com/modelcontextprotocol/go-sdk/blob/main/design/design.md#subscriptions

samthanawalla:
Almost have a CL ready for this, I'll assign to me

PR #138, opened 2025-07-15
mcp: add resource subscriptions

samthanawalla:
This CL adds the ability for clients to subscribe and receive updates for resources as described in https://modelcontextprotocol.io/specification/2025-06-18/server/resources#subscriptions

Fixes: #23

### DIFF
--- a/mcp/server.go
+++ b/mcp/server.go
@@ -13,6 +13,7 @@ import (
 	"fmt"
 	"iter"
 	"log"
+	"maps"
 	"net/url"
 	"path/filepath"
 	"slices"
@@ -43,7 +44,7 @@ type Server struct {
 	sessions                []*ServerSession
 	sendingMethodHandler_   MethodHandler[*ServerSession]
 	receivingMethodHandler_ MethodHandler[*ServerSession]
-	resourceSubscriptions   map[string]map[string]bool // uri -> session ID -> bool
+	resourceSubscriptions   map[string]map[*ServerSession]bool // uri -> session -> bool
 }
 
 // ServerOptions is used to configure behavior of the server.
@@ -109,7 +110,7 @@ func NewServer(impl *Implementation, opts *ServerOptions) *Server {
 		resourceTemplates:       newFeatureSet(func(t *serverResourceTemplate) string { return t.resourceTemplate.URITemplate }),
 		sendingMethodHandler_:   defaultSendingMethodHandler[*ServerSession],
 		receivingMethodHandler_: defaultReceivingMethodHandler[*ServerSession],
-		resourceSubscriptions:   make(map[string]map[string]bool),
+		resourceSubscriptions:   make(map[string]map[*ServerSession]bool),
 	}
 }
 
@@ -442,19 +443,13 @@ func fileResourceHandler(dir string) ResourceHandler {
 	}
 }
 
+// ResourceUpdated sends a notification to all clients that have subscribed to the
+// resource specified in params. This method is the primary way for a
+// server author to signal that a resource has changed.
 func (s *Server) ResourceUpdated(ctx context.Context, params *ResourceUpdatedNotificationParams) error {
 	s.mu.Lock()
-	subscribedSessionIDs := s.resourceSubscriptions[params.URI]
-	if len(subscribedSessionIDs) == 0 {
-		s.mu.Unlock()
-		return nil
-	}
-	var sessions []*ServerSession
-	for _, session := range s.sessions {
-		if _, ok := subscribedSessionIDs[session.ID()]; ok {
-			sessions = append(sessions, session)
-		}
-	}
+	subscribedSessions := s.resourceSubscriptions[params.URI]
+	sessions := slices.Collect(maps.Keys(subscribedSessions))
 	s.mu.Unlock()
 	notifySessions(sessions, notificationResourceUpdated, params)
 	return nil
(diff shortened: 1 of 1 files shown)


mcp/server.go
```diff
@@ -224,6 +236,12 @@ func (s *Server) capabilities() *serverCapabilities {
 	if s.resources.len() > 0 || s.resourceTemplates.len() > 0 {
 		caps.Resources = &resourceCapabilities{ListChanged: true}
 	}
+	if s.opts.SubscribeHandler != nil {
```
jba: Hmm.
I think that if there are no resources, it doesn't matter if there's a subscribe handler. There is no resource capability.
It's OK to provide a handler, it just doesn't do anything.

mcp/server.go
```diff
@@ -426,6 +444,55 @@ func fileResourceHandler(dir string) ResourceHandler {
...
+
+	if err := s.opts.UnsubscribeHandler(ctx, params); err != nil {
+		return nil, err
+	}
+
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	uri := params.URI
+	if sessions, ok := s.resourceSubscriptions[uri]; ok {
```
jba: Might as well skip the check and just call DeleteFunc.

mcp/server.go
```diff
@@ -97,6 +108,7 @@ func NewServer(impl *Implementation, opts *ServerOptions) *Server {
 		resourceTemplates:       newFeatureSet(func(t *serverResourceTemplate) string { return t.resourceTemplate.URITemplate }),
 		sendingMethodHandler_:   defaultSendingMethodHandler[*ServerSession],
 		receivingMethodHandler_: defaultReceivingMethodHandler[*ServerSession],
+		resourceSubscriptions:   make(map[string][]*ServerSession),
```
jba: I think this would be a little nicer if it was a map[string]map[string]bool. Adding and deleting would be trivial.

mcp/server.go
jba: (If you prefer to use a slice.)

jba (changes requested):
Nice.

mcp/server.go
```diff
@@ -614,6 +686,8 @@ var serverMethodInfos = map[string]methodInfo{
 	methodListResourceTemplates:  newMethodInfo(serverMethod((*Server).listResourceTemplates)),
 	methodReadResource:           newMethodInfo(serverMethod((*Server).readResource)),
 	methodSetLevel:               newMethodInfo(sessionMethod((*ServerSession).setLevel)),
+	methodSubscribe:              newMethodInfo(serverMethod((*Server).subscribe)),
+	methodUnsubscribe:            newMethodInfo(serverMethod((*Server).unsubscribe)),
```
RoyceLeonD, line 690: Would this ever be run in reality?

jba: These are run on the server side when the client calls Subscribe and Unsubscribe.

mcp/server.go
samthanawalla: Done

mcp/server.go
samthanawalla: Done. Used session ID

mcp/server.go
samthanawalla: Updated to just set false

mcp/server.go
```diff
@@ -426,6 +442,64 @@ func fileResourceHandler(dir string) ResourceHandler {
...
+
+	if err := s.opts.UnsubscribeHandler(ctx, params); err != nil {
+		return nil, err
+	}
+
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	uri := params.URI
+	if subscribedSessionIDs, ok := s.resourceSubscriptions[uri]; ok {
```
jba: Or just delete it. That way the map never has false entries and you can skip the `if !active` test above.

mcp/server.go
```diff
@@ -426,6 +442,64 @@ func fileResourceHandler(dir string) ResourceHandler {
...
 
+func (s *Server) ResourceUpdated(ctx context.Context, params *ResourceUpdatedNotificationParams) error {
+	s.mu.Lock()
+	subscribedSessionIDs := maps.Clone(s.resourceSubscriptions[params.URI])
+	s.mu.Unlock()
+	if len(subscribedSessionIDs) == 0 {
+		return nil
+	}
+	sessions := make([]*ServerSession, 0, len(subscribedSessionIDs))
+	for sessionID, active := range subscribedSessionIDs {
```
jba: I think this can be done more simply. First, keep the lock here. You only need to release it before calling notifySessions.

Then loop over s.sessions (directly, not via the iterator, which makes a copy) and ask if each ID is in the map before appending it to a new slices.

mcp/server.go
```diff
@@ -426,6 +442,64 @@ func fileResourceHandler(dir string) ResourceHandler {
...
 }
 
+func (s *Server) ResourceUpdated(ctx context.Context, params *ResourceUpdatedNotificationParams) error {
+	s.mu.Lock()
+	subscribedSessionIDs := maps.Clone(s.resourceSubscriptions[params.URI])
+	s.mu.Unlock()
+	if len(subscribedSessionIDs) == 0 {
+		return nil
+	}
+	sessions := make([]*ServerSession, 0, len(subscribedSessionIDs))
```
jba: `var sessions []*ServerSession`

The optimization isn't needed in this code: it's not performance-sensitive.

mcp/server.go
jba: Also if you keep the lock you won't need maps.Clone.

mcp/server.go
samthanawalla: Done

mcp/server.go
samthanawalla: Good point. Done

mcp/server.go
samthanawalla: Done

mcp/server.go
```diff
@@ -444,22 +443,18 @@ func fileResourceHandler(dir string) ResourceHandler {
...
-		if !active {
-			continue
-		}
-		for session := range s.Sessions() {
-			if session.ID() == sessionID {
-				sessions = append(sessions, session)
-			}
+	var sessions []*ServerSession
+	for _, session := range s.sessions {
+		if _, ok := subscribedSessionIDs[session.ID()]; ok {
```
jba: It's a bool map, so just `if m[k]`

jba (changes requested):
Thought for a followup: what if the inner map was `map[*ServerSession]bool`?
Then we wouldn't need the loop to collect the sessions, the map keys would _be_ the sessions.

mcp/server.go
samthanawalla: Done

samthanawalla: Changed to map[*ServerSession]bool

jba (approved):

Issue #136, opened 2025-07-15
Schema generator customisation


bc-tayvs:
**Is your feature request related to a problem? Please describe.**
Provide a way to customize the schema generator.

Possible customizations include:
- Setting a default `additionalProperties` schema section.
- Adding a schema registry to provide manually created schemas for interfaces and structures with custom scalar types.

Currently, schemas generated for structures with interfaces cause schema-related errors on the MCP client side (e.g., inspector, cursor).

**Describe the solution you'd like**
`infer.go` file contains public method:
 ```go
func For[T any]() (*Schema, error)
```

Introduce a new method that accepts customization options. For example:
```go
type CustomisationOptions struct {
  additionalProperties *Schema
  schemaRegistry map[reflect.Type]*Schema
}

// Used for regular 'For' method
var defaultCustomisationOptions := CustomisationOptions {falseSchema(), make(map[reflect.Type]*Schema)}

func CustomisedFor[T any](options CustomisationOptions) (*Schema, error)
```

**Describe alternatives you've considered**
Helper methods to traverse generated schema and update it

**Additional context**
In our project, we use [genqlient](https://github.com/Khan/genqlient) to generate methods and models based on a GraphQL schema and requests. GraphQL interfaces are represented in Go as `struct` types containing `interface` fields.

When generating schemas from such models, we encounter schemas with objects that lack fields and include `additionalProperties: {not: {}}`. This causes `structuredContent` validation to fail.

Allowing customization of the `additionalProperties` behavior would resolve these client-side validation errors.

jba:
Could you adjust the schemas that you get after calling `For[T]`? It seems like you could write your own function to do that, without adding to the jsonschema package.

bc-tayvs:
> Could you adjust the schemas that you get after calling For[T]?

Yes, I believe it's possible. Though I think `CustomisationOptions` might be a starting point of making generator more extendable and customisable, allowing to cover more complicated use cases.

jba:
Although you're right that we could cover more cases, we also want to keep things as simple as possible.

bc-tayvs:
**Update**

> Could you adjust the schemas that you get after calling For[T]

Tried possible implementation that does not required changes in `For[T]` function. Encountered problem of type erasing. In a generator we know a name of a struct/interface and we can provide modification for all type occurrences. In generated schema we have a field with type `object` (more tricky for arrays).

[...]

PR #139, opened 2025-07-15
Introduce json schema generator customisation

bc-tayvs:
Introduce json schema generator customisation via GeneratorOptions

For https://github.com/modelcontextprotocol/go-sdk/issues/136

### DIFF
--- a/jsonschema/infer.go
+++ b/jsonschema/infer.go
@@ -14,6 +14,23 @@ import (
 	"github.com/modelcontextprotocol/go-sdk/internal/util"
 )
 
+// GeneratorOptions contains options for the schema generator.
+// It allows defining custom AdditionalProperties for a specific type.
+// Also, SchemaRegistry can be used to provide pre-defined schemas for specific types (e.g., struct, interfaces)
+type GeneratorOptions struct {
+	AdditionalProperties func(reflect.Type) *Schema // input is type name
+	SchemaRegistry       map[reflect.Type]*Schema
+}
+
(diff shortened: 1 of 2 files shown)


jba:
As described in the [CONTRIBUTING doc](https://github.com/modelcontextprotocol/go-sdk/blob/main/CONTRIBUTING.md#contributing-code), you should wait until we approve an issue before submitting code.
We'll look at this PR once we've agreed that #136 is the right approach.

jsonschema/infer.go
```diff
@@ -14,6 +14,23 @@ import (
 	"github.com/modelcontextprotocol/go-sdk/internal/util"
 )
 
+// GeneratorOptions contains options for the schema generator.
+// It allows defining custom AdditionalProperties for a specific type.
+// Also, SchemaRegistry can be used to provide pre-defined schemas for specific types (e.g., struct, interfaces)
+type GeneratorOptions struct {
+	AdditionalProperties func(reflect.Type) *Schema // input is type name
```
jba, line 21: #185 implements the schema registry. Can you explain why AdditionalProperties is necessary also? It seems like SchemaRegistry is all that's needed.

bc-tayvs: Main issue was that interface field were not added to parent object. Because of this, MCP clients were reporting that structured response is not comply with schema. Registry should fix it.

BUT if there are another corner cases, additionalProperties should allow to resolve schema-related complains. At this point, I have implemented additionalProperties override after schema generation in our project

PR #145, opened 2025-07-17
mcp: export Event

jba:
Export the Event type in preparation for providing user-definable storage for resumable streams.

Also, move event code to a separate file.

This PR is just renaming and code motion.


RoyceLeonD (approved):
Just for my own understanding, we pulled out login in sse.go, brought it into the exported Event and then updated it to use the new event to handle write and response. The SSE file itself is also doing things with client connection but we don't care about that in this PR, and we just want to pull out the event and streamable response handlers.

jba:
@RoyceLeonD yes.
Don't depend on any of this yet. It's possible it won't be v0.3.0. Stick to v0.2.0 to avoid breakage.

PR #147, opened 2025-07-18
jsonschema: inference ignores invalid types

jba:
Add ForLax[T], which ignores invalid types in schema inference instead of returning an error.

This allows additional customization of a schema after inference does what it can.

For example, an interface type where all the possible implementations are known can be described with "oneof".

For #136.

### DIFF
--- a/jsonschema/infer.go
+++ b/jsonschema/infer.go
@@ -30,8 +30,10 @@ import (
 //     struct field JSON name. Fields that are marked "omitempty" are
 //     considered optional; all other fields become required properties.
 //
-// For returns an error if t contains (possibly recursively) any of the following Go
-// types, as they are incompatible with the JSON schema spec.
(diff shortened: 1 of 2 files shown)


jsonschema/infer.go
```diff
@@ -47,15 +47,30 @@ import (
...
 		return nil, fmt.Errorf("For[%T](): %w", z, err)
 	}
 	return s, nil
 }
 
-func forType(t reflect.Type, seen map[reflect.Type]bool) (*Schema, error) {
+// ForLax behaves like [For], except that it ignores struct fields with invalid types instead of
+// returning an error. That allows callers to adjust the resulting schema using custom knowledge.
+// For example, an interface type where all the possible implementations are known
+// can be described with "oneof".
```
RoyceLeonD, line 61: I like choices
freedom

PR #152, opened 2025-07-21
mcp: add EventStore

jba:
Introduce an EventStore interface to store events for resumable streams.

Provide an in-memory implementation.

Still to do: connect to streaming transports.

For #10

### DIFF
--- a/mcp/event.go
+++ b/mcp/event.go
@@ -22,6 +22,10 @@ import (
 	"sync"
 )
 
+// If true, MemoryEventStore will do frequent validation to check invariants, slowing it down.
+// Remove when we're confident in the code.
+const validateMemoryEventStore = true
+
 // An Event is a server-sent event.
 // See https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events#fields.
 type Event struct {
@@ -331,6 +335,9 @@ func (s *MemoryEventStore) StreamClosed(_ context.Context, sessionID string, str
 	dl := sm[streamID]
 	s.nBytes -= dl.size
 	delete(sm, streamID)
+	if len(sm) == 0 {
+		delete(s.store, sessionID)
+	}
 	s.validate()
 	return nil
 }
@@ -350,9 +357,6 @@ func (s *MemoryEventStore) SessionClosed(_ context.Context, sessionID string) er
 // purge removes data until no more than s.maxBytes bytes are in use.
 // It must be called with s.mu held.
 func (s *MemoryEventStore) purge() {
-	if s.maxBytes <= 0 {
-		panic("non-positive maxBytes")
-	}
 	// Remove the first element of every dataList until below the max.
 	for s.nBytes > s.maxBytes {
 		changed := false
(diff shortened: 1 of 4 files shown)


mcp/event.go
```diff
@@ -136,3 +140,278 @@ func scanEvents(r io.Reader) iter.Seq2[Event, error] {
...
+	}
+}
+
+// StreamClosed implements [EventStore.StreamClosed].
+func (s *MemoryEventStore) StreamClosed(_ context.Context, sessionID string, streamID StreamID) error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	sm := s.store[sessionID]
+	dl := sm[streamID]
```
samthanawalla, line 339: we should validate session ID exists otherwise this could panic

mcp/event.go
```diff
@@ -136,3 +140,278 @@ func scanEvents(r io.Reader) iter.Seq2[Event, error] {
...
+	s.nBytes -= dl.size
+	delete(sm, streamID)
+	s.validate()
```
samthanawalla, line 345: we could check if the session stream map is empty and clean up the session

```
if len(sm) == 0 {
		delete(s.store, sessionID)
}
```

mcp/event.go
```diff
@@ -136,3 +140,278 @@ func scanEvents(r io.Reader) iter.Seq2[Event, error] {
...
+		if !changed {
+			panic("no progress during purge")
+		}
+	}
+	s.validate()
+}
+
+// validate checks that the store's data structures are valid.
+// It must be called with s.mu held.
+func (s *MemoryEventStore) validate() {
```
samthanawalla, line 387: Should we move this into a test? This could introduce performance overhead

mcp/event.go
```diff
@@ -136,3 +140,278 @@ func scanEvents(r io.Reader) iter.Seq2[Event, error] {
...
+	}
+	delete(s.store, sessionID)
+	s.validate()
+	return nil
+}
+
+// purge removes data until no more than s.maxBytes bytes are in use.
+// It must be called with s.mu held.
+func (s *MemoryEventStore) purge() {
+	if s.maxBytes <= 0 {
```
samthanawalla: do we need to check maxBytes here when we already checked when SetMaxBytes was called?

samthanawalla (changes requested):

mcp/event.go
jba: I think we should panic: this would be a bug in the implementation.

mcp/event.go
jba: Done.

mcp/event.go
jba: true. deleted.

mcp/event.go
jba: I'd like to leave it as it is for a while to catch bugs, because I've seen bugs with this sort of thing before and it's easy to check. Eventually we can remove it.

I put it behind a const to make it easy to turn off.

mcp/event.go
samthanawalla: True then we should call panic("session ID does not exist") or something similar so it's easier to debug if that happens

jba: Done.

mcp/event.go
samthanawalla: SGTM, can you add a comment saying this? Things get forgotten 😅

samthanawalla (changes requested):

mcp/event.go
samthanawalla: Ah just saw the comment behind the const. nvmd

PR #156, opened 2025-07-23
mcp: incorporate EventStore into StreamableServerTransport

jba:
Use the EventStore interface to implement resumption on the server side.


mcp/streamable.go
```diff
@@ -301,16 +312,15 @@ func (t *StreamableServerTransport) ServeHTTP(w http.ResponseWriter, req *http.R
 
 func (t *StreamableServerTransport) serveGET(w http.ResponseWriter, req *http.Request) {
 	// connID 0 corresponds to the default GET request.
-	id, nextIdx := StreamID(0), 0
+	id, lastIdx := StreamID(0), -1
```
samthanawalla: Can you document why we set lastIdx to -1?

samthanawalla (approved):

PR #157, opened 2025-07-23
mcp: panic if AddTool is given a nil InputSchema

jba:
The spec requires an input schema for every tool.


mcp/server.go
```diff
@@ -123,8 +122,7 @@ func (s *Server) RemovePrompts(names ...string) {
 func (s *Server) AddTool(t *Tool, h ToolHandler) {
 	// TODO(jba): This is a breaking behavior change. Add before v0.2.0?
 	if t.InputSchema == nil {
-		log.Printf("mcp: tool %q has a nil input schema. This will panic in a future release.", t.Name)
-		// panic(fmt.Sprintf("adding tool %q: nil input schema", t.Name))
+		panic(fmt.Sprintf("adding tool %q: nil input schema", t.Name))
```
samthanawalla, line 145: Why not try to use &jsonschema.Schema{} instead?

If 99% of the time a person who supplied nil would supply &jsonschema.Schema{} then we should make that decision for them to simplify the UI.

jba: That is fine if they really mean that the tool has no inputs. But that is extremely unlikely. (Maybe a tool that returns the current time? Or a random number? But basically nothing else.)

More likely, there really are arguments, and the tool author forgot to write a schema. If we used the empty schema, then anything that the LLM passes to the tool will be validated. If that's what they really want, they can pass an empty schema themselves. But if we didn't panic, they might not detect a problem until much later, when the LLM passed bad input.

If we put this in, we can always take it out if we get enough complaints. But we can't go the other way: we can't add a panic if there wasn't one before. So I think now is the time to do this, and see if we get pushback.

samthanawalla: Ack, can you summarize this justification as a comment?

Also can you update the example in the README to provide a InputSchema to AddTool?

jba: Added a comment, but no input schema is needed for the top-level function AddTool: it infers it. It's only needed for the method.

PR #159, opened 2025-07-23
mcp: support stateless connections for streamable

### LINKED ISSUE #148: Proposal: Session recovery / shared session storage for distributed MCP env

Currently MCP SDK is using in-memory storage for sessions https://github.com/modelcontextprotocol/go-sdk/blob/dd49b7039b69498b242fe0ffea5016b734e6181c/mcp/streamable.go#L102 which become potentially a tricky place for session recovery scenarios in particular in distributed prod-like environment with multiple instances of MCP servers up and running.
 
As the result in the scenario when a client has proper session id but for some reason has lost or patially lost the connection in the downstream of an entrypoint  (client -> entripoint -> [no connection] downstream), the request may be routed through load balancer to various instances of MCP which may not have session context in memory. 

Additionally within the deployment in scalable multi instance enveronment session context due to being in mem only often also can be lost without a way to greacefully recover.

Potentially introduce shared session storage (e.g redis) for being able to share the session state for graceful recovery within connection loss / re-deployments. 

EDIT(@findleyr): The current concrete proposal is in https://github.com/modelcontextprotocol/go-sdk/issues/148#issuecomment-3156607256.

ln-12:
When the client lost the connection to the mcp server or the mcp server runs on distributed nodes (e.g. behind a loadbalancer), it can be desired to run the server in stateless mode. When doing so, the Mcp-Session-Id should not generated or verified to allow the execution of methods without prior initialization.

Or would it be better to still generate the session id and send the `Mcp-Session-Id` but just not validate it?

The [Rust sdk](https://github.com/modelcontextprotocol/rust-sdk/issues/212), [mcp-go](https://github.com/mark3labs/mcp-go/issues/370) and the [typescript sdk](https://github.com/modelcontextprotocol/typescript-sdk/issues/330) also support it for example. Additionally, this mode is mentioned in the [typescript documentation](https://github.com/modelcontextprotocol/typescript-sdk/tree/main/src/examples#stateless-mode) although the official documentation does not seem to contain any information about it. 

Fixes #148


ln-12:
One issue I found so far is that for example sending log messages from the server to the client is not working as the session id is needed for that. So maybe it would be better to still generate and send the session id but simply not validate it like I tested on [this branch](https://github.com/ln-12/go-sdk/compare/main...ln-12:go-sdk:stateless-http-fix?expand=1).

Please let me know what you think about it.

findleyr:
Thanks for this! I implemented stateless mode slightly differently in #277, because in the context of #148 we're talking about APIs for loading and initializing sessions, for which a 'stateless session' is just a special case. I also went the route of the typescript SDK and triggered stateless mode whenever a new user-provided GetSessionID function returns the empty string.

However, I used your test, and your logic was very helpful. I added you as a Co-Author on that commit.

ln-12:
Thanks for implementing it, I will then close this one.

Issue #162, opened 2025-07-23
KeepAlive Ping message send null as params


kzwang:
**Describe the bug**
https://github.com/modelcontextprotocol/go-sdk/blob/main/mcp/shared.go#L354
When keep alive send the ping message, it uses nil as params, when connecting to the server using MCP Inspector or Claude Code, it got error

```
ZodError: [
  {
    "code": "invalid_union",
    "unionErrors": [
      {
        "issues": [
          {
            "code": "invalid_type",
            "expected": "object",
            "received": "null",
            "path": [
              "params"
            ],
            "message": "Expected object, received null"
          }
        ],
        "name": "ZodError"
      },
      {
        "issues": [
          {
            "code": "invalid_type",
            "expected": "object",
            "received": "null",
            "path": [
              "params"
            ],
            "message": "Expected object, received null"
          },
          {
            "code": "unrecognized_keys",
            "keys": [
              "id"
            ],
            "path": [],
            "message": "Unrecognized key(s) in object: 'id'"
          }
        ],
        "name": "ZodError"
      },
      {
        "issues": [
          {
            "code": "invalid_type",
            "expected": "object",
            "received": "undefined",
            "path": [
              "result"
            ],
            "message": "Required"
          },
          {
            "code": "unrecognized_keys",
            "keys": [
              "method",
              "params"
            ],
            "path": [],
            "message": "Unrecognized key(s) in object: 'method', 'params'"
          }
        ],
        "name": "ZodError"
      },
      {
        "issues": [
          {
            "code": "invalid_type",
            "expected": "object",
            "received": "undefined",
            "path": [
              "error"
            ],
            "message": "Required"
          },
          {
            "code": "unrecognized_keys",
            "keys": [
              "method",
              "params"
            ],
            "path": [],
            "message": "Unrecognized key(s) in object: 'method', 'params'"
          }
        ],
        "name": "ZodError"
      }
    ],
    "path": [],
    "message": "Invalid input"
  }
]
```

**To Reproduce**
1. Add `KeepAlive: 10 * time.Second` in `ServerOptions`
2. Connect to the MCP server with MCP Inspector using Streamable HTTP and check MCP Inspector logs

PR #163, opened 2025-07-23
mcp/server: fix nil PingParams in server.Ping

samthanawalla:
Use OrZero in case PingParams is nil to avoid typed nil.

Fixes: #162


jba (approved):
Thanks for grabbing this.

Issue #165, opened 2025-07-23
Would you accept an examples/ addition with a practical MCP server?


wkoszek:
**Describe the bug**

Lack of a practical server that works with a popular MCP client like Claude Code makes this repo hard to understand. While STDIO might be a popular transport for tech users, I'm working on a product that must support  HTTP/SSE

**Additional context**

Would you accept an example that I'd write that shows how to use HTTP/SSE ?

jba:
Possibly.

If it imports anything that the main module does not, then it would need to be in its own module, like examples/rate-limiting.

It also shouldn't be too much code. For an example of something that's too much, see #70. It shouldn't need makefiles, dockerfiles, shell scripts, etc.

With those caveats, go ahead. If you're not sure, provide more detail.

Looking forward to seeing it!

wkoszek:
@jba https://github.com/modelcontextprotocol/go-sdk/pull/168 <- was thinking about something like this.

jba:
I like it. It is more comprehensive than our "hello" example. Additional comments on the PR.

findleyr:
This was fixed by https://github.com/modelcontextprotocol/go-sdk/pull/168.

PR #168, opened 2025-07-23
mcp/examples: HTTP server example with a simple client built-in

wkoszek:
I'm adding a practical example of an MCP with HTTP streaming: both client and server. Those are useful for testing real-world applications.

### DIFF
--- a/examples/http/main.go
+++ b/examples/http/main.go
@@ -17,8 +17,8 @@ import (
 )
 
 var (
-	host = flag.String("host", "localhost", "host to connect to/listen on")
-	port = flag.Int("port", 8000, "port number to connect to/listen on")
+	host  = flag.String("host", "localhost", "host to connect to/listen on")
+	port  = flag.Int("port", 8000, "port number to connect to/listen on")
 	proto = flag.String("proto", "http", "if set, use as proto:// part of URL (ignored for server)")
 )
 
@@ -46,7 +46,7 @@ func main() {
 
 	switch mode {
 	case "server":
-		if (*proto != "http") {
+		if *proto != "http" {
 			log.Fatalf("Server only works with 'http' (you passed proto=%s)", *proto)
 		}
 		runServer(fmt.Sprintf("%s:%d", *host, *port))
@@ -57,7 +57,6 @@ func main() {
 		flag.Usage()
 	}
 }
-
 
 // GetTimeParams defines the parameters for the cityTime tool.
 type GetTimeParams struct {


jba:
This fixes #165.

examples/http/README.md
```diff
@@ -0,0 +1,68 @@
+# MCP HTTP Example
+
+This example demonstrates how to use the Model Context Protocol (MCP) over HTTP using the streamable transport. It includes both a server and client implementation.
+
+## Overview
+
+The example implements:
+- A server that provides a `get_time` tool
```
jba: Let's use idiomatic Go naming throughout (even though technically this isn't language-specific).
So `getTime`.
Or even better (since we don't like "get"): cityTime

examples/http/main.go
```diff
@@ -0,0 +1,235 @@
...
+		fmt.Fprintf(os.Stderr, "\nExamples:\n")
+		fmt.Fprintf(os.Stderr, "  Run as server:  %s -server\n", os.Args[0])
+		fmt.Fprintf(os.Stderr, "  Run as client:  %s -client\n", os.Args[0])
+		fmt.Fprintf(os.Stderr, "  Custom host/port: %s -server -host 0.0.0.0 -port 9090\n", os.Args[0])
+		os.Exit(1)
+	}
+
+	flag.Parse()
+
+	if (*serverMode && *clientMode) || (!*serverMode && !*clientMode) {
```
jba: Since you have to pick one or the other, this shouldn't be a flag. The first command-line arg should be "client" or "server".

examples/http/main.go
```diff
@@ -0,0 +1,235 @@
...
+	// Format the response
+	cityNames := map[string]string{
+		"nyc":    "New York City",
+		"sf":     "San Francisco",
+		"boston": "Boston",
+	}
+
+	response := fmt.Sprintf("The current time in %s is %s", 
+		cityNames[city], 
+		now.Format("3:04:05 PM MST on Monday, January 2, 2006"))
```
jba: Use a standard time format (from the time package) so we don't have people wondering about the obscure format notation.

examples/http/main.go
```diff
@@ -0,0 +1,235 @@
...
+		Description: "Get the current time in NYC, San Francisco, or Boston",
+	}, getTime)
+
+	// Create the streamable HTTP handler
+	handler := mcp.NewStreamableHTTPHandler(func(req *http.Request) *mcp.Server {
+		return server
+	}, nil)
+
+	// Wrap the handler with logging middleware
+	loggingHandler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
jba: This is a nice logging wrapper, but something of a distraction from the main point. Please put it, and the responseWriter type, in a separate file. All we should see here is something like
```
handler = loggingWrapper(handler)
```

examples/http/main.go
```diff
@@ -0,0 +1,235 @@
...
+	session, err := client.Connect(ctx, transport)
+	if err != nil {
+		log.Fatalf("Failed to connect: %v", err)
+	}
+	defer session.Close()
+
+	log.Printf("Connected to server (session ID: %s)", session.ID())
+
+	// First, list available tools
+	log.Println("\nListing available tools...")
```
jba: Remove initial newlines from log messages.

examples/http/main.go
```diff
@@ -0,0 +1,235 @@
...
+	toolsResult, err := session.ListTools(ctx, &mcp.ListToolsParams{})
```
jba: Pass nil for the params.

examples/http/main.go
```diff
@@ -0,0 +1,235 @@
...
+
+	// Call the get_time tool for each city
+	cities := []string{"nyc", "sf", "boston"}
+	
+	log.Println("\nGetting time for each city...")
+	for _, city := range cities {
+		// Call the tool
+		result, err := session.CallTool(ctx, &mcp.CallToolParams{
+			Name: "get_time",
+			Arguments: map[string]interface{}{
```
jba: s/interface{}/any/

wkoszek:
@jba I think I addressed everything here.

examples/http/logging_middleware.go
```diff
@@ -0,0 +1,51 @@
...
+
+package main
+
+import (
+	"log"
+	"net/http"
+	"time"
+)
+
+// responseWriter wraps http.ResponseWriter to capture the status code
```
jba: end sentence with period (unlike this one)

examples/http/main.go
```diff
@@ -18,40 +18,45 @@ import (
...
+		fmt.Fprintf(os.Stderr, "  Run as client:  %s client\n", os.Args[0])
+		fmt.Fprintf(os.Stderr, "  Custom host/port: %s server -host 0.0.0.0 -port 9090\n", os.Args[0])
 		os.Exit(1)
 	}
 
-	flag.Parse()
-
-	if (*serverMode && *clientMode) || (!*serverMode && !*clientMode) {
-		fmt.Fprintf(os.Stderr, "Error: Must specify exactly one of -server or -client\n\n")
+	// Check if we have at least one argument
```
jba: period

examples/http/main.go
```diff
@@ -18,40 +18,45 @@ import (
...
+	if len(os.Args) < 2 {
```
jba: Call flag.Parse first, then access args as flag.Arg(0), etc. Use flag.NArg() for the count.

examples/http/main.go
```diff
@@ -132,47 +126,21 @@ func runServer(host, port string) {
...
-		duration := time.Since(start)
-		log.Printf("[RESPONSE] %s | %s | %s %s | Status: %d | Duration: %v",
-			time.Now().Format("2006-01-02 15:04:05.000"),
-			r.RemoteAddr,
-			r.Method,
-			r.URL.Path,
-			wrapped.statusCode,
-			duration)
-	})
+	handlerWithLogging := loggingHandler(handler)
```
jba, line 131: `handler = loggingHandler(handler)`

examples/http/main.go
```diff
@@ -103,27 +108,16 @@ func getTime(ctx context.Context, ss *mcp.ServerSession, params *mcp.CallToolPar
...
-
 func runServer(host, port string) {
 	// Create an MCP server
 	server := mcp.NewServer(&mcp.Implementation{
 		Name:    "time-server",
 		Version: "1.0.0",
 	}, nil)
 
-	// Add the get_time tool
+	// Add the cityTime tool
```
jba: period

jba: There are more below, I din't mark them all.

examples/http/main.go
wkoszek: @jba That won't work - they are different types.

examples/http/main.go
wkoszek: How about we simplify it even mode and make it  `./http <client|server> [http://localhost:9000]` 

This would be good because with this, I can test the server mode on a VPS with TLS cert issued to Caddy and proxying requests to:

./http server http://localhost:8000

and then I can connect to it with the client:

./http client https://mymcp.koszek.com:443

wkoszek: @jba fixed this part.

examples/http/main.go
```diff
@@ -0,0 +1,203 @@
...
+	"net/url"
+	"os"
+	"time"
+
+	"github.com/modelcontextprotocol/go-sdk/mcp"
+)
+
+func main() {
+	flag.Usage = func() {
+		fmt.Fprintf(os.Stderr, "Usage: %s <client|server> [proto://<host>:<port>]\n\n", os.Args[0])
```
jba:
```
out := CommandLine.Output()
fmt.Fprintf(out, ...)
```

examples/http/main.go
```diff
@@ -0,0 +1,203 @@
...
+		flag.PrintDefaults()
+		fmt.Fprintf(os.Stderr, "\nExamples:\n")
+		fmt.Fprintf(os.Stderr, "  Run as server:  %s server\n", os.Args[0])
+		fmt.Fprintf(os.Stderr, "  Run as client:  %s client\n", os.Args[0])
+		fmt.Fprintf(os.Stderr, "  Custom host/port: %s server https://0.0.0.0:8000\n", os.Args[0])
+		os.Exit(1)
+	}
+
+	if len(os.Args) < 2 {
+		fmt.Fprintf(os.Stderr, "Error: Must specify 'client' or 'server' as first argument\n\n")
```
jba: ditto

examples/http/main.go
```diff
@@ -0,0 +1,203 @@
...
+	}, getTime)
+
+	// Create the streamable HTTP handler.
+	handler := mcp.NewStreamableHTTPHandler(func(req *http.Request) *mcp.Server {
+		return server
+	}, nil)
+
+	handlerWithLogging := loggingHandler(handler)
+
+	laddr := fmt.Sprintf("%s:%s", url.Hostname(), url.Port())
```
jba: Allowing proto is deceptive: I can't create an HTTPS server if I write `https://example.com:8080`. 
I think you should go back to host and port flags. The host will essentially always be localhost, so that should definitely be a flag. If you want to have port be an arg, I'm OK with that, but I'm also fine with a flag and a default value.

jba (changes requested):

examples/http/main.go
wkoszek: @jba take a look -- I fixed the rest of stuff, and I want to keep the proto, because without a way to set it to https:// for the client, I can't really test the remote side, if it's behind the load balancer.

examples/http/main.go
```diff
@@ -10,53 +10,55 @@ import (
...
 	"os"
 	"time"
 
 	"github.com/modelcontextprotocol/go-sdk/mcp"
 )
 
+var (
+	host = flag.String("host", "localhost", "host to connect to/listen on")
+	port = flag.Int("port", 8000, "port number to connect to/listen on")
+	proto = flag.String("proto", "http", "if set, use as proto:// part of URL (ignored for server)")
```
jba, line 22: If the server is always http, what's the point of letting the client be something different?

wkoszek: @jba The very good use case of this example is testing with real-world MCP client. In my testing, I take this code and put it on the remote VPS server behind the HTTPS load-balancer, and then I can use our client to connect to this https. And then I can use real-world client like Claude.

In other words: server is listening on 8080 HTTP but the HTTPS is provided by Caddy. And then I can test different clients with our server code

jba (approved):

jba:
Address failing tests.

wkoszek:
@jba this is ready

PR #171, opened 2025-07-24
mcp: add stream type

jba:
Consolidate several maps into a single struct.

Simplifies the code, for the most part.

### DIFF
--- a/mcp/streamable.go
+++ b/mcp/streamable.go
@@ -255,7 +255,7 @@ type stream struct {
 
 	// streamRequests is the set of unanswered incoming RPCs for the stream.
 	//
-	// Lifecycle: requests values persist as until the requests have been
+	// Lifecycle: requests values persist until the requests have been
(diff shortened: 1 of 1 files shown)


mcp/streamable.go
```diff
@@ -579,30 +598,31 @@ func (t *StreamableServerTransport) Write(ctx context.Context, msg jsonrpc.Messa
 	t.mu.Lock()
 	defer t.mu.Unlock()
 	if t.isDone {
-		return fmt.Errorf("session is closed") // TODO: should this be EOF?
+		return errors.New("session is closed") // TODO: should this be EOF?
```
samthanawalla: I think we can remove the TODO comment, it should be an error to write to a closed session. EOF would suggest it succeeded.

jba: done

samthanawalla (approved):

Issue #166, opened 2025-07-23
Expose server capabilities as client


spachava753:
**Is your feature request related to a problem? Please describe.**
Currently, it looks like there is no way for a `ClientSession` to return the server's capabilities, leading to users of the `ClientSession` guessing whether a certain capability is supported by the server. For example, the [DeepWiki MCP server](https://docs.devin.ai/work-with-devin/deepwiki-mcp) only supports the tools capability, and trying to use the `List Prompts` method returns an error. 

**Describe the solution you'd like**
Instead, I think a method on `ClientSession` to expose the server's capabilities will allow users to check if method is supported before calling it, potentially saving the network call in cases of connecting to HTTP servers. We already store the server's capabilities in the initialize result:
https://github.com/modelcontextprotocol/go-sdk/blob/8dd9a819bb283849de91c839f6f1734b86a9fc1b/mcp/client.go#L158-L164
https://github.com/modelcontextprotocol/go-sdk/blob/8dd9a819bb283849de91c839f6f1734b86a9fc1b/mcp/protocol.go#L304-L322

**Describe alternatives you've considered**
A possible alternative is simply returning an exposed sentinel error, that the user can check with `error.Is` or `error.As`. This alternative can also be implemented _with_ the solution described above, offering both ways to understand a server's capabilities as a client.

**Additional context**

samthanawalla:
I agree that serverCapabilities should be accessible to the client. I think @findleyr added the serverCapabilities struct, maybe he knows why they were unexported originally?

jba:
I'm reluctant to expose all these types, but offhand I don't see another way. @findleyr ?

jba:
I think if we controlled the server, we could return a special error from the ListXXX functions that the client could inspect to see that feature wasn't supported. But we don't—the server could be anything.

jba:
Ugh. Based on looking at some other SDKs, I think we should go further: let the client see the `InitializationResult`. That still requires exporting serverCapabilities and everything it contains.

spachava753:
> Ugh. Based on looking at some other SDKs, I think we should go further: let the client see the `InitializationResult`. That still requires exporting serverCapabilities and everything it contains.

I agree, looking at the `InitializationResult`, it contains some other fields, like `Instructions` and `Protocol Version` that seems useful for the client to access.

[...]

PR #173, opened 2025-07-24
feat(mcp): expose server capabilities to client

spachava753:
This change introduces a new method, `ServerCapabilities()`, on the `ClientSession`. This allows a client to inspect the capabilities of the connected server after the initialization handshake.

Previously, the capability-related structs were unexported, and the client session did not provide a way to access them. This change exports the `ServerCapabilities` type and related structs, making them part of the public API. The `ServerCapabilities()` method returns a deep copy of the capabilities to prevent accidental modification of the session's internal state.

In addition, this commit also introduces a `.gitignore` file to prevent tracking of unwanted local files, such as generated files and IDE-specific configurations.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,27 @@
+### Go.AllowList template
+# Allowlisting gitignore template for GO projects prevents us
+# from adding various unwanted local files, such as generated
+# files, developer configurations or IDE-specific files etc.
+#
+# Recommended: Go.AllowList.gitignore
+
+# Ignore everything
+*
+
+# But not these files...
+!/.gitignore
+
+!*.go
+!go.*
+
+!*.sh
+
+!README.md
+!CONTRIBUTING.md
+!LICENSE
+
+!design/**
+!.github/**
+
+# ...even if they are in subdirectories
+!*/
--- a/mcp/client.go
+++ b/mcp/client.go
@@ -174,6 +174,59 @@ func (cs *ClientSession) ID() string {
 	return cs.mcpConn.SessionID()
 }
 
+// ServerCapabilities returns a copy of the server capabilities obtained during initialization.
+// If the session has not been initialized or capabilities are not available, it returns nil.
+func (cs *ClientSession) ServerCapabilities() *ServerCapabilities {
+	if cs.initializeResult == nil || cs.initializeResult.Capabilities == nil {
+		return nil
+	}
+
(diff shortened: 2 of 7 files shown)


spachava753:
I filed an issue (https://github.com/modelcontextprotocol/go-sdk/issues/166) earlier, and wanted to create a PR that could serve as a possible implementation, if the feature from the issue gets approved to be implemented

.gitignore
```diff
@@ -0,0 +1,27 @@
+### Go.AllowList template
```
jba, line 1: remove this file from PR

mcp/client.go
```diff
@@ -174,6 +174,59 @@ func (cs *ClientSession) ID() string {
 	return cs.mcpConn.SessionID()
 }
 
+// ServerCapabilities returns a copy of the server capabilities obtained during initialization.
+// If the session has not been initialized or capabilities are not available, it returns nil.
+func (cs *ClientSession) ServerCapabilities() *ServerCapabilities {
```
jba, line 179: I wonder if we can just use an exported field. The doc will say not to modify it, but even if someone does, they're only hurting themselves, as far as I can tell. 

Or if we're not sure, let's use the function, but not bother copying until we discover a case where it could matter. But really I think the field suffices.

/cc @samthanawalla @findleyr

mcp/protocol.go
```diff
@@ -907,46 +907,51 @@ type Implementation struct {
 	Version string `json:"version"`
 }
 
+// CompletionCapabilities represents server capabilities for argument autocompletion suggestions.
```
jba, line 910: We say "is" or "are" instead of represents.
Change throughout.

jba (changes requested):

mcp/client.go
samthanawalla: I agree with not copying right now.

I think we should go with the cs.ServerCapabilities() function because cs.InitializeResult.Capabilities is a bit cumbersome/less obvious to use. Plus, like you said, we can add copying to ServerCapabilities later.

spachava753:
@jba @samthanawalla @findleyr I've opened a new PR to address the comments and expose the entire initialized result rather than just the server capabilities via a `ClientSession` method at #228

jba:
Can we close this PR, since #228 is a replacement?

jba:
superseded

PR #176, opened 2025-07-25
mcp: add client-side OAuth flow (preliminary)

jba:
This is a preliminary implementation of OAuth 2.1 for the client.

When a StreamableClientTransport encounters a 401 Unauthorized response
from the server, it initiates the OAuth flow described in thec
authorization section of the MCP spec
(https://modelcontextprotocol.io/specification/2025-06-18/basic/authorization).
On success, the transport obtains an access token which it passes
to all subsequent requests.

Much remains to be done here:

- Dynamic client registration is not implemented. Since it is optional,
  we also need another way of supplying the client ID and secret to
  this code.

- Resource Indicators, as described in section 2.5.1 of the MCP spec.

- There is no way for the user to provide a redirect URL.

- All of this is unexported, so it is available only to our own
  StreamingClientTransport. We should add API so people can use it
  with their own transports.

- And, of course, tests. We should test against fake implementations
  but also, if we can find any, real reference implementations.


jba:
Ignore the first two commits. They should be in separate PRs.

mcp/streamable.go
```diff
@@ -966,7 +1008,7 @@ func (s *streamableClientConn) Close() error {
 				req.Header.Set(protocolVersionHeader, s.protocolVersion)
 			}
 			req.Header.Set(sessionIDHeader, s._sessionID)
-			if _, err := s.client.Do(req); err != nil {
+			if _, err := s.opts.HTTPClient.Do(req); err != nil {
```
samthanawalla: use s.authClient if non-nil?

mcp/streamable.go
```diff
@@ -992,7 +1034,7 @@ func (s *streamableClientConn) establishSSE(lastEventID string) (*http.Response,
 	}
 	req.Header.Set("Accept", "text/event-stream")
 
-	return s.client.Do(req)
+	return s.opts.HTTPClient.Do(req)
```
samthanawalla: use s.authClient if non-nil?

mcp/streamable.go
```diff
@@ -819,28 +829,59 @@ func (s *streamableClientConn) postMessage(ctx context.Context, sessionID string
...
+		if resp != nil && !bodyClosed {
+			resp.Body.Close()
+		}
+	}()
+
+	if resp.StatusCode == http.StatusUnauthorized {
+		if client == s.authClient {
+			return "", errors.New("got StatusUnauthorized when already authorized")
+		}
+		tokenSource, err := doOauth(ctx, resp.Header, s.opts.HTTPClient, s.opts.AuthHandler)
```
samthanawalla: s/s.opts.HttpClient/client ?

wagnerjt:
For testing, Google supports DCR and there is an example of Spotify MCP server to include client id and client secret for the fallback [here](https://github.com/TylerLeonhardt/spotify-remote-mcp). This was used to support the vscode's client (currently in insider) at the moment to do exactly this fallback mechanism when DCR does not exist at the IDP.

RoyceLeonD:
@jba Sorry It took so long for me to be active on this. Will take a deeper look at the code and revert back in the next day or so.

mcp/streamable.go

mcp/auth.go
```diff
@@ -0,0 +1,71 @@
...
+	// or find another way to get the client ID and secret.
+
+	// Get an access token from the auth server.
+	config := &oauth2.Config{
+		ClientID:     "TODO: from registration",
+		ClientSecret: "TODO: from registration",
+		Endpoint: oauth2.Endpoint{
+			AuthURL:  asm.AuthorizationEndpoint,
+			TokenURL: asm.TokenEndpoint,
+			// DeviceAuthURL: "",
```
ramongalate: adding auth style should be necessary depending on the server metadata
https://pkg.go.dev/golang.org/x/oauth2#section-readme:~:text=type-,AuthStyle,-%C2%B6

depending on the AS's token_endpoint_auth_methods_supported.

mcp/auth.go
```diff
@@ -0,0 +1,71 @@
...
+	asm, err := oauthex.GetAuthServerMeta(ctx, authServer, c)
+	if err != nil {
+		return nil, err
+	}
+	// TODO: register the client with the auth server if not registered yet,
+	// or find another way to get the client ID and secret.
+
+	// Get an access token from the auth server.
+	config := &oauth2.Config{
+		ClientID:     "TODO: from registration",
```
ramongalate: I think you should add two options, one for passed in client ids/secrets and one for PKCE for public clients, for a more AUTH2.1 compliant solution.

mcp/auth.go
```diff
@@ -0,0 +1,71 @@
...
+	prm, err := oauthex.GetProtectedResourceMetadataFromHeader(ctx, header, c)
+	if err != nil {
+		return nil, err
+	}
+	if len(prm.AuthorizationServers) == 0 {
+		return nil, fmt.Errorf("resource %s provided no authorization servers", prm.Resource)
+	}
+	// TODO: try more than one?
+	authServer := prm.AuthorizationServers[0]
+	// TODO: which scopes to ask for? All of them?
```
ramongalate: I don't think we should ask for all the scopes, I think the best solution would be for the client to specify in its options which scopes it wants and validate against the supported ones, failing if not supported.

mcp/auth.go
```diff
@@ -0,0 +1,71 @@
...
+		ClientID:     "TODO: from registration",
+		ClientSecret: "TODO: from registration",
+		Endpoint: oauth2.Endpoint{
+			AuthURL:  asm.AuthorizationEndpoint,
+			TokenURL: asm.TokenEndpoint,
+			// DeviceAuthURL: "",
+			// AuthStyle: "from auth meta?",
+		},
+		RedirectURL: "", // ???
+		Scopes:      scopes,
```
ramongalate: If you also request openid, it needs to generate and pass a nonce and validate it in the ID Token.

mcp/auth.go
ramongalate: and we could default to all, if no scope is passed.

besmiralia:
Hi, is there an expected timeline for this feature/pr to be reviewed and merged?

jba:
@besmiralia I will make progress on it this week, but client-side auth probably won't be ready until after v1 is out.

sword-jin:
would like to help if needed, I have much experience with oauth2 and MCP, we are waiting this feature.

jba:
See auth/client.go for where I'm currently at with this. Basically, we do very little for you, other than provide a hook. I'd like to see whether this is sufficient. We can certainly do more, and you can find the PRM and ASM implementations in internal/oauthex.

/cc @wagnerjt @ramongalate @sword-jin @RoyceLeonD.

wagnerjt:
Heya @jba -- In my opinion, the `type OAuthHandler func(context.Context, OAuthHandlerArgs) (oauth2.TokenSource, error)` is the minimum for everyone to implement the rest.

However, I can also see arguments that it would be great to have another option that supports only the DCR flow (creating and registering them). At the end, it still returns the `(oauth2.TokenSource, error)`.

findleyr:
I can review this, but don't have much experience with oauth2. It would be really helpful if someone could opine on whether this is necessary and sufficient. Thanks!

wagnerjt:
Here is a nice [writeup around DCR](https://www.scalekit.com/blog/dynamic-client-registration-oauth2). If anything additional would be added it would be this flow and ignoring the rest -- since there are many different ways the OAuth handshake could occur (device authorization, oidc client id/secret flow, needing to use PKCE over authorization flow) on top of the many different encryption algorithms.

auth/client.go
```diff
@@ -0,0 +1,102 @@
...
+		return resp, nil
+	}
+	// Try to authorize.
+	t.mu.Lock()
+	// If we don't have a token source, get one by following the OAuth flow.
+	// (We may have obtained one while t.mu was not held above.)
+	if _, ok := t.opts.Base.(*oauth2.Transport); !ok {
+		authHeaders := resp.Header[http.CanonicalHeaderKey("WWW-Authenticate")]
+		ts, err := t.handler(req.Context(), OAuthHandlerArgs{
+			ResourceMetadataURL: extractResourceMetadataURL(authHeaders),
```
rolandshoemaker, line 84: Per https://verialabs.com/blog/from-mcp-to-shell/, we may want to add some constraints to the type of URL returned in order to prevent certain attacks.

jba:
This had gotten quite stale, so I squashed and force-pushed. 

I will add commits to address Roland's recent comment, and to put this behind an experiment so we can land it and people can start using it.

auth/client.go
```diff
@@ -0,0 +1,104 @@
...
+	resp, err := base.RoundTrip(req)
+	if err != nil {
+		return nil, err
+	}
+	if resp.StatusCode != http.StatusUnauthorized {
+		return resp, nil
+	}
+	if haveTokenSource {
+		// We failed to authorize even with a token source; give up.
+		return resp, nil
```
rolandshoemaker, line 75: Should we return an error here explaining that we tried to authorize and it failed, or is that going to be handled higher in the stack?

auth/client.go
```diff
@@ -0,0 +1,104 @@
...
+	handler OAuthHandler
+	mu      sync.Mutex // protects opts.Base
+	opts    HTTPTransportOptions
+}
+
+// NewHTTPTransport returns a new [*HTTPTransport].
+// The handler is invoked when an HTTP request results in a 401 Unauthorized status.
+// It is called only once per transport. Once a TokenSource is obtained, it is used
+// for the lifetime of the transport; subsequent 401s are not processed.
+func NewHTTPTransport(handler OAuthHandler, opts *HTTPTransportOptions) (*HTTPTransport, error) {
```
rolandshoemaker, line 42: Check if handler is nil, otherwise we will panic when trying to initialize the token source in RoundTrip.

auth/client.go
jba: Higher up. The caller will see the unauthorized status and should perform the OAuth dance again.

samthanawalla:
@jba asked me to take over this PR: https://github.com/modelcontextprotocol/go-sdk/pull/544

It is easier to close and re-open. Added co-authored-by tag.

Currently in the draft state while I add some quick tests.

PR #181, opened 2025-07-28
mcp: support application/json in streamable client

jba:
The client handles POST responses with content type application/json.

TODO: tests

### DIFF
--- a/mcp/streamable.go
+++ b/mcp/streamable.go
@@ -800,6 +800,8 @@ func (s *streamableClientConn) Write(ctx context.Context, msg jsonrpc.Message) e
 	return nil
 }
 
+// postMessage POSTs msg to the server and reads the response.
+// It returns the session ID from the response.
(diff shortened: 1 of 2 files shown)


mcp/streamable.go
```diff
@@ -836,8 +838,17 @@ func (s *streamableClientConn) postMessage(ctx context.Context, sessionID string
 		// Section 2.1: The SSE stream is initiated after a POST.
 		go s.handleSSE(resp)
 	case "application/json":
-		// TODO: read the body and send to s.incoming (in a select that also receives from s.done).
+		body, err := io.ReadAll(resp.Body)
```
samthanawalla, line 841: Should we run this in its own goroutine?

jba: I don't think we need to. This case is synchronous, I believe. It's just an ordinary RPC.

samthanawalla: Sounds good

jba:
> Is it possible to add a test for this in this CL?

Done.

Issue #135, opened 2025-07-15
Enable advertising unused capabilities


SamMorrowDrums:
> Servers advertise that they have a capability only if the corresponding feature was added. For example, a server will advertise the "prompts" capability only if AddPrompt was called before the client and server exchange initialization messages.

The above is from release notes, and is fine for static servers (the most common case), however servers that support the various dynamic features of MCP (such as the GitHub MCP Server) can do things like add the first Prompt after the initialisation call.

This means that implicit feature enablement is fine, but the ability to manually enable features that can become available is also a requirement to support the dynamic features of the MCP spec properly.

jba:
Thanks, I was wondering if that use case was a real one. We'll add this ability. Perhaps `Has{Prompts,Resources,Tools}` booleans in `ServerOptions`?

samthanawalla:
I agree with the new ServerOption. It enables servers to pre-advertise and have more flexibility.
The only downside to this is if a Server advertises a feature that they may never add during the lifecycle, but I don't really see any problems with that.

I think clients could probably react to a list changed notification instead of having a server option but pre-advertising may create a better UI overall with little to no downside.

SamMorrowDrums:
> I think clients could probably react to a list changed notification instead of having a server option but pre-advertising may create a better UI overall with little to no downside.

Agreed that could be nice, however I think the approach of "all capabilities are negotiated up front" is closest to the current spec.

@jba your suggested method makes sense to me and seems clear. Thanks for considering.

PR #182, opened 2025-07-28
mcp/server: add initial capability server option

samthanawalla:
This CL adds an option for servers to advertise support for features that have not been added yet. This includes Prompts, Tools, and Resources as these could be dynamically added after initialization.

Fixes: #135

### DIFF
--- a/mcp/server.go
+++ b/mcp/server.go
@@ -46,6 +46,14 @@ type Server struct {
 	resourceSubscriptions   map[string]map[*ServerSession]bool // uri -> session -> bool
 }
 
+// InitialCapabilities allows a server to advertise support for capabilities
+// during initialization, even if no resources for that capability have been added yet.
+type InitialCapabilities struct {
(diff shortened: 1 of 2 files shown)


mcp/server.go
```diff
@@ -46,6 +46,14 @@ type Server struct {
 	resourceSubscriptions   map[string]map[*ServerSession]bool // uri -> session -> bool
 }
 
+// InitialCapabilities allows a server to advertise support for capabilities
+// during initialization, even if no resources for that capability have been added yet.
+type InitialCapabilities struct {
```
jba: I don't think we need a new type for these. Just add them to ServerOptions.
Let's call them `HasTools`, etc. unless we can think of a better name.

jba (changes requested):

mcp/server.go
samthanawalla: Sounds good, done.

PR #184, opened 2025-07-28
mcp: various cleanups to streamable code

jba:


jba:
You can review each commit separately: each does one logical change.

PR #185, opened 2025-07-28
jsonschema: options for inference

jba:
- Add ForOptions to hold options for schema inference.

- Replace ForLax with ForOptions.IgnoreBadTypes.

- Add an option to provide schemas for arbitrary types.

- Provide a default mapping from types to schemas that includes stdlib types with MarshalJSON methods.

- Add Schema.CloneSchemas. This is needed to make copies of the schemas in the above map: a schema cannot appear twice in a parent schema, because schema addresses matter when resolving internal references.

### DIFF
--- a/jsonschema/infer.go
+++ b/jsonschema/infer.go
@@ -19,12 +19,12 @@ import (
 
 // ForOptions are options for the [For] function.
 type ForOptions struct {
-	// If IgnoreBadTypes is true, fields that can't be represented as a JSON Schema
+	// If IgnoreInvalidTypes is true, fields that can't be represented as a JSON Schema
 	// are ignored instead of causing an error.
 	// This allows callers to adjust the resulting schema using custom knowledge.
 	// For example, an interface type where all the possible implementations are
 	// known can be described with "oneof".
-	IgnoreBadTypes bool
+	IgnoreInvalidTypes bool
 
 	// TypeSchemas maps types to their schemas.
 	// If [For] encounters a type equal to a type of a key in this map, the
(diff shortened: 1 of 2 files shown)


samthanawalla (approved):

jsonschema/infer.go
```diff
@@ -8,12 +8,33 @@ package jsonschema
...
+	// TypeSchemas maps types to their schemas.
+	// If [For] encounters a type equal to a type of a key in this map, the
+	// corresponding value is used as the resulting schema (after cloning to
+	// ensure uniqueness).
+	// Types in this map override the default translations, as described
+	// in [For]'s documentation.
+	TypeSchemas map[any]*Schema
+}
+
 // For constructs a JSON schema object for the given type argument.
```
findleyr, line 38: // If non-nil, the provided options configures certain aspects of this contruction, described below.

(we need to document somewhere that opts may be nil)

jsonschema/infer.go
```diff
@@ -8,12 +8,33 @@ package jsonschema
...
 	"reflect"
 	"regexp"
+	"time"
 
 	"github.com/modelcontextprotocol/go-sdk/internal/util"
 )
 
+// ForOptions are options for the [For] function.
+type ForOptions struct {
+	// If IgnoreBadTypes is true, fields that can't be represented as a JSON Schema
```
findleyr: Is there any better word than 'Bad'? I'd say 'Unsupported' is better, but it's too long.

Probably Bad is best, I just wanted to note this decision.

jba: replaced with "Invalid"

jsonschema/infer.go
jba: rewrote doc

findleyr (approved):

PR #191, opened 2025-07-29
Implement json.Unmarshaler to CreateMessageResult

francistm:
When calling `ServerSession.CreateMessage`, will get an json unmarshal error because `CreateMessageResult.Content` is an interface, and it doesn't implement `json.Unmarshaller`.

Implement the `json.Unmarshaller` to `CreateMessageResult`, let it able to unmarshal the client response without error

### DIFF
--- a/mcp/mcp_test.go
+++ b/mcp/mcp_test.go
@@ -125,7 +125,7 @@ func TestEndToEnd(t *testing.T) {
 	loggingMessages := make(chan *LoggingMessageParams, 100) // big enough for all logging
 	opts := &ClientOptions{
 		CreateMessageHandler: func(context.Context, *ClientSession, *CreateMessageParams) (*CreateMessageResult, error) {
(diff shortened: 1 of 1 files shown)


mcp/protocol.go
```diff
@@ -264,6 +264,23 @@ type CreateMessageResult struct {
 	StopReason string `json:"stopReason,omitempty"`
 }
 
+func (r *CreateMessageResult) UnmarshalJSON(data []byte) error {
+	type result CreateMessageResult
```
jba: add the comment "// avoid recursion"

jba:
Thanks, good catch!

mcp/protocol.go
francistm: added

jba (dismissed):

jba:
Please fix failing test.

francistm:
> Please fix failing test.

I've update the stub in unittest, should be working now

Issue #179, opened 2025-07-26
Obtain server response even with malformed json input


aryannr97:
You can make MCP server log error, send response and then abruptly exit, with a json input malformed at the end

**To Reproduce**
1. Start the MCP server
```
$ go run github.com/modelcontextprotocol/go-sdk/examples/hello
```

2. Send `initialize`
```
{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{"roots":{"listChanged":true},"sampling":{},"elicitation":{}},"clientInfo":{"name":"ExampleClient","title":"Example Client Display Name","version":"1.0.0"}}}
```

3. Send `notifications/initialized`
```
{"jsonrpc":"2.0","method":"notifications/initialized"}
```

4. Now send `tools/list` by malforming it at the end
```
{"jsonrpc":"2.0","id":2,"method":"tools/list"},
```

**Expected behavior**
Server should handle invalid json input and report it as an error without sending the valid tool/list response and exiting

**Logs**
```
$ go run github.com/modelcontextprotocol/go-sdk/examples/hello
{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{"roots":{"listChanged":true},"sampling":{},"elicitation":{}},"clientInfo":{"name":"ExampleClient","title":"Example Client Display Name","version":"1.0.0"}}}
read: {"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{"roots":{"listChanged":true},"sampling":{},"elicitation":{}},"clientInfo":{"name":"ExampleClient","title":"Example Client Display Name","version":"1.0.0"}}}
{"jsonrpc":"2.0","id":1,"result":{"capabilities":{"completions":{},"logging":{},"prompts":{"listChanged":true},"resources":{"listChanged":true},"tools":{"listChanged":true}},"protocolVersion":"2024-11-05","serverInfo":{"name":"greeter","version":""}}}
write: {"jsonrpc":"2.0","id":1,"result":{"capabilities":{"completions":{},"logging":{},"prompts":{"listChanged":true},"resources":{"listChanged":true},"tools":{"listChanged":true}},"protocolVersion":"2024-11-05","serverInfo":{"name":"greeter","version":""}}}
{"jsonrpc":"2.0","method":"notifications/initialized"}
[...]

PR #192, opened 2025-07-29
mcp: handle bad trailing stdio input graciously (#179)

aryannr97:
### Issue
You can make MCP server log error, send response and then abruptly exit, with a json input malformed at the end

### Why this change
- While using `json.RawMessage`, once a valid first json is decoded, trailing bad input is silently ignored.
- Without graciously handling this input, mcp server is currently sending response as well as encountering error
- It should just report error without further processing of request

### Status
- [x] Code changes
- [x] Unit tests

### DIFF
--- a/mcp/transport.go
+++ b/mcp/transport.go
@@ -308,6 +308,19 @@ func newIOConn(rwc io.ReadWriteCloser) *ioConn {
 		for {
 			var raw json.RawMessage
 			err := dec.Decode(&raw)
+			// If decoding was successful, check for trailing data at the end of the stream.
+			if err == nil {
+				// Read the next byte to check if there is trailing data.
+				var tr [1]byte
+				if n, readErr := dec.Buffered().Read(tr[:]); n > 0 {
+					// If read byte is not a newline, it is an error.
+					if tr[0] != '\n' {
+						err = fmt.Errorf("invalid trailing data at the end of stream")
+					}
+				} else if readErr != nil && readErr != io.EOF {
+					err = readErr
+				}
+			}
 			select {
 			case incoming <- msgOrErr{msg: raw, err: err}:
 			case <-closed:
@@ -419,17 +432,6 @@ func (t *ioConn) Read(ctx context.Context) (jsonrpc.Message, error) {
 
 	case <-t.closed:
 		return nil, io.EOF
-	}
-
-	// Read the next byte to check if there is trailing data.
-	var tr [1]byte
-	if n, err := in.Buffered().Read(tr[:]); n > 0 {
-		// If read byte is not a newline, it is an error.
-		if tr[0] != '\n' {
-			return nil, fmt.Errorf("invalid trailing data at the end of stream")
-		}
-	} else if err != nil && err != io.EOF {
-		return nil, err
 	}
 
 	msgs, batch, err := readBatch(raw)


mcp/transport.go
```diff
@@ -375,6 +375,18 @@ func (t *ioConn) read(ctx context.Context, in *json.Decoder) (jsonrpc.Message, e
 	if err := in.Decode(&raw); err != nil {
 		return nil, err
 	}
+	// Read remaining data in the buffer.
```
jba: I'd think this would be simpler. Doesn't it suffice to check that the next byte is a newline?

aryannr97: As per my debugging, `raw` is definitely holding only valid json bytes after `Decode` call. Anything beyond that, needs to be read from `in.Buffered`. We can check only for newline in the next byte as you are suggesting and return err. I just think, providing actual garbage trail may just be informative

jba: I see your point, but let's keep it simple for now.

mcp/transport.go
```diff
@@ -375,16 +375,17 @@ func (t *ioConn) read(ctx context.Context, in *json.Decoder) (jsonrpc.Message, e
 	if err := in.Decode(&raw); err != nil {
 		return nil, err
 	}
-	// Read remaining data in the buffer.
-	tr := &trail{}
-	buf := in.Buffered()
-	err := tr.load(buf)
+
+	// Read the next byte to check if there is trailing data.
+	tr := make([]byte, 1)
```
jba: var tr [1]byte

mcp/transport.go
```diff
@@ -375,16 +375,17 @@ func (t *ioConn) read(ctx context.Context, in *json.Decoder) (jsonrpc.Message, e
...
+	_, err := in.Buffered().Read(tr)
```
jba: if _, err := ...; err != nil

jba: Also, io.EOF is not an error.

jba: And you should check the first return value. From the io.Reader doc:
 > Callers should always process the n > 0 bytes returned before considering the error err. Doing so correctly handles I/O errors that happen after reading some bytes and also both of the allowed EOF behaviors.

jba: Also, it's possible that the reader is not at EOF, but there is no more data (yet) on the input stream. For example, the other end writes `{...}` with no newline and no more data. Then I think this would hang waiting for data, or maybe it would return 0; I'm not sure. Write a test case.

mcp/transport.go
aryannr97: Done

mcp/transport_test.go
```diff
@@ -51,3 +53,46 @@ func TestBatchFraming(t *testing.T) {
...
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			tr := newIOConn(tt.fields.rwc)
+			_, err := tr.Read(tt.args.ctx)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("ioConn.Read() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+			if !reflect.DeepEqual(err.Error(), tt.want) {
```
jba: You can just use `==`. These are strings.

mcp/transport_test.go
```diff
@@ -51,3 +53,46 @@ func TestBatchFraming(t *testing.T) {
 		}
 	}
 }
+
+func Test_ioConn_Read_BadTrailingData(t *testing.T) {
```
jba: TestIOConnRead

mcp/transport_test.go
```diff
@@ -51,3 +53,46 @@ func TestBatchFraming(t *testing.T) {
...
+	type fields struct {
+		rwc io.ReadWriteCloser
+	}
+	type args struct {
+		ctx context.Context
+	}
+	tests := []struct {
+		name    string
+		fields  fields
```
jba: Only the input would vary across tests, so this can be `input string`

mcp/transport_test.go
```diff
@@ -51,3 +53,46 @@ func TestBatchFraming(t *testing.T) {
...
+		args    args
```
jba: not needed

mcp/transport_test.go
```diff
@@ -51,3 +53,46 @@ func TestBatchFraming(t *testing.T) {
...
+		want    string
+		wantErr bool
```
jba: an empty `want` can show that you don't want an error, so remove `wantErr`

mcp/transport_test.go
```diff
@@ -51,3 +53,46 @@ func TestBatchFraming(t *testing.T) {
...
+	}{
```
jba, line 61: Add two cases for valid read. Same input without the trailing data. One with a newline, one that ends at the `}`.

aryannr97:
@jba could you please take a look at the changes again

aryannr97:
@jba Could you please take a look at this

jba (approved):

PR #206, opened 2025-07-30
mcp/streamable: add persistent SSE GET listener

samthanawalla:
This CL adds the optional persistent SSE GET listener as specified in section 2.2. https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#listening-for-messages-from-the-server

This enables server initiated SSE streams.

For #10

### DIFF
--- a/mcp/streamable.go
+++ b/mcp/streamable.go
@@ -735,6 +735,12 @@ func (t *StreamableClientTransport) Connect(ctx context.Context) (Connection, er
 		ctx:              connCtx,
 		cancel:           cancel,
 	}
+	// Start the persistent SSE listener right away.
+	// Section 2.2: The client MAY issue an HTTP GET to the MCP endpoint.
+	// This can be used to open an SSE stream, allowing the server to
+	// communicate to the client, without the client first sending data via HTTP POST.
+	go conn.handleSSE(nil, false)
+
 	return conn, nil
 }
 
@@ -859,7 +865,7 @@ func (s *streamableClientConn) postMessage(ctx context.Context, sessionID string
 	switch ct := resp.Header.Get("Content-Type"); ct {
 	case "text/event-stream":
 		// Section 2.1: The SSE stream is initiated after a POST.
-		go s.handleSSE(resp)
+		go s.handleSSE(resp, true)
 	case "application/json":
 		body, err := io.ReadAll(resp.Body)
 		resp.Body.Close()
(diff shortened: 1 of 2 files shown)


mcp/streamable.go
```diff
@@ -879,13 +885,11 @@ func (s *streamableClientConn) postMessage(ctx context.Context, sessionID string
 	return sessionID, nil
 }
 
-// handleSSE manages the entire lifecycle of an SSE connection. It processes
-// an incoming Server-Sent Events stream and automatically handles reconnection
-// logic if the stream breaks.
-func (s *streamableClientConn) handleSSE(initialResp *http.Response) {
+// handleSSE manages the lifecycle of an SSE connection. It can be either
+// temporary (for a POST response) or persistent (for the main GET listener).
+func (s *streamableClientConn) handleSSE(initialResp *http.Response, temporary bool) {
```
findleyr: I think it may be slightly clearer to invert the sense of this boolean, passing `persistent bool` for the persistent GET.

mcp/streamable_test.go
```diff
@@ -211,8 +211,40 @@ func TestClientReplay(t *testing.T) {
 	}
 }
 
-// Helper to read a specific number of progress notifications.
-func readProgressNotifications(t *testing.T, ctx context.Context, notifications chan string, count int) []string {
+// TestServerInitiatedSSE verifies that the persistent SSE connection remains
+// open and can receive multiple, non-consecutive, server-initiated events.
+func TestServerInitiatedSSE(t *testing.T) {
+	notifications := make(chan string)
+	server := NewServer(testImpl, &ServerOptions{})
```
findleyr: Pass nil rather than the empty struct.

findleyr (dismissed):
Thanks, this ended up being simpler than expected.

mcp/streamable.go

mcp/streamable_test.go

mcp/streamable_test.go
```diff
@@ -211,8 +211,40 @@ func TestClientReplay(t *testing.T) {
...
+	client := NewClient(testImpl, &ClientOptions{ToolListChangedHandler: func(ctx context.Context, cc *ClientSession, params *ToolListChangedParams) {
+		notifications <- "toolListChanged"
+	},
+	})
+	clientSession, err := client.Connect(ctx, NewStreamableClientTransport(httpServer.URL, nil))
+	if err != nil {
+		t.Fatalf("client.Connect() failed: %v", err)
+	}
+	defer clientSession.Close()
+	time.Sleep(50 * time.Millisecond)
```
jba: I don't believe you need this. (If you do, we have a race condition.) 
The server side will enqueue all messages for stream 0 until the GET occurs.

jba (changes requested):

mcp/streamable_test.go
samthanawalla: Good point, removed.

jba (approved):

findleyr (approved):

PR #208, opened 2025-07-30
examples: add a 'listeverything' client example, and reorganize

findleyr:
Add a client example that lists all features on a stdio server. Since this is our first client example, organize examples by client / server -- otherwise it is too difficult to find client examples.

For #33


examples/client/listeverything/main.go
```diff
@@ -0,0 +1,65 @@
+// Copyright 2025 The Go MCP SDK Authors. All rights reserved.
+// Use of this source code is governed by an MIT-style
+// license that can be found in the LICENSE file.
+
+// The listeverything command lists all features of a stdio MCP server.
```
samthanawalla: Since it only lists features, maybe listfeatures? or mcp-features or mcp-inspect?

findleyr: Good idea. Done.

jba (approved):

PR #213, opened 2025-07-31
mcp/server: advertise completions only if installed

samthanawalla:
Previously, completions would always be advertised as a capability even if the CompletionHandler was not installed by the server. This CL fixes that.


findleyr (approved):
Thanks!

PR #215, opened 2025-07-31
add and implement SessionStorage interface

joshwlewis:
This is a step toward enabling distributed MCP servers. It's a new interface that would allow additional types of session stores like:
- Database backed sessions
- Stateless sessions

This interface supports streamable HTTP, SSE, and custom Transports.

MemorySessionStore is the only current implementation. Other implementations are left to users for now.

This doesn't totally solve distributed MCP servers - routing messages between server instances still needs to be solved.

Related: #148, #212

### DIFF
--- a/examples/sse/main.go
+++ b/examples/sse/main.go
@@ -65,6 +65,6 @@ func main() {
 		default:
 			return nil
 		}
-	})
+	}, nil)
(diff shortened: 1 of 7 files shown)


jba:
See #232. In particular, we don't want to save transports, we want to save session state.

joshwlewis:
Thanks -- closing in favor of #232.

PR #220, opened 2025-08-01
support http://json-schema.org/draft-07/schema#

slimslenderslacks:
We can not connect to some MCP servers, such as mongodb and playwright, because we cannot validate against this draft.

```
panic: adding tool "switch-connection": cannot validate version http://json-schema.org/draft-07/schema#, 
only https://json-schema.org/draft/2020-12/schema
```

I heard from @jba that we will be moving this module out of the schema, but in the short term, the number of MCP servers that we can be limited.

### DIFF
--- a/jsonschema/doc.go
+++ b/jsonschema/doc.go
@@ -6,8 +6,8 @@
 Package jsonschema is an implementation of the [JSON Schema specification],
 a JSON-based format for describing the structure of JSON data.
 The package can be used to read schemas for code generation, and to validate
-data using the draft 2020-12 specification. Validation with other drafts
-or custom meta-schemas is not supported.
+data using the draft 2020-12 and draft-07 specifications. Validation with
+other drafts or custom meta-schemas is not supported.
 
 Construct a [Schema] as you would any Go struct (for example, by writing
 a struct literal), or unmarshal a JSON schema into a [Schema] in the usual
--- a/jsonschema/schema.go
+++ b/jsonschema/schema.go
@@ -20,10 +20,10 @@ import (
 )
 
 // A Schema is a JSON schema object.
-// It corresponds to the 2020-12 draft, as described in https://json-schema.org/draft/2020-12,
-// specifically:
-// - https://json-schema.org/draft/2020-12/draft-bhutton-json-schema-01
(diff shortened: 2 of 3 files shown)


jba:
Wow, this is amazing! Thanks so much. Can you tell me how you wrote this code? Did you have a draft 07 implementation lying around or did you just read the specs and diff them? Either way, I'm going to read the draft 07 spec carefully myself to check your implementation, and then we'd be very happy to accept it. I hope to get to this early next week.

slimslenderslacks:
Honestly, I just had Claude code read the two drafts, asked it to diff them, and then had it update the impl to support both.  I had to give it a new test (existing MCPs that were use draft 07), but then I asked it to loop until both the existing and new tests were passing.

I mostly just sanity checked the output but if this is on the right track then I can do some more work.

I think it's probably important for adoption because I'm seeing lots of MCP servers in the wild that will need this support.  Over the next few days, I'll have a test base of about 150 MCP servers that I can connect to with this client.

From my tests so far, there are 3 things that come up:
1. this draft07 schema thing
2. lots of servers still rely on stdio transport impls dropping non-json messages (when they erroneously log junk to stdout). Other SDKs have not pushed back very hard on this so we are where we are?
3. resource naming:  go-sdk is panicing on quite a few servers because it doesn't like some of the existing resource uri names (I'll put 2 and 3 in separate issues for discussion)

jba:
Sounds good!
I would want to make sure it passes the conformance tests. See https://github.com/modelcontextprotocol/go-sdk/blob/main/jsonschema/testdata/draft2020-12/README.md. That would go a long way towards convincing me, plus we want them for regression anyway.

jba:
@slimslenderslacks Please open this at github.com/google/jsonschema-go. If you don't want to, let me know and I'll do it. But then I will be the committer. (I will make you co-author of course.)

slimslenderslacks:
Cool!  Will open this now.

jba:
Thanks. And how about adding the test suite? We can't accept the PR without it.

slimslenderslacks:
Right, I'll try to get back to a test suite later today.
But for now, I've added a draft PR over in the other repo:  https://github.com/google/jsonschema-go/pull/6

I guess we'll just close the pull request, right?

jba:
Yup!

PR #231, opened 2025-08-04
update Subscribe/Unsubscribe handlers in ServerOpts

slimslenderslacks:
## Background

Since the Server has a map of ServersSessions->subscriptions, and takes care of routing ResourceUpdate messages to ServerSessions, I think the decision was made to leave *ServerSession out of the SubscribeHandler/UnsubscribeHandler signatures. However, resource subscriptions are not specific to the a particular client and this information is kept private in the Session today.

## Proposal

Align the Subscribe/Unsubscribe handlers with the other two handlers that are client specific (ProgressNotificationHandler, and RootListChangeHanlder), and include *ServerSession in the signature.

It comes in practice because because a gateway server that needs to forward subscriptions still needs this information. The routing logic in Server is convenient and means that most servers won't need to know which ServerSessions are subscribed. But I think there are still use cases where the user will need to know.

### DIFF
--- a/design/design.md
+++ b/design/design.md
@@ -776,9 +776,9 @@ If a server author wants to support resource subscriptions, they must provide ha
 type ServerOptions struct {
   ...
   // Function called when a client session subscribes to a resource.
-  SubscribeHandler func(context.Context, *SubscribeParams) error
+  SubscribeHandler func(context.Context, ss *ServerSession, *SubscribeParams) error
   // Function called when a client session unsubscribes from a resource.
-  UnsubscribeHandler func(context.Context, *UnsubscribeParams) error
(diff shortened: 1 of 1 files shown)


samthanawalla:
This change makes sense to me. @jba Do you see any problems with adding ServerSession to the Subscribe Handlers?

jba:
Agreed, this seems right.

jba (dismissed):

slimslenderslacks:
@jba @samthanawalla sorry, just realized there was a ref to these two handlers in design/design.md so updated them as well.

Since this would be a breaking change, do we have a change log that you want updated?

jba:
We don't need one; we run [apidiff](https://go.googlesource.com/exp/+/refs/heads/master/cmd/apidiff) to compare the previous and candidate releases.

Issue #225, opened 2025-08-02
mcp lifecycle: initialization phase is not completely followed


aryannr97:
**Issue**

- MCP server is responding to `tools/list` request, even before receiving `initialized` message. 
- As per the MCP specification
>The server SHOULD NOT send requests other than [pings](https://modelcontextprotocol.io/specification/2025-06-18/basic/utilities/ping) and [logging](https://modelcontextprotocol.io/specification/2025-06-18/server/utilities/logging) before receiving the initialized notification.

**To Reproduce**
1. Start MCP server
2. Send `initialize` message from client to server
3. Send `tools/list` message from client to server

**Expected behavior**
The `tools/list` call should result in error citing initialization phase is not fully complete

**Logs**
```
$ go run github.com/modelcontextprotocol/go-sdk/examples/hello
{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{"roots":{"listChanged":true},"sampling":{},"elicitation":{}},"clientInfo":{"name":"ExampleClient","title":"Example Client Display Name","version":"1.0.0"}}}
read: {"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{"roots":{"listChanged":true},"sampling":{},"elicitation":{}},"clientInfo":{"name":"ExampleClient","title":"Example Client Display Name","version":"1.0.0"}}}
{"jsonrpc":"2.0","id":1,"result":{"capabilities":{"completions":{},"logging":{},"prompts":{"listChanged":true},"resources":{"listChanged":true},"tools":{"listChanged":true}},"protocolVersion":"2024-11-05","serverInfo":{"name":"greeter","version":""}}}
write: {"jsonrpc":"2.0","id":1,"result":{"capabilities":{"completions":{},"logging":{},"prompts":{"listChanged":true},"resources":{"listChanged":true},"tools":{"listChanged":true}},"protocolVersion":"2024-11-05","serverInfo":{"name":"greeter","version":""}}}
{"jsonrpc":"2.0","id":2,"method":"tools/list"}
read: {"jsonrpc":"2.0","id":2,"method":"tools/list"}
[...]

Issue #190, opened 2025-07-29
unbounded memory in streamable transport


jba:
In the current implementation of the streamable HTTP transport, the server can consume unbounded memory.

1. `stream.outgoing` is a slice of messages ready to send. But they may be sent very slowly if the client connection is slow. What's worse, if the client never does a non-replay GET (and the spec says that it [doesn't have to](https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#listening-for-messages-from-the-server)), then messages for stream 0 will never be sent.

2. Since a stream can always be replayed, even after the response to a request is sent, data structures for the stream are kept forever.

For (1) and POST requests, we should bound the outgoing queue to exert backpressure on the server.

For (1) and the non-replay GET, bounding the queue will just lead to deadlock if the GET never arrives. However, dropping messages that are sent prior to receiving the GET is also wrong: it will cause a race condition where the server might send a message on a session just as it's established and before the client can make the GET request. One solution is to use a circular queue and drop old messages. These messages should be notifications for the most part (feature-changed, logging) so dropping them doesn't matter. If the server were to make requests outside of a client request, dropping those would be more problematic.

samthanawalla:
I can work on this.

To break it down:

- [ ] For POST request streams, replace stream.outgoing with bounded queue (buffered channel)
- [ ] Replace the notification stream (stream 0) with a circular buffer
- [ ] Enforce a configurable TTL eviction policy to purge inactive streams

jba:
> If the server were to make requests outside of a client request, dropping those would be more problematic.

We should also talk about this case. For now, I would error out if a request is sent to stream 0. It would be bad to drop requests.

I believe this is a flaw in the spec. It should either change the MAY to a MUST for the GET (and require that it happen ASAP, though even that wouldn't truly avoid the possibility of server memory overflow if the server sent millions of requests immediately); or the spec should say that server requests that aren't part of of a client request MAY fail (or better, MUST fail). If you have a better idea, let's discuss. Otherwise I'll file an issue against the spec.

jba:
> Enforce a configurable TTL eviction policy to purge inactive streams

Let's keep this as simple as possible: a single duration.

samthanawalla:
To implement the eviction policy for the MemoryEventStore, I was thinking of doing a LRU cache style system.

[...]

PR #232, opened 2025-08-04
server: support distributed sessions

jba:
- SessionState holds the state of a ServerSession.

- SessionStore allows arbitrary storage for SessionState.

- ServerSession.InitSession provides a session ID, store and state to a ServerSession.

### DIFF
--- a/mcp/server.go
+++ b/mcp/server.go
@@ -535,7 +535,7 @@ func (s *Server) Run(ctx context.Context, t Transport) error {
 // bind implements the binder[*ServerSession] interface, so that Servers can
 // be connected using [connect].
 func (s *Server) bind(conn *jsonrpc2.Connection) *ServerSession {
-	ss := &ServerSession{conn: conn, server: s}
+	ss := &ServerSession{conn: conn, server: s, state: &SessionState{}}
 	s.mu.Lock()
 	s.sessions = append(s.sessions, ss)
 	s.mu.Unlock()
@@ -596,25 +596,32 @@ func (ss *ServerSession) NotifyProgress(ctx context.Context, params *ProgressNot
 // Call [ServerSession.Close] to close the connection, or await client
 // termination with [ServerSession.Wait].
 type ServerSession struct {
-	server           *Server
-	conn             *jsonrpc2.Connection
-	mcpConn          Connection
-	mu               sync.Mutex
-	logLevel         LoggingLevel
-	initializeParams *InitializeParams
-	initialized      bool
-	keepaliveCancel  context.CancelFunc
+	server          *Server
+	conn            *jsonrpc2.Connection
+	mu              sync.Mutex
+	initialized     bool
+	keepaliveCancel context.CancelFunc
+
+	sessionID string
+	state     *SessionState
+	store     SessionStore
 }
 
 func (ss *ServerSession) setConn(c Connection) {
-	ss.mcpConn = c
 }
 
-func (ss *ServerSession) ID() string {
-	if ss.mcpConn == nil {
-		return ""
-	}
-	return ss.mcpConn.SessionID()
+func (ss *ServerSession) ID() string { return ss.sessionID }
+
+// InitSession initializes the session with a session ID, state, and store.
+// If called, it must be called immediately after the session is connected.
+// If never called, the session will begin with a zero SessionState and no session
+// ID or store.
+func (ss *ServerSession) InitSession(sessionID string, state *SessionState, store SessionStore) {
+	ss.mu.Lock()
+	defer ss.mu.Unlock()
+	ss.sessionID = sessionID
+	ss.state = state
+	ss.store = store
 }
 
 // Ping pings the client.
@@ -638,7 +645,7 @@ func (ss *ServerSession) CreateMessage(ctx context.Context, params *CreateMessag
 // is below that of the last SetLevel.
 func (ss *ServerSession) Log(ctx context.Context, params *LoggingMessageParams) error {
 	ss.mu.Lock()
-	logLevel := ss.logLevel
+	logLevel := ss.state.LogLevel
 	ss.mu.Unlock()
 	if logLevel == "" {
 		// The spec is unclear, but seems to imply that no log messages are sent until the client
@@ -752,8 +759,13 @@ func (ss *ServerSession) initialize(ctx context.Context, params *InitializeParam
 		return nil, fmt.Errorf("%w: \"params\" must be be provided", jsonrpc2.ErrInvalidParams)
 	}
 	ss.mu.Lock()
-	ss.initializeParams = params
+	ss.state.InitializeParams = params
 	ss.mu.Unlock()
+	if ss.store != nil {
+		if err := ss.store.Store(ctx, ss.sessionID, ss.state); err != nil {
+			return nil, fmt.Errorf("storing session state: %w", err)
+		}
+	}
 
 	// Mark the connection as initialized when this method exits.
 	// TODO: Technically, the server should not be considered initialized until it has
@@ -787,10 +799,15 @@ func (ss *ServerSession) ping(context.Context, *PingParams) (*emptyResult, error
 	return &emptyResult{}, nil
 }
 
-func (ss *ServerSession) setLevel(_ context.Context, params *SetLevelParams) (*emptyResult, error) {
+func (ss *ServerSession) setLevel(ctx context.Context, params *SetLevelParams) (*emptyResult, error) {
 	ss.mu.Lock()
 	defer ss.mu.Unlock()
-	ss.logLevel = params.Level
+	ss.state.LogLevel = params.Level
+	if ss.store != nil {
+		if err := ss.store.Store(ctx, ss.sessionID, ss.state); err != nil {
+			return nil, err
+		}
+	}
 	return &emptyResult{}, nil
 }
 
--- a/mcp/session.go
+++ b/mcp/session.go
@@ -1,0 +1,74 @@
+// Copyright 2025 The Go MCP SDK Authors. All rights reserved.
+// Use of this source code is governed by an MIT-style
+// license that can be found in the LICENSE file.
+
+package mcp
+
+import (
+	"context"
+	"io/fs"
+	"sync"
+)
+
+// SessionState is the state of a session.
+type SessionState struct {
+	// InitializeParams are the parameters from the initialize request.
+	InitializeParams *InitializeParams `json:"initializeParams"`
+
+	// LogLevel is the logging level for the session.
+	LogLevel LoggingLevel `json:"logLevel"`
+
+	// TODO: resource subscriptions
+}
+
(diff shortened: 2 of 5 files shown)


joshwlewis:
This looks right to me -- session state is the correct thing to store rather than a transport.

Should this be implemented for the SSE transport as well?

mcp/streamable.go
```diff
@@ -133,28 +140,49 @@ func (h *StreamableHTTPHandler) ServeHTTP(w http.ResponseWriter, req *http.Reque
...
+		session, err := server.Connect(req.Context(), transport)
+		if err != nil {
+			http.Error(w, fmt.Sprintf("failed connection: %v", err), http.StatusInternalServerError)
 			return
 		}
-		h.sessionsMu.Lock()
-		h.sessions[s.sessionID] = s
-		h.sessionsMu.Unlock()
-		session = s
+		session.InitSession(sessionID, state, h.opts.SessionStore)
```
findleyr: I'd prefer if we can avoid the split initialization of session. Can we instead pass something in to `Server.Connect`? Perhaps a `SessionOptions` or `ConnectOptions` struct?

mcp/streamable.go
```diff
@@ -133,28 +140,49 @@ func (h *StreamableHTTPHandler) ServeHTTP(w http.ResponseWriter, req *http.Reque
...
+		h.transportMu.Lock()
+		h.transports[transport.sessionID] = transport
```
findleyr: This is racy. You need to check once more whether the transport exists.

mcp/streamable.go

mcp/server.go
```diff
@@ -556,14 +558,33 @@ func (s *Server) disconnect(cc *ServerSession) {
 	}
 }
 
+type ServerSessionOptions struct {
```
findleyr: Nit: the other types you introduce aren't `ServerSessionState` and `ServerSessionStore`, so I think this should just be `SessionOptions`, to be consistent.

Here's a (perhaps weak) argument for it: a 'Session' is the name for the logical interaction between server and client. These options relate to that abstract interaction; it just so happens that we pass them to the server's view of the session.

mcp/server.go
```diff
@@ -513,7 +514,8 @@ func (s *Server) unsubscribe(ctx context.Context, ss *ServerSession, params *Uns
 // If no tools have been added, the server will not have the tool capability.
 // The same goes for other features like prompts and resources.
 func (s *Server) Run(ctx context.Context, t Transport) error {
-	ss, err := s.Connect(ctx, t)
+	// TODO: provide a way to pass ServerSessionOptions?
```
findleyr, line 518: Should `SessionStore` be part of ServerOptions?

findleyr: To be clear: *instead of* SessionOptions. It relates to the server as a whole, not the individual session.

findleyr: Nevermind, I see that it must be part of `SessionOptions`, since the session needs to be able to update it.

jba: I think as a convenience we could have ServerOptions.SessionStore, which we would just pass along. That would resolve this TODO.

mcp/streamable.go
```diff
@@ -133,28 +140,52 @@ func (h *StreamableHTTPHandler) ServeHTTP(w http.ResponseWriter, req *http.Reque
...
+			// The session might be in the store.
+			state, err = h.opts.SessionStore.Load(req.Context(), sessionID)
+			if errors.Is(err, fs.ErrNotExist) {
+				http.Error(w, fmt.Sprintf("no session with ID %s", sessionID), http.StatusNotFound)
+				return
+			} else if err != nil {
+				http.Error(w, fmt.Sprintf("SessionStore.Load(%q): %v", sessionID, err), http.StatusInternalServerError)
+				return
+			}
+			transport = NewStreamableServerTransport(sessionID, nil)
```
BC-ACherednichenko: If I read it correctly, issue described here https://github.com/modelcontextprotocol/go-sdk/issues/190 is potentially possible

in simple request/response flows (without the actual streaming), using StreamableServerTransport bound to every session creates buffered channels, maps, and a MemoryEventStore. If these are not explicitly cleaned up, then over time, memory will accumulate — if actual streaming never occurs.

This becomes more visible in multi-container environments:
For example, if session abc123 initially hits container A, a transport is created and retained in memory. If a subsequent request for the same session hits container B, container B will also create a new transport instance.

But in a simple, stateless request/response model, there's no clear point where these transports are explicitly cleaned up once the request is served — especially when streaming isn't used, right?

BC-ACherednichenko (commented):
Session change LGTM, just one minor question about NewStreamableServerTransport

mcp/streamable.go
jba: That's true. But is that different from any other network protocol, like TCP? If a peer keeps acknowledging keep-alives, it can keep a connection around forever. 

I think we may need to do some work to propagate a close from keep-alive failure to the transport layer. But I think that's orthogonal to this PR.

BC-ACherednichenko: makes sense to keep it separate from this PR

mcp/session.go
```diff
@@ -0,0 +1,74 @@
...
+// SessionState is the state of a session.
+type SessionState struct {
+	// InitializeParams are the parameters from the initialize request.
+	InitializeParams *InitializeParams `json:"initializeParams"`
+
+	// LogLevel is the logging level for the session.
+	LogLevel LoggingLevel `json:"logLevel"`
+
+	// TODO: resource subscriptions
+}
```
omgitsads: Do we want to store the Initialized state of the session along side the `InitializeParams`, or is there another plan for tackling that?

While the `InitializeParams` and `LogLevel` can be loaded from an external store, the `ServerSession.initialized` bool is still private. If the request comes into a server that did not initialize the session, any subsequent requests can't be fulfilled until the session goes through the Initialization process again.

https://github.com/omgitsads/go-sdk/commit/9f888732d5b98cf7291ddf4b6bc72e3db33929c2 moves the intialized bool into the `SessionState`, but not sure how you'd want to tackle that happening before the response has been sent?

mcp/streamable.go
jba: See #258.

mcp/session.go
jba: That bool isn't necessary, I think: it's equivalent to InitializedParams != nil.

findleyr: The bool IS necessary, because right now its incorrectly set: we should only set it once we've received the 'initialized' notification: https://github.com/modelcontextprotocol/go-sdk/issues/225

mcp/server.go
revanthpobala: Should it be something like this?

```golang

sessionStore := shared.NewDynamoDBSessionStore("mcp-sessions" , client)
mcp.NewServer(&mcp.Implementation {Name : " ABC"},  &mcp.ServerOptions {sessionStore: sessionStore})

```

mcp/session.go
BC-ACherednichenko: I can confirm having the same issue after trying quick implementation on top of the PR new interfaces. 

If we emulate the scenario when a client initialised a connection and retrieved session id, if on a subsequent request, let's say hitting a particular tool,  when it hits a new instance (assuming it lost the connection, or it works in simple request/response mode behind load balancers + mesh, without the actual persistent connection) the new instance has no awereness on initialised state = it equals false, so regardless of successful session retrieval a new instance will reject the request with the following error.

https://github.com/modelcontextprotocol/go-sdk/blob/0375535db46aa5b2117f4065a20a5d594ce8860d/mcp/server.go#L755

mcp/server.go

mcp/server.go
jba: We decided that this should be an argument to Connect.

mcp/streamable.go

mcp/session.go
jba: Added the initialized bool to SessionState.

jba:
I've converted this to draft status. I don't think we fully understand what it means to have a distributed server.

For example, if a server receives a request and then sends a nested request back, how can we be sure that the same server process will receive the reply? I don't think having a session store helps with that problem. So we're going to wait until this gets resolved in the specification.

BC-ACherednichenko:
@jba what do you mean by sends nested request back? my understanding of the spec is that it is potentially possible to operate in both streamable and simple request/response modes, as long as session is distributed and transport does not assume streams only that should be compatible, right?

jba:
> what do you mean by sends nested request back?

1. client -> CallTool -> server (a POST)
2. server -> Sampling -> client (via the SSE stream that's part of the POST connection)
3. client -> Sampling response -> server (separate POST)

What if the server in step 1 is in a different process from the server in step 3?

BC-ACherednichenko:
> server -> Sampling -> client (via the SSE stream that's part of the POST connection)

I think these are essentially two different modes, stremableHTTP and stremableHTTP + SSE, SSE is optional according to the spec, hence, the concrete servers using SDK which dont utilise SSE may omit that part completely.

when it comes to the stremableHTTP + SSE and the provided use-case, regardless of session storage there is also resumability aspect, hence the server needs to have the ability to replay messages based on "Last-event-id" when resume request from the client arrives, as the result in multi container server setup (where is more than one instance of MCP server in a cluster), I think we will need to provide an interface or a fallback so it is possible to replay these messages, something like RMQ.

The protocol itself omits the details on mechanism for message replay, but on the interface level they assume resumability to be functional if I read it correctly, expecting the request to land on the same server limits the implemtation options to sticky session on the edge.

BC-ACherednichenko (commented):
FYI I checked most recent updates on the branch - it looks fully functional now 👍

6juara9:
I'd like to include a small flow diagram as well, and I'm curious if you agree with its representation 🙂

image

cc: @jba

jba:
> > server -> Sampling -> client (via the SSE stream that's part of the POST connection)
> 
> I think these are essentially two different modes, stremableHTTP and stremableHTTP + SSE, SSE is optional according to the spec, hence, the concrete servers using SDK which dont utilise SSE may omit that part completely.

You could say that distributed servers must not initiate SSE streams, but then they couldn't even send things like progress notifications and log messages. So you'd have to say that they don't send nested requests. That seems like a very strange restriction, and I'm not sure how you'd enforce it.

All the discussion about resumability is irrelevant to this point. There may be separate issues there, but I think this one is enough of a concern that I'd like to wait for clarity in the specification before pursuing this.

jba:
[See #148](https://github.com/modelcontextprotocol/go-sdk/issues/148#issuecomment-3176743221).

findleyr:
See also my comment here: https://github.com/modelcontextprotocol/go-sdk/issues/148#issuecomment-3176743221

(Sorry that the conversation is split between that issue and this PR).

jba:
Closing. There are many problems with distributed sessions. A full solution will have to wait for a clearer specification.

PR #234, opened 2025-08-05
mcp: unblock reads of ioConn as soon as Close is called

### LINKED ISSUE #224: mcp/Server.Run does not stop when the context is cancelled
**Describe the bug**
[Per the documentation](https://github.com/modelcontextprotocol/go-sdk/blob/b34ba2149ddefc5ea0318240fc096d032b472216/mcp/server.go#L508-L509), `mcp/Server.Run`:

> // Run blocks until the client terminates the connection or the provided
> // context is cancelled. If the context is cancelled, Run closes the connection.

Closing the context does NOT terminate the close the connection.

**To Reproduce**

1. Create a simple MCP server:

    ```go
		package main

		import (
			"context"
			"errors"
			"log"
			"os"
			"os/signal"
			"syscall"

		)

		func main() {
			ctx, done := signal.NotifyContext(context.Background(),
				syscall.SIGINT, syscall.SIGTERM)
			defer done()

			log.SetOutput(os.Stderr)

			server := mcp.NewServer(&mcp.Implementation{Name: "repro", Version: "0.1.0"}, nil)

			t := mcp.NewLoggingTransport(mcp.NewStdioTransport(), os.Stderr)

			errCh := make(chan error, 1)
			doneCh := make(chan struct{})
			go func() {
				defer close(doneCh)

				log.Print("[INFO] stdio server is starting")
				defer log.Print("[INFO] stdio server stopped")

				if err := server.Run(ctx, t); err != nil && !errors.Is(err, mcp.ErrConnectionClosed) {
					log.Print("[INFO] server is terminated")
					select {
					case errCh <- err:
					default:
					}
				}
			}()

			select {
			case err := <-errCh:
				log.Printf("[ERROR] failed to run stdio server: %s", err)
				done()
				os.Exit(1)
			case <-ctx.Done():
				log.Print("[INFO] provided context was closed, triggering shutdown")
			}

			// Wait for goroutine to exit
			log.Print("[INFO] waiting for server to stop")
			<-doneCh
		}

    ```

2. Run the server:

    ```shell
    go run ./main.go
    ```

3. Press `ctrl+c` (or send INT to the PID)

4. Context will close (logs confirm), but the server does not stop

**Expected behavior**
The server stops when the context is cancelled

**Logs**

```text
./bin/cancel-stdio
2025/08/01 12:44:01 [INFO] stdio server is starting
^C2025/08/01 12:44:02 [INFO] provided context was closed, triggering shutdown
2025/08/01 12:44:02 [INFO] waiting for server to stop
(hang)
```

**Additional context**
N/A

findleyr:
Use a read loop and incoming channel so that we can select on incoming messages and unblock calls of ioConn.Read as soon as Close is called. This avoids the scenario of #224, where a close of the StdioTransport does not gracefully shut down the JSON-RPC connection.

Fixes #224

### DIFF
--- a/mcp/transport.go
+++ b/mcp/transport.go
@@ -265,7 +265,9 @@ func (r rwc) Close() error {
 // See [msgBatch] for more discussion of message batching.
 type ioConn struct {
 	rwc io.ReadWriteCloser // the underlying stream
-	in  *json.Decoder      // a decoder bound to rwc
+
+	// incoming receives messages from the read loop started in [newIOConn].
+	incoming <-chan msgOrErr
 
 	// If outgoiBatch has a positive capacity, it will be used to batch requests
 	// and notifications before sending.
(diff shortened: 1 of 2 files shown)


slimslenderslacks:
Had the same issue.  Confirming that this change fixes the issue.

mcp/transport.go
```diff
@@ -279,12 +281,47 @@ type ioConn struct {
...
 
 func newIOConn(rwc io.ReadWriteCloser) *ioConn {
+	var (
+		incoming = make(chan msgOrErr)
+		closed   = make(chan struct{})
+	)
+	// Start a goroutine for reads, so that we can select on the incoming channel
+	// in [ioConn.Read] and unblock the read as soon as Close is called (see #224).
+	//
+	// This leaks a goroutine, but that is unavoidable since AFAIK there is no
```
jba, line 303: There are returns in this function, so it doesn't _always_ leak a goroutine. When does it?

mcp/transport.go
```diff
@@ -478,7 +525,11 @@ func (t *ioConn) Write(ctx context.Context, msg jsonrpc.Message) error {
 }
 
 func (t *ioConn) Close() error {
-	return t.rwc.Close()
+	t.closeOnce.Do(func() {
```
jba, line 528: Does OnceValue make this nicer?

findleyr: I don't think so, because it requires an additional layer of indirection.

mcp/transport.go
findleyr: Ah, it leaks when the decoder is blocked in Read. Updated the comment.

jba: Did you push your change? I don't see it. 
Approving anyway.

Issue #244, opened 2025-08-06
proposal: jsonschema: marshal bool-equivalent schemas as bools


jba:
Marshal the empty schema as true and the negation of the empty schema as false.

We should be able to do this unconditionally, since they are equivalent.  And it will fix #230. 

If we needed to make it conditional, We could add a `MarshalAsBool` global function, or add it as a method on `Schema`. Would the global suffice? Perhaps not, if the same process talked to two different systems which each required a different setting.

Perhaps a better idea, though more clumsy, is to add an `Encoder` type, which could then be configured as needed. That would be the best way to handle the case of two incompatible systems. That will be familiar to users of `encoding/json`.

jba:
Done in #247.

PR #247, opened 2025-08-06
jsonschema: marshal to bools

### LINKED ISSUE #230: additionalProperties is boolean
additionalProperties is boolean for java-sdk
io.modelcontextprotocol.spec.McpSchema.JsonSchema

jba:
Marshal the empty schema to true, and its negation to false.

This is technically a breaking behavior change, but a properly written consumer of JSON Schema will not notice.

For #244.

Fixes #230.

### DIFF
--- a/jsonschema/schema_test.go
+++ b/jsonschema/schema_test.go
@@ -54,9 +54,9 @@ func TestJSONRoundTrip(t *testing.T) {
 	for _, tt := range []struct {
 		in, want string
 	}{
-		{`true`, `{}`}, // boolean schemas become object schemas
-		{`false`, `{"not":{}}`},
-		{`{"type":"", "enum":null}`, `{}`}, // empty fields are omitted
+		{`true`, `true`},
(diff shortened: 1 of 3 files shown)


jsonschema/schema_test.go
```diff
@@ -54,9 +54,9 @@ func TestJSONRoundTrip(t *testing.T) {
 	for _, tt := range []struct {
 		in, want string
 	}{
-		{`true`, `{}`}, // boolean schemas become object schemas
-		{`false`, `{"not":{}}`},
-		{`{"type":"", "enum":null}`, `{}`}, // empty fields are omitted
+		{`true`, `true`},
```
findleyr, line 57: Do we want to make both of these changes, or just 'false'?
I've not seen many examples of a schema being 'true', but have seen examples of a schema being 'false'.

jba: It seems odd to do one without the other.

findleyr (approved):

PR #249, opened 2025-08-06
jsonschema: move jsonschema to google repo script

samthanawalla:
Only for viewing.

DO NOT SUBMIT
DO NOT REVIEW

For #28


findleyr:
Looks approximately right--if it works, it works!

PR #253, opened 2025-08-06
mcp: add syntax and scheme validation to AddResourceTemplate

cryo-zd:
### change:
To avoid silent failures caused by invalid URI template, this PR:
1. Add template validation and scheme check in `AddResourceTemplate`, panic if the template is invalid or lacks a scheme
2. Add test cases
---
We recently encountered an internal issue where a malformed template was mistakenly added. It could never match any URI, and since the current `Matches` implementation returns `false` on parse errors without logging, it silently failed and made debugging harder.
If this validation is not aligned with intended behavior (e.g. if template correctness is considered caller responsibility), happy to close this PR. Just wanted to raise it after hitting a subtle failure in practice.
https://github.com/modelcontextprotocol/go-sdk/blob/b392875dc5aeb6da971fcf3da4b15179dd25d71c/mcp/resource.go#L157-L164


findleyr:
@samthanawalla could you help review?

mcp/server.go
```diff
@@ -220,7 +221,19 @@ func (s *Server) RemoveResources(uris ...string) {
 func (s *Server) AddResourceTemplate(t *ResourceTemplate, h ResourceHandler) {
 	s.changeAndNotify(notificationResourceListChanged, &ResourceListChangedParams{},
 		func() bool {
-			// TODO: check template validity.
+			// Validate the URI template syntax
+			_, err := uritemplate.New(t.URITemplate)
+			if err != nil {
+				panic(fmt.Errorf("URI template %s is invalid: %v", t.URITemplate, err))
```
samthanawalla: wrap the error with %w instead of %v

mcp/server.go
```diff
@@ -220,7 +221,19 @@ func (s *Server) RemoveResources(uris ...string) {
...
+			}
+			// Ensure the URI template has a valid scheme
+			u, err := url.Parse(t.URITemplate)
+			if err != nil {
+				panic(err) // url.Parse includes the URI in the error
+			}
+			if !u.IsAbs() {
+				panic(fmt.Errorf("URI template %s needs a scheme", t.URITemplate))
```
samthanawalla: use %q instead of %s to quote the string

mcp/server.go
samthanawalla: use %q instead of %s to quote the string

cryo-zd: All done. Thank you for the review!

samthanawalla (approved):

mcp/server_test.go
```diff
@@ -373,6 +373,42 @@ func TestServerCapabilities(t *testing.T) {
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			rt := ResourceTemplate{URITemplate: tt.template}
+
+			defer func() {
+				if r := recover(); r != nil {
+					if !tt.expectPanic {
+						t.Errorf("%s: unexpected panic: %v", tt.name, r)
+					}
+				} else {
```
jba, line 399: else if

jba (approved):
Thanks!

Issue #211, opened 2025-07-31
Proposal: add StreamableServerTransport.JSONResponse to support application/json format messages in StreamableHTTP Transport


sushi-small:
When I use Cursor to get the tools from my MCP server, the server responds with a text/event-stream format message. Cursor can't parse this message and then displays no tools. The Python SDK responds with an application/json format message, which Cursor can parse and display tools.

findleyr:
We can absolutely support returning application/json, but if Cursor behaves that was I don't think it is compliant with the spec, which [states](https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#sending-messages-to-the-server):

> The client MUST include an Accept header, listing both application/json and text/event-stream as supported content types.

The question is how to choose between application/json and text/event-stream. I suppose we'd need an option to indicate preference for one over the other.

findleyr:
Aha, this is exactly how the typescript SDK does it: an 'enableJsonResponse' option. 
https://github.com/modelcontextprotocol/typescript-sdk/blob/31acdcbb189056ec83d14a4f7a37ae2b1c67680e/src/server/streamableHttp.ts#L69

sushi-small:
Thanks for looking into this! The `enableJsonResponse` option from the typescript SDK sounds like exactly what we need. Would it be possible to add a similar option to the go-sdk?

findleyr:
Yes, I am adding such an option, though I'm calling it `JSONResponse`. Working on it today.

findleyr:
Concretely, the proposal is this:

```go
type StreamableServerTransportOptions struct {
   ...

  // When set, JSONResponse causes the transport to return responses to POST requests with a single 
  // application/json payload rather than text/event stream. This is incompatible with batched requests,
  // supported in the 2025-03-16 version of the spec.
  //
  // Any non-response messages from the server are delivered to the hanging GET stream.
  JSONResponse bool
}
```

samthanawalla:
I'd like to push back on this proposal. I think our API should abstract away the underlying transport details and this flag exposes a choice that I don't think users should need to make.

Moreover, I think accommodating a non-complaint client by deviating from the spec sets a bad precedent.

findleyr:
@samthanawalla to be clear, the spec allows for this behavior (returning application/json), and we're simply adding a way to activate it, and the typescript SDK provides an equivalent option.

findleyr:
[...]

PR #257, opened 2025-08-07
mcp: refactor streamable serving and support JSON responses

findleyr:
Add support for JSON responses in the streamable HTTP server, and simplify the streamable client. See commit messages for details.

Fixes https://github.com/modelcontextprotocol/go-sdk/issues/211

### DIFF
--- a/internal/jsonrpc2/frame.go
+++ b/internal/jsonrpc2/frame.go
@@ -12,12 +12,14 @@ import (
 	"io"
 	"strconv"
 	"strings"
+	"sync"
 )
 
 // Reader abstracts the transport mechanics from the JSON RPC protocol.
 // A Conn reads messages from the reader it was provided on construction,
 // and assumes that each call to Read fully transfers a single message,
 // or returns an error.
+//
 // A reader is not safe for concurrent use, it is expected it will be used by
 // a single Conn in a safe manner.
 type Reader interface {
@@ -29,8 +31,9 @@ type Reader interface {
 // A Conn writes messages using the writer it was provided on construction,
 // and assumes that each call to Write fully transfers a single message,
 // or returns an error.
-// A writer is not safe for concurrent use, it is expected it will be used by
-// a single Conn in a safe manner.
+//
+// A writer must be safe for concurrent use, as writes may occur concurrently
+// in practice: libraries may make calls or respond to requests asynchronously.
 type Writer interface {
 	// Write sends a message to the stream.
 	Write(context.Context, Message) error
@@ -62,7 +65,10 @@ func RawFramer() Framer { return rawFramer{} }
 
 type rawFramer struct{}
 type rawReader struct{ in *json.Decoder }
-type rawWriter struct{ out io.Writer }
+type rawWriter struct {
+	mu  sync.Mutex
+	out io.Writer
+}
 
 func (rawFramer) Reader(rw io.Reader) Reader {
 	return &rawReader{in: json.NewDecoder(rw)}
@@ -92,10 +98,14 @@ func (w *rawWriter) Write(ctx context.Context, msg Message) error {
 		return ctx.Err()
 	default:
 	}
+
 	data, err := EncodeMessage(msg)
 	if err != nil {
 		return fmt.Errorf("marshaling message: %v", err)
 	}
+
+	w.mu.Lock()
+	defer w.mu.Unlock()
 	_, err = w.out.Write(data)
 	return err
 }
@@ -107,7 +117,10 @@ func HeaderFramer() Framer { return headerFramer{} }
 
 type headerFramer struct{}
 type headerReader struct{ in *bufio.Reader }
-type headerWriter struct{ out io.Writer }
+type headerWriter struct {
+	mu  sync.Mutex
+	out io.Writer
+}
 
 func (headerFramer) Reader(rw io.Reader) Reader {
 	return &headerReader{in: bufio.NewReader(rw)}
@@ -180,6 +193,9 @@ func (w *headerWriter) Write(ctx context.Context, msg Message) error {
 		return ctx.Err()
 	default:
 	}
+	w.mu.Lock()
+	defer w.mu.Unlock()
+
 	data, err := EncodeMessage(msg)
 	if err != nil {
 		return fmt.Errorf("marshaling message: %v", err)
--- a/mcp/shared.go
+++ b/mcp/shared.go
@@ -275,6 +275,8 @@ func sessionMethod[S Session, P Params, R Result](f func(S, context.Context, P) 
 
 // Error codes
 const (
+	// TODO: should these be unexported?
+
 	CodeResourceNotFound = -32002
 	// The error code if the method exists and was called properly, but the peer does not support it.
 	CodeUnsupportedMethod = -31001
--- a/mcp/sse.go
+++ b/mcp/sse.go
@@ -155,7 +155,7 @@ func (t *SSEServerTransport) Connect(context.Context) (Connection, error) {
 	if err != nil {
 		return nil, err
 	}
-	return sseServerConn{t}, nil
+	return &sseServerConn{t: t}, nil
 }
 
 func (h *SSEHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
@@ -244,10 +244,10 @@ type sseServerConn struct {
 }
 
 // TODO(jba): get the session ID. (Not urgent because SSE transports have been removed from the spec.)
-func (s sseServerConn) SessionID() string { return "" }
+func (s *sseServerConn) SessionID() string { return "" }
 
 // Read implements jsonrpc2.Reader.
-func (s sseServerConn) Read(ctx context.Context) (jsonrpc.Message, error) {
+func (s *sseServerConn) Read(ctx context.Context) (jsonrpc.Message, error) {
 	select {
 	case <-ctx.Done():
 		return nil, ctx.Err()
@@ -259,7 +259,7 @@ func (s sseServerConn) Read(ctx context.Context) (jsonrpc.Message, error) {
 }
 
 // Write implements jsonrpc2.Writer.
-func (s sseServerConn) Write(ctx context.Context, msg jsonrpc.Message) error {
+func (s *sseServerConn) Write(ctx context.Context, msg jsonrpc.Message) error {
 	if ctx.Err() != nil {
 		return ctx.Err()
 	}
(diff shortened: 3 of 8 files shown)


mcp/streamable.go
```diff
@@ -390,47 +443,91 @@ func (t *StreamableServerTransport) servePOST(w http.ResponseWriter, req *http.R
...
 
+	var msgs []json.RawMessage
+	ctx := req.Context()
+	for msg, ok := range t.messages(ctx, stream, false) {
+		if !ok {
+			if ctx.Err() != nil {
+				w.WriteHeader(http.StatusNoContent)
+				return
+			} else {
+				http.Error(w, http.StatusText(http.StatusGone), http.StatusGone)
```
findleyr, line 498: return here!

mcp/streamable.go
```diff
@@ -464,65 +564,82 @@ func (t *StreamableServerTransport) streamResponse(stream *stream, w http.Respon
...
-			if err := t.opts.EventStore.Append(req.Context(), t.SessionID(), stream.id, data); err != nil {
-				return http.StatusInternalServerError, err.Error()
-			}
-			if !write(data) {
-				return 0, ""
+	ctx := req.Context()
+	for msg, ok := range t.messages(ctx, stream, persistent) {
+		if !ok {
+			if ctx.Err() != nil && writes == 0 {
+				// This probably doens't matter, but respond with NoContent if the client disconnected.
```
findleyr: doesn't

mcp/streamable.go
```diff
@@ -464,65 +564,82 @@ func (t *StreamableServerTransport) streamResponse(stream *stream, w http.Respon
...
+				w.WriteHeader(http.StatusNoContent)
+			} else {
+				errorf(http.StatusGone, "stream terminated")
 			}
+			return
+		}
+		if err := t.opts.EventStore.Append(req.Context(), t.SessionID(), stream.id, msg); err != nil {
+			errorf(http.StatusInternalServerError, "storing event: %v", err.Error())
```
findleyr, line 589: return

mcp/shared.go
```diff
@@ -275,6 +275,8 @@ func sessionMethod[S Session, P Params, R Result](f func(S, context.Context, P)
 
 // Error codes
 const (
+	// TODO: should these be unexported?
```
jba, line 278: I don't recall why they were exported, but we can certainly unexport them for now.

mcp/streamable.go
```diff
@@ -44,6 +47,10 @@ type StreamableHTTPHandler struct {
 type StreamableHTTPOptions struct {
 	// TODO: support configurable session ID generation (?)
 	// TODO: support session retention (?)
+
+	// TransportOptions sets the streamable server transport options to use when
+	// establishing a new session.
+	TransportOptions func(*http.Request) *StreamableServerTransportOptions
```
jba: Why is this a function on the request? (Use cases?)

mcp/streamable.go
```diff
@@ -390,47 +443,91 @@ func (t *StreamableServerTransport) servePOST(w http.ResponseWriter, req *http.R
...
+	var msgs []json.RawMessage
+	ctx := req.Context()
+	for msg, ok := range t.messages(ctx, stream, false) {
+		if !ok {
+			if ctx.Err() != nil {
+				w.WriteHeader(http.StatusNoContent)
+				return
+			} else {
+				http.Error(w, http.StatusText(http.StatusGone), http.StatusGone)
+			}
```
jba, line 500: return

mcp/streamable.go
```diff
@@ -390,47 +443,91 @@ func (t *StreamableServerTransport) servePOST(w http.ResponseWriter, req *http.R
...
+	var data []byte
+	if len(msgs) == 1 {
+		data = []byte(msgs[0])
+	} else {
+		// TODO: add tests for batch responses, or disallow them entirely.
+		var err error
+		data, err = json.Marshal(msgs)
+		if err != nil {
+			http.Error(w, fmt.Sprintf("internal error marshalling response: %v", err), http.StatusInternalServerError)
+		}
```
jba, line 514: return

mcp/streamable.go
findleyr: Just defensiveness. We found that parameterizing getServer was useful, and I guessed that it might be useful here, too. WDYT? YAGNI?

mcp/shared.go
findleyr: Will do in a separate CL.

mcp/streamable.go
jba: Let's think for a moment if there's a possible use. You can remove for now to make progress.

mcp/streamable.go
```diff
@@ -738,38 +862,78 @@ func (t *StreamableClientTransport) Connect(ctx context.Context) (Connection, er
...
-	c.protocolVersion = s
+	c.initializedResult = res
+	sessionID := c._sessionID
+	if sessionID == "" {
+		c.fail(fmt.Errorf("internal error: initialized without session id"))
+	}
+	c.mu.Unlock()
+
+	if sessionID != "" {
+		go c.handleSSE(sessionID, nil, true)
```
samthanawalla: One thing I am confused about is if we can only _start_ the initial SSE after a sessionID is set. 

See 2.5 Session Management: 
It says "A server using the Streamable HTTP transport MAY assign a session ID at initialization time"

So a session ID could technically be unset?

mcp/streamable.go
```diff
@@ -738,38 +862,78 @@ func (t *StreamableClientTransport) Connect(ctx context.Context) (Connection, er
 		ReconnectOptions: reconnOpts,
 		ctx:              connCtx,
 		cancel:           cancel,
+		failed:           make(chan struct{}),
 	}
+
 	// Start the persistent SSE listener right away.
 	// Section 2.2: The client MAY issue an HTTP GET to the MCP endpoint.
 	// This can be used to open an SSE stream, allowing the server to
 	// communicate to the client, without the client first sending data via HTTP POST.
```
samthanawalla: Keep this comment block with the initial handleSSE call

mcp/streamable.go
```diff
@@ -161,6 +176,15 @@ type StreamableServerTransportOptions struct {
...
 	EventStore EventStore
+
+	// JSONResponse, if set, tells the server to prefer to respond to requests
+	// using application/json responses rather than text/event-stream.
+	//
+	// Specifically, responses will be application/json whenever incoming POST
+	// request contain only a single message. In this case, notifications or
+	// requests made within the context of a server request will be sent to the
+	// hanging GET request, if any.
+	JSONResponse bool
```
samthanawalla: unexport

jba (approved):

mcp/streamable.go
```diff
@@ -797,56 +960,58 @@ func (s *streamableClientConn) Read(ctx context.Context) (jsonrpc.Message, error
...
 	}
 
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	sessionID := c._sessionID
 	if sessionID == "" {
-		// locked
-		s._sessionID = gotSessionID
+		sessionID, err := c.postMessage(ctx, "", data)
```
samthanawalla: This logic is a bit confusing:

It waits for a sessionID to be returned from postMessage but it could block in the case that handleSSE is called and is handling a stream.

It seems like we only _need_ to do this logic to wait for sessionID but the way it is written we are also waiting for the stream to complete.

mcp/streamable.go
```diff
@@ -390,47 +443,93 @@ func (t *StreamableServerTransport) servePOST(w http.ResponseWriter, req *http.R
...
-	// the response.
-	// (But how would we know there is only a single message? For example, couldn't
-	//  a progress notification be sent before a response on the same context?)
-	return t.streamResponse(stream, w, req, -1)
+	if stream == nil {
+		w.WriteHeader(http.StatusAccepted)
+		return
+	}
+
+	if stream.jsonResponse {
```
samthanawalla, line 478: Always use respondJSON here for the optimization

findleyr:
I believe I've figured out how to simplify this significantly. However, it's too hard to review. I'm going to break this up into smaller, logical PRs.

mcp/streamable.go
findleyr: I unexported.

mcp/streamable.go
findleyr: This is now simplified, with commentary.

mcp/streamable.go

mcp/streamable.go
findleyr: Should be a lot simpler now.

mcp/streamable.go
findleyr: FWIW, that's now quite right: we actually want to do this in handleSSE, if and only if we only get one response message. We don't know that yet, here.

I'll do this as a separate PR.

internal/jsonrpc2/frame.go
```diff
@@ -92,6 +98,9 @@ func (w *rawWriter) Write(ctx context.Context, msg Message) error {
 		return ctx.Err()
 	default:
 	}
+	w.mu.Lock()
```
jba: You don't need to lock the encode, I believe; only the write.

mcp/sse.go
```diff
@@ -259,7 +260,10 @@ func (s sseServerConn) Read(ctx context.Context) (jsonrpc.Message, error) {
 }
 
 // Write implements jsonrpc2.Writer.
-func (s sseServerConn) Write(ctx context.Context, msg jsonrpc.Message) error {
+func (s *sseServerConn) Write(ctx context.Context, msg jsonrpc.Message) error {
+	s.writeMu.Lock()
```
jba: Ditto.

mcp/streamable.go
```diff
@@ -738,164 +738,200 @@ func (t *StreamableClientTransport) Connect(ctx context.Context) (Connection, er
...
+	//
+	// § 2.5: A server using the Streamable HTTP transport MAY assign a session
+	// ID at initialization time, by including it in an Mcp-Session-Id header
+	// on the HTTP response containing the InitializeResult.
+	go c.handleSSE(nil, true)
+}
+
+// fail handles an asynchronous error while reading.
+//
+// If err is non-nil, it is terminal, and the connection will fault.
```
jba: Not sure what "fault" means here.

jba (approved):
Thanks for splitting it up!

mcp/sse.go
findleyr: D'oh. This made me realize that the write was *already* guarded.

jba (approved):

samthanawalla:
LGTM 👍

Issue #237, opened 2025-08-05
server-side auth support


jba:
Provide ways for an MCP server to authenticate incoming requests, and pass the resulting information on to handlers.

For authentication, add this to `ServerOptions`:
```
RequestVerifier func(*http.Request) (*AuthInfo, error)
```
The function can, for example, get a bearer token from the request and verify it using OAuth. It can then parse it or obtain user information another way. The `AuthInfo` type is TBD, but it should be flexible enough to allow extra information to be added during parsing.

Our HTTP transports can look at the error to decide whether to serve a resource URL with the appropriate status so that clients know to conduct the flow to obtain an access token.

Can the resource URL depend on the resource being requested? I don't think that's possible, because other SDKs' implementations like [TS's](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/server/auth/middleware/bearerAuth.ts) and [Python's](https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/server/auth/middleware/bearer_auth.py)   use middleware for this. So in that case, the resource URL can also be a server option.

Once we have the authentication info, how do we pass it down to handlers? Currently, handlers only take a context, a server session, and the request parameter struct. We could add a field to every request parameter struct that includes that information. Or we could add another parameter to handlers for request-scoped metadata like this.

ChesterTI:
I notice that the design doc mentions that "options to hook into low-level request handling for the purposes of authentication or context injection" are absent. Is this still the intention given that context injection seems to solve this issue quite nicely on the surface?

jba:
That's a good question. I don't think we need that for auth, and there are no other use cases I know of. But I'd like to get more eyes on this.

jba:
I ended up writing a traditional piece of middleware, as the TS SDK does. I edited the top post to reflect that.

jba:
Closing this; I think it's done.

PR #261, opened 2025-08-07
auth: add OAuth authenticating middleware for server

jba:
Add auth.RequireBearerToken and associated types.

This piece of middleware authenticates clients using OAuth 2.0, as specified in the MCP spec.

For #237.

Usage:
```
st := mcp.NewStreamableServerTransport(...)
http.Handle(path, auth.RequireBearerToken(verifier, nil)(st.ServeHTTP))
```

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -1,0 +1,96 @@
+// Copyright 2025 The Go MCP SDK Authors. All rights reserved.
+// Use of this source code is governed by an MIT-style
+// license that can be found in the LICENSE file.
+
+package auth
+
+import (
+	"context"
+	"errors"
+	"net/http"
+	"slices"
+	"strings"
+	"time"
+)
+
+type TokenInfo struct {
+	Scopes     []string
+	Expiration time.Time
+}
+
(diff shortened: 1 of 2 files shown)


auth/auth.go
```diff
@@ -0,0 +1,96 @@
...
+		// See typescript-sdk/src/server/auth/middleware/bearerAuth.ts.
+		return nil, err.Error(), http.StatusInternalServerError
+	}
+
+	// Check scopes.
+	if opts != nil {
+		// Note: quadratic, but N is small.
+		for _, s := range opts.Scopes {
+			if !slices.Contains(tokenInfo.Scopes, s) {
+				return nil, "insufficient scope", http.StatusUnauthorized
```
samthanawalla: Should a valid token + insufficient scope return StatusForbidden instead of StatusUnauthorized?

jba: Yes! Good catch. Fixed.

samthanawalla (dismissed):

samthanawalla:
Ah you have to update the tests

PR #266, opened 2025-08-08
mcp: changed streamID's from int64 to random strings

### LINKED ISSUE #259: stream IDs should be random strings
Now they are integers that are incremented per-process. This might not work well when servers are distributed: two servers can get the same stream ID, and then the event store (assuming it is global to all servers) can get confused.

2bitburrito:
- Changed StreamID's to store randomly generated strings. 
- Created a defaultStreamID field on the StreamableServerTransport struct which is generated and saved on instansiation.
- Updated all tests.

Fixes #259

### DIFF
--- a/mcp/streamable.go
+++ b/mcp/streamable.go
@@ -273,7 +273,7 @@ func (t *StreamableServerTransport) Connect(context.Context) (Connection, error)
 	//
 	// It is always text/event-stream, since it must carry arbitrarily many
 	// messages.
-	t.connection.streams[0] = newStream(0, false)
+	t.connection.streams[""] = newStream("", false)
 	if t.connection.eventStore == nil {
 		t.connection.eventStore = NewMemoryEventStore(nil)
 	}
@@ -331,7 +331,7 @@ func (c *streamableServerConn) SessionID() string {
 // at any time.
 type stream struct {
 	// id is the logical ID for the stream, unique within a session.
-	// ID 0 is used for messages that don't correlate with an incoming request.
+	// an empty string is used for messages that don't correlate with an incoming request.
 	id StreamID
 
 	// jsonResponse records whether this stream should respond with application/json
@@ -379,9 +379,9 @@ func signalChanPtr() *chan struct{} {
 	return &c
 }
 
-// A StreamID identifies a stream of SSE events. It is unique within the stream's
+// A StreamID identifies a stream of SSE events. It is globally unique.
 // [ServerSession].
-type StreamID int64
+type StreamID string
 
 // We track the incoming request ID inside the handler context using
 // idContextValue, so that notifications and server->client calls that occur in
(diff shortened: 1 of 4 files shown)


mcp/streamable.go
```diff
@@ -199,12 +200,11 @@ func (t *StreamableServerTransport) SessionID() string {
...
 // single session.
 type StreamableServerTransport struct {
-	nextStreamID atomic.Int64 // incrementing next stream ID
-
-	sessionID string
-	opts      StreamableServerTransportOptions
-	incoming  chan jsonrpc.Message // messages from the client to the server
-	done      chan struct{}
+	sessionID       string
+	defaultStreamID StreamID
```
jba: I think we'd just use the empty string for this.

mcp/streamable.go
```diff
@@ -283,9 +283,10 @@ func signalChanPtr() *chan struct{} {
 	return &c
 }
 
-// A StreamID identifies a stream of SSE events. It is unique within the stream's
+// A StreamID identifies a stream of SSE events. It is a random string
+// 26 characters in length
```
jba: Remove the sentence about the random string—that's an implementation detail. Add a sentence saying that it's globally unique.

mcp/streamable_test.go
```diff
@@ -222,9 +222,10 @@ func TestServerInitiatedSSE(t *testing.T) {
 
 	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
 	defer cancel()
-	client := NewClient(testImpl, &ClientOptions{ToolListChangedHandler: func(ctx context.Context, cc *ClientSession, params *ToolListChangedParams) {
-		notifications <- "toolListChanged"
-	},
+	client := NewClient(testImpl, &ClientOptions{
```
jba: This change seems unrelated, and it's a no-op; revert it.

jba (changes requested):

2bitburrito:
Ok - thanks for the notes. I apprecaite the feedback and apologies if there are any further issues. I haven't done much open-source stuff before. 

I have: 
- Removed the defaultStreamID implementation and just used an empty string for the default stream.
- Updated the comment to not include implementation details
- Fixed the auto-formatted issue
- Added an extra test in TestEventID that ensure's StreamID's can be empty strings.

Thanks, let me know how that looks and whether there's anything else you'd recommend. 

Hugh

mcp/streamable.go
```diff
@@ -182,7 +182,7 @@ func NewStreamableServerTransport(sessionID string, opts *StreamableServerTransp
 		streams:        make(map[StreamID]*stream),
 		requestStreams: make(map[jsonrpc.ID]StreamID),
 	}
-	t.streams[0] = newStream(0)
+	t.streams[StreamID("")] = newStream(StreamID(""))
```
jba: You shouldn't need to write StreamID in either place.

mcp/streamable.go
```diff
@@ -242,7 +240,7 @@ type StreamableServerTransport struct {
 // at any time.
 type stream struct {
 	// id is the logical ID for the stream, unique within a session.
-	// ID 0 is used for messages that don't correlate with an incoming request.
+	// defaultStreamID is used for messages that don't correlate with an incoming request.
```
jba: The empty string is used...

2bitburrito:
Hi Jonathan, I've made those changes now.

jba:
Now please resolve conflicts.

2bitburrito:
Hi Jonathan, 
All conflicts are resolved.
I have also sqashed that down to a single commit to keep it clean too. Let me know if there are any issues.

Issue #243, opened 2025-08-05
proposal: Use `XXXRequest` types in server handlers.


jba:
We decided early on to use `XXXParams` on the client and server, because the enclosing request is just the method name + the params, and asking client users to provide the method name "foo" and then call `Foo` seemed cruelly redundant. And we naturally did the same on the server, so server handlers like a ToolHandler take CallToolParams.

But what we missed is that our server infrastructure might want to add request-scoped information besides the params. Auth information is one example (#237). We could also expose all the HTTP headers. 

So while the client methods should continue to accept XXXParams, we should add the corresponding XXXRequest types that can include more, and have server handlers take XXXRequests. An XXXRequest struct will include the method name (just to match the spec) and the params, but may include more.

findleyr:
I'm not necessarily opposed to this, but I'm skeptical of adding so many more types, and of stashing information in the request objects (which are really just part of the JSON-RPC spec, not MCP).

Isn't the request context the correct place to transmit things like auth headers?

jba:
It could be a single generic type, parameterized by the Params type. 

But yes, it would share a name with the JSON-RPC request, even though it holds more.

I think it's generally better to avoid putting things in the context unless they really may not be needed until deep in the call stack. Logging is a good example. So is tracing, where you don't need the trace information until you make a subsequent RPC. I'm not sure auth is: authorizing to the MCP server is probably not related to authorizing to some downstream server. (Although I suppose it might be; I don't fully understand how it's all supposed to work.) In other words, a ToolHandler would probably want look at the AuthInfo early on.

I could see the context becoming a grab-bag of stuff (AuthInfo, request headers, a logger, ...). 

All that said, I'm not wedded to this and if you prefer using the context, I would at least suggest that we have a single struct for request-scoped things rather than a bunch of different context values.

findleyr:
> I think it's generally better to avoid putting things in the context unless they really may not be needed until deep in the call stack

I've thought about it more, and I generally agree. I guess the confusing part for me is that this object doesn't really represent the JSON-RPC request--it's more of an abstraction of the MCP request.

Do we keep the `context.Context` argument, or make it a field / method on the request. What about the `ServerSession`, which is infrequently used?

Here's one possible API:

```go
type Request[T Params] struct {
  Session *ServerSession
  Params T
  // whatever else we need
}

[...]

PR #267, opened 2025-08-08
mcp: introduce Requests

jba:
Handlers and client/server methods take a single request argument combining the session and params. This slightly simplifies the signature for handlers, since there is no longer an often-ignored session argument.

But more importantly, it opens the door to adding more information in requests, such as auth info and HTTP request headers.

For #243.

### DIFF
--- a/design/design.md
+++ b/design/design.md
@@ -431,12 +431,9 @@ We provide a mechanism to add MCP-level middleware on the both the client and se
 
 ```go
 // A MethodHandler handles MCP messages.
-// The params argument is an XXXParams struct pointer, such as *GetPromptParams.
-// For methods, a MethodHandler must return either an XXResult struct pointer and a nil error, or
-// nil with a non-nil error.
-// For notifications, a MethodHandler must return nil, nil.
-type MethodHandler[S Session] func(
-	ctx context.Context, _ *S, method string, params Params) (result Result, err error)
+// For methods, exactly one of the return values must be nil.
+// For notifications, both must be nil.
+type MethodHandler[S Session] func(ctx context.Context, method string, req *Request[S]) (result Result, err error)
 
 // Middleware is a function from MethodHandlers to MethodHandlers.
 type Middleware[S Session] func(MethodHandler[S]) MethodHandler[S]
@@ -620,8 +617,9 @@ type Tool struct {
 	Name string                    `json:"name"`
 }
 
-type ToolHandlerFor[In, Out any] func(context.Context, *ServerSession, *CallToolParamsFor[In]) (*CallToolResultFor[Out], error)
-type ToolHandler = ToolHandlerFor[map[string]any, any]
+// A ToolHandlerFor handles a call to tools/call with typed arguments and results.
+type ToolHandlerFor[In, Out any] func(context.Context, *ServerRequest[*CallToolParamsFor[In]]) (*CallToolResultFor[Out], error)
+
 ```
 
 Add tools to a server with the `AddTool` method or function. The function is generic and infers schemas from the handler
@@ -671,8 +669,8 @@ type AddParams struct {
     Y int `json:"y"`
 }
 
-func addHandler(ctx context.Context, ss *mcp.ServerSession, params *mcp.CallToolParamsFor[AddParams]) (*mcp.CallToolResultFor[int], error) {
-    return &mcp.CallToolResultFor[int]{StructuredContent: params.Arguments.X + params.Arguments.Y}, nil
+func addHandler(ctx context.Context, req *mcp.ServerRequest[*mcp.CallToolParamsFor[AddParams]]) (*mcp.CallToolResultFor[int], error) {
+    return &mcp.CallToolResultFor[int]{StructuredContent: req.Params.Arguments.X + req.Params.Arguments.Y}, nil
 }
 ```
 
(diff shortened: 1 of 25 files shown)


mcp/shared.go
```diff
@@ -333,6 +340,28 @@ func setProgressToken(p Params, pt any) {
 	m[progressTokenKey] = pt
 }
 
+// A Request is a method request with parameters and additional information, such as the session.
+// The parameters are untyped; see [RequestFor] for the typed version.
+type Request[S Session] struct {
```
findleyr: Just a thought, but would it be cleaner to have ClientRequest and ServerRequest?

Not only will it be slightly shorter, but there may be fields on request that only relate to the client, or server.

mcp/shared.go
```diff
@@ -333,6 +340,28 @@ func setProgressToken(p Params, pt any) {
...
+	Session S
+	Params  Params
+}
+
+// A RequestFor is a method request with parameters and additional information, such as the session.
+// The parameters are typed; see [Request] for the untyped version.
+type RequestFor[S Session, P Params] struct {
+	Session S
+	Params  P
+	// TODO: HTTPHeader http.Header
```
findleyr: Isn't this transport specific metadata? Do we want to hard-code it in this way?
It's also server-specific, and doesn't make sense for the client.

I think it should be an arbitrary annotation API.

mcp/shared.go
jba: This is the type that is used in the middleware system. To split into two types would mean splitting middleware.
Current state:
```
type MethodHandler[S Session] func(ctx context.Context, method string, req *Request[S]) (result Result, err error)
type Middleware[S Session] func(MethodHandler[S]) MethodHandler[S] 
func (c *Client) AddSendingMiddleware(middleware ...Middleware[*ClientSession])
func (c *Client) AddReceivingMiddleware(middleware ...Middleware[*ClientSession]) 
func (s *Server) AddSendingMiddleware(middleware ...Middleware[*ServerSession])
func (s *Server) AddReceivingMiddleware(middleware ...Middleware[*ServerSession])
```
You're asking to double that.

mcp/shared.go
jba: An "arbitrary annotation API" is a `map[string]any`. You give up compile-time name and type checking. And you also give up the ability to read the godoc and know what might be there and what definitely isn't there.

I'd rather have fields that are just empty if unused. The header is only non-nil on servers that use an HTTP transport. That's fine, users will just check. That's still better than a grab-bag IMO.

mcp/shared.go
jba: Oh, wait, you're not actually splitting all of it, just MethodHandler and Middleware. That still seems like a lot. But we could land this first, then I'll try that?

jba: I'll try splitting RequestFor. That will be what most people will see anyway; not many people will write middleware.

jba: Latest commit splits RequestFor into two parts. Leaving Request as it is.

jba: I like the separate Client and Server requests.

mcp/shared.go
jba: This is partly obsolete: with separate client and server requests, HTTP headers will not appear in client requests.

design/design.md
```diff
@@ -671,8 +669,8 @@ type AddParams struct {
     Y int `json:"y"`
 }
 
-func addHandler(ctx context.Context, ss *mcp.ServerSession, params *mcp.CallToolParamsFor[AddParams]) (*mcp.CallToolResultFor[int], error) {
-    return &mcp.CallToolResultFor[int]{StructuredContent: params.Arguments.X + params.Arguments.Y}, nil
+func addHandler(ctx context.Context, req *mcp.RequestFor[*mcp.ServerSession, *mcp.CallToolParamsFor[AddParams]]) (*mcp.CallToolResultFor[int], error) {
```
findleyr: My only objection at this point is that we still have `mcp.Request` and `mcp.RequestFor`. What would be the cost of eliminating them?

jba: And replacing them with what?

jba: Ah, that only exists in the design doc. Fixed.

mcp/shared.go
```diff
@@ -333,6 +358,36 @@ func setProgressToken(p Params, pt any) {
...
+	Params  P
+}
+
+// A ServerRequest is a request to a server.
+type ServerRequest[P Params] struct {
+	Session *ServerSession
+	Params  P
+}
+
+func (*ClientRequest[P]) isRequest() {}
```
findleyr, line 402: I don't think these are used, given that you settled on a concrete Request type.

mcp/shared.go
findleyr: Ok, here's the problem that I see: eventually, we will want to encode information such as auth headers in the request, which means that it will be the responsibility of ServerSession.handle to map this information into the handler. But that means that ServerSession.handle will need to pack everything into a ServerRequest, and that needs to be transmitted through the method handler. But the method handler  does not handle a ServerRequest, it handles a Request[*ServerSession]. In other words, you're necessarily piping this information through the Request type, which would lose any specialization for servers.

Therefore, I think we need to either make the *entire stack* of ServerXXX objects, or revert back to RequestFor[Session, Params], or make Request an interface.

findleyr (commented):
Sorry this is so difficult. Perhaps we should meet to discuss.

mcp/shared.go
findleyr: One advantage of simply having Request be an interface is that users can reuse middleware for both the client and server, without needing a wrapper.

We can remove the `*ClientSession | *ServerSession` from the session interface, so that it's an ordinary interface, and add an accessor. Note that Session is already closed since it has unexported methods, so we don't need to explicitly mark it as a union.

mcp/shared.go
```diff
@@ -194,7 +195,10 @@ type methodInfo struct {
 // - R: results
 
 // A typedMethodHandler is like a MethodHandler, but with type information.
-type typedMethodHandler[S Session, P Params, R Result] func(context.Context, S, P) (R, error)
+type (
+	typedClientMethodHandler[P Params, R Result] func(context.Context, *ClientRequest[P]) (R, error)
```
findleyr, line 199: These seem like they could be one type with a third type parameter.

findleyr (approved):
LGTM, thanks!

mcp/shared.go
jba: I think there is a lot of cleanup of unexported code I could do. Tomorrow I'll follow up on this and others.

PR #274, opened 2025-08-09
mcp: clarify Server.Run use

jba:
Make it clear that it's not needed for an HTTP server.

Fixes #273.


andig:
Perfect, thank you 👍🏻

Issue #205, opened 2025-07-30
test Content unmarshal on nil


jba:
We have a couple of UnmarshalJSON methods for Content. Check that they don't panic on nil.

MrGossett:
I’m interested in helping with this.

MrGossett:
I took the liberty of putting up a draft PR here: https://github.com/modelcontextprotocol/go-sdk/pull/275

Happy to adjust or close the PR at your direction.

MrGossett:
This should probably be closed now that #275 is merged.

findleyr:
Indeed, thanks!

PR #275, opened 2025-08-09
mcp: add tests for UnmarshalJSON handling of nil Content pointers

MrGossett:
Check that UnmarshalJSON methods for Content don't panic on nil.

- Introduces a new test file `content_nil_test.go` which verifies that `UnmarshalJSON` methods for various `Content` types do not panic when unmarshaling onto `nil` pointers.
- Adds a `nil` check in `contentFromWire` function to guard against a `nil` `wire.Content` parameter.
- Tests cover different scenarios, including valid and invalid content types, as well as cases with empty or missing content fields.

For [modelcontextprotocol#205](https://github.com/modelcontextprotocol/go-sdk/issues/205)

### DIFF
--- a/mcp/content.go
+++ b/mcp/content.go
@@ -252,6 +252,9 @@ func contentsFromWire(wires []*wireContent, allow map[string]bool) ([]Content, e
 }
 
 func contentFromWire(wire *wireContent, allow map[string]bool) (Content, error) {
+	if wire == nil {
+		return nil, fmt.Errorf("content wire is nil")
+	}
 	if allow != nil && !allow[wire.Type] {
 		return nil, fmt.Errorf("invalid content type %q", wire.Type)
 	}
--- a/mcp/content_nil_test.go
+++ b/mcp/content_nil_test.go
@@ -1,0 +1,251 @@
+// Copyright 2025 The Go MCP SDK Authors. All rights reserved.
+// Use of this source code is governed by an MIT-style
+// license that can be found in the LICENSE file.
+
+// This file contains tests to verify that UnmarshalJSON methods for Content types
+// don't panic when unmarshaling onto nil pointers, as requested in GitHub issue #205.
+//
+// NOTE: The contentFromWire function has been fixed to handle nil wire.Content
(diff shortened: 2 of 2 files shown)


mcp/content_nil_test.go
```diff
@@ -0,0 +1,251 @@
...
+				}
+			}()
+
+			err := json.Unmarshal([]byte(tt.json), tt.content)
+			if err != nil {
+				t.Errorf("UnmarshalJSON failed: %v", err)
+			}
+
+			// Verify that the Content field was properly populated
+			switch v := tt.content.(type) {
```
findleyr: I think you can simplify this significantly using cmp.Diff: just compare the expected and actual value.
https://pkg.go.dev/github.com/google/go-cmp/cmp is already required by this module.

mcp/content.go
```diff
@@ -252,6 +252,9 @@ func contentsFromWire(wires []*wireContent, allow map[string]bool) ([]Content, e
 }
 
 func contentFromWire(wire *wireContent, allow map[string]bool) (Content, error) {
+	if wire == nil {
```
findleyr, line 255: The test failed without this change, right?

MrGossett: That's correct. One of the tests that I added failed without this change.

The test is unmarshaling `{"model":"test","role":"user"}` (missing `"content"`), which causes the `*wireContent` parameter here to be nil.

Below is the test output without this change:

```text
$ go test ./mcp   
--- FAIL: TestContentUnmarshalNilWithEmptyContent (0.00s)
    --- FAIL: TestContentUnmarshalNilWithEmptyContent/Missing_Content_field (0.00s)
        content_nil_test.go:170: UnmarshalJSON panicked: runtime error: invalid memory address or nil pointer dereference
FAIL
FAIL    github.com/modelcontextprotocol/go-sdk/mcp      7.532s
FAIL
```

mcp/content_nil_test.go
MrGossett: Done in a56a3cc

findleyr (approved):
Thanks!

PR #276, opened 2025-08-09
jsonrpc: paper over race in jsonrpc2 test

jba:
In jsonprc2_test.go, binder.Bind starts a goroutine. That goroutine begins to run during jsonrpc2.bindConnection, and can race with the setting of conn.write in bindConnection.

This PR adds a sleep, which is a poor way to deal with the race, but the least invasive change. Better ones include running the test function after Dial returns, or adding a Connection.Ready method to detect when initialization is complete.


findleyr:
Ok for now, thanks. I'll investigate a proper fix.

PR #277, opened 2025-08-09
mcp: support stateless streamable sessions

findleyr:
Support stateless streamable sessions by adding a GetSessionID function to StreamableHTTPOptions. If GetSessionID returns "", the session is stateless, and no validation is performed. This is implemented by providing the session a trivial initialization state.

To implement this, some parts of #232 (distributed sessions) are copied over, since they add an API for creating an already-initialized session.

In total, the following new API is added:
- StreamableHTTPOptions.GetSessionID
- ServerSessionOptions (a new parameter to Server.Connect)
- ServerSessionState
- ClientSessionOptions (a new parameter to Client.Connect, for symmetry)

For #10

### DIFF
--- a/mcp/server.go
+++ b/mcp/server.go
@@ -301,7 +301,7 @@ func (s *Server) listPrompts(_ context.Context, _ *ServerSession, params *ListPr
 	})
 }
 
-func (s *Server) getPrompt(ctx context.Context, cc *ServerSession, params *GetPromptParams) (*GetPromptResult, error) {
+func (s *Server) getPrompt(ctx context.Context, ss *ServerSession, params *GetPromptParams) (*GetPromptResult, error) {
 	s.mu.Lock()
 	prompt, ok := s.prompts.get(params.Name)
 	s.mu.Unlock()
@@ -309,7 +309,7 @@ func (s *Server) getPrompt(ctx context.Context, cc *ServerSession, params *GetPr
 		// TODO: surface the error code over the wire, instead of flattening it into the string.
 		return nil, fmt.Errorf("%s: unknown prompt %q", jsonrpc2.ErrInvalidParams, params.Name)
 	}
-	return prompt.handler(ctx, cc, params)
+	return prompt.handler(ctx, ss, params)
 }
 
 func (s *Server) listTools(_ context.Context, _ *ServerSession, params *ListToolsParams) (*ListToolsResult, error) {
@@ -518,7 +518,7 @@ func (s *Server) unsubscribe(ctx context.Context, ss *ServerSession, params *Uns
 // It need not be called on servers that are used for multiple concurrent connections,
 // as with [StreamableHTTPHandler].
 func (s *Server) Run(ctx context.Context, t Transport) error {
-	ss, err := s.Connect(ctx, t)
+	ss, err := s.Connect(ctx, t, nil)
 	if err != nil {
 		return err
 	}
@@ -539,8 +539,12 @@ func (s *Server) Run(ctx context.Context, t Transport) error {
 
 // bind implements the binder[*ServerSession] interface, so that Servers can
 // be connected using [connect].
-func (s *Server) bind(conn *jsonrpc2.Connection) *ServerSession {
-	ss := &ServerSession{conn: conn, server: s}
+func (s *Server) bind(mcpConn Connection, conn *jsonrpc2.Connection, state *ServerSessionState) *ServerSession {
+	assert(mcpConn != nil && conn != nil, "nil connection")
+	ss := &ServerSession{conn: conn, mcpConn: mcpConn, server: s}
+	if state != nil {
+		ss.state = *state
+	}
 	s.mu.Lock()
 	s.sessions = append(s.sessions, ss)
 	s.mu.Unlock()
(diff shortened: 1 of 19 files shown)


mcp/server.go
```diff
@@ -561,32 +562,46 @@ func (s *Server) disconnect(cc *ServerSession) {
...
+
 // Connect connects the MCP server over the given transport and starts handling
 // messages.
 //
 // It returns a connection object that may be used to terminate the connection
 // (with [Connection.Close]), or await client termination (with
 // [Connection.Wait]).
-func (s *Server) Connect(ctx context.Context, t Transport) (*ServerSession, error) {
+//
+// If state is non-nil, it is the initial state for the server.
```
jba: opts.State

mcp/server.go
```diff
@@ -561,32 +562,46 @@ func (s *Server) disconnect(cc *ServerSession) {
...
+func (s *Server) Connect(ctx context.Context, t Transport, opts *ServerSessionOptions) (*ServerSession, error) {
```
jba, line 581: You're not using opts?

mcp/streamable.go
```diff
@@ -45,6 +45,12 @@ type StreamableHTTPHandler struct {
 // StreamableHTTPOptions is a placeholder options struct for future
 // configuration of the StreamableHTTP handler.
 type StreamableHTTPOptions struct {
+	// GetSessionID provides the next session ID to use for an incoming request.
+	//
+	// If GetSessionID returns an empty string, the session is 'stateless',
+	// meaning it is not persisted and no session validation is performed.
+	GetSessionID func() string
```
jba, line 51: Why do we want the user to control this? (We could do statelessness with just a bool.)

mcp/transport.go
```diff
@@ -60,11 +60,28 @@ type Connection interface {
...
+// If client connections implement this interface, they may receive information
+// about changes to the client session.
+//
+// TODO: should this interface be exported?
 type clientConnection interface {
 	Connection
-	initialized(*InitializeResult)
+
+	// SessionUpdated is called whenever the client session state changes.
+	sessionUpdated(clientSessionState)
```
jba, line 73: Why do we need this general mechanism? It's only used for initialized.

mcp/streamable.go
```diff
@@ -45,6 +45,12 @@ type StreamableHTTPHandler struct {
 // StreamableHTTPOptions is a placeholder options struct for future
```
jba: no longer just a placeholder

jba (commented):
I'm confused about why the server is choosing statelessness. I thought there is some mechanism by which the client forces it, for example by sending a tool call without initialization.

mcp/transport.go
findleyr: Just for symmetry. Since this isn't exported, I thought it was simpler.

mcp/server.go
findleyr: done.

mcp/server.go
findleyr: D'oh! Fixed now, and I fixed the test to actually test!

mcp/streamable.go
```diff
@@ -42,9 +42,14 @@ type StreamableHTTPHandler struct {
...
 }
 
-// StreamableHTTPOptions is a placeholder options struct for future
-// configuration of the StreamableHTTP handler.
+// StreamableHTTPOptions configures the StreamableHTTPHandler.
 type StreamableHTTPOptions struct {
+	// GetSessionID provides the next session ID to use for an incoming request.
+	//
+	// If GetSessionID returns an empty string, the session is 'stateless',
+	// meaning it is not persisted and no session validation is performed.
```
jba, line 50: Describe default.

mcp/transport.go
jba: OK.

BC-ACherednichenko:
@findleyr @jba  sorry a bit late on this PR, i see that session ID is still getting verified in case header is provided, is it intentional?
My thinknig process is that in case we are in a stateless mode we can skip validation regardless of headers and etc, and treat every request as a new one, fetching the header seems to be limiting this capability
 
```
	var session *StreamableServerTransport
	if id := req.Header.Get(sessionIDHeader); id != "" {
		h.sessionsMu.Lock()
		session, _ = h.sessions[id]
		h.sessionsMu.Unlock()
		if session == nil {
			http.Error(w, "session not found", http.StatusNotFound)
			return
		}
	}
```

Wanted to check with you here before jumping into the issue

findleyr:
@BC-ACherednichenko but why would the client provide a session ID if the server didn't return one? I guess I'm not understanding the use-case.

BC-ACherednichenko:
@findleyr it is still related to the https://github.com/modelcontextprotocol/go-sdk/issues/148, since we put on ice the distributed sessions interface we have been wondering if it is possible to solve this problem via stateless mode

In particular, if stateless mode could be completely without session validation or etc (sort of you pass the flag, and sdk acts as if sessions did not exist at all), the mcp-session-id emission, management and distributed storage could have been moved to tools middleware layer (upper layer of the server). 

Currently it reacts right away on the service header `mcp-session-id`, which i feel can be more flexible and could be managed at the concrete server implementation discretion

Issue #281, opened 2025-08-11
lint issues in codebase


krtkvrm:
running `golangci-lint` fetched these lint errors
```
$ golangci-lint run

.
.
.

62 issues:
* errcheck: 50
* ineffassign: 2
* staticcheck: 7
* unused: 3
```
we should ideally solve this, not urgently but in trenches

raw log:
```sh
internal/jsonrpc2/conn.go:517:20: Error return value of `c.internalErrorf` is not checked (errcheck)
                        c.internalErrorf("Read returned an unexpected message of type %T", msg)
                                        ^
internal/jsonrpc2/conn.go:572:18: Error return value of `c.processResult` is not checked (errcheck)
                c.processResult("acceptRequest", req, nil, err)
                               ^
internal/jsonrpc2/conn.go:585:19: Error return value of `c.processResult` is not checked (errcheck)
                        c.processResult("Preempt", req, result, err)
                                       ^
internal/jsonrpc2/conn.go:625:18: Error return value of `c.processResult` is not checked (errcheck)
                c.processResult("acceptRequest", req, nil, err)
                               ^
internal/jsonrpc2/conn.go:677:19: Error return value of `c.internalErrorf` is not checked (errcheck)
                c.internalErrorf("%#v returned a non-nil result with a non-nil error for %s:\n%v\n%#v", from, req.Method, err, result)
                                ^
internal/jsonrpc2/jsonrpc2_test.go:139:17: Error return value of `listener.Close` is not checked (errcheck)
                listener.Close()
                              ^
internal/jsonrpc2/jsonrpc2_test.go:140:14: Error return value of `server.Wait` is not checked (errcheck)
                server.Wait()
                           ^
internal/jsonrpc2/jsonrpc2_test.go:149:24: Error return value of `h.conn.Close` is not checked (errcheck)
                                        defer h.conn.Close()
                                                          ^
internal/jsonrpc2/jsonrpc2_test.go:159:15: Error return value of `client.Wait` is not checked (errcheck)
                        client.Wait()
                                   ^
internal/jsonrpc2/jsonrpc2_test.go:378:19: Error return value of `h.conn.Respond` is not checked (errcheck)
                                h.conn.Respond(req.ID, true, nil)
                                              ^
internal/jsonrpc2/jsonrpc2_test.go:380:19: Error return value of `h.conn.Respond` is not checked (errcheck)
                                h.conn.Respond(req.ID, nil, ctx.Err())
                                              ^
[...]

PR #282, opened 2025-08-11
ci: add Makefile and integrate with GitHub workflows

krtkvrm:
Summary
 - Add Makefile with common CI tasks (fmt, test, race, lint, ci)
 - Replace manual commands in GitHub workflows with make targets
 - Provides consistent development experience between local and CI

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,14 +18,7 @@ jobs:
     - name: Set up Go
       uses: actions/setup-go@v5
     - name: Check formatting
-      run: |
-        unformatted=$(gofmt -l .)
-        if [ -n "$unformatted" ]; then
-          echo "The following files are not properly formatted:"
-          echo "$unformatted"
-          exit 1
-        fi
-        echo "All Go files are properly formatted"
+      run: make fmt
 
   test:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


krtkvrm:
partially helping solve https://github.com/modelcontextprotocol/go-sdk/issues/281

jba:
I'm sorry, but this feels like overkill. You've got a complex makefile that people must understand and maintain, but there is no compelling reason for make—no real dependencies other than what the go command already handles.

If you want to add something to our workflow, add `go vet` and `staticcheck` to our lint step.

krtkvrm:
Thanks for the feedback! I understand the complexity concern. The Makefile is just shortcuts so contributors don't have to remember `go test -v ./...` every time or figure out the race detector command. Makes it easier to run the same stuff that CI does locally.

Will definitely add `go vet` and `staticcheck` to the lint step. Mind if `golangci-lint` gets added too? It catches a lot of issues that the others miss.

Also, would you be open to keeping just the basic targets like `make test` and `make lint`?

jba:
Sorry, maybe I wasn't clear. No makefile.

Also, no golangci-lint. It has too many false positives.

wkoszek:
@jba having some ci.sh or a make shortcut would be good for contributors. Instead of wasting time at #168 because I forgot some space or a tab, it'd be nice to just run a simple script to build/test everything for me.

krtkvrm:
@jba closing this as I see `staticcheck` is added in CI

Issue #272, opened 2025-08-09
Proposal: make all transports open structs


findleyr:
It may be late for this change, but I think we can simplify the transport API, and I'd like to strongly consider doing so.

Unlike other SDKs, we made the decision to have a `Connect` method on the `Transport` interface, rather than the stateful `Start`.  This means that all of our transports are inherently stateless: when `Connect` is called, they create a stateful connection.

But that means there's really no reason for having constructors with options. Our transports could just be open structs. For example:

```go
type CommandTransport struct {
  Command *exec.Cmd
}

type StreamableClientTransport struct {
  URL string
  HTTPClient *http.Client
  ReconnectOptions *StreamableReconnectOptions
}
```

etc. etc.

By my count, this could eliminate 10 symbols.
- NewCommandTransport
- NewLoggingTransport
- NewSSEServerTransport
- NewSSEClientTransport
- SSEClientTransportOptions
- NewStreamableServerTransport
- StreamableServerTransportOptions
- NewStreamableClientTransport
- StreamableClientTransportOptions
- NewStdioTransport

Furthermore, we've been inconsistent about which transports get options, and I can foresee us wanting to add options to *all of them*. Making this change now allows us to be future-proof, without having to further pollute the namespace.

v0.3.0 is necessarily going to contain other breaking changes. I think we should consider these as well. If we want to ease the breakage, we can deprecate the NewXXX constructors and remove them in a later release (they are not incompatible with this change).

CC @jba @samthanawalla

jba:
Sounds good to me. This seems like a relatively minor change.

I'd remove the constructors  in v0.3.0

findleyr:
Since this is done, and no objection to the proposal, I'm going to mark it as accepted and fixed.
I'll open a separate issue for removing the transport constructors prior to the release.

PR #286, opened 2025-08-12
mcp: make transports open structs

findleyr:
As described in https://github.com/modelcontextprotocol/go-sdk/issues/272, there is really no reason for transports to be closed structs with constructors, since their state must be established by a call to Connect. Making them open structs simplifies their APIs, and means that all transports can be extended in the future: we don't have to create empty Options structs just for the purpose of future compatibility.

For now, the related constructors and options structs are simply deprecated (with go:fix directives where possible). A future CL will remove them prior to the v1.0.0 release.

For https://github.com/modelcontextprotocol/go-sdk/issues/272

### DIFF
--- a/design/design.md
+++ b/design/design.md
@@ -100,11 +100,7 @@ The `CommandTransport` is the client side of the stdio transport, and connects b
 ```go
 // A CommandTransport is a [Transport] that runs a command and communicates
 // with it over stdin/stdout, using newline-delimited JSON.
-type CommandTransport struct { /* unexported fields */ }
-
-// NewCommandTransport returns a [CommandTransport] that runs the given command
-// and communicates with it over stdin/stdout.
-func NewCommandTransport(cmd *exec.Command) *CommandTransport
+type CommandTransport struct { Command *exec.Command }
 
 // Connect starts the command, and connects to it over stdin/stdout.
 func (*CommandTransport) Connect(ctx context.Context) (Connection, error) {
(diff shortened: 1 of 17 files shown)


findleyr:
PTAL and let me know what you think about this change.

design/design.md
```diff
@@ -257,8 +230,10 @@ func NewInMemoryTransports() (*InMemoryTransport, *InMemoryTransport)
 
 // A LoggingTransport is a [Transport] that delegates to another transport,
 // writing RPC logs to an io.Writer.
-type LoggingTransport struct { /* ... */ }
-func NewLoggingTransport(delegate Transport, w io.Writer) *LoggingTransport
+type LoggingTransport struct { 
+    Delegate Transport
```
jba, line 234: I never liked `Delegate` because we don't commonly use it. And now it is part of the API, not just an arg name. Consider `Base` (https://pkg.go.dev/golang.org/x/oauth2#Transport) or just `Transport` (my pref).

mcp/transport.go
```diff
@@ -208,24 +211,28 @@ func call(ctx context.Context, conn *jsonrpc2.Connection, method string, params
 // A LoggingTransport is a [Transport] that delegates to another transport,
 // writing RPC logs to an io.Writer.
 type LoggingTransport struct {
-	delegate Transport
-	w        io.Writer
+	Delegate Transport
```
jba: ditto

jba:
LGTM.

mcp/streamable.go
```diff
@@ -212,22 +234,36 @@ type StreamableServerTransportOptions struct {
 	// requests made within the context of a server request will be sent to the
 	// hanging GET request, if any.
 	jsonResponse bool
+
+	//
```
samthanawalla: blank comment

design/design.md
findleyr: Done.

mcp/transport.go
findleyr: Done.

mcp/streamable.go
findleyr: Done

findleyr:
PTAL: I renamed LoggingTransport.Delegate to Transport, and updated all callsites.

samthanawalla (approved):

PR #295, opened 2025-08-13
mcp/tool: duplicate tools should not error

### LINKED ISSUE #217: Adding the same tool results in a jsonschema: Resolve: root already resolved error
**Describe the bug**

Adding the same tool results in a jsonschema: Resolve: root already resolved error

Despite the code saying that adding the same tool is a noop. https://github.com/modelcontextprotocol/go-sdk/blob/b34ba2149ddefc5ea0318240fc096d032b472216/mcp/server.go#L166

**To Reproduce**
1. Add the same tool twice to the server

**Expected behavior**
No error should occur.

**Logs**
If applicable, add logs to help explain your problem.

**Additional context**

KamdynS:
mcp/tool: duplicate tools should not error

These changes touch SetSchema, since that was the best place that I could find resolve this issue. 

This is my first time contributing to open source, I hope that I've followed expected formatting and if not I would happy to fix whatever you all need :)

All tests are passing and I duplicated tools in one of the examples(was hello/) and it ran without error. 

Fixes #217

### DIFF
--- a/mcp/tool.go
+++ b/mcp/tool.go
@@ -97,31 +97,11 @@ func setSchema[T any](sfield **jsonschema.Schema, rfield **jsonschema.Resolved) 
 	var err error
 	if *sfield == nil {
 		*sfield, err = jsonschema.For[T](nil)
-		if err != nil {
-			return err
-		}
 	}
-	// Resolve operates with internal state and cannot be called twice on the same
-	// Schema instance. Deep-copy the schema before resolving to avoid mutating the
-	// original (which may be reused when adding the same tool again).
-	cloned, err := func(orig *jsonschema.Schema) (*jsonschema.Schema, error) {
-		if orig == nil {
-			return nil, nil
-		}
-		b, err := json.Marshal(orig)
-		if err != nil {
-			return nil, err
-		}
-		var cp jsonschema.Schema
-		if err := json.Unmarshal(b, &cp); err != nil {
-			return nil, err
-		}
-		return &cp, nil
-	}(*sfield)
 	if err != nil {
 		return err
 	}
-	*rfield, err = cloned.Resolve(&jsonschema.ResolveOptions{ValidateDefaults: true})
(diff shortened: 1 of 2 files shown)


mcp/mcp_test.go
```diff
@@ -938,4 +938,30 @@ func TestKeepAliveFailure(t *testing.T) {
 	t.Errorf("expected connection to be closed by keepalive, but it wasn't. Last error: %v", err)
 }
 
+func TestAddTool_DuplicateNoPanicAndNoDuplicate(t *testing.T) {
+	// Adding the same tool pointer twice should not panic and should not
+	// produce duplicates in the server's tool list.
+	_, cs := basicConnection(t, func(s *Server) {
+		tool := &Tool{Name: "dup", InputSchema: &jsonschema.Schema{}}
+		s.AddTool(tool, nopHandler)
+		s.AddTool(tool, nopHandler)
```
findleyr: The bug was only in the json schema resolution, right? Meaning, it would also be reprouced if we use a different Tool instance? In that case, use a different description and assert that the tool value is the last tool that was added, by checking its description.

mcp/tool.go
```diff
@@ -97,11 +97,31 @@ func setSchema[T any](sfield **jsonschema.Schema, rfield **jsonschema.Resolved)
...
+		}
 	}
+	// Resolve operates with internal state and cannot be called twice on the same
+	// Schema instance. Deep-copy the schema before resolving to avoid mutating the
+	// original (which may be reused when adding the same tool again).
+	cloned, err := func(orig *jsonschema.Schema) (*jsonschema.Schema, error) {
+		if orig == nil {
+			return nil, nil
+		}
+		b, err := json.Marshal(orig)
```
findleyr: I'm not sure if this is the best way to handle this, but @jba will know. Added him as a reviewer.

findleyr (commented):
Thanks!

mcp/mcp_test.go
KamdynS: Good suggestion. Made the change and tests are passing.

PR #296, opened 2025-08-13
internal/testing: fake auth server

jba:
This is a fake OAuth authentication server, for use in testing.

### DIFF
--- a/internal/testing/fake_auth_server.go
+++ b/internal/testing/fake_auth_server.go
@@ -34,6 +34,8 @@ type state struct {
 	authCodes map[string]authCodeInfo
 }
 
+// NewFakeAuthMux constructs a ServeMux that implements an OAuth 2.1 authentication
+// server. It should be used with [httptest.NewTLSServer].
 func NewFakeAuthMux() *http.ServeMux {
 	s := &state{authCodes: make(map[string]authCodeInfo)}
 	mux := http.NewServeMux()
@@ -44,6 +46,7 @@ func NewFakeAuthMux() *http.ServeMux {
 }
 
 func (s *state) handleMetadata(w http.ResponseWriter, r *http.Request) {
+	issuer := "https://localhost:" + r.URL.Port()
 	metadata := map[string]any{
 		"issuer":                                issuer,
 		"authorization_endpoint":                issuer + "/authorize",
@@ -71,11 +74,15 @@ func (s *state) handleAuthorize(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	if redirectURI == "" {
-		http.Error(w, "invalid_request", http.StatusBadRequest)
+		http.Error(w, "invalid_request (no redirect_uri)", http.StatusBadRequest)
 		return
 	}
 	if codeChallenge == "" || codeChallengeMethod != "S256" {
-		http.Error(w, "invalid_request", http.StatusBadRequest)
+		http.Error(w, "invalid_request (code challenge is not S256)", http.StatusBadRequest)
+		return
+	}
+	if query.Get("client_id") == "" {
+		http.Error(w, "invalid_request (missing client_id)", http.StatusBadRequest)
 		return
 	}
 
@@ -93,8 +100,9 @@ func (s *state) handleToken(w http.ResponseWriter, r *http.Request) {
 	r.ParseForm()
 	grantType := r.Form.Get("grant_type")
 	code := r.Form.Get("code")
-	redirectURI := r.Form.Get("redirect_uri")
 	codeVerifier := r.Form.Get("code_verifier")
+	// Ignore redirect_uri; it is not required in 2.1.
+	// https://www.ietf.org/archive/id/draft-ietf-oauth-v2-1-13.html#redirect-uri-in-token-request
 
 	if grantType != "authorization_code" {
 		http.Error(w, "unsupported_grant_type", http.StatusBadRequest)
@@ -107,11 +115,6 @@ func (s *state) handleToken(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	delete(s.authCodes, code)
-
-	if authCodeInfo.redirectURI != redirectURI {
-		http.Error(w, "invalid_grant", http.StatusBadRequest)
-		return
-	}
 
 	// PKCE verification
 	hasher := sha256.New()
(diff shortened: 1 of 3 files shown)


findleyr:
Do you want me to try to review?

internal/testing/fake_auth_server.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+	mux := http.NewServeMux()
+	mux.HandleFunc("/.well-known/oauth-authorization-server", server.handleMetadata)
+	mux.HandleFunc("/authorize", server.handleAuthorize)
+	mux.HandleFunc("/token", server.handleToken)
+	server.server = &http.Server{
+		Addr:    authServerPort,
+		Handler: mux,
+	}
+	return server
```
neild: Instead of a complete fake server, would it be simpler to have a this package provide a ServeMux? In the case where you want a server, you can use httptest:

```
srv := httptest.NewServer(fakeauth.NewMux())
```

This pushes the start-a-server logic off to httptest, which already handles it, and doesn't seem substantially less convenient to use in tests.

internal/testing/fake_auth_server.go
```diff
@@ -0,0 +1,165 @@
...
+
+func (s *FakeAuthServer) Stop() {
+	if err := s.server.Close(); err != nil {
+		log.Printf("Failed to stop server: %v", err)
+	}
+}
+
+func (s *FakeAuthServer) handleMetadata(w http.ResponseWriter, r *http.Request) {
+	metadata := map[string]any{
+		"issuer":                                issuer,
```
neild, line 51: Instead of hardcoding http://localhost:8080 as the issuer, should this use r.URL to find either the name or the port?

RFC 8414 says the issuer is present to prevent "mix-up" attacks, but that's probably not much of a concern for a test-only implementation. To be extra cautious, however, maybe this could set the issuer to `"http://localhost:" + r.URL.Port()`.

neild, line 51: RFC 8414 requires most/all of the URLs here to have an "https" scheme. The test server uses HTTP. Not sure if this is going to be a problem in practice, but this might be something else using httptest would help with, since it handles setting up a test server with TLS support.

internal/testing/fake_auth_server.go
```diff
@@ -0,0 +1,165 @@
...
+	codeChallengeMethod := query.Get("code_challenge_method")
+
+	if responseType != "code" {
+		http.Error(w, "unsupported_response_type", http.StatusBadRequest)
+		return
+	}
+	if redirectURI == "" {
+		http.Error(w, "invalid_request", http.StatusBadRequest)
+		return
+	}
```
neild, line 79: I might be missing something, but I believe redirect_uri is an optional parameter and only needs to be provided if a client has registered more than one redirect URI.

I'm going to guess that this fake auth server will only be used with clients that always provide a redirect_uri?

internal/testing/fake_auth_server.go
```diff
@@ -0,0 +1,165 @@
...
+	if codeChallenge == "" || codeChallengeMethod != "S256" {
+		http.Error(w, "invalid_request", http.StatusBadRequest)
+		return
+	}
```
neild, line 87: Should this check for the required client_id field? (It doesn't use it, but perhaps it would be useful to return an error to a client which fails to provide it?)

internal/testing/fake_auth_server.go
```diff
@@ -0,0 +1,165 @@
...
+
+	authCode := "fake-auth-code-" + fmt.Sprintf("%d", time.Now().UnixNano())
```
neild, line 89: Unlikely though it is for a collision to happen here, perhaps this should include a counter or random component?

internal/testing/fake_auth_server.go
```diff
@@ -0,0 +1,165 @@
...
+
+	redirectURL := fmt.Sprintf("%s?code=%s&state=%s", redirectURI, authCode, query.Get("state"))
+	http.Redirect(w, r, redirectURL, http.StatusFound)
+}
+
+func (s *FakeAuthServer) handleToken(w http.ResponseWriter, r *http.Request) {
+	r.ParseForm()
+	grantType := r.Form.Get("grant_type")
+	code := r.Form.Get("code")
+	redirectURI := r.Form.Get("redirect_uri")
```
neild: redirect_uri was removed in OAuth 2.1:
https://www.ietf.org/archive/id/draft-ietf-oauth-v2-1-13.html#redirect-uri-in-token-request

If this needs to support OAuth 2.0 clients, I believe it should validate redirect_uri when present but not complain when it isn't present. I suspect there's more to be done to support OAuth 2.0, though--I haven't checked to see what else changes. If this is only supporting OAuth 2.1 (https://modelcontextprotocol.io/specification/draft/basic/authorization only mentions 2.1) then I think it can ignore redirect_uri.

internal/testing/fake_auth_server.go
jba: Done.

internal/testing/fake_auth_server.go
jba: done.

jba: using https

internal/testing/fake_auth_server.go
jba: Yes, let's assume there's a redirect_uri.

jba: also note: in https://www.ietf.org/archive/id/draft-ietf-oauth-v2-1-13.html#redirect-uri-in-token-request it says:
"in practice, many authorization server implementations required the redirect_uri parameter in the authorization request even if only one was registered."

internal/testing/fake_auth_server.go
jba: Done.

internal/testing/fake_auth_server.go
jba: Done.

internal/testing/fake_auth_server.go
```diff
@@ -0,0 +1,151 @@
...
+type authCodeInfo struct {
+	codeChallenge string
+	redirectURI   string
+}
+
+// // FakeAuthServer is a fake OAuth2 authorization server.
+// type FakeAuthServer struct {
+// 	server    *http.Server
+// 	authCodes map[string]authCodeInfo
+// }
```
neild, line 31: Leftover commented-out code?

neild (approved):

PR #297, opened 2025-08-13
mcp: allow correct 'Accept' header values (#290)

### LINKED ISSUE #290: 'Accept' Content-Types don't support wildcards
**Describe the bug**
When making calls to the Streamable HTTP Transport JSON RPC endpoint, wildcards aren't supported as valid content types. 

**To Reproduce**
1. Call StreamableHTTP Transport MCP Server with `Accept: */*` (e.g. using Postman default).
2. MCP Server responds with `401` status and the following body: `Accept must contain both 'application/json' and 'text/event-stream'`.

**Expected behavior**
I would expect the `*/*`, `text/*`, and `application/*` values for the `Accept` header to imply the respective `application/json` and `text/event-stream` values. See the Mozilla docs on this header's syntax [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Accept#syntax).

**Additional context**
I'd love to make a short PR adding this to the switch statement at [mcp/streamable.go:100](https://github.com/modelcontextprotocol/go-sdk/blob/6e03217c831b1d142ed8a29c5d5916f47884008b/mcp/streamable.go#L100). It would be my first open source contribution. 😄

RileyNorton:
Allow other header values for the 'Accept' header that imply application/json and text/event-stream for requests to the Streamable HTTP Transport.

Fixes #290

Unit tests here seemed like unnecessary complexity and wouldn't fit cleanly with the style of the `streamable_test.go` tests. Happy to add those if desired though.

### DIFF
--- a/mcp/streamable.go
+++ b/mcp/streamable.go
@@ -98,9 +98,12 @@ func (h *StreamableHTTPHandler) ServeHTTP(w http.ResponseWriter, req *http.Reque
 	var jsonOK, streamOK bool
 	for _, c := range accept {
 		switch strings.TrimSpace(c) {
-		case "application/json":
+		case "application/json", "application/*":
 			jsonOK = true
-		case "text/event-stream":
+		case "text/event-stream", "text/*":
+			streamOK = true
+		case "*/*":
+			jsonOK = true
 			streamOK = true
 		}
 	}


findleyr (approved):
Thanks! I'll add a test.

PR #304, opened 2025-08-14
mcp/streamable: fix unbounded memory in streamable transport

samthanawalla:
This CL adds a boundedBuffer and a circularBuffer to prevent the server from consuming unbounded memory.
The circularBuffer is used for notifications to be consumed by the client-side hanging GET and will drop the oldest notification if any. The boundedBuffer is used to prevent writing additional messages related to POST requests if the client is too slow.

For #190

### DIFF
--- a/mcp/streamable.go
+++ b/mcp/streamable.go
@@ -324,7 +324,7 @@ type stream struct {
 
 	// outgoing is the list of outgoing messages, enqueued by server methods that
 	// write notifications and responses, and dequeued by streamResponse.
-	outgoing [][]byte
+	outgoing outgoingBuffer
 
 	// streamRequests is the set of unanswered incoming RPCs for the stream.
 	//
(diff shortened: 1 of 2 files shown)


mcp/streamable.go
```diff
@@ -324,19 +324,137 @@ type stream struct {
...
 	//
 	// Requests persist until their response data has been added to outgoing.
 	requests map[jsonrpc.ID]struct{}
 }
 
+// outgoingBuffer defines a common interface for storing outgoing messages.
+// This allows us to use either a circular or bounded buffer.
+type outgoingBuffer interface {
+	// tryAdd attempts to add a message to the buffer.
+	// It returns true if the message was added successfully.
```
jba, line 339: "It reports whether the message was added." I think that's sufficient.

mcp/streamable.go
```diff
@@ -324,19 +324,137 @@ type stream struct {
...
+
+	// close signals that no more messages will be sent.
+	close()
+}
+
+// boundedBuffer implements buffer using a buffered channel.
+type boundedBuffer struct {
+	ch chan []byte
+}
+
```
jba, line 354: Add `func newBoundedBuffer(capacity int)`.

mcp/streamable.go
```diff
@@ -324,19 +324,137 @@ type stream struct {
...
 
 	// streamRequests is the set of unanswered incoming RPCs for the stream.
 	//
 	// Requests persist until their response data has been added to outgoing.
 	requests map[jsonrpc.ID]struct{}
 }
 
+// outgoingBuffer defines a common interface for storing outgoing messages.
+// This allows us to use either a circular or bounded buffer.
+type outgoingBuffer interface {
```
jba, line 337: Just `buffer`?

mcp/streamable.go
```diff
@@ -324,19 +324,137 @@ type stream struct {
...
+	b.mu.Lock()
+	defer b.mu.Unlock()
+
+	// Place the new item at the current head position.
+	b.messages[b.head] = msg
+
+	// Move the head pointer forward, wrapping around.
+	b.head = (b.head + 1) % b.capacity
+
+	if b.size == b.capacity {
```
jba, line 403: Why isn't the capacity the same as len(messages)?

mcp/streamable.go
```diff
@@ -324,19 +324,137 @@ type stream struct {
...
+func (b *circularBuffer) drain() [][]byte {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+	if b.size == 0 {
+		return nil
+	}
+	out := make([][]byte, b.size)
+
+	// Copy items in the correct FIFO order (from tail to head).
+	for i := 0; i < b.size; i++ {
```
jba, line 423: for i := range b.size

mcp/streamable.go
```diff
@@ -324,19 +324,137 @@ type stream struct {
...
+// circularBuffer implements buffer using a circular buffer.
+type circularBuffer struct {
+	mu       sync.Mutex
+	messages [][]byte
+	head     int // Index to write next
+	tail     int // Index to read next
+	size     int // Current number of items in buffer
+	capacity int
+}
+
```
jba, line 392: A `newCircularBuffer` would be useful here too.

mcp/streamable.go
```diff
@@ -777,10 +893,9 @@ func (t *StreamableServerTransport) Write(ctx context.Context, msg jsonrpc.Messa
 	if len(stream.requests) == 0 && forStream != 0 || stream.jsonResponse && !isResponse {
 		stream = t.streams[0]
 	}
-
-	// TODO: if there is nothing to send these messages to (as would happen, for example, if forConn == 0
-	// and the client never did a GET), then memory will grow without bound. Consider a mitigation.
-	stream.outgoing = append(stream.outgoing, data)
+	if ok := stream.outgoing.tryAdd(data); !ok {
```
jba, line 896: if !stream...

mcp/streamable.go
```diff
@@ -777,10 +893,9 @@ func (t *StreamableServerTransport) Write(ctx context.Context, msg jsonrpc.Messa
...
+		return fmt.Errorf("stream is full: %d", forStream)
```
jba, line 897: It's important to think about where this error ends up. How can the server act on it? This may get back to having an ErrorHandler. 
Not sure there is anything to do here, except trace the control flow to see where the error ends up and add a TODO here.

mcp/streamable.go
```diff
@@ -324,19 +324,137 @@ type stream struct {
...
+
+	// Place the new item at the current head position.
+	b.messages[b.head] = msg
+
+	// Move the head pointer forward, wrapping around.
+	b.head = (b.head + 1) % b.capacity
+
+	if b.size == b.capacity {
+		// If the buffer was already full, the tail also moves forward
+		// because we just overwrote the oldest item.
```
jba, line 405: I think overwriting a notification is OK, but overwriting a request or response would be very bad. We may have to distinguish those cases.

mcp/streamable_test.go
```diff
@@ -891,3 +891,203 @@ func TestStreamableStateless(t *testing.T) {
...
+					if !strings.Contains(err.Error(), "stream is full") {
+						t.Errorf("NotifyProgress error mismatch: expected %v but got %q", wantErr, err)
+					}
+					return &CallToolResult{}, err
+				}
+			}
+			return &CallToolResult{}, nil
+		})
+
+		// 2. Set up a client to receive the notifications.
```
jba, line 926: Clarify what exactly is happening here. Does the client signal an error somewhere because it has no ProgressHandler? That is not quite the case I was worried about, where there is a handler but it takes a long time.

samthanawalla:
I think that would be okay if EventStore was in-memory but what if a user provides a slower EventStore that requires I/O? Wouldn't this cause a performance hit on the main server logic if we had to push to eventstore on each write?

findleyr:
But we *do* write to the EventStore before writing back to the http request in any case, since that's the only way we can guarantee delivery.

samthanawalla:
Understood- that makes sense. I will remove outgoing and update this CL once that is done

findleyr:
Sounds good. There is some subtlety about the application/json serving path: we may need to unpack events and repack them as pure JSON (this should be trivial).

I suspect that this will significantly simplify the code, actually.

findleyr:
Of course, we still need to have an eviction policy for the in-memory eventstore :)

mcp/streamable_test.go
```diff
@@ -891,3 +891,203 @@ func TestStreamableStateless(t *testing.T) {
...
+		var wg sync.WaitGroup
+		receivedNotifications := make(chan string, outgoingBufferLimit)
+		defer close(receivedNotifications)
+
+		// 1. Set up a server with a tool that triggers many background notifications.
+		server := NewServer(testImpl, nil)
+		AddTool(server, &Tool{Name: "backgroundSpam"}, func(ctx context.Context, ss *ServerSession, params *CallToolParamsFor[any]) (*CallToolResult, error) {
+			defer wg.Done()
+			// Send more messages than the circular buffer can hold.
+			for i := range outgoingBufferLimit + 5 {
```
jba, line 963: make 5 a const, since it appears several times

jba:
Sorry, I missed the discussion about EventStore. Maybe ignore my comments?

jba:
I'm not sure I understand something. Doesn't the streamableServerConn still need to buffer messages because there is another goroutine writing them? I assume that's what's going on in the loop at the end of streamableServerConn.respondSSE.

PR #306, opened 2025-08-15
mcp/server.go: implement server-side logging throughout codebase

### LINKED ISSUE #170: server-side logging
Add logging to the server to enhance debuggability.

Thread a slog.Logger through server-side structures and log important actions and errors.

KamdynS:
mcp/server.go: implement server‑side logging

- Thread a `*slog.Logger` through server paths to improve debuggability
- Introduce no‑op discard handler + `ensureLogger` so loggers are never nil
- Use ensured loggers in `Server`, `SSEHandler`/`SSEServerTransport`; remove nil checks
- Route jsonrpc2 internal errors to an internal logger

All tests are passing and no functionality should change with this PR

Fixes #170

### DIFF
--- a/mcp/logging.go
+++ b/mcp/logging.go
@@ -87,6 +87,27 @@ type LoggingHandler struct {
 	handler         slog.Handler
 }
 
+// discardHandler is a slog.Handler that drops all logs.
+type discardHandler struct{}
+
+func (discardHandler) Enabled(context.Context, slog.Level) bool  { return false }
+func (discardHandler) Handle(context.Context, slog.Record) error { return nil }
+func (discardHandler) WithAttrs([]slog.Attr) slog.Handler        { return discardHandler{} }
+func (discardHandler) WithGroup(string) slog.Handler             { return discardHandler{} }
+
+// ensureLogger returns l if non-nil, otherwise a discard logger.
+func ensureLogger(l *slog.Logger) *slog.Logger {
+	if l != nil {
+		return l
+	}
+	return slog.New(discardHandler{})
+}
+
+// internalLogger is used for package-internal logging where we don't have a
+// specific server/handler context. It defaults to a discard logger to avoid
+// unsolicited output from library code.
+var internalLogger = slog.New(discardHandler{})
+
 // NewLoggingHandler creates a [LoggingHandler] that logs to the given [ServerSession] using a
 // [slog.JSONHandler].
 func NewLoggingHandler(ss *ServerSession, opts *LoggingHandlerOptions) *LoggingHandler {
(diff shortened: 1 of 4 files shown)


mcp/logging.go
```diff
@@ -87,6 +87,27 @@ type LoggingHandler struct {
 	handler         slog.Handler
 }
 
+// discardHandler is a slog.Handler that drops all logs.
+type discardHandler struct{}
```
jba, line 91: slog.DiscardHandler

mcp/server.go
```diff
@@ -50,6 +52,8 @@ type Server struct {
 type ServerOptions struct {
 	// Optional instructions for connected clients.
 	Instructions string
+	// Logger is used for server-side logging. If nil, slog.Default() is used.
```
jba, line 55: nil should mean no logging.

mcp/server.go
```diff
@@ -108,9 +112,11 @@ func NewServer(impl *Implementation, opts *ServerOptions) *Server {
 	if opts.UnsubscribeHandler != nil && opts.SubscribeHandler == nil {
 		panic("UnsubscribeHandler requires SubscribeHandler")
 	}
+	l := ensureLogger(opts.Logger)
```
jba, line 115: Inline this into L119

mcp/sse.go
```diff
@@ -181,6 +198,7 @@ func (t *SSEServerTransport) Connect(context.Context) (Connection, error) {
 }
 
 func (h *SSEHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
+	h.ensureLogger()
```
jba, line 201: If this only occurs in one place, inline it.

mcp/logging.go
KamdynS: This was added in Go 1.24 (https://tip.golang.org/doc/go1.24#logslogpkglogslog), but we're currently on Go 1.23. Leave as is?

jba: Yes, good point. Just leave a TODO so we update it when we drop 1.23.

ramongalate:
could this also extend to streamable http? 
I can add it to this, if needed

jba:
@ramongalate I think that should be separate, since the transport is lower level. Maybe a field on StreamableHTTPOptions and StreamableServerTransport?

I think you can make a separate PR, not dependent on this one.

IAmSurajBobade:
@KamdynS there is no activity on this PR for over a month. Do you mind if I take this up? Will address open review comments.

KamdynS:
@IAmSurajBobade Yes please

IAmSurajBobade:
Hi @KamdynS I have addressed review comments and made minor tweaks. To update the same PR , could you please provide contributer access to your repo/branch. Otherwise, I can create new PR.

jba:
Superseded by #501.

PR #307, opened 2025-08-15
Improvements to the stateless streamable transport

### LINKED ISSUE #10: Streamable HTTP support

### LINKED ISSUE #284: Fail requests immediately in 'stateless' mode
When the streamable transport is in 'stateless' mode, we need to fail requests immediately (because there will be no way for the client to respond!).

Right now, requests will just hang, waiting for a response that will never come.

We should also consider reasonable timeouts for all requests, since this transport is less reliable than others.

findleyr:
Several improvements for the stateless streamable mode, plus support for
a 'distributed' (or rather, distributable) version of the stateless
server.

- Add a 'Stateless' option to StreamableHTTPOptions and
  StreamableServerTransport, which controls stateless behavior.
  GetSessionID may still return a non-empty session ID.
- Audit validation of stateless mode to allow requests with a session
  id. Propagate this session ID to the temporary connection.
- Peek at requests to allow 'initialize' and 'initialized' requests to go through to the
  session, so that version negotiation can occur (FIXME: add tests).

Fixes #284
Fixes #10
Updates #148

### DIFF
--- a/internal/jsonrpc2/conn.go
+++ b/internal/jsonrpc2/conn.go
@@ -742,8 +742,8 @@ func (c *Connection) write(ctx context.Context, msg Message) error {
 	var err error
 	// Fail writes immediately if the connection is shutting down.
 	//
-	// TODO(rfindley): should we allow cancellation notifations through? It could
-	// be the case that writes can still succeed.
+	// TODO(rfindley): should we allow cancellation notifications through? It
+	// could be the case that writes can still succeed.
 	c.updateInFlight(func(s *inFlightState) {
 		err = s.shuttingDown(ErrServerClosing)
 	})
--- a/internal/jsonrpc2/conn.go
+++ b/internal/jsonrpc2/conn.go
@@ -739,7 +739,23 @@ func (c *Connection) processResult(from any, req *incomingRequest, result any, e
 // write is used by all things that write outgoing messages, including replies.
 // it makes sure that writes are atomic
 func (c *Connection) write(ctx context.Context, msg Message) error {
-	err := c.writer.Write(ctx, msg)
+	var err error
+	// Fail writes immediately if the connection is shutting down.
+	//
+	// TODO(rfindley): should we allow cancellation notifations through? It could
+	// be the case that writes can still succeed.
+	c.updateInFlight(func(s *inFlightState) {
+		err = s.shuttingDown(ErrServerClosing)
+	})
+	if err == nil {
+		err = c.writer.Write(ctx, msg)
+	}
+
+	// For rejected requests, we don't set the writeErr (which would break the
+	// connection). They can just be returned to the caller.
+	if errors.Is(err, ErrRejected) {
+		return err
+	}
 
 	if err != nil && ctx.Err() == nil {
 		// The call to Write failed, and since ctx.Err() is nil we can't attribute
--- a/mcp/streamable.go
+++ b/mcp/streamable.go
@@ -49,6 +49,9 @@ type StreamableHTTPOptions struct {
 	// GetSessionID provides the next session ID to use for an incoming request.
 	// If nil, a default randomly generated ID will be used.
 	//
+	// Session IDs should be globally unique across the scope of the server,
+	// which may span multiple processes in the case of distributed servers.
+	//
 	// As a special case, if GetSessionID returns the empty string, the
 	// Mcp-Session-Id header will not be set.
 	GetSessionID func() string
@@ -58,7 +61,9 @@ type StreamableHTTPOptions struct {
 	// A stateless server does not validate the Mcp-Session-Id header, and uses a
 	// temporary session with default initialization parameters. Any
 	// server->client request is rejected immediately as there's no way for the
-	// client to respond.
+	// client to respond. Server->Client notifications may reach the client if
+	// they are made in the context of an incoming request, as described in the
+	// documentation for [StreamableServerTransport].
 	Stateless bool
 
 	// TODO: support session retention (?)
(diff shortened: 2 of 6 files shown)


BC-ACherednichenko (commented):
LGTM, I checked a quick implementation on top of this PR, works as expected!

internal/jsonrpc2/conn.go
```diff
@@ -725,7 +725,23 @@ func (c *Connection) processResult(from any, req *incomingRequest, result any, e
 // write is used by all things that write outgoing messages, including replies.
 // it makes sure that writes are atomic
 func (c *Connection) write(ctx context.Context, msg Message) error {
-	err := c.writer.Write(ctx, msg)
+	var err error
+	// Fail writes immediately if the connection is shutting down.
+	//
+	// TODO(rfindley): should we allow cancellation notifations through? It could
```
jba: notifications

mcp/streamable.go
```diff
@@ -38,18 +38,29 @@ type StreamableHTTPHandler struct {
...
 }
 
 // StreamableHTTPOptions configures the StreamableHTTPHandler.
 type StreamableHTTPOptions struct {
 	// GetSessionID provides the next session ID to use for an incoming request.
+	// If nil, a default randomly generated ID will be used.
 	//
-	// If GetSessionID returns an empty string, the session is 'stateless',
-	// meaning it is not persisted and no session validation is performed.
+	// As a special case, if GetSessionID returns the empty string, the
```
jba, line 55: Mention that it needs to be globally unique for distributed sessions.

mcp/streamable.go
```diff
@@ -38,18 +38,29 @@ type StreamableHTTPHandler struct {
...
+	// Mcp-Session-Id header will not be set.
 	GetSessionID func() string
 
+	// Stateless controls whether the session is 'stateless'.
+	//
+	// A stateless server does not validate the Mcp-Session-Id header, and uses a
+	// temporary session with default initialization parameters. Any
+	// server->client request is rejected immediately as there's no way for the
```
jba, line 63: What about server client notifications? Presumably, just dropped on the floor.

mcp/streamable.go
```diff
@@ -118,36 +129,39 @@ func (h *StreamableHTTPHandler) ServeHTTP(w http.ResponseWriter, req *http.Reque
...
 	var transport *StreamableServerTransport
-	if id := req.Header.Get(sessionIDHeader); id != "" {
+	if sessionID != "" {
 		h.mu.Lock()
-		transport, _ = h.transports[id]
+		transport, _ = h.transports[sessionID]
 		h.mu.Unlock()
-		if transport == nil {
+		if transport == nil && !h.opts.Stateless {
+			// In stateless mode we allow a missing transport.
```
jba: Comment feels weird here, because this is the opposite case.

mcp/streamable.go
```diff
@@ -118,36 +129,39 @@ func (h *StreamableHTTPHandler) ServeHTTP(w http.ResponseWriter, req *http.Reque
...
+			transport.connection.Close()
+		}
 		w.WriteHeader(http.StatusNoContent)
 		return
 	}
 
 	switch req.Method {
 	case http.MethodPost, http.MethodGet:
-		if req.Method == http.MethodGet && transport == nil {
+		if req.Method == http.MethodGet && sessionID == "" {
```
jba, line 170: This is the client's hanging get? You should document that above at GetSessionID. Also, this will break clients that do the get unconditionally, like our own.

mcp/streamable.go
```diff
@@ -164,37 +178,83 @@ func (h *StreamableHTTPHandler) ServeHTTP(w http.ResponseWriter, req *http.Reque
...
+		if h.opts.Stateless {
+			// Peek at the body to see if it is initialize or initialized.
+			// We want those to be handled as usual.
+			var hasInitialize, hasInitialized bool
+			{
+				// TODO: verify that this allows protocol version negotiation for
+				// stateless servers.
+				body, err := io.ReadAll(req.Body)
+				if err != nil {
+					http.Error(w, "failed to read body", http.StatusBadRequest)
```
jba: nit: internal server error? Otherwise hard to see how this could fail.

mcp/streamable.go
```diff
@@ -225,6 +285,13 @@ type StreamableServerTransport struct {
 	// generator to produce one, as with [crypto/rand.Text].)
 	SessionID string
 
+	// Stateless controls whether the eventstore is 'Stateless'. Servers sessions
```
jba: Server sessions

mcp/streamable.go

mcp/streamable.go
findleyr: Woah, you found a big bug. The spec says that this is what the server is supposed to do, but we weren't handling it correctly from the client side.
The only reason tests were passing is that our initial request delay calculation was off, and so we were never even getting to the hanging GET.

Fixed the delay, and fixed our handling of 405 (as well as other status codes) in the third commit on this PR. I had to rebase to pick up other fixes I'd made for the client transport, but I didn't touch the other commits except to rebase them, so the third commit contains all my edits based on your comments.

mcp/streamable.go
findleyr: Updated the comment to invert the sense. PTAL.

mcp/streamable.go

internal/jsonrpc2/conn.go
findleyr: Done.

mcp/streamable.go
findleyr: Done.

mcp/streamable.go
findleyr: Oh, they may be received if they're in the context of a hanging POST. Updated the commentary, and explained how this works in the doc comment for StreamableServerTransport.

jba (approved):

samthanawalla (approved):

PR #312, opened 2025-08-17
mcp: replace custom randText implementation with crypto/rand.Text

yasomaru:
- Remove custom base32alphabet constant and manual implementation
- Use Go 1.24's crypto/rand.Text() function for secure random text generation
- Simplify code by removing 26 lines of custom implementation
- Maintain same functionality with standard library implementation

This change resolves TODO comments about using crypto/rand.Text once Go 1.24 is assured. The standard library implementation provides the same security guarantees with better maintainability and consistency.

Fixes: #TODO (use crypto/rand.Text)


jba:
Thanks. I don't know if we're ready to give up on 1.23 just yet. 1.25 just came out. But we can keep this around.

PR #314, opened 2025-08-17
mcp: ensure keepalive goroutine is started exactly once

### LINKED ISSUE #313: Excessive goroutine creation on duplicate 'initialized' notification
**Describe the bug**
The `ServerSession.initialized` method's current implementation calls `ss.startKeepalive` before checking if the session has already been initialized. This leads to `startKeepalive` being called on every `initialized` notification, regardless of the session state.

**To Reproduce**
```Go
func TestServerSessionkeepaliveCancelOverwritten(t *testing.T) {
	server := NewServer(testImpl, &ServerOptions{KeepAlive: 5 * time.Second})
	ss := &ServerSession{server: server}

	// 1. Initialize the session.
	_, err := ss.initialize(context.Background(), &InitializeParams{})
	if err != nil {
		t.Fatalf("ServerSession initialize failed: %v", err)
	}

	// 2. Call 'initialized' for the first time. This should start the keepalive mechanism.
	_, err = ss.initialized(context.Background(), &InitializedParams{})
	if err != nil {
		t.Fatalf("First initialized call failed: %v", err)
	}
	if ss.keepaliveCancel == nil {
		t.Fatalf("expected ServerSession.keepaliveCancel to be set after the first call of initialized")
	}

	firstCancel := ss.keepaliveCancel
	defer firstCancel()
	// 3. Manually set the field to nil.
	// Do this to facilitate the test's core assertion. The goal is to verify that
	// 'ss.keepaliveCancel' is not assigned a second time. By setting it to nil,
	// we can easily check after the next call if a new keepalive goroutine was started.
	ss.keepaliveCancel = nil

	// 4. Call 'initialized' for the second time. This should return an error.
	_, err = ss.initialized(context.Background(), &InitializedParams{})
	if err == nil {
		t.Fatalf("Expected 'duplicate initialized received' error on second call, got nil")
	}

	time.Sleep(2 * time.Second)
	// 5. Re-check the field to ensure it remains nil.
	// Since 'initialized' correctly returned an error and did not call
	// 'startKeepalive', the field should remain unchanged.
	if ss.keepaliveCancel != nil {
		t.Fatal("expected ServerSession.keepaliveCancel to be nil after we manually niled it and re-initialized")
	}
}
```
Image

**Expected behavior**
`ss.startKeepalive` had better to be set exactly once as described in
https://github.com/modelcontextprotocol/go-sdk/blob/1afdb1f27309c888be8a4bbb91deb584ede164be/mcp/server.go#L870-L880

cryo-zd:
### change:
Reorder the logic within `ServerSession.initialized` to ensure `ss.startKeepalive` is called only after the session's state is validated. 
The `startKeepalive` function is now only invoked when the session is first transitioning from `initialize` to `initialized` state.

Fixes #313

### DIFF
--- a/mcp/server.go
+++ b/mcp/server.go
@@ -602,9 +602,6 @@ func (ss *ServerSession) initialized(ctx context.Context, params *InitializedPar
 		// params are non-nil.
 		params = new(InitializedParams)
 	}
-	if ss.server.opts.KeepAlive > 0 {
-		ss.startKeepalive(ss.server.opts.KeepAlive)
-	}
 	var wasInit, wasInitd bool
 	ss.updateState(func(state *ServerSessionState) {
 		wasInit = state.InitializeParams != nil
@@ -619,6 +616,9 @@ func (ss *ServerSession) initialized(ctx context.Context, params *InitializedPar
 	}
 	if wasInitd {
 		return nil, fmt.Errorf("duplicate %q received", notificationInitialized)
+	}
+	if ss.server.opts.KeepAlive > 0 {
+		ss.startKeepalive(ss.server.opts.KeepAlive)
 	}
 	if h := ss.server.opts.InitializedHandler; h != nil {
 		h(ctx, serverRequestFor(ss, params))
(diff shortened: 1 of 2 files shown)


findleyr (approved):
Thanks!

PR #319, opened 2025-08-18
mcp/streamable: remove StreamableReconnectOptions

### LINKED ISSUE #308: mcp: redo retry options
Move them into the parent struct.
For MaxRetries, treat 0 as default and -1 as "never retry".

samthanawalla:
Moves MaxRetries into the parent struct and assume a negative integer signifies to not retry.

Fixes #308

### DIFF
--- a/mcp/streamable.go
+++ b/mcp/streamable.go
@@ -849,26 +849,15 @@ func (c *streamableServerConn) Close() error {
 // endpoint serving the streamable HTTP transport defined by the 2025-03-26
 // version of the spec.
 type StreamableClientTransport struct {
-	Endpoint         string
-	HTTPClient       *http.Client
-	ReconnectOptions *StreamableReconnectOptions
-}
-
-// StreamableReconnectOptions defines parameters for client reconnect attempts.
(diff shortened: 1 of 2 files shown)


mcp/streamable.go
```diff
@@ -901,7 +892,7 @@ func NewStreamableClientTransport(url string, opts *StreamableClientTransportOpt
 	t := &StreamableClientTransport{Endpoint: url}
 	if opts != nil {
 		t.HTTPClient = opts.HTTPClient
```
jba, line 894: I think we should do our usual trick of storing the opts in the struct, making a shallow copy.
E.g. https://github.com/modelcontextprotocol/go-sdk/blob/main/mcp/client.go#L48.

samthanawalla: It looks like we are deprecating StreamableClientTransportOptions in favor of using the StreamableClientTransport literal

samthanawalla: Additionally NewStreamableClientTransport will be deleted soon

jba: Oh, right.

PR #325, opened 2025-08-19
mcp: change tool handler design

jba:
WIP
Description TBD.

### DIFF
--- a/examples/server/memory/kb.go
+++ b/examples/server/memory/kb.go
@@ -431,12 +431,12 @@ func (k knowledgeBase) openNodes(names []string) (KnowledgeGraph, error) {
 	}, nil
 }
 
-func (k knowledgeBase) CreateEntities(ctx context.Context, req *mcp.ServerRequest[*mcp.CallToolParamsFor[CreateEntitiesArgs]]) (*mcp.CallToolResultFor[CreateEntitiesResult], error) {
-	var res mcp.CallToolResultFor[CreateEntitiesResult]
+func (k knowledgeBase) CreateEntities(ctx context.Context, req *mcp.ServerRequest[*mcp.CallToolParams], args CreateEntitiesArgs) (*mcp.CallToolResult, CreateEntitiesResult, error) {
+	var res mcp.CallToolResult
 
-	entities, err := k.createEntities(req.Params.Arguments.Entities)
+	entities, err := k.createEntities(args.Entities)
 	if err != nil {
-		return nil, err
+		return nil, CreateEntitiesResult{}, err
 	}
 
 	res.Content = []mcp.Content{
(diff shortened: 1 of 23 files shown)


mcp/mcp_test.go
```diff
@@ -946,8 +946,8 @@ func TestAddTool_DuplicateNoPanicAndNoDuplicate(t *testing.T) {
 		// This case was written specifically to reproduce a bug where duplicate tools where causing jsonschema errors
 		t1 := &Tool{Name: "dup", Description: "first", InputSchema: &jsonschema.Schema{}}
 		t2 := &Tool{Name: "dup", Description: "second", InputSchema: &jsonschema.Schema{}}
-		s.AddTool(t1, nopHandler)
-		s.AddTool(t2, nopHandler)
+		s.AddRawTool(t1, nopHandler)
```
findleyr: This should just be AddTool. People won't understand what a 'RawTool' is, so let's not invent this distinction. We should instead explain in the documentation.

mcp/server.go
```diff
@@ -138,59 +140,100 @@ func (s *Server) RemovePrompts(names ...string) {
...
-// AddTool adds a [Tool] to the server, or replaces one with the same name.
 // If the tool's input schema is nil, it is set to the schema inferred from the In
 // type parameter, using [jsonschema.For].
 // If the tool's output schema is nil and the Out type parameter is not the empty
 // interface, then the output schema is set to the schema inferred from Out.
-// The Tool argument must not be modified after this call.
-func AddTool[In, Out any](s *Server, t *Tool, h ToolHandlerFor[In, Out]) {
-	if err := addToolErr(s, t, h); err != nil {
-		panic(err)
+func RawToolHandlerFor[In, Out any](t *Tool, h ToolHandlerFor[In, Out]) RawToolHandler {
```
findleyr: When used like this, there's no way to chain the call, which seemed critical.

Isn't it better not to mutate the input?

examples/server/memory/kb.go
```diff
@@ -447,136 +447,127 @@ func (k knowledgeBase) CreateEntities(ctx context.Context, req *mcp.ServerReques
...
 		&mcp.TextContent{Text: "Entities deleted successfully"},
 	}
 
-	return &res, nil
+	return &res, struct{}{}, nil
 }
 
-func (k knowledgeBase) DeleteObservations(ctx context.Context, req *mcp.ServerRequest[*mcp.CallToolParamsFor[DeleteObservationsArgs]]) (*mcp.CallToolResultFor[struct{}], error) {
-	var res mcp.CallToolResultFor[struct{}]
+func (k knowledgeBase) DeleteObservations(ctx context.Context, req *mcp.ServerRequest[*mcp.CallToolParams], args DeleteObservationsArgs) (*mcp.CallToolResult, struct{}, error) {
```
findleyr, line 500: Can we make this 'any'? The struct{}{} results are awkward.

Unless it's necessary for this to be struct{}, in which case we should discuss why that's the case.

examples/server/memory/kb.go
```diff
@@ -447,136 +447,127 @@ func (k knowledgeBase) CreateEntities(ctx context.Context, req *mcp.ServerReques
...
+	var res mcp.CallToolResult
```
findleyr, line 501: Nit: preexisting: just inline into the literal below.

examples/server/memory/kb_test.go
```diff
@@ -427,203 +427,203 @@ func TestFileFormatting(t *testing.T) {
...
-			if err != nil {
-				t.Fatalf("Final MCP ReadGraph failed: %v", err)
-			}
-			if len(finalRead.StructuredContent.Entities) != 0 {
-				t.Errorf("expected empty graph after deletion, got %d entities", len(finalRead.StructuredContent.Entities))
-			}
-		})
-	}
-}
+// func TestMCPServerIntegration(t *testing.T) {
```
findleyr, line 430: Please explain what specifically is broken about this. I can fix as soon as you land this.

mcp/streamable_test.go
```diff
@@ -48,18 +48,18 @@ func TestStreamableTransports(t *testing.T) {
...
 				case <-time.After(5 * time.Second):
-					return nil, nil
+					return nil, nil, nil
 				}
-				return nil, nil
+				return nil, nil, nil
 			}
 			AddTool(server, &Tool{Name: "hang"}, hang)
-			AddTool(server, &Tool{Name: "sample"}, func(ctx context.Context, req *ServerRequest[*CallToolParams]) (*CallToolResult, error) {
+			AddTool(server, &Tool{Name: "sample"}, func(ctx context.Context, req *ServerRequest[*CallToolParams], args map[string]any) (*CallToolResult, any, error) {
```
findleyr, line 62: I think args here should just be 'any'.

findleyr (approved):

examples/server/memory/kb.go
jba: Done in a followup.

examples/server/memory/kb.go
jba: I don't understand.

mcp/streamable_test.go
jba: done

andig:
Not sure if I'm commenting on the right PR and fwiw, but I've had some trouble understanding the API changes. It seems kinda clumsy now having to specify input and output types both for tool handlers when those are never ever used and/or input and output schemas have already been defined as part of the `mcp.Tool`.

findleyr:
@andig the right issue is #302.

The motivation is that in the common case, you only need to access the args, and also want access to the actual incoming params. I agree that the Out parameter is somewhat awkward, but once you get used to it it is also convenient.

findleyr:
Sorry, the issue is #327

PR #334, opened 2025-08-20
mcp: replace type struct{} with any for user-defined fields

h9jiang:
This change adopts a clearer convention:
- any is used for fields where the structure is defined by the client or server implementer.
- an named empty struct type is the place holder type to be defined by the MCP spec in the future. 

The "Experimental" fields in "ClientCapabilities" and "ServerCapabilities" now use "map[string]any". This allows clients and servers to negotiate custom capabilities that contain complex data, rather than being restricted to a simple on/off flag.

The "Metadata" field in "CreateMessageParams" was also changed to "any" to properly serve its purpose as a flexible container for server-defined data.

### DIFF
--- a/mcp/protocol.go
+++ b/mcp/protocol.go
@@ -126,7 +126,7 @@ func (x *CancelledParams) SetProgressToken(t any) { setProgressToken(x, t) }
 // additional capabilities.
 type ClientCapabilities struct {
 	// Experimental, non-standard capabilities that the client supports.
-	Experimental map[string]struct{} `json:"experimental,omitempty"`
+	Experimental map[string]any `json:"experimental,omitempty"`
 	// Present if the client supports listing roots.
 	Roots struct {
 		// Whether the client supports notifications for changes to the roots list.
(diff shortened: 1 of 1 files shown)


h9jiang:
Some reference that might be helpful: 

LSP Client Capability Experimental definition in [typescript](https://github.com/microsoft/vscode-languageserver-node/blob/d90a87f9557a0df9142cfb33e251cfa6fe27d970/protocol/src/common/protocol.ts#L1122)

```
	/**
	 * Experimental client capabilities.
	 */
	experimental?: LSPAny;
```

The go LSP client used a [slice](https://go.googlesource.com/vscode-go/+/refs/heads/master/extension/src/language/goLanguageServer.ts#362) (array) for some feature.


```
	// Experimental client capabilities.
	Experimental any `json:"experimental,omitempty"`
```

LSP Client Capability Experimental definition in [go](https://go.googlesource.com/tools/+/refs/heads/master/gopls/internal/protocol/tsprotocol.go#252)

jba:
You should remove the "PR Guideline" section from the commit message. (Can wait until you commit.)

h9jiang:
Just remove anything below the PR Guideline, does this look good to you?

PR #335, opened 2025-08-20
mcp/streamable: use event store to fix unbounded memory issues

samthanawalla:
This CL utilizes the event store to write outgoing messages and removes the unbounded outgoing data structure.

For #190

### DIFF
--- a/mcp/event.go
+++ b/mcp/event.go
@@ -274,12 +274,6 @@ func (s *MemoryEventStore) Append(_ context.Context, sessionID string, streamID 
 	// Purge before adding, so at least the current data item will be present.
 	// (That could result in nBytes > maxBytes, but we'll live with that.)
 	s.purge()
-
-	// An empty data slice signals that a stream has been registered.
-	// We ignore it since it contains no content and shouldn't affect the total size.
-	if data == nil {
-		return nil
-	}
 	dl.appendData(data)
 	s.nBytes += len(data)
 	return nil
--- a/mcp/event.go
+++ b/mcp/event.go
@@ -153,6 +153,11 @@ func scanEvents(r io.Reader) iter.Seq2[Event, error] {
 //
 // All of an EventStore's methods must be safe for use by multiple goroutines.
 type EventStore interface {
+	// Open prepares the event store for a given session. It ensures that the
+	// underlying data structure for the sessionID is initialized, making it
+	// ready to store event streams.
+	Open(_ context.Context, sessionID string, streamID StreamID) error
+
 	// Append appends data for an outgoing event to given stream, which is part of the
 	// given session.
 	Append(_ context.Context, sessionID string, _ StreamID, data []byte) error
@@ -256,11 +261,21 @@ func NewMemoryEventStore(opts *MemoryEventStoreOptions) *MemoryEventStore {
 	}
 }
 
-// Append implements [EventStore.Append] by recording data in memory.
-func (s *MemoryEventStore) Append(_ context.Context, sessionID string, streamID StreamID, data []byte) error {
+// Open implements [EventStore.Open]. It ensures that the underlying data
+// structures for the given session and stream IDs are initialized and ready
+// for use.
+func (s *MemoryEventStore) Open(_ context.Context, sessionID string, streamID StreamID) error {
 	s.mu.Lock()
 	defer s.mu.Unlock()
+	s.init(sessionID, streamID)
+	return nil
+}
 
+// init is an internal helper function that ensures the nested map structure for a
+// given sessionID and streamID exists, creating it if necessary. It returns the
+// dataList associated with the specified IDs.
+// This function must be called within a locked context.
+func (s *MemoryEventStore) init(sessionID string, streamID StreamID) *dataList {
 	streamMap, ok := s.store[sessionID]
 	if !ok {
 		streamMap = make(map[StreamID]*dataList)
@@ -271,6 +286,14 @@ func (s *MemoryEventStore) Append(_ context.Context, sessionID string, streamID 
 		dl = &dataList{}
 		streamMap[streamID] = dl
 	}
+	return dl
+}
+
+// Append implements [EventStore.Append] by recording data in memory.
+func (s *MemoryEventStore) Append(_ context.Context, sessionID string, streamID StreamID, data []byte) error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	dl := s.init(sessionID, streamID)
 	// Purge before adding, so at least the current data item will be present.
 	// (That could result in nBytes > maxBytes, but we'll live with that.)
 	s.purge()
--- a/mcp/event.go
+++ b/mcp/event.go
@@ -283,6 +283,12 @@ func (s *MemoryEventStore) Append(_ context.Context, sessionID string, streamID 
 // index is no longer available.
 var ErrEventsPurged = errors.New("data purged")
 
+// ErrUnknownSession is the error that [EventStore.After] should return if the session ID is unknown.
+var ErrUnknownSession = errors.New("unknown session ID")
+
+// ErrUnknownSession is the error that [EventStore.After] should return if the stream ID is unknown.
+var ErrUnknownStream = errors.New("unknown stream ID")
+
 // After implements [EventStore.After].
 func (s *MemoryEventStore) After(_ context.Context, sessionID string, streamID StreamID, index int) iter.Seq2[[]byte, error] {
 	// Return the data items to yield.
@@ -292,11 +298,11 @@ func (s *MemoryEventStore) After(_ context.Context, sessionID string, streamID S
 		defer s.mu.Unlock()
 		streamMap, ok := s.store[sessionID]
 		if !ok {
-			return nil, fmt.Errorf("MemoryEventStore.After: unknown session ID %q", sessionID)
+			return nil, fmt.Errorf("MemoryEventStore.After: session ID %v: %w", sessionID, ErrUnknownSession)
 		}
 		dl, ok := streamMap[streamID]
 		if !ok {
-			return nil, fmt.Errorf("MemoryEventStore.After: unknown stream ID %v in session %q", streamID, sessionID)
+			return nil, fmt.Errorf("MemoryEventStore.After: stream ID %v in session %q: %w", streamID, sessionID, ErrUnknownStream)
 		}
 		start := index + 1
 		if dl.first > start {
--- a/mcp/event.go
+++ b/mcp/event.go
@@ -274,6 +274,12 @@ func (s *MemoryEventStore) Append(_ context.Context, sessionID string, streamID 
 	// Purge before adding, so at least the current data item will be present.
 	// (That could result in nBytes > maxBytes, but we'll live with that.)
 	s.purge()
+
+	// An empty data slice signals that a stream has been registered.
+	// We ignore it since it contains no content and shouldn't affect the total size.
+	if data == nil {
+		return nil
+	}
 	dl.appendData(data)
 	s.nBytes += len(data)
 	return nil
@@ -282,12 +288,6 @@ func (s *MemoryEventStore) Append(_ context.Context, sessionID string, streamID 
 // ErrEventsPurged is the error that [EventStore.After] should return if the event just after the
 // index is no longer available.
 var ErrEventsPurged = errors.New("data purged")
-
-// ErrUnknownSession is the error that [EventStore.After] should return if the session ID is unknown.
-var ErrUnknownSession = errors.New("unknown session ID")
-
-// ErrUnknownSession is the error that [EventStore.After] should return if the stream ID is unknown.
-var ErrUnknownStream = errors.New("unknown stream ID")
 
 // After implements [EventStore.After].
 func (s *MemoryEventStore) After(_ context.Context, sessionID string, streamID StreamID, index int) iter.Seq2[[]byte, error] {
@@ -298,11 +298,11 @@ func (s *MemoryEventStore) After(_ context.Context, sessionID string, streamID S
 		defer s.mu.Unlock()
 		streamMap, ok := s.store[sessionID]
 		if !ok {
-			return nil, fmt.Errorf("MemoryEventStore.After: session ID %v: %w", sessionID, ErrUnknownSession)
+			return nil, fmt.Errorf("MemoryEventStore.After: unknown session ID %q", sessionID)
 		}
 		dl, ok := streamMap[streamID]
 		if !ok {
-			return nil, fmt.Errorf("MemoryEventStore.After: stream ID %v in session %q: %w", streamID, sessionID, ErrUnknownStream)
+			return nil, fmt.Errorf("MemoryEventStore.After: unknown stream ID %v in session %q", streamID, sessionID)
 		}
 		start := index + 1
 		if dl.first > start {
--- a/mcp/event.go
+++ b/mcp/event.go
@@ -153,8 +153,8 @@ func scanEvents(r io.Reader) iter.Seq2[Event, error] {
 //
 // All of an EventStore's methods must be safe for use by multiple goroutines.
 type EventStore interface {
-	// Open prepares the event store for a given session. It ensures that the
-	// underlying data structure for the sessionID is initialized, making it
+	// Open prepares the event store for a given stream. It ensures that the
+	// underlying data structure for the stream is initialized, making it
 	// ready to store event streams.
 	Open(_ context.Context, sessionID string, streamID StreamID) error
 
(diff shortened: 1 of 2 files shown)


mcp/streamable.go
```diff
@@ -675,41 +652,48 @@ func (c *streamableServerConn) respondSSE(stream *stream, w http.ResponseWriter,
...
-			stream.outgoing = nil
-			nOutstanding := len(stream.requests)
-			c.mu.Unlock()
-
-			for _, data := range outgoing {
-				if !yield(data, true) {
+			for data, err := range c.eventStore.After(ctx, c.SessionID(), stream.id, lastIndex) {
+				if err != nil {
+					// Wait for session initialization before yielding.
+					if errors.Is(err, ErrUnknownSession) || errors.Is(err, ErrUnknownStream) {
```
jba: Instead of doing it this way, I would avoid calling After at all if there is no session or stream.

If there is a session and stream and After returns one of these errors, I think it is a real error and should be yielded.

samthanawalla: I am not sure there is an easy way to do that because the session and stream may exist but it may not exist in the event store yet. 

Append could happen before or after the After call which is why we need After to report the error to us.

jba: You should only call After when the client sends Last-Event-ID. If they send it too early, the server should return an error. I don't understand the state where After is called before Append.

mcp/event.go
```diff
@@ -283,6 +283,12 @@ func (s *MemoryEventStore) Append(_ context.Context, sessionID string, streamID
 // index is no longer available.
 var ErrEventsPurged = errors.New("data purged")
 
+// ErrUnknownSession is the error that [EventStore.After] should return if the session ID is unknown.
```
findleyr: Nit: s/return/wrap: we don't return this error value exactly, but one that wraps it.

mcp/streamable.go
samthanawalla: >>> You should only call After when the client sends Last-Event-ID.

We can call After with an index of -1 to start writing from the beginning of the stream which allows us to simplify the logic even if last-event-id is not sent.

>>> I don't understand the state where After is called before Append.

After is called when respondSSE is called which is disjoint from when Append is called in Write. These events can happen in any order which is why we case on ErrUnknownSession and ErrUnknownStream to skip to the logic below which waits for a stream signal.

mcp/streamable.go
```diff
@@ -675,41 +652,48 @@ func (c *streamableServerConn) respondSSE(stream *stream, w http.ResponseWriter,
...
+			nOutstanding := len(stream.requests)
+			c.mu.Unlock()
 
 			// If all requests have been handled and replied to, we should terminate this connection.
 			// "After the JSON-RPC response has been sent, the server SHOULD close the SSE stream."
 			// §6.4, https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#sending-messages-to-the-server
 			// We only want to terminate POSTs, and GETs that are replaying. The general-purpose GET
 			// (stream ID 0) will never have requests, and should remain open indefinitely.
-			if nOutstanding == 0 && !persistent {
+			if nOutstanding == 0 && !persistent && lastIndex >= int(stream.lastWriteIndex.Load()) {
```
findleyr: I originally thought that this commit reintroduced the bug that was fixed in https://github.com/findleyr/go-sdk/commit/f4a9396942843dbbe20cb3d4d1de962e720ce887. However, I see that it probably doesn't, because of this atomic check.

I think it would be simpler to just move the check for nOutstanding *above* the After loop above. Then you don't need lastWriteIndex. WDYT?

I prefer to avoid atomics when there's already a synchronization mechanism (mu), because it's hard to reason about the relationship between the atomics and critical sections.

mcp/streamable.go
findleyr: Why don't we just call Append with nil data when we create the stream?

Here's the problem: I'd like the eventstore to be able to completely clean up the stream or session at will, and so when we get an unknown session or stream, we should fail this connection because it will never be recoverable.

mcp/event.go
```diff
@@ -274,6 +274,12 @@ func (s *MemoryEventStore) Append(_ context.Context, sessionID string, streamID
 	// Purge before adding, so at least the current data item will be present.
 	// (That could result in nBytes > maxBytes, but we'll live with that.)
 	s.purge()
+
+	// An empty data slice signals that a stream has been registered.
```
jba: "...signals that the stream has just been created."

This behavior should be documented in EventStore.After.

jba: Actually, I'd rather have a new method for this, instead of overloading append. `Start(ctx, sessionID, streamID)`?

IIUC, the purpose of this is so that After will not return an error if it is called before the first Append, is that right?

mcp/streamable.go
```diff
@@ -675,20 +653,27 @@ func (c *streamableServerConn) respondSSE(stream *stream, w http.ResponseWriter,
 // If the stream did not terminate normally, it is either because ctx was
 // cancelled, or the connection is closed: check the ctx.Err() to differentiate
 // these cases.
-func (c *streamableServerConn) messages(ctx context.Context, stream *stream, persistent bool) iter.Seq2[json.RawMessage, bool] {
-	return func(yield func(json.RawMessage, bool) bool) {
+func (c *streamableServerConn) messages(ctx context.Context, stream *stream, persistent bool, lastIndex int) iter.Seq2[json.RawMessage, error] {
```
jba, line 811: doc the persistent and lastIndex args.

mcp/streamable.go
```diff
@@ -675,20 +653,27 @@ func (c *streamableServerConn) respondSSE(stream *stream, w http.ResponseWriter,
...
-			c.mu.Unlock()
-
-			for _, data := range outgoing {
-				if !yield(data, true) {
+			for data, err := range c.eventStore.After(ctx, c.SessionID(), stream.id, lastIndex) {
+				if err != nil {
+					yield(nil, err)
 					return
 				}
+				// The stream exists, but does not contain any messages on the stream.
```
jba: s/on the stream//

mcp/streamable.go
```diff
@@ -675,20 +653,27 @@ func (c *streamableServerConn) respondSSE(stream *stream, w http.ResponseWriter,
...
+				// Do not yield this data.
+				if data == nil {
```
jba: I'm confused. If there are no messages, how did we get inside this loop? Shouldn't After yield nothing at all?

samthanawalla:
> Sorry that this PR is confusing me so much. I think I need to see the algorithm as a whole, without diffs. So if you think it is correct after addressing my comments that it is correct, I'll approve and then read it separately.

Sorry I prematurely sent a new commit- not ready for review yet! (The flow is very confusing, I agree)

mcp/streamable.go
```diff
@@ -816,27 +788,30 @@ func (c *streamableServerConn) respondSSE(stream *stream, w http.ResponseWriter,
...
-			outgoing := stream.outgoing
-			stream.outgoing = nil
-			nOutstanding := len(stream.requests)
-			c.mu.Unlock()
-
-			for _, data := range outgoing {
-				if !yield(data, true) {
+			for data, err := range c.eventStore.After(ctx, c.SessionID(), stream.id, lastIndex) {
+				if err != nil {
+					break
```
samthanawalla: yielding this error results in test failures since the stream may or may not exist in the eventstore at this point in time.

Not sure how to fix this.

mcp/streamable.go
samthanawalla: Removed lastWriteIndex

mcp/event.go
samthanawalla: Yes- but removed this additional Append.

mcp/streamable.go
samthanawalla: fixed

samthanawalla:
Ready for review. The challenge is now with properly yielding the _eventstore.After_ error.
Yielding this error immediately is not quite right since the _eventstore.Append_ could happen after _messages_ is called.

So we can either:
1. skip streamid/session does not exist errors but yield other errors (PurgedError)
2. skip yielding all eventstore errors and wait for an Append
3. add some other synchronization methods such as eventstore.Start()

We can discuss IRL if that's better.

mcp/streamable.go
samthanawalla: See latest message

findleyr:
@jba and I were discussing yesterday, and he also suggested adding something like EventStore.Start. What should it be called though?

We could use this to allow customizing the stream id...
Here are a few options
- the name could be `NewStream`, `StartStream`, `OpenStream`
- The signature could be `func(StreamID)` (accept a stream id) or `func() StreamID` (return a stream id).

samthanawalla:
> @jba and I were discussing yesterday, and he also suggested adding something like EventStore.Start. What should it be called though?
> 
> We could use this to allow customizing the stream id... Here are a few options
> 
> * the name could be `NewStream`, `StartStream`, `OpenStream`
> * The signature could be `func(StreamID)` (accept a stream id) or `func() StreamID` (return a stream id).

Sounds good. Latest commit adds the EventStore.Open

mcp/event.go
```diff
@@ -153,6 +153,11 @@ func scanEvents(r io.Reader) iter.Seq2[Event, error] {
 //
 // All of an EventStore's methods must be safe for use by multiple goroutines.
 type EventStore interface {
+	// Open prepares the event store for a given session. It ensures that the
+	// underlying data structure for the sessionID is initialized, making it
+	// ready to store event streams.
+	Open(_ context.Context, sessionID string, streamID StreamID) error
```
jba, line 159: Does this need to be called once per session, or once per stream? If the latter, the doc is wrong.

mcp/event.go
```diff
@@ -153,6 +153,11 @@ func scanEvents(r io.Reader) iter.Seq2[Event, error] {
...
+
 	// Append appends data for an outgoing event to given stream, which is part of the
 	// given session.
 	Append(_ context.Context, sessionID string, _ StreamID, data []byte) error
```
jba, line 163: "The session (stream?) must have been opened previously (see [EventStore.Open]).

mcp/event.go
```diff
@@ -256,11 +261,21 @@ func NewMemoryEventStore(opts *MemoryEventStoreOptions) *MemoryEventStore {
...
 	s.mu.Lock()
 	defer s.mu.Unlock()
+	s.init(sessionID, streamID)
+	return nil
+}
 
+// init is an internal helper function that ensures the nested map structure for a
+// given sessionID and streamID exists, creating it if necessary. It returns the
+// dataList associated with the specified IDs.
+// This function must be called within a locked context.
```
jba: "locked context" sounds like it has something to do with context.Context.
"with the lock held"
or better, "Requires s.mu."

jba: (we should have a standard phrase for this)

mcp/streamable.go
```diff
@@ -706,8 +713,8 @@ func (c *streamableServerConn) respondJSON(stream *stream, w http.ResponseWriter
 
 	var msgs []json.RawMessage
 	ctx := req.Context()
-	for msg, ok := range c.messages(ctx, stream, false) {
-		if !ok {
+	for msg, err := range c.messages(ctx, stream, false, -1) {
+		if err != nil {
```
jba, line 717: Explain the logic here. Why does a canceled context => no content? Why StatusGone on error? What if it's an error from the EventStore implementation, like a disk read error? that should be StatusInternal.

mcp/streamable.go
```diff
@@ -770,44 +777,20 @@ func (c *streamableServerConn) respondSSE(stream *stream, w http.ResponseWriter,
...
 	// Repeatedly collect pending outgoing events and send them.
 	ctx := req.Context()
-	for msg, ok := range c.messages(ctx, stream, persistent) {
-		if !ok {
+	for msg, err := range c.messages(ctx, stream, persistent, lastIndex) {
+		if err != nil {
 			if ctx.Err() != nil && writes == 0 {
 				// This probably doesn't matter, but respond with NoContent if the client disconnected.
 				w.WriteHeader(http.StatusNoContent)
+			} else if errors.Is(err, ErrEventsPurged) {
```
jba: Why doesn't this match the above logic?

mcp/streamable.go
```diff
@@ -816,27 +799,31 @@ func (c *streamableServerConn) respondSSE(stream *stream, w http.ResponseWriter,
 
 // messages iterates over messages sent to the current stream.
 //
+// persistent indicates if it is the main GET listener.
```
jba: ", which should never be terminated" ("closed"?)

mcp/streamable.go
```diff
@@ -816,27 +799,31 @@ func (c *streamableServerConn) respondSSE(stream *stream, w http.ResponseWriter,
...
+// lastIndex is the index of the last seen event.
+//
 // The first iterated value is the received JSON message. The second iterated
-// value is an OK value indicating whether the stream terminated normally.
+// value is an error value indicating whether the stream terminated normally.
 //
```
jba, line 807: Mention that the iteration ends at the first non-nil error.

mcp/streamable.go
```diff
@@ -816,27 +799,31 @@ func (c *streamableServerConn) respondSSE(stream *stream, w http.ResponseWriter,
 
 // messages iterates over messages sent to the current stream.
 //
+// persistent indicates if it is the main GET listener.
+// lastIndex is the index of the last seen event.
```
jba: iteration begins at lastIndex+1

mcp/event.go
jba: Looks like once per stream.

samthanawalla: Done

mcp/event.go
samthanawalla: Done

mcp/event.go
samthanawalla: Done

mcp/streamable.go

mcp/streamable.go
samthanawalla: Done

mcp/streamable.go

mcp/streamable.go
samthanawalla: It should. Updated it.

mcp/streamable.go
samthanawalla: Changed to status internal. I do not know why canceled context means no content. I think that decision was made previously.

findleyr: It doesn't matter what the http status is, since the HTTP request is terminated from the client and they'll likely never see the status code.

mcp/event.go
findleyr: Maybe we should call this OpenStream or StartStream to make that clearer? I'd be fine with either.

findleyr (commented):
LGTM, just want to revisit the name 'Open'.

mcp/event.go
jba: Everything is XXXStream. Append to stream, get messages from stream. The only session method has "Session" in the name. So "Open" is right.

jba: I said that badly. I meant that EventStore.M is about streams unless M explicitly says otherwise.

jba: Also, Open takes a stream ID, so of course it is opening a stream.

jba:
> The challenge is now with properly yielding the _eventstore.After_ error. Yielding this error immediately is not quite right since the _eventstore.Append_ could happen after _messages_ is called.
> 
> So we can either:
> 
> 1. skip streamid/session does not exist errors but yield other errors (PurgedError)
> 2. skip yielding all eventstore errors and wait for an Append
> 3. add some other synchronization methods such as eventstore.Start()
> 
> We can discuss IRL if that's better.

Not sure I understand. An error from `After` should mean that things are broken: the session or stream is unknown, the data you need is gone, the underlying storage has failed, etc. It's not an error to exhaust the available messages (even there are zero). The outer loop in `messages` takes care of `Append`s that come later. I think `Open` fixes the "unknown" case, right?

samthanawalla:
> > The challenge is now with properly yielding the _eventstore.After_ error. Yielding this error immediately is not quite right since the _eventstore.Append_ could happen after _messages_ is called.
> > So we can either:
> > 
> > 1. skip streamid/session does not exist errors but yield other errors (PurgedError)
> > 2. skip yielding all eventstore errors and wait for an Append
> > 3. add some other synchronization methods such as eventstore.Start()
> > 
> > We can discuss IRL if that's better.
> 
> Not sure I understand. An error from `After` should mean that things are broken: the session or stream is unknown, the data you need is gone, the underlying storage has failed, etc. It's not an error to exhaust the available messages (even there are zero). The outer loop in `messages` takes care of `Append`s that come later. I think `Open` fixes the "unknown" case, right?

Yes Open fixes the unknown case now.

jba (approved):

PR #336, opened 2025-08-20
mcp/server: expose InitializeParams to ServerSession

### LINKED ISSUE #141: Reachable client capabilities in Server Session
MCP Servers may want to check the capabilities of the client and decide actions based on it. For instance, MCP Server checks whether client supports elicitation or not. If client supports this, MCP Server initiates an elicitation request.

Client capabilities are already embedded and stored in https://github.com/modelcontextprotocol/go-sdk/blob/78a66a438a4ef06fa50c8e5d6c559669f26ccd3b/mcp/server.go#L523 ServerSession. However, MCP Servers can't access to this field. I recommend to introduce a new function such as `GetInitializeParams`, or whatever that is used in tool calls to check the client capabilities.

Nothing comes to my mind as an alternative. But I appreciate if there is already a way and we can close this issue.

**Additional context**
I've opened this https://github.com/mark3labs/mcp-go/pull/491 to embed capabilities in session. However, they are already stored in go-sdk. We just need a way to expose it.

samthanawalla:
This CL enables the server session to see what capabilities the client has by introducing the InitializeParams() function.

This CL also adds a test to ensure ClientCapabilities is accurate.

Fixes #141

### DIFF
--- a/mcp/client.go
+++ b/mcp/client.go
@@ -105,6 +105,15 @@ func (e unsupportedProtocolVersionError) Error() string {
 // ClientSessionOptions is reserved for future use.
 type ClientSessionOptions struct{}
 
+func (c *Client) capabilities() *ClientCapabilities {
+	caps := &ClientCapabilities{}
+	caps.Roots.ListChanged = true
+	if c.opts.CreateMessageHandler != nil {
+		caps.Sampling = &SamplingCapabilities{}
+	}
+	return caps
+}
+
 // Connect begins an MCP session by connecting to a server over the given
 // transport, and initializing the session.
 //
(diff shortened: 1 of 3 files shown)


mcp/client.go
```diff
@@ -105,6 +105,15 @@ func (e unsupportedProtocolVersionError) Error() string {
 // ClientSessionOptions is reserved for future use.
 type ClientSessionOptions struct{}
 
+func (c *Client) capabilities() *ClientCapabilities {
```
jba, line 108: Why did you pull this out into a function if it's only called once?

samthanawalla: Mainly so I can test it in client_test.go. This is the same pattern in server.go

PR #340, opened 2025-08-21
chore(linter): Fix golangci-lint issues

ptsgr:
Fix https://github.com/modelcontextprotocol/go-sdk/issues/281

Resolved linter issues:
- Properly handled unused return values from functions where applicable
- Added `//nolint:staticcheck` directives in justified cases to suppress specific warnings
- Added CI `golangci/golangci-lint-action@v8.0.0` step

Result running `golangci-lint`
```
$ golangci-lint run                         
0 issues.
```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,6 +19,8 @@ jobs:
         uses: actions/setup-go@v5
         with:
           go-version: "^1.23"
+      - name: Run golangci-lint
+        uses: golangci/golangci-lint-action@v8.0.0
       - name: Check formatting
         run: |
           unformatted=$(gofmt -l .)
--- a/examples/client/listfeatures/main.go
+++ b/examples/client/listfeatures/main.go
@@ -45,7 +45,9 @@ func main() {
 	if err != nil {
 		log.Fatal(err)
 	}
-	defer cs.Close()
+	defer func() {
+		_ = cs.Close()
+	}()
 
 	printSection("tools", cs.Tools(ctx, nil), func(t *mcp.Tool) string { return t.Name })
 	printSection("resources", cs.Resources(ctx, nil), func(r *mcp.Resource) string { return r.Name })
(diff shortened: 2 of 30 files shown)


ptsgr:
Hi @jba @findleyr

I would like to start contributing to this project and submitted a PR where I fixed all linter issues and added the linter to the CI pipeline. Could you please take a look when you have a chance?

I’m planning to work on more significant improvements in the future, and this PR is a first step toward that.

Thank you for your time and for maintaining this project!

examples/client/listfeatures/main.go
```diff
@@ -45,7 +45,9 @@ func main() {
 	if err != nil {
 		log.Fatal(err)
 	}
-	defer cs.Close()
+	defer func() {
```
jba, line 48: We don't write in this style. It's clear enough from `defer cs.Close()` that we're ignoring the error.
All the other similar changes in these files have the same problem.

examples/server/everything/main.go
```diff
@@ -56,7 +56,7 @@ func main() {
 			return server
 		}, nil)
 		log.Printf("MCP handler listening at %s", *httpAddr)
-		http.ListenAndServe(*httpAddr, handler)
+		_ = http.ListenAndServe(*httpAddr, handler)
```
jba, line 59: log.Fatal(http.ListenAndServe(...)) is reasonable.

internal/jsonrpc2/conn.go
```diff
@@ -365,7 +365,8 @@ func (c *Connection) Call(ctx context.Context, method string, params any) *Async
 			if s.outgoingCalls[ac.id] == ac {
 				delete(s.outgoingCalls, ac.id)
 				ac.retire(&Response{ID: id, Error: err})
-			} else {
+			} else { //nolint:staticcheck
```
jba, line 368: What does the directive do?
I thought staticcheck passed on our code?

.github/workflows/test.yml
```diff
@@ -19,6 +19,8 @@ jobs:
         uses: actions/setup-go@v5
         with:
           go-version: "^1.23"
+      - name: Run golangci-lint
```
jba, line 22: We don't use this linter because it has too many false positives.

ptsgr:
@jba Thank you for the feedback! 🙏
Sorry for taking up your time

PR #342, opened 2025-08-21
all: remove ServerRequest[T] for concrete T

jba:
Replace all occurrences of ServerRequest[*CallToolParams] and other concrete instantiations with CallToolRequest and the like.

Make the XXXRequest types aliases, to preserve the convenience of generics for the internal machinery (see shared.go, for example.)

I will expand the aliases in a followup PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,7 +115,7 @@ type HiParams struct {
 	Name string `json:"name" jsonschema:"the name of the person to greet"`
 }
 
-func SayHi(ctx context.Context, req *mcp.ServerRequest[*mcp.CallToolParams], args HiParams) (*mcp.CallToolResult, any, error) {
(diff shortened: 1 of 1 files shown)


mcp/requests.go
```diff
@@ -0,0 +1,24 @@
+// Copyright 2025 The Go MCP SDK Authors. All rights reserved.
+// Use of this source code is governed by an MIT-style
+// license that can be found in the LICENSE file.
+
+// This file holds the request types.
+
+package mcp
+
+// TODO: expand the aliases
```
findleyr, line 9: To be clear, we want to keep ServerRequest as an embedded field, right?

jba: no, I was going to flatten. Is there value to that?

PR #352, opened 2025-08-22
mcp: validate tool output

### LINKED ISSUE #301: typed tool handlers should verify against OutputSchema

jba:
Validate the handler's returned output against the output schema of a tool.

Fixes #301.

### DIFF
--- a/mcp/mcp_test.go
+++ b/mcp/mcp_test.go
@@ -270,7 +270,7 @@ func TestEndToEnd(t *testing.T) {
 		AddTool(s, badout, func(_ context.Context, _ *CallToolRequest, arg map[string]any) (*CallToolResult, map[string]any, error) {
 			return nil, map[string]any{"x": 1}, nil
 		})
-		gotFail, err = cs.CallTool(ctx, &CallToolParams{Name: "badout"})
+		_, err = cs.CallTool(ctx, &CallToolParams{Name: "badout"})
 		wantMsg := `has type "integer", want "string"`
 		if err == nil || !strings.Contains(err.Error(), wantMsg) {
 			t.Errorf("\ngot  %q\nwant error message containing %q", err, wantMsg)


mcp/mcp_test.go
```diff
@@ -1154,19 +1154,19 @@ func TestPointerArgEquivalence(t *testing.T) {
...
-		}
-		r1, err := cs.CallTool(ctx, &CallToolParams{Name: t1.Name, Arguments: args})
-		if err != nil {
-			t.Fatal(err)
-		}
-		if diff := cmp.Diff(r0, r1); diff != "" {
-			t.Errorf("CallTool(%v) with no arguments mismatch (-%s +%s):\n%s", args, t0.Name, t1.Name, diff)
-		}
-	}
+	// for _, args := range []any{nil, struct{}{}} {
```
findleyr: Why is this commented out?

mcp/tool.go
```diff
@@ -42,13 +43,18 @@ func unmarshalSchema(data json.RawMessage, resolved *jsonschema.Resolved, v any)
 	if err := dec.Decode(v); err != nil {
 		return fmt.Errorf("unmarshaling: %w", err)
 	}
-	// TODO: test with nil args.
+	return validateSchema(resolved, v)
+}
+
+// TODO: test with nil args.
+func validateSchema(resolved *jsonschema.Resolved, value any) error {
+	// log.Printf("validating %s against %#v %[2]T", schemaJSON(resolved.Schema()), value)
```
findleyr: Cruft?

mcp/mcp_test.go
jba: fixed

mcp/tool.go
jba: fixed

mcp/server.go
```diff
@@ -240,8 +240,12 @@ func toolForErr[In, Out any](t *Tool, h ToolHandlerFor[In, Out]) (*Tool, ToolHan
 			}, nil
 		}
 
-		// TODO(v0.3.0): Validate out.
-		_ = outputResolved
+		// Validate output schema, if any.
+		// Skip if out is nil: we've removed "null" from the output schema, so nil won't validate.
+		if v := reflect.ValueOf(out); v.Kind() == reflect.Pointer && v.IsNil() {
```
findleyr, line 252: But `reflect.ValueOf(untyped nil)` is invalid, no? Should you also add a v.IsValid() check here?
In that case, I'm concerned as to why our tests pass.

mcp/server.go
```diff
@@ -240,8 +240,12 @@ func toolForErr[In, Out any](t *Tool, h ToolHandlerFor[In, Out]) (*Tool, ToolHan
...
+		} else if err := validateSchema(outputResolved, &out); err != nil {
+			return nil, err
```
findleyr: Wrap this error to say that the tool output failed to validate.

mcp/server.go
jba: You can call v.Kind on an invalid reflect.Value, which is all we do with it. (It will fail the reflect.Pointer check and we'll never call IsNil).

mcp/server.go

jba:
rebased, apologies

mcp/tool.go
```diff
@@ -42,13 +43,17 @@ func unmarshalSchema(data json.RawMessage, resolved *jsonschema.Resolved, v any)
 	if err := dec.Decode(v); err != nil {
 		return fmt.Errorf("unmarshaling: %w", err)
 	}
-	// TODO: test with nil args.
+	return validateSchema(resolved, v)
+}
+
+// TODO: test with nil args.
```
findleyr: I don't really understand what this TODO is saying. Do you? If so, can we clarify it?

mcp/server.go
```diff
@@ -260,8 +260,12 @@ func toolForErr[In, Out any](t *Tool, h ToolHandlerFor[In, Out]) (*Tool, ToolHan
 			}, nil
 		}
 
-		// TODO(v0.3.0): Validate out.
-		_ = outputResolved
+		// Validate output schema, if any.
+		// Skip if out is nil: we've removed "null" from the output schema, so nil won't validate.
+		if v := reflect.ValueOf(out); v.Kind() == reflect.Pointer && v.IsNil() {
```
findleyr, line 252: This needs a test. Just add a tool that returns an invalid result. (for example, by setting max/min constraints on the output schema).

The test should assert that the client gets a meaningful error message.

jba: Done.

mcp/tool.go
jba: I don't recall. Removed.

jba: If I meant "test that you can pass in a nil Resolved," that's obviously true, so no test required.

findleyr (approved):

PR #360, opened 2025-08-22
mcp: cleanup transport after keepalive ping fails

### LINKED ISSUE #258: streamable session persists past client disconnect
StreamableHTTPHandler holds a map of StreamableServerTransports. Currently, an entry from that map is freed only when the client sends a DELETE. We should also free it when the client stops responding to keepalive pings. If clients don't send DELETEs, the server memory will grow without bound.

h9jiang:
An onClose function is passed to the ServerSession and ClientSession to help cleanup resources from the caller.

The onClose function will be executed as part of the ServerSession and ClientSession closure.

Fixes #258

### DIFF
--- a/mcp/streamable.go
+++ b/mcp/streamable.go
@@ -40,6 +40,8 @@ type StreamableHTTPHandler struct {
 	getServer func(*http.Request) *Server
 	opts      StreamableHTTPOptions
 
+	onTransportDeletion func(sessionID string) // for testing only
+
 	mu sync.Mutex
 	// TODO: we should store the ServerSession along with the transport, because
 	// we need to cancel keepalive requests when closing the transport.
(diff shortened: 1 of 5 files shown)


mcp/streamable.go
```diff
@@ -40,6 +40,8 @@ type StreamableHTTPHandler struct {
 	getServer func(*http.Request) *Server
 	opts      StreamableHTTPOptions
 
+	onTransportDeletion func(sessionID string) // for testing only
```
h9jiang, line 43: I would prefer not to add this field only for testing purpose. But the test need some signal from the production indicating `transportations` have changed. 

Let me know if you have any better solution to this.

mcp/streamable_test.go
```diff
@@ -332,6 +333,78 @@ func testClientReplay(t *testing.T, test clientReplayTest) {
...
+			if id == nClient {
+				t.Errorf("creating more than %v session", nClient)
+			}
+			chans[fmt.Sprint(id)] = make(chan struct{}, 1)
+			return fmt.Sprint(id)
+		},
+	})
+
+	handler.onTransportDeletion = func(sessionID string) {
+		mu.Lock()
```
findleyr: There is no need to lock here: channel sends are concurrency safe.

mcp/transport.go
```diff
@@ -122,15 +122,15 @@ func NewInMemoryTransports() (*InMemoryTransport, *InMemoryTransport) {
 }
 
 type binder[T handler, State any] interface {
-	bind(Connection, *jsonrpc2.Connection, State) T
+	bind(Connection, *jsonrpc2.Connection, State, func()) T
```
findleyr, line 126: You can add `// TODO(rfindley): the bind API has gotten too complicated. Simplify.`

findleyr (commented):
Thanks! This essentially LGTM, but a couple superficial comments.

mcp/streamable_test.go
h9jiang: Done.

mcp/transport.go
h9jiang: Done.

findleyr (approved):

PR #363, opened 2025-08-23
mcp: allow configurable terminate duration for CommandTransport

### LINKED ISSUE #322: Feature request: make stdio shutdown grace periods configurable in `CommandTransport.Close()`
`CommandTransport.Close()` currently hardcodes a 5s wait after closing stdin before escalating. This follows the MCP spec's "reasonable time" requirement, but the fixed value leads to noticeable delays (e.g., when testing with processes like `sleep` that never exit quickly).

Make this timeout configurable via a `TermGrace time.Duration` field on `CommandTransport`. Default should remain `5 * time.Second` to preserve current behavior.

* Keeping the hardcoded 5s wait (current behavior) — but this hurts short-lived or test scenarios.
* Manually killing the process from outside — works, but breaks the intended MCP shutdown sequence.

**Additional context**
The MCP spec requires a shutdown sequence of closing stdin → waiting a reasonable time → sending SIGTERM, but does not prescribe exact durations. Making `TermGrace` configurable would keep spec compliance while offering flexibility for different environments and use cases.

cruffinoni:
Make the process termination timeout configurable instead of using a hardcoded 5-second delay. This allows applications to customize the termination behavior based on their specific needs.

Fixes #322

### DIFF
--- a/mcp/cmd.go
+++ b/mcp/cmd.go
@@ -15,7 +15,6 @@ import (
 
 const (
 	defaultTerminateDuration = 5 * time.Second
-	minTerminateDuration     = 1 * time.Second
 )
 
 // A CommandTransport is a [Transport] that runs a command and communicates
@@ -24,7 +23,7 @@ type CommandTransport struct {
 	Command *exec.Cmd
 	// TerminateDuration controls how long Close waits after closing stdin
 	// for the process to exit before sending SIGTERM.
-	// If less than 1 second (including zero or negative), the default of 5s is used.
+	// If zero or negative, the default of 5s is used.
 	TerminateDuration time.Duration
 }
 
@@ -55,11 +54,11 @@ func (t *CommandTransport) Connect(ctx context.Context) (Connection, error) {
 	if err := t.Command.Start(); err != nil {
 		return nil, err
 	}
-	terminateDuration := t.TerminateDuration
-	if terminateDuration < minTerminateDuration {
-		terminateDuration = defaultTerminateDuration
+	td := t.TerminateDuration
+	if td <= 0 {
+		td = defaultTerminateDuration
 	}
-	return newIOConn(&pipeRWC{t.Command, stdout, stdin, terminateDuration}), nil
+	return newIOConn(&pipeRWC{t.Command, stdout, stdin, td}), nil
 }
 
 // A pipeRWC is an io.ReadWriteCloser that communicates with a subprocess over
--- a/mcp/cmd_test.go
+++ b/mcp/cmd_test.go
@@ -273,7 +273,7 @@ func TestCommandTransportTerminateDuration(t *testing.T) {
 			defer cancel()
 
 			// Use a command that won't exit when stdin is closed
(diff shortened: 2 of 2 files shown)


mcp/cmd.go
```diff
@@ -13,10 +13,19 @@ import (
...
+	minTerminateDuration     = 1 * time.Second
+)
+
 // A CommandTransport is a [Transport] that runs a command and communicates
 // with it over stdin/stdout, using newline-delimited JSON.
 type CommandTransport struct {
 	Command *exec.Cmd
+	// TerminateDuration controls how long Close waits after closing stdin
+	// for the process to exit before sending SIGTERM.
+	// If less than 1 second (including zero or negative), the default of 5s is used.
```
jba: I don't think there should be a minimum. 0 should be 5s, but 1 (nanosecond) should be allowed, in case someone doesn't want to wait.

mcp/cmd.go
```diff
@@ -46,15 +55,20 @@ func (t *CommandTransport) Connect(ctx context.Context) (Connection, error) {
 	if err := t.Command.Start(); err != nil {
 		return nil, err
 	}
-	return newIOConn(&pipeRWC{t.Command, stdout, stdin}), nil
+	terminateDuration := t.TerminateDuration
```
jba: Shorter name for the local: its scope is small. `td` is fine.

mcp/cmd_test.go
```diff
@@ -239,6 +239,73 @@ func createServerCommand(t *testing.T, serverName string) *exec.Cmd {
...
+				t.Fatal(err)
+			}
+
+			start := time.Now()
+			err = conn.Close()
+			elapsed := time.Since(start)
+
+			// Close() may return "signal: terminated" when the subprocess is killed,
+			// which is expected behavior for our test with a non-responsive subprocess
+			if err != nil && err.Error() != "signal: terminated" {
```
jba: Is the returned error an exec.ExitError, or something that wraps it? If so, you should be able to get the exit code.

mcp/cmd_test.go
```diff
@@ -239,6 +239,73 @@ func createServerCommand(t *testing.T, serverName string) *exec.Cmd {
...
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			ctx, cancel := context.WithCancel(context.Background())
+			defer cancel()
+
+			// Use a command that won't exit when stdin is closed
+			cmd := exec.Command("sleep", "3600")
```
jba: If something goes wrong, this test could wait an hour (in theory; I believe tests time out after 10 min by default). What about 20 seconds?

jba (changes requested):

findleyr:
@cruffinoni just checking in on this: did we miss something? Usually we wait for responses to review comments before re-reviewing, but it looks like you addressed comments. Should I re-review?

jba:
It would probably help to resolve comments that were addressed.

jba:
Anyway, thanks for the PR!

PR #364, opened 2025-08-25
make mcp.handleNotify public

slimslenderslacks:
mcp: make handleNotify function public

When implementing an MCP proxy use case, the implementations of `ResourceListChangedHandler`, `ToolListChangeHandler`, or `PromptListChangeHandler` are simpler if these handlers can re-use this function from the mcp package.  For example, to forward ToolListChangeRequests from a ClientSession to a ServerSession, a simple handler could be written like this:

```
    ToolListChangedHandler: func(ctx context.Context, req *mcp.ToolListChangedRequest) {
        _ = mcp.HandleNotify(ctx, "notifications/tools/list_changed", newServerRequest(serverSession, req.Params))
    }
```

Alternatively, since `NotifyProgress`, `Log`, `Elicit`, and `ResourceUpdated` are on the ServerSession or Session, should an analogous `Notify` be available on ServerSession?

### DIFF
--- a/mcp/client.go
+++ b/mcp/client.go
@@ -154,7 +154,7 @@ func (c *Client) Connect(ctx context.Context, t Transport, _ *ClientSessionOptio
 		hc.sessionUpdated(cs.state)
 	}
 	req2 := &initializedClientRequest{Session: cs, Params: &InitializedParams{}}
-	if err := handleNotify(ctx, notificationInitialized, req2); err != nil {
+	if err := HandleNotify(ctx, notificationInitialized, req2); err != nil {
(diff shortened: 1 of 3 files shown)


jba:
We've never thought about a case where an MCP client is a proxy to an MCP server. I suspect other things would have to be different as well. Also, handleNotify has a low-level API (it requires the name of the JSON-RPC method, for example). We can probably support this use case better.

Please open a proposal; I think we are some distance from accepting PRs on this topic.
I am going to gently close this. We can re-open it if we decide to go this way.

PR #372, opened 2025-08-26
mcp/streamable: fixes broken DELETE request on session close

mathetake:
This commit fixes the accept header check for DELETE request on connection close. Previously, the accept content type was checked unconditionally regardless of whether it is DELETE or not in StreamableHTTPHandler.ServeHTTP.

As a result any DELETE request on session close resulted in `"Accept must contain both 'application/json' and 'text/event-stream'` response with the status code 400 (invalid request) as opposed to 204 (no content).

### DIFF
--- a/mcp/streamable.go
+++ b/mcp/streamable.go
@@ -133,7 +133,7 @@ func (h *StreamableHTTPHandler) ServeHTTP(w http.ResponseWriter, req *http.Reque
 			http.Error(w, "Accept must contain 'text/event-stream' for GET requests", http.StatusBadRequest)
 			return
 		}
-	} else if !jsonOK || !streamOK {
+	} else if (!jsonOK || !streamOK) && req.Method != http.MethodDelete {
 		http.Error(w, "Accept must contain both 'application/json' and 'text/event-stream'", http.StatusBadRequest)
 		return
 	}
--- a/mcp/streamable_test.go
+++ b/mcp/streamable_test.go
@@ -687,6 +687,10 @@ func TestStreamableServerTransport(t *testing.T) {
 					},
 					wantSessionID: true,
 				},
+				{
+					method:         "DELETE",
+					wantStatusCode: http.StatusNoContent,
+				},
 			},
 		},
 		{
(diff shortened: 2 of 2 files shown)


mcp/streamable.go
```diff
@@ -133,7 +133,7 @@ func (h *StreamableHTTPHandler) ServeHTTP(w http.ResponseWriter, req *http.Reque
 			http.Error(w, "Accept must contain 'text/event-stream' for GET requests", http.StatusBadRequest)
 			return
 		}
-	} else if !jsonOK || !streamOK {
+	} else if (!jsonOK || !streamOK) && req.Method != http.MethodDelete {
```
findleyr: Sorry, this is turned into a bit of a mess (not because of this CL, but because of a combination of many CLs adding handing for http method.

Can you add a `// TODO: consolidate with handling of http method below`? Thanks.

mcp/streamable_test.go
```diff
@@ -945,7 +949,9 @@ func (s streamableRequest) do(ctx context.Context, serverURL, sessionID string,
 		req.Header.Set("Mcp-Session-Id", sessionID)
 	}
 	req.Header.Set("Content-Type", "application/json")
-	req.Header.Set("Accept", "application/json, text/event-stream")
+	if s.method != http.MethodDelete { // DELETE expects "No Content" response.
```
findleyr: For your test please set this in s.headers, which will overwrite the default.
Let's not add this special handling here.

mathetake: so are you suggesting to add `"Accept", "application/json, text/event-stream"` every single one of streamableRequest instantiation for non-DELETE methods? That could be a relatively large change across many places in this test file but if that's ok with you, then i will do that

findleyr: No, I mean just set the `headers` field in your new test case. On line 955 below, it overrides req.Headers. That way, you don't need to add any more logic in the test body: it is already supported by the test case.

mathetake: ah ok now i see it

mcp/streamable.go
mathetake: done

mcp/streamable_test.go
mathetake: done

mathetake:
@findleyr thanks for the comment. I addressed both comments.

findleyr (approved):

PR #374, opened 2025-08-28
fix: complete

zhaohuabing:
### PR Guideline

You **must** follow the form:

```
net/http: handle foo when bar

Fixes #12345
```

- there is no trailing period

Additionally:

- Markdown is allowed.

### DIFF
--- a/mcp/server.go
+++ b/mcp/server.go
@@ -407,7 +407,7 @@ func (s *Server) capabilities() *ServerCapabilities {
 	return caps
 }
 
-func (s *Server) complete(ctx context.Context, req *CompleteRequest) (Result, error) {
+func (s *Server) complete(ctx context.Context, req *CompleteRequest) (*CompleteResult, error) {
 	if s.opts.CompletionHandler == nil {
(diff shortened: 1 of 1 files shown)


jba:
Can you add a test? Our tests should have caught this.

jba:
Also the commit message should be something like:
```
mcp: change return value of Server.complete to concrete type

This makes it consistent with other methods that implement RPCs.

Fixes #375.
```

zhaohuabing:
> See comments in `Conversation`.

@jba this is my first PR on MCP Go SDK - could you suggest where is the best place to add this test?

zhaohuabing:
> > See comments in `Conversation`.
> 
> @jba this is my first PR on MCP Go SDK - could you suggest where is the best place to add this test?

Guess it should be added to mcp_test.go.

zhaohuabing:
> Can you add a test? Our tests should have caught this.

Done.

findleyr:
Thank you for the fix!

findleyr (approved):

PR #380, opened 2025-08-29
examples/server/distributed: a stateless distributed server

findleyr:
Add an example of a server that proxies HTTP requests to multiple stateless streamable MCP backends.

For #148


examples/server/distributed/main.go
```diff
@@ -0,0 +1,166 @@
...
+		cmd := exec.CommandContext(ctx, exe, os.Args[1:]...)
+		cmd.Env = append(os.Environ(), fmt.Sprintf("%s=%s", childPortVar, port))
+		cmd.Stderr = os.Stderr
+		go func() {
+			defer wg.Done()
+			log.Printf("starting child %d at %s", i, childURL)
+			if err := cmd.Run(); err != nil && ctx.Err() == nil {
+				log.Printf("child %d failed: %v", i, err)
+			} else {
+				log.Printf("child %d exited gracefully", i)
```
jba, line 98: even if err != nil and ctx.Err() != nil?

examples/server/distributed/main.go
```diff
@@ -0,0 +1,166 @@
...
+	// Ensure that children are cleaned up on CTRL-C
+	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt)
+	defer stop()
+
+	// Start the child processes.
+	ports := strings.Split(*childPorts, ",")
+	var wg sync.WaitGroup
+	childURLs := make([]*url.URL, len(ports))
+	for i, port := range ports {
+		wg.Add(1)
```
jba, line 83: nit: I think it's clearer if this is just before the `go` statement.

jba (approved):

examples/server/distributed/main.go
findleyr: Yes, I think so. We typically follow the convention that if the context is cancelled, all bets are off. (This is how we do it in gopls, and it seems to work well).

You could try to check whether the error is actually a context cancellation error, but this tends to be brittle and not worthwhile.

examples/server/distributed/main.go
findleyr: Done.

PR #381, opened 2025-08-30
examples/http: update to v0.3.0

jba:
Fix build breakage.

PR #389, opened 2025-09-01
mcp/streamable: reuse transport if sessionID is given in stateless mode

### LINKED ISSUE #387: Server notifications are not sent out in stateless mode
**Describe the bug**
With https://github.com/modelcontextprotocol/go-sdk/pull/307 the `Stateless` flag was introduced. The documentation for the flag says that `Server->Client notifications may reach the client if they are made in the context of an incoming request`. However, I cannot get them to work again after the update to 0.3.0 of this library.

I explicitly set the logging level to debug in the inspector:
Image

The state update also reaches the server:
Image

But if I look at the stored `logLevel` when sending a notification, it is empty:
Image

So I think the state is not correctly handled in stateless mode. Or am I missing something?

**To Reproduce**
Minimal example:
```go
package main

import (
	"context"
	"flag"
	"log"
	"net/http"
	"os"

)

var httpAddr = flag.String("http", "", "if set, use streamable HTTP at this address, instead of stdin/stdout")

func main() {
	flag.Parse()

	server := mcp.NewServer(&mcp.Implementation{Name: "everything"}, nil)

	mcp.AddTool(server, &mcp.Tool{Name: "greet", Description: "say hi"}, contentTool)

	if *httpAddr != "" {
		handler := mcp.NewStreamableHTTPHandler(
			func(*http.Request) *mcp.Server {
				return server
			}, &mcp.StreamableHTTPOptions{Stateless: true},
		)
		http.ListenAndServe(*httpAddr, handler)
	} else {
		t := &mcp.LoggingTransport{Transport: &mcp.StdioTransport{}, Writer: os.Stderr}
		if err := server.Run(context.Background(), t); err != nil {
			log.Printf("Server failed: %v", err)
		}
	}
}

type args struct {
	SomeString string `json:"someString" jsonschema:"A random string"`
}

func contentTool(ctx context.Context, req *mcp.CallToolRequest, args args) (*mcp.CallToolResult, any, error) {
	req.Session.Log(ctx, &mcp.LoggingMessageParams{Logger: "logger", Level: "debug", Data: "Some data"})

	return &mcp.CallToolResult{
		Content: []mcp.Content{
			&mcp.TextContent{Text: "Something"},
		},
	}, nil, nil
}
```

1. Run the code above
2. Connect to the server and set logging level to debug
3. Perform a request

**Expected behavior**
The example above should work as the notification is sent in the context of the incoming request.

ln-12:
This change fixes logging not working in stateless mode as the transport (and therefore the configured log level) is not used across requests. With this change, the transport is reused in the case that a non empty sessionID was given. Fixes #387.


ln-12:
Build pipeline is broken until https://github.com/modelcontextprotocol/go-sdk/pull/381 is merged.

Thaikoma:
192.168.1.1

findleyr:
Hi, I don't understand this change. If the server is distributed, there's no way that state changes (such as the configured log level) can be shared across multiple backends.

Log messages in the context of a single request should work, but they must be at the default log level.

ln-12:
What would be your suggestion how to solve it then? Currently notifications directly fail [here](https://github.com/modelcontextprotocol/go-sdk/blob/main/mcp/server.go#L880-L885) because the `logLevel` is always empty.

jba:
The default log level is "no logging". The spec is unclear about this. I filed an [issue](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/546) about that some time ago, but it's never been addressed.

I looked at a couple of other SDKs:
- TypeScript: the SDK doesn't provide any help for logging.
- C#: it seems to behave like ours: no logging if no level is set.

I'm fine if we want to treat the default level as INFO. We may have to change that if the spec ever clarifies things, but at this point that doesn't seem very likely.

I don't see how this PR can fix that in general. If you're in stateless mode, it's probably because you have a distributed server. In that case, even if the client sets the log level, that request will go to only one server process. A better fix is to change the logic at the [code you linked to](https://github.com/modelcontextprotocol/go-sdk/blob/main/mcp/server.go#L880-L885).

ln-12:
Alright, thanks for the clarification. Then I'll close this one and we can continue discussion in the issue.

PR #392, opened 2025-09-02
cleanup: remove reference to deleted jsonschema package

bradhoekstra:
Signed-off-by: Brad Hoekstra <bhoekstra@google.com>


jba (approved):

jba:
The change is fine, but it has to happen in internal/readme/README.src.md and then you have to run the makefile.

bradhoekstra:
> The change is fine, but it has to happen in internal/readme/README.src.md and then you have to run the makefile.

Done.

findleyr:
Thanks!

PR #397, opened 2025-09-03
fix: export jsonResponse field

CCpro10:
StreamableHTTPOptions.JsonResponse should be exported.

### DIFF
--- a/mcp/streamable.go
+++ b/mcp/streamable.go
@@ -72,8 +72,8 @@ type StreamableHTTPOptions struct {
 
 	// TODO: support session retention (?)
 
-	// jsonResponse is forwarded to StreamableServerTransport.jsonResponse.
-	jsonResponse bool
+	// JSONResponse is forwarded to StreamableServerTransport.jsonResponse.
+	JSONResponse bool
 }
 
 // NewStreamableHTTPHandler returns a new [StreamableHTTPHandler].
@@ -233,7 +233,7 @@ func (h *StreamableHTTPHandler) ServeHTTP(w http.ResponseWriter, req *http.Reque
 		transport = &StreamableServerTransport{
 			SessionID:    sessionID,
 			Stateless:    h.opts.Stateless,
-			jsonResponse: h.opts.jsonResponse,
+			jsonResponse: h.opts.JSONResponse,
 		}
 
 		// To support stateless mode, we initialize the session with a default
@@ -487,7 +487,7 @@ type stream struct {
 	// jsonResponse records whether this stream should respond with application/json
 	// instead of text/event-stream.
 	//
-	// See [StreamableServerTransportOptions.jsonResponse].
+	// See [StreamableServerTransportOptions.JSONResponse].
 	jsonResponse bool
 
 	// signal is a 1-buffered channel, owned by an incoming HTTP request, that signals
--- a/mcp/streamable_test.go
+++ b/mcp/streamable_test.go
@@ -84,7 +84,7 @@ func TestStreamableTransports(t *testing.T) {
 			// Start an httptest.Server with the StreamableHTTPHandler, wrapped in a
 			// cookie-checking middleware.
 			handler := NewStreamableHTTPHandler(func(req *http.Request) *Server { return server }, &StreamableHTTPOptions{
-				jsonResponse: useJSON,
+				JSONResponse: useJSON,
 			})
 
 			var (


CCpro10:
@jba  @findleyr  Could you please take a look at this PR when you have a moment? Thanks!

findleyr:
Why should JsonResponse be exported? What is the use-case?

The reason it is unexported is that we didn't understand why it is necessary.
(I think one potential use case is proxy support for text/event-stream. Is that your use-case?)

CCpro10:
> Why should JsonResponse be exported? What is the use-case?
> 
> The reason it is unexported is that we didn't understand why it is necessary. (I think one potential use case is proxy support for text/event-stream. Is that your use-case?)

Thanks for asking for clarification.
JsonResponse should be exported because it's an optional field that needs to be assigned during initialization. For example, if JsonResponse is a boolean flag or a specific configuration related to how the server should format its messages, I need to be able to set that value from outside the package when initializing the StreamableHTTPOptions.
If it remains unexported, how would the server be able to reply in JSON format for the SSE stream, or how could I enable/disable that behavior? Without being able to assign it, the feature it represents becomes unusable from an external package.

findleyr:
@CCpro10 my question is really why the server needs to reply with application/json.

I'm not sure what you mean by "reply in JSON format for the SSE stream". Do you mean "reply with application/json rather than text/event-stream"?

Can you clarify why you need to have your server reply with application/json?

CCpro10:
@findleyr    "Yes, that's correct. My intention is for the server to reply with application/json rather than text/event-stream. If the server supports application/json for these responses, it would make the MCP client much easier to handle."

"As per the official MCP documentation, it states:
'If the input contains any number of JSON-RPC requests, the server MUST either return Content-Type: text/event-stream, to initiate an SSE stream, or Content-Type: application/json, to return one JSON object. The client MUST support both these cases.'"

"Therefore, enabling the server to respond with application/json (by allowing configuration via the JsonResponse field) is a valid and explicitly supported option according to the protocol."

findleyr:
@CCpro10 yes, the spec states that the server must return _one of_ text/event-stream or application/json. The SDK, by default, returns text/event-stream. When I added jsonResponse, I was asked why it was needed, when always preferring text/event-stream also conforms to the spec.

What is your client? If you are using a spec-conformant client, then it must support text/event-stream, and therefore it also doesn't matter from the client's perspective. Put differently, if your client needs to connect to servers that you don't control, then it must support text/event-stream.

I'm not opposed to exposing JSONResponse, but I'd like to have a real reason to do so. Again, it was originally exported in the CL that added it, but I didn't have a good reason for it. So I'd like to understand your use-case better.

FWIW, I do hear that some proxies have trouble with text/event-stream, so maybe that's a good enough reason, but it sounds like that's not the reason you need it.

CCpro10:
@findleyr image
"This is how mark3labs/mcp-go handles requests: it defaults to application/json, and only upgradeToSSE when sending notifications before a response. I believe this behavior is very reasonable."
"Regarding the reasons you mentioned, simply put, the MCP server needs full control over its behavior. Therefore, this field (JsonResponse) should either be **exported** or **removed entirely**. An unexported field in this context is both confusing and aesthetically poor (or 'ugly' in a less formal sense)."

findleyr:
@CCpro10 I think defaulting to application/json and upgrading to SSE would be fine, though it's not trivial. I can send a CL implementing that.

I'll file a proposal for exporting JSONResponse, to discuss if there are use cases. If not, we can remove it.

CCpro10:
@findleyr  OK, that's great

findleyr:
Actually, in https://github.com/modelcontextprotocol/go-sdk/issues/211, there was some feedback about the necessity of this for an MCP gateway, which I think is the same feedback as we'd heard about proxies.

Therefore, I reopened that proposal, and we can merge this (we'll leave the proposal open for some time).

mcp/streamable.go
```diff
@@ -72,8 +72,8 @@ type StreamableHTTPOptions struct {
 
 	// TODO: support session retention (?)
 
-	// jsonResponse is forwarded to StreamableServerTransport.jsonResponse.
-	jsonResponse bool
+	// JsonResponse is forwarded to StreamableServerTransport.jsonResponse.
```
findleyr: Since JSON is an initialism, when it is exported it should be JSONResponse.

mcp/streamable.go
```diff
@@ -487,7 +487,7 @@ type stream struct {
 	// jsonResponse records whether this stream should respond with application/json
 	// instead of text/event-stream.
 	//
-	// See [StreamableServerTransportOptions.jsonResponse].
+	// See [StreamableServerTransportOptions.JsonResponse].
```
findleyr: Nit: this should be `[StreamableServerTransport.JSONResponse]`. The existing comment was wrong in referencing `StreamableServerTransportOptions`. Mind fixing, as well as updating to 'JSONResponse'?

CCpro10: ok， I will fix it

CCpro10:
@findleyr I have revised the code as you requested. Json --> JSON

Issue #393, opened 2025-09-02
`failed to reconnect: Not Found` after initalize in stateless HTTP `POST` server


jpmcb:
I have an MCP server that only servers JSON over HTTP `POST` in a stateless fashion: it works like so:

```
❯ curl http://localhost:3000/mcp \
      -X POST \
      -H 'accept: application/json, text/event-stream' \
      -H 'content-type: application/json' \
       -d '{
    "jsonrpc": "2.0",
    "id": "1",
    "method": "ping"
  }' -vvv
Note: Unnecessary use of -X or --request, POST is already inferred.
* Host localhost:3000 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying [::1]:3000...
* Connected to localhost (::1) port 3000
> POST /mcp HTTP/1.1
> Host: localhost:3000
> User-Agent: curl/8.7.1
> accept: application/json, text/event-stream
> content-type: application/json
> Content-Length: 55
>
* upload completely sent off: 55 bytes
< HTTP/1.1 200 OK
< Content-Type: application/json
< Content-Length: 38
< Date: Tue, 02 Sep 2025 19:49:45 GMT
< Connection: keep-alive
< Keep-Alive: timeout=5
<
* Connection #0 to host localhost left intact
{"jsonrpc":"2.0","id":"1","result":{}}
```

But when attempting to us a very simple `go-sdk` with this like so:

```go
package main

import (
	"context"
	"log"

	"github.com/modelcontextprotocol/go-sdk/mcp"
)

func main() {
	ctx := context.Background()

	// Create a new client, with no features.
	client := mcp.NewClient(&mcp.Implementation{Name: "mcp-client", Version: "v1.0.0"}, nil)

	// Connect to a server over stdin/stdout
	transport := &mcp.StreamableClientTransport{Endpoint: "http://localhost:3000/mcp"}
	session, err := client.Connect(ctx, transport, nil)
	if err != nil {
		log.Fatal(err)
	}
	defer session.Close()

	err = session.Ping(ctx, nil)
	if err != nil {
		log.Fatalf("could not ping: %v", err)
	}
}
```

```
❯ go run main.go
2025/09/02 15:54:35 could not ping: calling "ping": failed to reconnect: Not Found
exit status 1
```

It seems that this client successfully connects and initializes but does _not_ reuse the connection:

[...]

PR #407, opened 2025-09-03
mcp: correctly disallow GET requests on stateless servers

findleyr:
The condition sessionID == "" was not quite right for disallowing GET requests. Since we decided to differentiate "stateless" vs "sessionless" servers, we need to disallow GET requests for both.

For #393

### DIFF
--- a/mcp/streamable.go
+++ b/mcp/streamable.go
@@ -171,7 +171,7 @@ func (h *StreamableHTTPHandler) ServeHTTP(w http.ResponseWriter, req *http.Reque
 
 	switch req.Method {
 	case http.MethodPost, http.MethodGet:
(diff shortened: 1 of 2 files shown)


jba (approved):

mcp/streamable.go
```diff
@@ -171,7 +171,7 @@ func (h *StreamableHTTPHandler) ServeHTTP(w http.ResponseWriter, req *http.Reque
 
 	switch req.Method {
 	case http.MethodPost, http.MethodGet:
-		if req.Method == http.MethodGet && sessionID == "" {
+		if req.Method == http.MethodGet && (h.opts.Stateless || sessionID == "") {
```
jpmcb, line 174: Thanks!!! Looks good!

PR #411, opened 2025-09-04
mcp: set default logging level to info

### LINKED ISSUE #387: Server notifications are not sent out in stateless mode

ln-12:
During initialization, the default logging level is set to info allowing servers to emit logs immediately without waiting for a client to set a log level.
The updated example shows how to send a notification in the context of the request.

Fixes #387

### DIFF
--- a/examples/server/distributed/main.go
+++ b/examples/server/distributed/main.go
@@ -144,7 +144,7 @@ func parent() {
 }
 
 func child(port string) {
-	// Create a server with a single tool that increments a counter.
+	// Create a server with a single tool that increments a counter and sends a notification.
 	server := mcp.NewServer(&mcp.Implementation{Name: "counter"}, nil)
 
 	var count atomic.Int64
@@ -153,6 +153,9 @@ func child(port string) {
 		if *verbose {
 			log.Printf("request %d (session %s)", n, req.Session.ID())
 		}
+		// Send a notification in the context of the request
+		// Hint: in stateless mode, at least log level 'info' is required to send notifications
+		req.Session.Log(ctx, &mcp.LoggingMessageParams{Data: fmt.Sprintf("request %d (session %s)", n, req.Session.ID()), Level: "info"})
 		return nil, struct{ Count int64 }{n}, nil
 	}
 	mcp.AddTool(server, &mcp.Tool{Name: "inc"}, inc)
--- a/mcp/streamable.go
+++ b/mcp/streamable.go
@@ -282,6 +282,7 @@ func (h *StreamableHTTPHandler) ServeHTTP(w http.ResponseWriter, req *http.Reque
 			if !hasInitialized {
 				state.InitializedParams = new(InitializedParams)
 			}
+			state.LogLevel = "info"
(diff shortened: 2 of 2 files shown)


examples/server/distributed/main.go
```diff
@@ -144,15 +144,19 @@ func parent() {
 }
 
 func child(port string) {
-	// Create a server with a single tool that increments a counter.
-	server := mcp.NewServer(&mcp.Implementation{Name: "counter"}, nil)
+	// Create a server with a single tool that increments a counter and sends a notification.
+	opts := &mcp.ServerOptions{
```
jba: I'd rather not introduce a new exported symbol. Let's hardcode the default level.
And let's make it Info, not Debug. Otherwise servers are likely to spam clients who don't care.
(In my experience, debug logs are often 10x the size of info logs.)

ln-12: Alright, changed it.

jba:
Thanks!

Issue #395, opened 2025-09-03
Claude Desktop Not Able To Detect The Tool


ohmpatel1997:
I have created a basic MCP server using Go SDK as given in the example, updated the configuration file of claude_desktop_config.json file as per below:

```
{
  "mcpServers": {
    "cars": {
      "command": "/Users/ohmpatel/c-mcp-servers/showroomv2/showroom",
      "env": {}
    }
  },
  "isUsingBuiltInNodeForMcp": false
}
```

but claude can not show detect the tool in conversation panel:

<img width="476" height="400" alt="Image" src="https://github.com/user-attachments/assets/bcb97ed8-4a36-4bdd-87b6-d595121179b1" />

<img width="476" height="400" alt="Image" src="https://github.com/user-attachments/assets/54539c9e-2534-4ce9-92bd-91b537031df2" />

<img width="528" height="487" alt="Image" src="https://github.com/user-attachments/assets/c78904d3-b94b-4fa1-9574-bfba725a6a2c" />

 Below is what i do:

```
func main() {
	server := mcp.NewServer(&mcp.Implementation{Name: "showroom", Version: "v1.0.0"}, nil)

	mcp.AddTool(server, &mcp.Tool{
		Name:        "list-showroom-cars",
		Description: "Retrieves a list of cars from the showroom with optional filtering by limit, currency, state, and offset."},
		getshowroomcars)
     
     // Run the server over stdin/stdout, until the client disconnects
	if err := server.Run(context.Background(), mcp.NewStdioTransport()); err != nil {
		log.Fatal(err)
	}

}
func getshowroomcars(ctx context.Context, session *mcp.ServerSession, req *mcp.CallToolParamsFor[any]) (*mcp.CallToolResultFor[mcp.Content], error) {
	return &mcp.CallToolResultFor[mcp.Content]{StructuredContent: &mcp.TextContent{
		Text: "welcome to example mcp tool",
	}}, nil
}
```

jba:
You should upgrade to v0.3.0. Perhaps we fixed whatever bug this is.

You don't need to specify the input schema when you use `mcp.AddTool`. It will infer it from the argument to `getshowroomcars`.

I think there is some confusion over Content vs. StructuredContent. Here is an example of a v0.3.0 tool handler: https://github.com/modelcontextprotocol/go-sdk/blob/v0.3.1/examples/server/basic/main.go#L19.

[...]

PR #419, opened 2025-09-05
mcp: allow requests before notifications/initialized

findleyr:
Upon a closer reading of the spec, it is OK for the server to response to client requests before it has received notifications/initialized (as long as it has received initialize). This effectively rolls back the fix from #225.

Some hooks are left for enforcing strictness around server->client requests prior to initialized. These will be revisited in subsequent CLs.

For #395


h9jiang (approved):
LGTM.

PR #427, opened 2025-09-07
fix(docs): correct number in README

flc1125:


jba (changes requested):
You have to change https://github.com/modelcontextprotocol/go-sdk/blob/main/internal/readme/README.src.md and run the Makefile there.

flc1125:
> You have to change https://github.com/modelcontextprotocol/go-sdk/blob/main/internal/readme/README.src.md and run the Makefile there.

Resolved

jba (approved):

PR #294, opened 2025-08-13
mcp/internal/oauthex: auth server metadata

jba:
Implement the Authorization Server Metadata spec.

PR #431, opened 2025-09-08
all: add a test that checks for missing or mismatching file copyright

findleyr:
In #294, we accidentally checked files with the default Go project copyright header. Avoid this in the future with a copyright test.

### DIFF
--- a/copyright_test.go
+++ b/copyright_test.go
@@ -1,0 +1,57 @@
+// Copyright 2025 The Go MCP SDK Authors. All rights reserved.
+// Use of this source code is governed by an MIT-style
+// license that can be found in the LICENSE file.
+
+package gosdk
+
+import (
+	"fmt"
+	"go/parser"
+	"go/token"
+	"io/fs"
+	"path/filepath"
+	"regexp"
+	"strings"
+	"testing"
+)
+
+func TestCopyrightHeaders(t *testing.T) {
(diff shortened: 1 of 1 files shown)


copyright_test.go
```diff
@@ -0,0 +1,56 @@
...
+	err := filepath.WalkDir(".", func(path string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+		if d.IsDir() && d.Name() != "." &&
+			(strings.HasPrefix(d.Name(), ".") ||
+				strings.HasPrefix(d.Name(), "_") ||
+				filepath.Base(d.Name()) == "testdata") {
+
+			// Skip directories starting with "." or "_", and testdata directories.
```
jba: nit: put this above L27

copyright_test.go
```diff
@@ -0,0 +1,56 @@
...
+			return filepath.SkipDir
+		}
+
+		if !strings.HasSuffix(path, ".go") {
+			// Skip non-go files.
```
jba: ditto

copyright_test.go
findleyr: Done.

copyright_test.go
findleyr: Done.

PR #439, opened 2025-09-09
async call handleNotify in notifySessions

RHzzz666:
Following previous todo comment, simply implement the async notification.

### DIFF
--- a/mcp/shared.go
+++ b/mcp/shared.go
@@ -346,18 +346,23 @@ func notifySessions[S Session, P Params](sessions []S, method string, params P) 
 	if sessions == nil {
 		return
 	}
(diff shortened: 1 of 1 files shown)


jba:
Even though I wrote that todo, I’m not convinced it’s OK to continue trying to notify asynchronously even after the function returns. Let’s notify each session concurrently, but wait until they all finish or there is a timeout before returning.

findleyr:
@jba I think this is redundant with your debouncing change, right?

jba:
Yes.

findleyr:
Ok, thanks. Let's close this then. Thanks @RHzzz666 for your patience.

Issue #21, opened 2025-06-24
Restrict JSON-RPC batching support 


findleyr:
Remove JSON-RPC batching support when the protocol version is >= 2025-06-18:
https://github.com/modelcontextprotocol/modelcontextprotocol/pull/416

ankitm123:
I started looking at this issue and will spend some time this week to get a draft PR ready.

findleyr:
Thanks @ankitm123. Assigning to you.

findleyr:
@ankitm123 I think we may need to pick this up, as it's technically a breaking change and we want to go to 1.0. If you don't think you'll be able to get to it in the next couple days, please let us know.

ankitm123:
I will get this done in the next couple of days, I have some time now for this.

findleyr:
Reopening as technically we need to do the same for stdio. Sam is on it.

PR #453, opened 2025-09-11
transport.go: disable stdio batching for newer protocols

samthanawalla:
This PR disables batching support for stdio by storing the negotiated protocol version and comparing it to protocolVersion20250618. 

Fixes: #21

### DIFF
--- a/mcp/streamable.go
+++ b/mcp/streamable.go
@@ -624,7 +624,6 @@ func (c *streamableServerConn) servePOST(w http.ResponseWriter, req *http.Reques
 	if protocolVersion == "" {
 		protocolVersion = protocolVersion20250326
 	}
-	protocolVersion = negotiatedVersion(protocolVersion)
 
 	if isBatch && protocolVersion >= protocolVersion20250618 {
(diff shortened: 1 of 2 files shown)


mcp/streamable.go
```diff
@@ -624,6 +624,7 @@ func (c *streamableServerConn) servePOST(w http.ResponseWriter, req *http.Reques
 	if protocolVersion == "" {
 		protocolVersion = protocolVersion20250326
 	}
+	protocolVersion = negotiatedVersion(protocolVersion)
```
findleyr: I'm not sure this is necessary: the header is supposed to report what we negotiated.
Changing it here is more confusing: the client said X, but we handled it like Y.

mcp/transport.go
```diff
@@ -283,6 +283,8 @@ func (r rwc) Close() error {
 //
 // See [msgBatch] for more discussion of message batching.
 type ioConn struct {
+	protocolVersion string
```
findleyr: // negotiated version, set during session initialization

mcp/streamable.go

mcp/transport.go
samthanawalla: Done

Issue #17, opened 2025-06-24
Audit security best practices


findleyr:
Audit SDK semantics to ensure that it implements (or facilitates) best practices as defined in
https://modelcontextprotocol.io/specification/2025-06-18/basic/security_best_practices

findleyr:
Done in https://github.com/modelcontextprotocol/go-sdk/blob/main/docs/protocol.md#security.

PR #457, opened 2025-09-12
auth: add integration tests for security best practices conformance

frenchi:
This PR adds integration tests to catch [Security Best Practices 2.2 Token Passthrough](https://modelcontextprotocol.io/specification/2025-06-18/basic/security_best_practices#token-passthrough).

Add HTTP middleware integration tests to validate MCP Security Best Practices conformance:
- Invalid tokens (e.g., wrong audience/unknown issuer) return `401`, set `WWW-Authenticate`, and do not invoke the handler
- Missing required scopes, return `403` and set `WWW-Authenticate`. Valid tokens succeed (`200`)
- No token passthrough: downstream requests do not receive the client Authorization header

Contributes towards [#12345](https://github.com/modelcontextprotocol/go-sdk/issues/17)

### DIFF
--- a/auth/auth_test.go
+++ b/auth/auth_test.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"errors"
 	"net/http"
+	"net/http/httptest"
 	"testing"
 	"time"
 )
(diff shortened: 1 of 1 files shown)


jba:
No objections, I just want to read the doc first. It may be a couple of days.

auth/auth_test.go
```diff
@@ -76,3 +77,105 @@ func TestVerify(t *testing.T) {
...
+		if token != "client-token" {
+			return nil, ErrInvalidToken
+		}
+		return &TokenInfo{Expiration: time.Now().Add(time.Hour)}, nil
+	}
+
+	wrapped := RequireBearerToken(verifier, nil)(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		// Simulate proxy-like behavior: perform a downstream request without
+		// forwarding the client's Authorization header.
+		resp, err := http.Get(downstream.URL)
```
jhrozek, line 162: Not saying anything is wrong with the PR, just pointing out that the SDK doesn't (and maybe shouldn't try to) prevent the pattern that I've seen when people forward tokens downstream which would roughly be:

```go
		req, err := http.NewRequest(http.MethodGet, downstream.URL, nil)
		if err != nil {
			t.Fatalf("failed to create downstream request: %v", err)
		}
		// Read and forward the Authorization header from the incoming request
		if auth := r.Header.Get("Authorization"); auth != "" {
			req.Header.Set("Authorization", auth)
		}
		resp, err := http.DefaultClient.Do(req)
		if err != nil {
			t.Fatalf("downstream request failed: %v", err)
		}
		resp.Body.Close()
```

jhrozek: (sorry clicked submit too soon)

what I meant to say is that it's fine to test that the SDK doesn't do the wrong thing on its own, but the SDK can't prevent the user from doing the wrong thing if they wish to.

findleyr:
@samthanawalla, could you please help with this review?

auth/auth_test.go
```diff
@@ -76,3 +77,105 @@ func TestVerify(t *testing.T) {
...
+
+	tests := []struct {
+		name       string
+		token      string
+		scopes     []string
+		wantCode   int
+		wantCalled bool
+	}{
+		{name: "invalid-aud", token: "bad-aud", wantCode: http.StatusUnauthorized, wantCalled: false},
+		{name: "unknown-issuer", token: "unknown-issuer", wantCode: http.StatusUnauthorized, wantCalled: false},
```
samthanawalla, line 104: Is invalid-aud and unknown-issuer the same test case?

auth/auth_test.go
```diff
@@ -76,3 +77,105 @@ func TestVerify(t *testing.T) {
...
+			if rw.Code != tt.wantCode {
+				t.Fatalf("got status %d, want %d", rw.Code, tt.wantCode)
+			}
+			if called != tt.wantCalled {
+				t.Fatalf("handler called=%v, want %v", called, tt.wantCalled)
+			}
+			if tt.wantCode == http.StatusUnauthorized || tt.wantCode == http.StatusForbidden {
+				want := "Bearer resource_metadata=" + resourceMetadata
+				if rw.Header().Get("WWW-Authenticate") != want {
+					t.Fatalf("unexpected WWW-Authenticate header: %q", rw.Header().Get("WWW-Authenticate"))
```
samthanawalla, line 135: print got and want and store w.Header().Get("WWW-Authenticate") as got

samthanawalla (changes requested):

jba:
I don't understand something. The spec says "MCP servers MUST NOT accept any tokens that were not explicitly issued for the MCP server." But the decision on whether to accept the token or not is left up to the TokenVerifier and is out of scope for our SDK. So how can we test this behavior?

maciej-kisiel:
Cleaning up older PRs. I believe this one doesn't introduce tests covering more than we already have.

Issue #460, opened 2025-09-12
Proposal: progress reporting (analogue to FastMCP)


unbekanntes-pferd:
**Is your feature request related to a problem? Please describe.**
For long running requests, I understand clients can provide a progress token - this can be used to notify about progress.
I'd like to see a convenience method as part of e.g. a tool to provide updates on progress, e.g. for larger up- and downloads.

**Describe the solution you'd like**
In FastMCP there's a way to do this with very convenient DX (see description: https://gofastmcp.com/servers/progress#progress-reporting) - essentially providing the current state and an optional total value in a function and updating it.

**Describe alternatives you've considered**
I've seen there's a notify func but I'm not sure if this covers what I need.

**Additional context**
As stated above, this is about long running requests (e.g. a download or upload of a larger file, processing many items etc.)

findleyr:
Agreed, our progress support is raw.

Please see https://github.com/modelcontextprotocol/go-sdk/pull/462 for a potential improvement. What do you think of that API?

findleyr:
Concrete proposal, implemented in #462. I'm not tied to the name 'Progress', we could call it 'ReportProgress', but I opted for the shorter name.

```go
var ErrNoProgressToken = errors.New("no progress token")

// Progress reports progress on the current request.
//
// An error is returned if sending progress failed. If there was no progress
// token, this error is ErrNoProgressToken.
func (r *ServerRequest[P]) Progress(ctx context.Context, msg string, progress, total float64) error
```

As an alternative, we could have a concept of a progress reporter that closes over the progress token and session:

```go
func (ServerRequest[P]) Progress() *ProgressReporter
func (*ProgressReporter) Report(context.Context, msg string, progress, total float64) error
```

This would be more complicated, but has a couple minor advantages: (1) it's easier to pass around the `ProgressReporter` than the generic `ServerRequest`, and (2) we could have additional helper methods on the `ProgressReporter` in the future.

unbekanntes-pferd:
> Agreed, our progress support is raw.
> 
> Please see [#462](https://github.com/modelcontextprotocol/go-sdk/pull/462) for a potential improvement. What do you think of that API?

Nice - many thanks for the very quick response and proposal - this would certainly help implement the use case I described as an example (simple up- or download). 
I'm not sure about the arguments though: Should a message `msg` always be required? Same goes for the total amount - it might be unknown at times.

I'd also vote for keeping the shorter name (`Progress`).

[...]

PR #462, opened 2025-09-12
mcp: add an ergonomic progress API

findleyr:
Add a more ergonomic API for reporting progress from server requests.

Requested in #460

DO NOT SUBMIT: needs a proposal


findleyr:
Closing this to keep tidy. We can reopen if we think this is worth doing.

Issue #444, opened 2025-09-10
Proposal: add IOTransport, a generalized transport over a Reader/Writer


atomAltera:
I run an MCP server in docker container. And I use docker go library to run container and attach to it. After attach, i have io.ReadCloser for stdout and io.WriteCloser for stdin. And i need a transport which accepts io.ReadCloser and io.WriteCloser. Same as StdioTransport but with exposed reader and writers.

I think this transport may be useful for other cases as well since it is more general than StdioTransport

findleyr:
This makes sense to me: it is generally useful, has a simple API, and must necessarily be implemented internally anyway (StdioTransport and CommandTransport are just specializations).

Once I cut v0.5.0 (today or tomorrow, I expect), I think we can merge this API addition.

Thanks.

jba:
Is it worth getting rid of StdioTransport and using an empty IOTransport in its place?

findleyr:
@jba I don't think so:
1. That would be a rather disruptive change, since StdioTransport is probably used throughout the user's code.
2. StdioTransport is discoverable, since `stdio` is used in the spec.
3. Maybe we'd want to change the behavior of stdiotransport in the future, since it is part of the spec and therefore may be subject to evolution.

On the other hand, I think we could change `NewInMemoryTransports` to return `*IOTransports`.

Thoughts?

findleyr:
@jba if you agree, please remove the release blocker and breaking change labels.

findleyr:
This is now done in #376.

PR #465, opened 2025-09-14
mcp: allow injecting custom streams in StdioTransport

robbyt:
Previously, `StdioTransport` was hardcoded to use `os.Stdin` and `os.Stdout`, making it difficult to test transport behavior with controlled input/output. With this change, users can inject custom streams (like `io.Pipe()`) for testing while maintaining backward compatibility.

This change modifies the `Connect` method to check the fields, defaulting back to the standard streams as defaults if nil.

### DIFF
--- a/mcp/transport.go
+++ b/mcp/transport.go
@@ -86,11 +86,23 @@ type serverConnection interface {
 
 // A StdioTransport is a [Transport] that communicates over stdin/stdout using
 // newline-delimited JSON.
-type StdioTransport struct{}
+type StdioTransport struct {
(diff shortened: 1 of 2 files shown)


findleyr:
This is redundant with #376, which introduces a general IOTransport. I'd like to keep StdioTransport and CmdTransport as distinct types.

Also, for testing in memory, use NewInMemoryTransports, which creates two transports connected to eachother.

findleyr:
Closing in favor of #376. 
Please comment on issue #444 if you disagree with the proposed new API.

PR #474, opened 2025-09-16
mcp: add a streamable serving benchmark

findleyr:
Add a benchmark using the streamable HTTP handler, and a more realistic
input/output signature. Also expose this handler as an example (both a
go example and a server in the examples/ directory).

### DIFF
--- a/mcp/streamable_bench_test.go
+++ b/mcp/streamable_bench_test.go
@@ -15,6 +15,8 @@ import (
 )
 
 func BenchmarkStreamableServing(b *testing.B) {
+	// This benchmark measures how fast we can handle a single tool on a
+	// streamable server, including tool validation and stream management.
 	customSchemas := map[any]*jsonschema.Schema{
 		Probability(0):  {Type: "number", Minimum: jsonschema.Ptr(0.0), Maximum: jsonschema.Ptr(1.0)},
 		WeatherType(""): {Type: "string", Enum: []any{Sunny, PartlyCloudy, Cloudy, Rainy, Snowy}},
--- a/mcp/tool_example_test.go
+++ b/mcp/tool_example_test.go
@@ -189,6 +189,8 @@ func ExampleAddTool_complexSchema() {
 		if err != nil {
 			log.Fatal(err)
 		}
+		// Formatting the entire schemas would be too much output.
+		// Just check that our customizations were effective.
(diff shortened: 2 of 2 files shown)


mcp/event_test.go
```diff
@@ -287,7 +287,7 @@ func BenchmarkMemoryEventStore(b *testing.B) {
 			payload := make([]byte, test.datasize)
 			start := time.Now()
 			b.ResetTimer()
-			for i := 0; i < b.N; i++ {
+			for i := range b.N {
```
jba, line 290: b.Loop? Or is that in too recent a version?

mcp/streamable_bench_test.go
```diff
@@ -0,0 +1,64 @@
...
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/google/jsonschema-go/jsonschema"
+	"github.com/modelcontextprotocol/go-sdk/mcp"
+)
+
+func BenchmarkStreamableServing(b *testing.B) {
```
jba, line 17: doc what you're benchmarking

mcp/tool_example_test.go
```diff
@@ -83,6 +83,122 @@ func ExampleAddTool_customTypeSchema() {
...
+		log.Fatal(err)
+	}
+	defer session.Close()
+
+	// Check that the client observes the correct schemas.
+	for t, err := range session.Tools(ctx, nil) {
+		if err != nil {
+			log.Fatal(err)
+		}
+		fmt.Println("max days:", *t.InputSchema.Properties["days"].Maximum)
```
jba, line 194: The best way to print the whole schema is to marshal it and print the string.
Just an idea; you may prefer this.

jba (changes requested):

mcp/tool_example_test.go
findleyr: Yeah, that's what the example above this one does, but it's too much output for here.

mcp/streamable_bench_test.go
findleyr: Done.

mcp/event_test.go
findleyr: Too recent, until we update our go directive.

jba (approved):

Issue #64, opened 2025-06-27
Proposal: Give receiving middleware access to tool errors.


jba:
**Is your feature request related to a problem? Please describe.**
Middleware for tool calls won't see the actual error (value of type `error`) that the ToolHandler returns, because it is turned into text with `IsError: true` by the ToolHandler. (At least, when the ToolHandler is created with NewServerTool.) Therefore, middleware can't make decisions based on the error value.

**Describe the solution you'd like**
Middleware can see the error value.

Ideas:
1. Use an interface: 
```
type CallToolResult interface {
	ToResult() *CallToolResult
}
```
See https://github.com/orgs/modelcontextprotocol/discussions/364#discussioncomment-13574919.

2. Add a method or field to CallToolResult holding the error. The field wouldn't serialize, but could be inspected by middleware.
```
// SetError sets the Content and IsError fields to reflect err. It also stores err itself for retrieval by GetError.
// SetError is a no-op if err == nil.
func (c *CallToolResult[In, Out]) SetError(err error) 

// GetError returns the error set by SetError, or nil if none.
func (c *CallToolResult[In, Out]) GetError() error
```

SamMorrowDrums:
Im our work with mcp-go we did initially wish for option 2. It would certainly suffice. 

I think either approach works. 2 is simpler and likely sufficient, 1 enables the original type to be upcast should that be useful, and allows for the ability to wrap a handler in another handler so you can compose 2nd order handler stuff or use another server as a library etc.

jba:
Can you give examples of "enables the original type to be upcast" and "ability to wrap and handler in another handler" using 1?

SamMorrowDrums:
So, as an example if a server exported the following:
```
	ServerTool := mcp.NewServerTool[*HiArgs, *HiResult]("greet", "say hi", SayHi, mcp.Input(
		mcp.Property("name", mcp.Description("the name to say hi to")),
	))
```

I was imagining something like: 

```
originalHandler := ServerTool.Handler
ServerTool.Handler := func(ctx context.Context, ss *mcp.ServerSession, params *mcp.CallToolParamsFor[json.RawMessage]) (ToolOutput, error) {
	result, err := originalHantdler(ctx, ss, params)
	if err != nil {
		return nil, err
	}

	typedResult, ok := result.(*HiResult)
	if !ok {
		return nil, fmt.Errorf("unexpected result type: %T", result)
	}

	// do something with the result, before returning it
[...]

PR #481, opened 2025-09-17
mcp: add CallToolResult.getError

jba:
Provide a way for middleware to get the error from a tool call, demonstrating that we can add this functionality without making a breaking change.

Leave getError unexported for now; we can export it (and setError) at any time.

For #64.

### DIFF
--- a/mcp/mcp_test.go
+++ b/mcp/mcp_test.go
@@ -1837,3 +1837,68 @@ func TestEmbeddedStructResponse(t *testing.T) {
 		t.Errorf("CallTool() failed: %v", err)
 	}
 }
+
+func TestToolErrorMiddleware(t *testing.T) {
+	ctx := context.Background()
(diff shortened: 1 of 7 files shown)


mcp/mcp_test.go
```diff
@@ -1837,3 +1837,68 @@ func TestEmbeddedStructResponse(t *testing.T) {
...
+		t.Fatal(err)
+	}
+	client := NewClient(&Implementation{Name: "test-client"}, nil)
+	clientSession, err := client.Connect(ctx, ct, nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer clientSession.Close()
+
+	res, err := clientSession.CallTool(ctx, &CallToolParams{
```
findleyr: _, err :=

findleyr (dismissed):
LGTM but for the staticcheck error.

Issue #478, opened 2025-09-17
Proposal: Move GetSessionID closure into ServerOptions


fgrosse:
### Is your feature request related to a problem? Please describe.

I'm writing an MCP server using the `StreamableHTTPHandler` and need more control over generating session IDs. I'm aware of the `StreamableHTTPOptions.GetSessionID() string` setting, but the problem with this closure is that it needs to be assigned at the wrong time in the lifecycle of the server.

In my use case, I'm creating a custom `mcp.Server` for each new session (using the `getServer(...)` function argument of `NewStreamableHTTPHandler(...)`). This enables me to expose different MCP tools to different sessions. Now I also want to be able to use a custom per-server logic for assigning session IDs. For instance, I would like to copy the session ID from a downstream service. Currently, this is not possible because the `StreamableHTTPOptions.GetSessionID()` closure is assigned to the HTTP handler only _once_ at application startup, rather than being an option of the `mcp.Server`, which can be created _per-session_.

### Describe the solution you'd like

Moving the `GetSessionID()` closure out of the `StreamableHTTPOptions` into the `ServerOptions` would enable me to assign a function that can use _per-session_ information instead of a closure that would only have information available at startup.

<details>

<summary>Details: Git diff</summary>

```diff
diff --git a/mcp/server.go b/mcp/server.go
index 69808ac..1621e4b 100644
--- a/mcp/server.go
+++ b/mcp/server.go
@@ -83,6 +83,16 @@ type ServerOptions struct {
 	// If true, advertises the tools capability during initialization,
 	// even if no tools have been registered.
 	HasTools bool
+
+	// GetSessionID provides the next session ID to use for an incoming request.
+	// If nil, a default randomly generated ID will be used.
+	//
+	// Session IDs should be globally unique across the scope of the server,
+	// which may span multiple processes in the case of distributed servers.
+	//
+	// As a special case, if GetSessionID returns the empty string, the
+	// Mcp-Session-Id header will not be set.
+	GetSessionID func() string
 }
 
 // NewServer creates a new MCP server. The resulting server has no features:
@@ -114,6 +124,11 @@ func NewServer(impl *Implementation, options *ServerOptions) *Server {
 	if opts.UnsubscribeHandler != nil && opts.SubscribeHandler == nil {
 		panic("UnsubscribeHandler requires SubscribeHandler")
 	}
+
+	if opts.GetSessionID == nil {
+		opts.GetSessionID = randText
+	}
+
[...]

PR #480, opened 2025-09-17
Export callback to StreamableHTTPHandler for closed transports

fgrosse:
This PR implements the changes suggested via #479 

We are introducing a new callback that callers can optionally set to receive a notification when a `StreamableHTTPHandler` connection is closed. Connections can be closed by the MCP client as part of the regular connection lifecycle (happy path) or when there was a connection error (e.g., a timeout).

---

Fixes #479

PR #488, opened 2025-09-18
Move GetSessionID closure into ServerOptions

fgrosse:
Fixes: #478

### DIFF
--- a/mcp/streamable_test.go
+++ b/mcp/streamable_test.go
@@ -409,15 +409,14 @@ func testClientReplay(t *testing.T, test clientReplayTest) {
 }
 
 func TestServerTransportCleanup(t *testing.T) {
-	server := NewServer(testImpl, &ServerOptions{KeepAlive: 10 * time.Millisecond})
-
 	nClient := 3
 
 	var mu sync.Mutex
 	var id int = -1 // session id starting from "0", "1", "2"...
 	chans := make(map[string]chan struct{}, nClient)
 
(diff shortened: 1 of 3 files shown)


mcp/streamable_test.go
```diff
@@ -409,15 +409,14 @@ func testClientReplay(t *testing.T, test clientReplayTest) {
...
-
 	nClient := 3
 
 	var mu sync.Mutex
 	var id int = -1 // session id starting from "0", "1", "2"...
 	chans := make(map[string]chan struct{}, nClient)
 
-	handler := NewStreamableHTTPHandler(func(*http.Request) *Server { return server }, &StreamableHTTPOptions{
+	server := NewServer(testImpl, &ServerOptions{
+		KeepAlive: 10 * time.Millisecond,
```
jba, line 419: Why is KeepAlive being set here?

jba:
Never mind, I did it.

mcp/streamable_test.go
fgrosse: I see you already resolved this but just for future reference: this is an artifact of moving code around (i.e., the keep alive was there already). 

I was wondering about this separately when debugging leaking goroutines and flakyness and it took me some time to realize that the keep alive here is used as a way to make the server close the connection with the client (because it would fail the keep alive after 10ms). This is necessary in this test because of the bug that transport cleanup does not happen even when the client closes the connection itself (which does happen in this test).

When we merge #480 we can actually improve this test and remove the keepalive.

PR #496, opened 2025-09-18
mcp: fix goroutine leaks in unit tests

fgrosse:
This PR fixes goroutine leaks in all unit tests.

To find the leaks I integrated https://github.com/uber-go/goleak and I suggest to keep using it to catch any future regressions. 

I used the folowing script to more easily find individual tests which were leaking goroutines:

```
for t in $(go test -run=Nothing -list=. ./mcp | grep -v ok); do go test ./mcp -run="$t" > /dev/null && echo -n . || echo -e "\nDetected leak: $t"; done
```


mcp/cmd_test.go
```diff
@@ -122,35 +137,30 @@ func TestServerInterrupt(t *testing.T) {
...
 	if err != nil {
 		t.Fatal(err)
 	}
 
-	// get a signal when the server process exits
-	onExit := make(chan struct{})
-	go func() {
-		cmd.Process.Wait()
-		close(onExit)
-	}()
```
fgrosse, line 141: This is just simplified to `cmd.Process.Wait()` further down below

mcp/mcp_test.go
```diff
@@ -118,16 +118,6 @@ func TestEndToEnd(t *testing.T) {
...
-	// Wait for the server to exit after the client closes its connection.
-	var clientWG sync.WaitGroup
-	clientWG.Add(1)
-	go func() {
-		if err := ss.Wait(); err != nil {
-			t.Errorf("server failed: %v", err)
-		}
-		clientWG.Done()
-	}()
-
```
fgrosse, line 130: Simplyfied to calling `ss.Wait()` below

mcp/mcp_test.go
```diff
@@ -750,11 +745,7 @@ func TestMiddleware(t *testing.T) {
 		t.Fatal(err)
 	}
 	// Wait for the server to exit after the client closes its connection.
-	defer func() {
-		if err := ss.Wait(); err != nil {
-			t.Errorf("server failed: %v", err)
-		}
-	}()
```
fgrosse, line 757: Closing the server session during the cleanup will already do this wait.

mcp/server.go
```diff
@@ -728,6 +729,7 @@ func (s *Server) Run(ctx context.Context, t Transport) error {
 	select {
 	case <-ctx.Done():
 		ss.Close()
+		<-ssClosed // wait until waiting go routine above actually completes
```
fgrosse, line 779: Added this condition so we are sure when `Server.Run(...)` exits, it has completed its started goroutine. This is useful for deterministic testing and finding issues (e.g., hanging Wait) faster.

mcp/server.go
```diff
@@ -1123,13 +1125,21 @@ func (ss *ServerSession) Close() error {
...
-	err := ss.conn.Close()
+
+	var connErr, mcpConnErr error
+	if err := ss.conn.Close(); err != nil {
+		connErr = fmt.Errorf("failed to close connection: %w", err)
+	}
+
+	if err := ss.mcpConn.Close(); err != nil {
+		connErr = fmt.Errorf("failed to close mcp connection: %w", err)
+	}
```
fgrosse: I checked the code, and it seemed like an oversight to me that the `mcpConn` was never closed, but please double-check if I missed something here.

mcp/streamable_test.go
```diff
@@ -316,7 +316,10 @@ func testClientReplay(t *testing.T, test clientReplayTest) {
 		})
 
 	realServer := httptest.NewServer(NewStreamableHTTPHandler(func(*http.Request) *Server { return server }, nil))
-	defer realServer.Close()
+	t.Cleanup(func() {
+		t.Log("Closing real HTTP server")
```
fgrosse, line 324: Given this function currently does not pass the leak check because of #499 , I decided to leave these logs in here which I added during debugging.

fgrosse:
@findleyr let me know what you think when you have time :)

mcp/server.go
findleyr: I think this should just be `ss.Close()`.

The mcpConn is closed automatically when the jsonrpc2 conn is closed (this happens in the overly complex `connect` function).

mcp/transport.go
```diff
@@ -284,7 +284,14 @@ func (r rwc) Write(p []byte) (n int, err error) {
 }
 
 func (r rwc) Close() error {
-	return errors.Join(r.rc.Close(), r.wc.Close())
+	rcErr := r.rc.Close()
+
+	var wcErr error
+	if r.wc != nil {
```
findleyr: Let's add a comment that we only allow a nil writer for tests.

mcp/streamable_test.go
findleyr: Thanks, makes sense. How about also adding a comment referencing #499. Something like:

```
// Until we have a way to clean up abandoned sessions, this test will leak goroutines (see #499)
```

mcp/server.go
findleyr: Yeah, nice catch!

mcp/cmd_test.go
findleyr: But the previous logic was serving a purpose: the test would wait at most 5s.
_Generally speaking_, it's bad form to depend on timing in tests, but pragmatically it can be useful, and in any case we probably should separate this simplification into a separate CL, if we really care about it.

fgrosse: I'm fine with removing this if you prefer the more complex setup.

Just sharing my motivation: In my experience it's typically acceptable to let the runtime / CI enforce a timeout in these situations (even if the default timeout of 10m can feel long). For the happy path (no test failures) it makes no difference but you get slightly less complexity in your test. When the test starts to hang, I would notice during local development, where I would expect developers to run it in isolation and make use of the `-timeout` flag of `go test`

mcp/server.go
fgrosse: Fixed via a43fb6ff32b6c00bcf015d8bc76be5f01f0a1b15

mcp/streamable_test.go
fgrosse: Done via f41256c20656c0ba1de063cf13d12749234529b0

mcp/transport.go
fgrosse: Done via 415ff5665186a860a3c7097d09e0b51ff36f5fc6

fgrosse:
I have addressed all comments and merged the most recent `main` into this branch. Good to go?

mcp/client_example_test.go
```diff
@@ -42,15 +42,17 @@ func Example_roots() {
 
 	// Connect the server and client...
 	t1, t2 := mcp.NewInMemoryTransports()
-	if _, err := s.Connect(ctx, t1, nil); err != nil {
+	sess1, err := s.Connect(ctx, t1, nil)
```
findleyr: s/sess1/serverSession (or ss)
s/sess2/clientSession (or cs)

sess1 and sess2 obscures the fact that these variables have different types.

mcp/cmd_test.go
findleyr: Ok, this is ifne.

mcp/streamable_example_test.go
```diff
@@ -26,7 +26,7 @@ func ExampleStreamableHTTPHandler() {
 	server := mcp.NewServer(&mcp.Implementation{Name: "server", Version: "v0.1.0"}, nil)
 	handler := mcp.NewStreamableHTTPHandler(func(r *http.Request) *mcp.Server {
 		return server
-	}, &mcp.StreamableHTTPOptions{JSONResponse: true})
+	}, &mcp.StreamableHTTPOptions{JSONResponse: true, Stateless: true})
```
findleyr: Why make this Stateless?

This example is demonstrating how to use the API, so using 'Stateless' here may be distracting. Would prefer to leave a comment that this test may leak goroutines, as you've done below.

go.mod
```diff
@@ -7,6 +7,7 @@ require (
 	github.com/google/go-cmp v0.7.0
 	github.com/google/jsonschema-go v0.3.0
 	github.com/yosida95/uritemplate/v3 v3.0.2
+	go.uber.org/goleak v1.3.0
```
findleyr: Hmm, I don't think we should add an additional dependency just for this purpose. It seems like handling these as a one-off, every once in a while, is sufficient for now.

findleyr:
Hi Friedrich, we really appreciate this contribution and would like to land it. Let me know if you'd like me to take it over (mea culpa for the review latency--it has been a busy few weeks!).

go.mod
fgrosse: I see your point and removed this from the PR so we can move ahead.

It would have been beneficial to keep an automated test because it will be very easy to introduce regressions without it, but of course that decision is up to you.

fgrosse:
Sorry it took so long to respond. I finally took some time and addressed the remaining comments, so this should be good to merge now. If you would like to see more changes and want to move quickly, feel free to make them directly.

findleyr:
Thanks!

PR #501, opened 2025-09-19
mcp/server: enable server logging

IAmSurajBobade:
mcp/server.go: implement server‑side logging

    Thread a *slog.Logger through server paths to improve debuggability
    Introduce no‑op discard handler + ensureLogger so loggers are never nil
    Use ensured loggers in Server

All tests are passing and no functionality should change with this PR

Fixes https://github.com/modelcontextprotocol/go-sdk/issues/170

### DIFF
--- a/mcp/logging.go
+++ b/mcp/logging.go
@@ -104,11 +104,6 @@ func ensureLogger(l *slog.Logger) *slog.Logger {
 	return slog.New(discardHandler{})
 }
 
-// internalLogger is used for package-internal logging where we don't have a
-// specific server/handler context. It defaults to a discard logger to avoid
-// unsolicited output from library code.
-var internalLogger = slog.New(discardHandler{})
-
 // NewLoggingHandler creates a [LoggingHandler] that logs to the given [ServerSession] using a
 // [slog.JSONHandler].
 func NewLoggingHandler(ss *ServerSession, opts *LoggingHandlerOptions) *LoggingHandler {
--- a/mcp/logging.go
+++ b/mcp/logging.go
@@ -87,6 +87,28 @@ type LoggingHandler struct {
 	handler         slog.Handler
 }
 
+// discardHandler is a slog.Handler that drops all logs.
+// TODO: use slog.NewNopHandler when we require Go 1.24+.
+type discardHandler struct{}
+
+func (discardHandler) Enabled(context.Context, slog.Level) bool  { return false }
+func (discardHandler) Handle(context.Context, slog.Record) error { return nil }
+func (discardHandler) WithAttrs([]slog.Attr) slog.Handler        { return discardHandler{} }
+func (discardHandler) WithGroup(string) slog.Handler             { return discardHandler{} }
+
+// ensureLogger returns l if non-nil, otherwise a discard logger.
+func ensureLogger(l *slog.Logger) *slog.Logger {
+	if l != nil {
+		return l
+	}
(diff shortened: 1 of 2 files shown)


mcp/logging.go
```diff
@@ -87,6 +87,23 @@ type LoggingHandler struct {
 	handler         slog.Handler
 }
 
+// discardHandler is a slog.Handler that drops all logs.
```
jba, line 91: Use slog.DiscardHandler.

mcp/server.go
```diff
@@ -35,8 +36,9 @@ const DefaultPageSize = 1000
 // sessions by using [Server.Run].
 type Server struct {
 	// fixed at creation
-	impl *Implementation
-	opts ServerOptions
+	impl   *Implementation
+	opts   ServerOptions
+	logger *slog.Logger
```
jba: Use opts.Logger?

mcp/server.go
```diff
@@ -53,6 +55,8 @@ type Server struct {
 type ServerOptions struct {
 	// Optional instructions for connected clients.
 	Instructions string
+	// Logger is used for server-side logging. If nil, disable logging.
```
jba: // If non-nil, log server activity.

mcp/server.go
```diff
@@ -804,9 +829,11 @@ func (ss *ServerSession) initialized(ctx context.Context, params *InitializedPar
 	})
 
 	if !wasInit {
+		ss.server.logger.Warn("initialized before initialize")
```
jba: If we are returning an error, shouldn't we log these as errors?

mcp/streamable.go
```diff
@@ -39,6 +40,7 @@ const (
 type StreamableHTTPHandler struct {
 	getServer func(*http.Request) *Server
 	opts      StreamableHTTPOptions
+	logger    *slog.Logger
```
jba: use opts.Logger

mcp/logging.go
IAmSurajBobade: slog.DiscardHandler requires go1.24 or later. I have updated TODO comment to use slog.DiscardHandler, once we move to 1.24+

mcp/server.go
IAmSurajBobade: updated code to use opts.Logger

mcp/server.go
IAmSurajBobade: updated comment

mcp/server.go
IAmSurajBobade: changed from Warn to Error

mcp/streamable.go
IAmSurajBobade: updated code to use opts.Logger

IAmSurajBobade:
@jba addressed all of your review comments, can you please have a look.

dmarkhas:
Could you use a logging interface instead of `*slog.Logger` ? We want to be able to provide our own logging implementation, and we don't use `slog` in our codebase.

jba:
@dmarkhas You can write a slog.Handler to wrap your logger.
`log/slog` is the _lingua franca_ for logging in the Go ecosystem. Adding another interface just for this package would further fragment the logging ecosystem.

jba:
This is ready to merge, but I recommend we do it after v1.0.0 to avoid last-minute changes. @findleyr

jba:
@findleyr @samthanawalla just sign off on the public API.

PR #504, opened 2025-09-19
SSE: Add support for SSE handler options (#503)

### LINKED ISSUE #503: SSE: Add support for SSE handler options
Currently NewSSEHandler does not have support to pass Options, and adding support later will be a breaking change.

For the actual options to be provided, we can start with #170 and address TODO mentioned in /mcp/sse.go file

This will also do the ground work for any new options we intend to add later. And that time, it won't be a breaking change!!

I am thinking of using same code symanctis we are following for StreamableHTTPHandler.

```
type SSEOptions struct {
	// Logger specifies the logger to use.
	// If nil, do not log.
	Logger *slog.Logger
}
```

```
type SSEHandler struct {
       // existing fields
	opts         SSEOptions
}
```

And signature change would be like below. Which is same code standard as we do for StreamableHTTPHandler

```
func NewSSEHandler(getServer func(request *http.Request) *Server, opts *SSEOptions) *SSEHandler {
	s := &SSEHandler{
		getServer: getServer,
		sessions:  make(map[string]*SSEServerTransport),
	}

	if opts != nil {
		s.opts = *opts
	}

	if s.opts.Logger == nil { // ensure we have a logger
		s.opts.Logger = ensureLogger(nil)
	}
	s.logger = s.opts.Logger

	return s
}
```

IAmSurajBobade:
mcp/sse: add support to provide for options for SSE transport

- Add SSEOptions struct to define SSE handler options.
- Update Signature of NewSSEHandler to accept SSEOptions
- Ensure Logger is never nill, using ensureLogger helper method
- update unit tests
- update example tests

Fixes #503


jba:
The SSEOptions work has to be separate from the logging.
We want the options in today, but the logging can wait another day or two.
Please send a separate PR.

jba:
Refer to #507.

IAmSurajBobade:
closing as duplicate of #508

PR #509, opened 2025-09-20
docs: adjust example code to remove horizontal scroll bar

xieyuschen:
This change adjusts the code example in readme so if you view it in github it won't show a horizontal scroll bar in code example.

It's trivial change, but I believe it's still an improvement.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,10 @@ type Output struct {
 	Greeting string `json:"greeting" jsonschema:"the greeting to tell to the user"`
 }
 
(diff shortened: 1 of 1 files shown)


xieyuschen:
specifically talks about this. It may differ in different devices but in my own device it always shows the scroll bar.

image

jba (approved):
That's reasonable. The line is over 100 characters long.

jba (changes requested):
...except you have to modify internal/readme/server/server.go and run go generate.

Issue #449, opened 2025-09-10
Input validation doesn't catch missing fields


iwyrkore:
**Describe the bug**
The input validation for a tool call doesn't seem to catch missing required fields. It does catch invalid fields that are not in the schema.

**To Reproduce**
1. I added a second field the the basic example (and also output the tool info):

Run this file:

```
// Copyright 2025 The Go MCP SDK Authors. All rights reserved.
// Use of this source code is governed by an MIT-style
// license that can be found in the LICENSE file.

package main

import (
	"context"
	"fmt"
	"log"

	"github.com/modelcontextprotocol/go-sdk/mcp"
)

type SayHiParams struct {
	Name  string `json:"name"`
	Title string `json:"title"`
}

func SayHi(ctx context.Context, req *mcp.CallToolRequest, args SayHiParams) (*mcp.CallToolResult, any, error) {
	return &mcp.CallToolResult{
		Content: []mcp.Content{
			&mcp.TextContent{Text: "Hi " + args.Title + " " + args.Name},
		},
	}, nil, nil
}

func main() {
	ctx := context.Background()
	clientTransport, serverTransport := mcp.NewInMemoryTransports()

	server := mcp.NewServer(&mcp.Implementation{Name: "greeter", Version: "v0.0.1"}, nil)
	mcp.AddTool(server, &mcp.Tool{Name: "greet", Description: "say hi"}, SayHi)

	serverSession, err := server.Connect(ctx, serverTransport, nil)
	if err != nil {
		log.Fatal(err)
	}

	client := mcp.NewClient(&mcp.Implementation{Name: "client"}, nil)
	clientSession, err := client.Connect(ctx, clientTransport, nil)
	if err != nil {
		log.Fatal(err)
	}

	res, err := clientSession.CallTool(ctx, &mcp.CallToolParams{
		Name:      "greet",
		Arguments: map[string]any{"name": "user"},
	})
	if err != nil {
		log.Fatal(err)
	}
	fmt.Println(res.Content[0].(*mcp.TextContent).Text)

	clientSession.Close()
	serverSession.Wait()

	// Output: 
        // {"tools":[{"description":"say hi","inputSchema":{"type":"object","required":["name","title"],"properties":{"name":{"type":"string"},"title":{"type":"string"}},"additionalProperties":false},"name":"greet"}]}
Hi  user
        // Hi  user
}
```

**Expected behavior**
[...]

PR #510, opened 2025-09-21
mcp/tool: Using Reflection For ApplySchema

RyoKusnadi:
**Requirement**
// TODO: use reflection to create the struct type to unmarshal into.
// Separate validation from assignment.

// Use default JSON marshalling for validation.
//
// This avoids inconsistent representation due to custom marshallers, such as
// time.Time (issue #449).
//
// Additionally, unmarshalling into a map ensures that the resulting JSON is
// at least {}, even if data is empty. For example, arguments is technically
// an optional property of callToolParams, and we still want to apply the
// defaults in this case.
//
// TODO(rfindley): in which cases can resolved be nil?

**New Files:**
- mcp/reflection_validator.go - Validates JSON using Go struct types
- mcp/schema_type_builder.go - Converts JSON schemas to Go types

**Modified:**
- mcp/tool.go - Enhanced applySchema with reflection validation + fallback
- mcp/tool_test.go - Added tests and benchmarks

### DIFF
--- a/mcp/reflection_validator.go
+++ b/mcp/reflection_validator.go
@@ -1,0 +1,147 @@
+// Copyright 2025 The Go MCP SDK Authors. All rights reserved.
+// Use of this source code is governed by an MIT-style
+// license that can be found in the LICENSE file.
+
+package mcp
+
+import (
+	"encoding/json"
+	"fmt"
+	"reflect"
+
+	"github.com/google/jsonschema-go/jsonschema"
+)
+
+// ReflectionValidator handles validation using dynamically created types.
+// It uses SchemaTypeBuilder to create appropriate struct types for validation.
+type ReflectionValidator struct {
+	builder *SchemaTypeBuilder
+}
+
+// NewReflectionValidator creates a new ReflectionValidator with a SchemaTypeBuilder.
+func NewReflectionValidator() *ReflectionValidator {
+	return &ReflectionValidator{
+		builder: NewSchemaTypeBuilder(),
+	}
+}
+
+// SchemaValidationError represents an error that occurred during schema validation.
+type SchemaValidationError struct {
+	Operation string               // The operation that failed
+	Schema    *jsonschema.Schema   // The schema being processed
+	Resolved  *jsonschema.Resolved // The resolved schema
+	Data      json.RawMessage      // The data being validated
+	Cause     error                // The underlying error
+}
+
+// Error returns a formatted error message.
+func (e *SchemaValidationError) Error() string {
+	return fmt.Sprintf("schema validation failed during %s: %v", e.Operation, e.Cause)
+}
+
+// Unwrap returns the underlying error.
+func (e *SchemaValidationError) Unwrap() error {
+	return e.Cause
+}
+
+// ValidateAndApply validates data and applies defaults using reflection.
+// It creates a typed struct from the schema, unmarshals directly into it for type validation,
+// then converts to map format for applying defaults and final validation.
+func (v *ReflectionValidator) ValidateAndApply(data json.RawMessage, resolved *jsonschema.Resolved) (json.RawMessage, error) {
+	if resolved == nil {
+		// If no schema is provided, return data as-is
+		return data, nil
+	}
+
+	// Get the schema from the resolved schema
+	schema := resolved.Schema()
+	if schema == nil {
+		return nil, &SchemaValidationError{
+			Operation: "schema_extraction",
+			Resolved:  resolved,
+			Data:      data,
+			Cause:     fmt.Errorf("resolved schema contains no schema definition"),
+		}
+	}
+
+	// Build the struct type from the schema for type validation
+	structType, err := v.builder.BuildType(schema)
+	if err != nil {
+		return nil, &SchemaValidationError{
+			Operation: "schema_conversion",
+			Schema:    schema,
+			Resolved:  resolved,
+			Data:      data,
+			Cause:     err,
+		}
+	}
+
+	var mapData map[string]any
+	// Handle empty data case
+	if len(data) == 0 {
+		mapData = make(map[string]any)
+	} else {
+		// First, unmarshal into a map to preserve the original structure
+		if err := json.Unmarshal(data, &mapData); err != nil {
+			return nil, &SchemaValidationError{
+				Operation: "unmarshaling",
+				Schema:    schema,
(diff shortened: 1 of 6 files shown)


mcp/reflection_validator.go
```diff
@@ -0,0 +1,147 @@
...
+				Data:      data,
+				Cause:     fmt.Errorf("unmarshaling into map: %w", err),
+			}
+		}
+
+		// Create a new instance of the struct type for type validation
+		structValue := reflect.New(structType)
+		structPtr := structValue.Interface()
+
+		// Unmarshal directly into the typed struct for reflection-based type validation
```
jba, line 99: This is the part I don't understand. What is the value of the struct? Wouldn't resolved.Validate(mapData) catch all the errors that this would catch?

jba:
Thanks for this big and impressive effort!

You should always file an issue before attempting this scale of work, even if there is a TODO in the code.

In this case, I'm not really sure what problem this is solving. Perhaps @findleyr has a better sense.

RyoKusnadi:
> Thanks for this big and impressive effort!
> 
> You should always file an issue before attempting this scale of work, even if there is a TODO in the code.
> 
> In this case, I'm not really sure what problem this is solving. Perhaps @findleyr has a better sense.

ups sorry @jba @findleyr, this is my first time i contribute so i miss this
so in this case do i need to file an issue for this TODO?

jba:
> so in this case do i need to file an issue for this TODO?

Let's start with my comment in mcp/reflection_validator.go. I don't understand what this PR provides.

RyoKusnadi:
the TODO "use reflection to create the struct type to unmarshal into"
so i'm thinking to remove map validation as below and change it into using reflection

```
v := make(map[string]any)
if len(data) > 0 {
	if err := json.Unmarshal(data, &v); err != nil {
		return nil, fmt.Errorf("unmarshaling arguments: %w", err)
	}
}
if err := resolved.ApplyDefaults(&v); err != nil {
	return nil, fmt.Errorf("applying schema defaults:\n%w", err)
}
```

So I create ReflectionValidator as the base/main wrapper for schema validation. 
The ValidateAndApply function works like this: 
1. Extracts the schema from the provided *jsonschema.Resolved. 
     - Returns a SchemaValidationError if no schema is found. 
2. Builds a Go type dynamically from the schema using SchemaTypeBuilder. 
3. Unmarshals the JSON into the typed struct for reflection-based type checking. 
4. Applies default values defined in the schema.
5. Runs full schema validation on the enriched data. 
6. Returns the final JSON

For SchemaTypeBuilder:
1. Input: *jsonschema.Schema
     - Returns an error if the schema is nil.
2. Generates a unique cache key based on the schema structure.
     - Uses this key to reuse previously built types and avoid regeneration.
3. Builds a Go reflect.Type from the schema definition:
    - Maps primitive type like string,  integer, bool to their Go equivalents.
    - Recursively create struct types for object schemas.
    - Recursively create slice types for array schemas.
 4. Creates the final reflect.Type using reflect.StructOf (for objects) or reflect.SliceOf (for arrays).
 5. Stores the built type in the cache for future reuse.

here's the idea of the PR, let me know if need to change something or remove the cache key implementation @jba .

jba:
It seems like you just summarized what this PR does without answering my question. The question is: Wouldn't resolved.Validate(mapData) catch all the errors that this would catch? (see my comment on [reflection_validator.go](https://github.com/modelcontextprotocol/go-sdk/pull/510/files/49af3aaa29a9406f10dfb89dad125adcfaaa372a#diff-aed1e2d23bfdfa0b55c9d9a962d5835f85c1d0b861f239096ce6022c2a972dbf))

RyoKusnadi:
> It seems like you just summarized what this PR does without answering my question. The question is: Wouldn't resolved.Validate(mapData) catch all the errors that this would catch? (see my comment on [reflection_validator.go](https://github.com/modelcontextprotocol/go-sdk/pull/510/files/49af3aaa29a9406f10dfb89dad125adcfaaa372a#diff-aed1e2d23bfdfa0b55c9d9a962d5835f85c1d0b861f239096ce6022c2a972dbf))

Ups Sorry, I couldn’t find your comment—maybe it wasn’t submitted yet? @jba 
image

jba (changes requested):

mcp/reflection_validator.go
jba: Sorry about that!

RyoKusnadi: What i'm understand is current implementation is apply to a generic map and re-marshalled

meaning that:
it not considering Go struct field tags such as ",string". As a result, a numeric default remains a JSON number, which cannot be unmarshalled into a field expecting a string-encoded number.

Can Try Below Unit Test Or Refer To Image
```
func TestApplySchema_DefaultTypeMismatchWithStringTag(t *testing.T) {
	schema := &jsonschema.Schema{
		Type: "object",
		Properties: map[string]*jsonschema.Schema{
			// Note: default is a JSON number (3), not a quoted string.
			"n": {Type: "integer", Default: json.RawMessage("3")},
		},
	}
	resolved, err := schema.Resolve(&jsonschema.ResolveOptions{ValidateDefaults: true})
	if err != nil {
		t.Fatalf("resolve schema: %v", err)
	}

	// No arguments provided; expect defaults to be applied.
	var input json.RawMessage = json.RawMessage(`{}`)
	input, err = applySchema(input, resolved)
	if err != nil {
		t.Fatalf("applySchema: %v", err)
	}

	// Struct expects a string-encoded number because of ",string" tag.
	type In struct {
		N int `json:"n,string"`
	}
	var got In
	// Current behavior: json.Unmarshal will fail with:
	// cannot unmarshal number into Go struct field In.n of type string
	// because the JSON contains {"n":3} after defaults, but the field expects a string.
	if err := json.Unmarshal(input, &got); err == nil {
		t.Fatalf("expected unmarshal to fail due to ,string tag with numeric default; got success with value: %+v", got)
	} else {
		// Tighten expectation to document the exact failure mode.
		if !strings.Contains(err.Error(), "cannot unmarshal number into Go struct field") {
			t.Fatalf("unexpected unmarshal error: %v", err)
		}
	}
}
```
image

That's why we need reflection

or can we close this PR first? let me clean up all this first into a proposal and raise an issue for this @jba

RyoKusnadi:
will raise a issue first before we proceed into PR later on

PR #511, opened 2025-09-22
mcp: update to jsonschema that does not validate structs

jba:
Update to the latest jsonschema commit, which removes the ability to validate structs and changes the type of ForOptions.TypeSchemas.

### DIFF
--- a/examples/server/toolschemas/main.go
+++ b/examples/server/toolschemas/main.go
@@ -136,6 +136,9 @@ func main() {
 	// We don't need to do all this work: below, we use jsonschema.For to start
 	// from the default schema.
 	manual, err := newManualGreeter()
+	if err != nil {
+		log.Fatal(err)
+	}
 	server.AddTool(&mcp.Tool{
 		Name:         "manual greeting",
 		InputSchema:  inputSchema,
--- a/docs/server.md
+++ b/docs/server.md
@@ -364,9 +364,9 @@ the `Tool.InputSchema` explicitly:
 
 ```go
 // Distinguished Go types allow custom schemas to be reused during inference.
-customSchemas := map[any]*jsonschema.Schema{
-	Probability(0):  {Type: "number", Minimum: jsonschema.Ptr(0.0), Maximum: jsonschema.Ptr(1.0)},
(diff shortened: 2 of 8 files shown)


findleyr:
Do we need to update `examples/server/toolschemas`? I believe it was validating structs.

jba:
Yes indeed. Added a test.

examples/server/toolschemas/main.go
```diff
@@ -58,11 +58,11 @@ func (t *manualGreeter) greet(_ context.Context, req *mcp.CallToolRequest) (*mcp
 	if err := json.Unmarshal(req.Params.Arguments, &input); err != nil {
 		return errf("failed to unmarshal arguments: %v", err), nil
 	}
-	if err := t.inputSchema.Validate(input); err != nil {
+	if err := validateStruct(input, t.inputSchema); err != nil {
```
findleyr: This isn't right: once we re-marshal, we're no longer validating the original JSON.

For example, suppose the input contains `"foo": 0`, and the struct has `omitempty` for the `Foo` field, and `"foo"` is required. Then the original input is valid, but the re-marshalled input is not.

This is why we should expose helpers for this...

jba: Good point!

examples/server/toolschemas/main.go
```diff
@@ -54,27 +54,72 @@ func (t *manualGreeter) greet(_ context.Context, req *mcp.CallToolRequest) (*mcp
...
+	}
+
 	return &mcp.CallToolResult{
 		Content:           []mcp.Content{&mcp.TextContent{Text: string(outputJSON)}},
 		StructuredContent: output,
 	}, nil
 }
 
+// unmarshalAndValidate unmarshals data to a map[string]any, then validates that against res.
+func unmarshalAndValidate(data []byte, res *jsonschema.Resolved) error {
```
findleyr, line 85: Don't you need to ApplyDefaults also? Not necessarily for this CL.

findleyr (approved):

examples/server/toolschemas/main.go
jba: That wasn't in the original, so I didn't add it. 
But yes, that seems right. Will do a followup (possibly tomorrow).

PR #514, opened 2025-09-22
mcp: fix context propagation in StreamableClientTransport

### LINKED ISSUE #513: Context propagation is broken in StreamableClientTransport
# 🐛 Describe the Bug
The StreamableClientTransport breaks Go's standard context propagation pattern. Inside its `Connect` method, it creates a new, detached context with `context.Background()` instead of deriving from the parent context provided by the caller.

This oversight prevents request-scoped values—such as authentication tokens, tracing IDs, and other middleware-injected data—from being propagated to background HTTP operations managed by the transport, specifically the SSE (Server-Sent Events) connection and the final **DELETE** request on close.

# Reproduce the Behavior
1. Create a parent context containing a value:
```go
ctx := context.WithValue(context.Background(), "my-key", "my-value")
```

2. Initialize and connect the client using this context:
```go
client.Connect(ctx, streamableTransport, nil)
```

3. In a middleware or handler, attempt to access "my-key" from the context within the HTTP requests initiated by `StreamableClientTransport`.
4. Observe that the value is missing because the original context was discarded.

# Root Cause Analysis
The issue stems from three distinct violations of context handling within streamable.go.

## Violation 1: Detached Context Creation

The context chain is broken at the very beginning of the connection process. A new context is created from context.Background(), discarding the parent ctx and all its associated values and deadlines.

`streamable.go:1021`
```diff
-- connCtx, cancel := context.WithCancel(context.Background()) // ❌ BREAKS CONTEXT CHAIN

++ connCtx, cancel := context.WithCancel(ctx) // ✅ PRESERVES CONTEXT CHAIN
```

## Violation 2: SSE Requests Use the Detached Context

The establishSSE method correctly uses `http.NewRequestWithContext`, but it receives the detached c.ctx created in Violation #1, which lacks the necessary request-scoped values.

`streamable.go:1422`
```go
// This code is functionally correct, but c.ctx is the wrong context.
req, err := http.NewRequestWithContext(c.ctx, http.MethodGet, c.url, nil)
```

## Violation 3: DELETE Request Ignores Context Entirely

The Close method creates a DELETE request using http.NewRequest instead of its context-aware counterpart, http.NewRequestWithContext. This means it can't be cancelled and carries no scoped data.

`streamable.go:1404`
```diff
-- req, err := http.NewRequest(http.MethodDelete, c.url, nil) // ❌ NO CONTEXT
++ req, err := http.NewRequestWithContext(ctx, http.MethodDelete, c.url, nil) // ✅ USES CONTEXT
```

# Expected Behavior
All background HTTP operations initiated by the client transport should honor the context provided during the Connect call. This ensures that deadlines, cancellation signals, and request-scoped values are propagated correctly, adhering to standard Go patterns.

# 💥 Impact
This bug can silently break critical application functionality that relies on context propagation, including:
- Authentication: Middleware-injected auth tokens won't be sent on background requests.
- Distributed Tracing: Trace and span IDs are lost, breaking request correlation in observability platforms.
- Request Scoping: Any request-specific data (e.g., user IDs, tenant info) becomes unavailable.
- Cancellation: Operations cannot be properly cancelled via the parent context.

# ✅ Proposed Solution
The fix involves correcting the three violations identified above to ensure the parent context is respected throughout the connection's lifecycle.

1. Preserve the parent context in Connect (line 1021):

```go
// streamable.go:1021
connCtx, cancel := context.WithCancel(ctx)
```

2. No change is needed for SSE requests (line 1422), as fixing #1 ensures c.ctx is the correct, derived context.

3. Pass the context to the DELETE request in Close (line 1404). Note: The original ctx from Connect must be stored on the struct to be accessible here.

```go
// streamable.go:1404
req, err := http.NewRequestWithContext(c.ctx, http.MethodDelete, c.url, nil)
```

# Additional Context
This behavior is particularly subtle because standard RPC-style calls on the client work as expected—they accept a context at call time. The bug is confined to the long-lived, client-initiated background operations (GET for SSE and DELETE for cleanup) that incorrectly manage the context provided at initialization.

manuelibar:
## Problem
Context propagation was broken in `StreamableClientTransport` because:
1. `Connect()` used `context.Background()` instead of the parent context
2. `Close()` created a race condition where DELETE requests were cancelled before completion

## Solution  
- Use parent context when creating the connection context in `Connect()`
- Reorder `Close()` operations to perform cleanup DELETE before cancelling context

## Impact
- Request-scoped values (auth tokens, trace IDs) now propagate correctly to background HTTP operations
- Eliminates race condition in cleanup operations  
- Maintains proper Go context semantics

Fixes #513

### DIFF
--- a/mcp/streamable.go
+++ b/mcp/streamable.go
@@ -1018,7 +1018,7 @@ func (t *StreamableClientTransport) Connect(ctx context.Context) (Connection, er
 	// Create a new cancellable context that will manage the connection's lifecycle.
 	// This is crucial for cleanly shutting down the background SSE listener by
 	// cancelling its blocking network operations, which prevents hangs on exit.
-	connCtx, cancel := context.WithCancel(context.Background())
+	connCtx, cancel := context.WithCancel(ctx)
 	conn := &streamableClientConn{
 		url:        t.Endpoint,
 		client:     client,
@@ -1394,14 +1394,10 @@ func (c *streamableClientConn) reconnect(lastEventID string) (*http.Response, er
 // Close implements the [Connection] interface.
 func (c *streamableClientConn) Close() error {
 	c.closeOnce.Do(func() {
-		// Cancel any hanging network requests.
-		c.cancel()
-		close(c.done)
-
 		if errors.Is(c.failure(), errSessionMissing) {
 			// If the session is missing, no need to delete it.
 		} else {
-			req, err := http.NewRequest(http.MethodDelete, c.url, nil)
+			req, err := http.NewRequestWithContext(c.ctx, http.MethodDelete, c.url, nil)
 			if err != nil {
 				c.closeErr = err
 			} else {
@@ -1411,6 +1407,10 @@ func (c *streamableClientConn) Close() error {
 				}
 			}
 		}
+
+		// Cancel any hanging network requests after cleanup.
+		c.cancel()
+		close(c.done)
 	})
 	return c.closeErr
 }


samthanawalla:
Thanks for the PR!

There seems to be a race failure. Could you also add a test to streamable_test.go that demonstrates that the context is now propagated correctly?

manuelibar:
Hi @samthanawalla, thanks for the review. I've updated the PR with the following:

- Feedback Addressed: The changes you requested have been implemented.
- Additional Fix: I also resolved a pre-existing race condition on the testAuth flag to improve test stability.

Ready for another look when you get a chance 👍 .

mcp/streamable_test.go
```diff
@@ -1430,3 +1431,76 @@ func TestStreamableGET(t *testing.T) {
...
+	}
+
+	if err := session.Close(); err != nil {
+		t.Errorf("session.Close() failed: %v", err)
+	}
+
+	mu.Lock()
+	defer mu.Unlock()
+
+	if getCtx != nil && getCtx.Done() == nil {
```
samthanawalla: This test only checks that the context is cancellable but doesn't check if the values are passed. Can we change this test to verify that the value of some testkey is preserved?

manuelibar: I addressed this comment on [this commit ](https://github.com/modelcontextprotocol/go-sdk/pull/514/commits/4475e7f0659f6d829234e8c4eef76692f67b2d7c)

PR #515, opened 2025-09-22
jsonschema: experiment with a proxy package

findleyr:
Use a proxy package to limit the visibility of the jsonschema package and guard against incompatible API changes.

DO NOT REVIEW: experiment
DO NOT SUBMIT


findleyr:
This is not worth doing, and doesn't really provide much.

PR #525, opened 2025-09-24
mcp: decouple HTTPClient concrete implementation from client transports

### LINKED ISSUE #522: Decouple client transports from `*http.Client` by accepting an interface
# Problem
Currently, the `StreamableClientTransport` and `SSEClientTransport` requires a concrete `*http.Client`, which restricts its integration into systems with established HTTP client abstractions. This tight coupling forces users to implement workarounds when integrating the MCP client into environments with standardized middleware for concerns like authentication, tracing, and metrics.

As seen in issues #367 and #373, users are already creating custom `RoundTripper` implementations to inject headers and other middleware. This becomes particularly cumbersome in enterprise settings where HTTP clients are expected to handle numerous cross-cutting concerns, such as:

- **Authentication**: OAuth2, API keys, mTLS, and other custom schemes.
- **Observability**: Distributed tracing (OpenTelemetry), metrics collection (Prometheus), and structured logging.
- **Resilience**: Circuit breakers, retry logic, and custom timeout policies.
- **Security & Compliance**: Request/response auditing and rate limiting.

Forcing the use of a concrete `*http.Client` makes it difficult to reuse existing, instrumented clients, leading to duplicated logic or complex wrapper patterns around the MCP client.

# Solution
To enhance flexibility and composability, `StreamableClientTransport` should accept an interface that satisfies the Do method, rather than a concrete `*http.Client`.

A standard HTTPClient interface could be defined as:

```go
type HTTPClient interface {
    Do(req *http.Request) (*http.Response, error)
}
```

This change would be fully backward-compatible, as `*http.Client` already implements this method signature. Users could continue to initialize the transport as before:

```go
// Existing usage remains unchanged and valid
&mcp.StreamableClientTransport{
    Endpoint:   endpoint,
    HTTPClient: &http.Client{},
}
```

However, it would also empower users to inject their own custom or instrumented HTTP client implementations:

```go
// Enables integration with custom, standardized clients
&mcp.StreamableClientTransport{
    Endpoint:   endpoint,
    HTTPClient: myOrg.NewInstrumentedHTTPClient(), // Custom implementation
}
```

This approach aligns with Go's design principle of "accept interfaces, return structs," promoting loose coupling and greater flexibility.

# Alternatives
1. **Custom `http.RoundTripper`**: While this is the current workaround (as demonstrated in #367), it is suboptimal. Middleware implemented at the transport layer is less flexible and harder to compose than middleware applied to a client abstraction.
2. **Wrapper Pattern**: Creating a wrapper around the entire MCP client introduces unnecessary complexity, adds a layer of indirection, and may have performance implications.

The proposed interface-based approach is cleaner and more idiomatic than these alternatives.

# Additional context
Adopting an interface for the HTTP client is a standard best practice in the Go ecosystem, used by major libraries to ensure interoperability and extensibility:

- AWS SDK for Go (v2)
- Google Cloud Go libraries
- OpenTelemetry instrumentation libraries
- Popular HTTP clients like Resty and go-retryablehttp

Implementing this change would make the go-sdk more enterprise-friendly and easier to test, as it would allow users to mock the HTTPClient in their unit tests. It would also provide a more robust and idiomatic solution to the use cases described in issues #367 and #373.

manuelibar:
# Description
This change decouples client transports from the concrete `net/http.Client` by accepting any HTTP client that implements `Do(http.Request) (http.Response, error)`. This enables middleware (auth, tracing, retries), improves testability, and unifies behavior across streamable and SSE transports.

# What changed
- Introduce `HTTPClient` interface in the `mcp` package.
- `StreamableClientTransport`: change `HTTPClient` field type to the interface; default to `http.DefaultClient` when nil.
- SSEClientTransport: same change as streamable.

# Why
- Flexibility to inject custom HTTP clients with middleware.
- Better testability via interface-based dependencies.
- Consistency across transports (streamable and SSE).

# Compatibility
- Backward compatible: existing code passing `http.Client` continues to work (it satisfies the interface).
- No behavior changes; defaults remain the same when HTTPClient is nil.

# Testing
- `go test ./...` passes locally.
- Existing transport tests validate streamable/SSE flows.

Fixes #522

### DIFF
--- a/mcp/shared.go
+++ b/mcp/shared.go
@@ -541,3 +541,11 @@ func startKeepalive(session keepaliveSession, interval time.Duration, cancelPtr 
 		}
 	}()
 }
+
+// HTTPClient is the minimal interface required by client transports for issuing HTTP requests.
+//
+// It matches the method set of (*http.Client).Do, allowing callers to provide custom HTTP clients
+// that incorporate middleware (auth, tracing, retries), or test doubles.
+type HTTPClient interface {
+	Do(*http.Request) (*http.Response, error)
+}
--- a/mcp/sse.go
+++ b/mcp/sse.go
@@ -326,9 +326,9 @@ type SSEClientTransport struct {
 	// Endpoint is the SSE endpoint to connect to.
 	Endpoint string
 
-	// HTTPClient is the client to use for making HTTP requests. If nil,
-	// http.DefaultClient is used.
-	HTTPClient *http.Client
+	// HTTPClient performs HTTP requests. If nil, http.DefaultClient is used.
+	// Any type implementing a Do(*http.Request) (*http.Response, error) method is supported.
+	HTTPClient HTTPClient
 }
 
 // Connect connects through the client endpoint.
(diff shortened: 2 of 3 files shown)


manuelibar:
@findleyr I'll close this PR given that the proposal was declined.
Thanks!

PR #539, opened 2025-09-29
internal/oauthex: address URL attacks in PRM

jba:
Fix Protected Resource Metadata to prevent URL attacks, as described in the issue.

For #526

### DIFF
--- a/internal/oauthex/oauth2.go
+++ b/internal/oauthex/oauth2.go
@@ -65,3 +65,21 @@ func getJSON[T any](ctx context.Context, c *http.Client, url string, limit int64
 	}
 	return &t, nil
 }
+
+// checkURLScheme ensures that its argument is a valid URL with a scheme
+// that prevents XSS attacks.
+// See #526.
+func checkURLScheme(u string) error {
+	if u == "" {
+		return nil
+	}
+	uu, err := url.Parse(u)
+	if err != nil {
+		return err
+	}
+	scheme := strings.ToLower(uu.Scheme)
+	if scheme == "javascript" || scheme == "data" || scheme == "vbscript" {
+		return fmt.Errorf("URL has disallowed scheme %q", scheme)
+	}
+	return nil
+}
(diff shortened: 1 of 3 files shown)


samthanawalla (approved):

findleyr (commented):
This LGTM (though I have only superficial knowledge of the problem).

However, changes like this should include a test.

internal/oauthex/oauth2.go
```diff
@@ -65,3 +65,21 @@ func getJSON[T any](ctx context.Context, c *http.Client, url string, limit int64
...
+func checkURLScheme(u string) error {
+	if u == "" {
+		return nil
+	}
+	uu, err := url.Parse(u)
+	if err != nil {
+		return err
+	}
+	scheme := strings.ToLower(uu.Scheme)
+	if scheme == "javascript" || scheme == "data" || scheme == "vbscript" {
```
rolandshoemaker, line 81: My guess is that we can't just blanket block everything except http(s), because it's plausible that some servers may want to authenticate using a custom app protocol, but that is a little worrying, since it's not immediately clear to me that those couldn't similarly be used for attacks.

I see this is what other SDKs have done though, so perhaps this is the best option we have. There should at least be a comment here explaining why we block these particular schemes, and why we cannot just only allow http(s).

findleyr:
Can we land this today? Would be good to include in the next release (planned soon).

internal/oauthex/oauth2.go
jba: I got the list of schemes from https://github.com/modelcontextprotocol/typescript-sdk/pull/877/files#diff-a0fe23c86bd448c6fa4c60fe4d380daa65451fea86a9e760673a5a7c65292e88, line 20. There's no explanation in that PR for where that list comes from. Clearly, they are all suspect, but I don't know if there are other suspect schemes as well. However, the article you cited, https://verialabs.com/blog/from-mcp-to-shell, seems to think that the above PR is sufficient. (See the section called the most impactful fix.}

rolandshoemaker: Seems fine with me. I think we should document that somewhere though.

IAmSurajBobade:
@jba I have added tests for your changes in your fork. If you have already added please ignore.

https://github.com/jba/go-sdk/pull/5

jba:
I merged this, but the comments should be addressed in a follow up. @samthanawalla

Issue #19, opened 2025-06-24
OAuth Support


findleyr:
Add the hooks and utilities necessary to implement the OAuth aspects of the spec, including the OAuth resource server classification and client-side resource indicators added in the 6/18 version of the spec.

See https://github.com/modelcontextprotocol/go-sdk/issues/19#issuecomment-3352959772

gannaramu:
is there an existing spec discussion on how we plan to implement this? would be happy to contribute if not already work in progress.

jba:
I have some work in progress. I should write a doc. Thanks for offering to help.
I will probably get to this next week (Thursday and Friday are holidays).

RoyceLeonD:
@jba Quick check on oauth since you're already working on this.

I see you already have a PR on the .well-known endpoints. As part of the OAuth 2.1 spec implementation, I'd like to suggest a few considerations for the SDK:

**Token Validation**
- First-class support for validating tokens via JWKs (which you may already be considering)
- A custom validator interface, since tokens from external auth servers aren't always JWTs - they could be JWEs, custom strings/UUIDs, or other formats
- The custom validator pattern would also enable validation of custom headers beyond just bearer, technically not in scope, but could fall under the Authorization component if people implement non-standard secondary headers such as csrf or transactionID we should make it easy for them.

**Client vs Server Flows**
Since this SDK supports both client and server modes, it would be valuable to consider:
- **As a client**: Bearer auth injection when making requests to downstream servers
- **As a server**: Passthrough OAuth 2.1 detection when go-sdk is a proxy for upstream requests

Let me know if this is scope creep for the current work - totally fine to push back. Otherwise, I'd be happy to take a crack at proposing some of these features.

jba:
I owe you a fuller response to this and on the other issue, but the idea is to use [golang.org/x/oauth2](https://pkg.go.dev/golang.org/x/oauth2), which hopefully has everything you're asking for. What's missing are the three extensions—Protected Resource Metadata, Auth Server Metadata, and Dynamic Client Registration—which we'll implement here.

We've mostly been thinking of it as client authorizing to MCP server, not passthrough. I think we should stick to what the [MCP spec asks for](https://modelcontextprotocol.io/specification/2025-06-18/basic/authorization) first, before we consider any extensions. That is already a fair bit of work.

[...]

PR #544, opened 2025-09-29
mcp: add client-side OAuth flow

samthanawalla:
This is a preliminary implementation of OAuth 2.1 for the client.

When a StreamableClientTransport encounters a 401 Unauthorized response
from the server, it initiates the OAuth flow described in thec
authorization section of the MCP spec
(https://modelcontextprotocol.io/specification/2025-06-18/basic/authorization).
On success, the transport obtains an access token which it passes
to all subsequent requests.

Much remains to be done here:

* Resource Indicators, as described in section 2.5.1 of the MCP spec.

* All of this is unexported, so it is available only to our own
StreamingClientTransport. We should add API so people can use it
with their own transports.

* And, of course, tests. We should test against fake implementations
but also, if we can find any, real reference implementations.

For #19
Co-authored-by: Jonathan Amsterdam <jba@google.com>


auth/client.go
```diff
@@ -0,0 +1,107 @@
...
+
+	resp.Body.Close()
+	// Try to authorize.
+	t.mu.Lock()
+	defer t.mu.Unlock()
+	// If we don't have a token source, get one by following the OAuth flow.
+	// (We may have obtained one while t.mu was not held above.)
+	if _, ok := t.opts.Base.(*oauth2.Transport); !ok {
+		authHeaders := resp.Header[http.CanonicalHeaderKey("WWW-Authenticate")]
+		ts, err := t.handler(req.Context(), OAuthHandlerArgs{
```
jba, line 92: it's unfortunate that we hold the lock for the entire OAuth flow. That's a long time to hold a lock. It's worth adding a comment here so someday we can think of a better way.

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/modelcontextprotocol/go-sdk
 
-go 1.23.0
+go 1.24.0
```
jba: This chunk should probably not be in this PR.

jba (changes requested):

auth/client.go
samthanawalla: Done

go.mod
samthanawalla: Done

jba (dismissed):

jba:
Let's not submit until after the release.

jba (approved):

auth/client.go
```diff
@@ -0,0 +1,109 @@
...
+			ResourceMetadataURL: extractResourceMetadataURL(authHeaders),
+		})
+		if err != nil {
+			return nil, err
+		}
+		t.opts.Base = &oauth2.Transport{Base: t.opts.Base, Source: ts}
+	}
+	return t.opts.Base.RoundTrip(req.Clone(req.Context()))
```
rumpl, line 100: In cagent we use the same trick for OAuth. I just spent the morning debugging and fixing our code for streamable mcp servers and thought I'd tell you what was wrong, might be useful for you.

This second request that comes after the oauth flow has a half-broken request, the body of the request was already read by the first call on line 70, so you would get a 400 Bad Request back from the server. You need to hold on the original body and put it back here before sending a second RoundTrip.

https://github.com/docker/cagent/pull/547

PR #546, opened 2025-09-30
mcp: add ModifyRequest to HTTP client transports

### LINKED ISSUE #533: Proposal: add `ModifyRequest func(*http.Request)` to HTTP client transports
Following up on #522, where we discussed making it easier to modify HTTP client transport requests, for example by adding headers, we propose the following two new options:

```go
type StreamableClientTransport struct {
  ...
  // If set, ModifyRequest is called before each outgoing HTTP request made by the client
  // connection. It can be use to, for example, add headers to outgoing requests.
  ModifyRequest func(*http.Request)
}

type SSEClientTransport struct {
  ...
  // <ditto>
  ModifyRequest func(*http.Request)
}
```

We believe these options will make it easier to do things like add auth headers. Currently, this can be achieved using the `HTTPClient` field, but only in the roundtripper, which requires cloning the request.

manuelibar:
## Problem
Users need to modify HTTP requests (add headers, authentication, etc.) but currently must implement custom `RoundTripper` implementations which requires cloning requests and is more complex than necessary.

## Solution  
Add `ModifyRequest func(*http.Request)` field to both `SSEClientTransport` and `StreamableClientTransport`. This callback is invoked before each outgoing HTTP request, providing a simple way to modify requests without needing custom `RoundTripper` implementations.

## Changes
- Add `ModifyRequest` field to `SSEClientTransport` (called before GET and POST requests)
- Add `ModifyRequest` field to `StreamableClientTransport` (called before GET, POST, and DELETE requests)
- Add comprehensive tests covering all request types
- Add example usage in documentation

## Impact
- Simpler API for common request modification use cases
- No breaking changes (new optional field)
- Backward compatible (nil ModifyRequest is safe)

Fixes #533


jba (approved):

manuelibar:
@jba please check again 🙏 , it appeared to have some merge conflicts prior to merging

findleyr:
@manuelibar I think we'll discuss this in the proposal meeting tomorrow and likely approve the proposal (unless anyone objects), so we may as well wait to merge until then.

fgrosse:
Should we wait for the outcome of https://github.com/golang/go/issues/75814 before merging this?

findleyr:
Yes, I think we're waiting clarity on the end-to-end client oauth flow, before we commit to more API. See also #591.

PR #559, opened 2025-10-06
oauthex: move internal/oauthex to oauthex

### LINKED ISSUE #558: make oauthex visible
Move the contents of internal/oauthex into oauthex, gated by the experiment tag mcp_go_client_oauth.

samthanawalla:
This PR relocates oauthex and moves DCR logic into dcr.go.

Fixes #558


jba:
It's important that the symbol oauthex/ProtectedResourceMetadata still exist, even without the experiment tag. I can't tell if that's true from this PR.

samthanawalla:
Ah good point. Updated

PR #562, opened 2025-10-07
mcp: don't close stdout when shutting down the stdio transport

### LINKED ISSUE #548: Using `mcp.StdioTransport{}` breaks code coverage reporting during tests
**Describe the bug**

Using `mcp.StdioTransport{}` for the server transport protocol causes `os.Stout` to be closed when the server is stopped which prevents go test code coverage from being reported:

```bash
$ go test -coverprofile cover.out ./...
error generating coverage report: write /dev/stdout: file already closed
```

**To Reproduce**

Create the following main and test files:

```go
package main

import (
	"context"
	"time"

)

func main() {
	RunServer()
}

func RunServer() {
	impl := &mcp.Implementation{}
	options := &mcp.ServerOptions{}
	server := mcp.NewServer(impl, options)

	ctx, cancel := context.WithTimeout(context.Background(), 100*time.Millisecond)
	defer cancel()

	server.Run(ctx, &mcp.InMemoryTransport{})
}
```

```go
package main

import "testing"

func Test_RunServer(t *testing.T) {
	RunServer()
}
```

Then run the following to test the code:
```bash 
$ go test -coverprofile cover.out ./...
```

**Expected behavior**

Coverage should be reported and the tests shouldn't fail

```bash
$ go test -coverprofile cover.out ./...
ok      mcp-test        0.114s  coverage: 87.5% of statements
```

findleyr:
We need to close stdin when the stdio transport is shut down, to unblock its read loop. However, we don't technically need to close stdout.

Fixes #548

### DIFF
--- a/mcp/transport.go
+++ b/mcp/transport.go
@@ -115,8 +115,8 @@ func (t *IOTransport) Connect(context.Context) (Connection, error) {
 // An InMemoryTransport is a [Transport] that communicates over an in-memory
 // network connection, using newline-delimited JSON.
 //
-// InMemoryTransports should be connected using [NewInMemoryTranports], which
-// returns two transports connected to each other.
+// InMemoryTransports should be constructed using [NewInMemoryTransports],
+// which returns two transports connected to each other.
 type InMemoryTransport struct {
 	rwc io.ReadWriteCloser
 }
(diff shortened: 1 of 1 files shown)


samthanawalla (dismissed):

findleyr:
Sorry @samthanawalla, I had a couple typos which I've now fixed. Would you mind re-reviewing?

Issue #473, opened 2025-09-16
clients should check PKCE support


jba:
The client auth implementation for clients should check that the auth server implements PKCE, as described in https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/docs/specification/draft/basic/authorization.mdx?plain=1#L350-L356.

RyoKusnadi:
Hi @jba, for this issue meaning that in auth_meta, we need to use **GetAuthServerMeta** and check using CodeChallengeMethodsSupported right? if yes, can i assign this task to me, willing to raise the PR?

RyoKusnadi:
Hi @jba, allow me to double confirming
1. In auth_meta.go, Use **GetAuthServerMeta** and check using CodeChallengeMethodsSupported
2. provide example in client_example_test.go

let me know if i missed something.
if already no issue, can assign this task to me? willing to raise the PR

jba:
Yes, CodeChallengeMethodsSupported must be a non-empty slice.

maciej-kisiel:
This looks done (PR was merged).

PR #563, opened 2025-10-07
Add Check PCKE Support

RyoKusnadi:
Refer To [Issue 473](https://github.com/modelcontextprotocol/go-sdk/issues/473)
Changes:
- Add check for PKCE support via server metadata and return an error if not supported.

### DIFF
--- a/oauthex/auth_meta_test.go
+++ b/oauthex/auth_meta_test.go
@@ -39,19 +39,16 @@ func TestGetAuthServerMetaPKCESupport(t *testing.T) {
 	tests := []struct {
 		name           string
 		hasPKCESupport bool
-		expectError    bool
-		expectedError  string
+		wantError      string
 	}{
 		{
 			name:           "server_with_pkce_support",
 			hasPKCESupport: true,
-			expectError:    false,
 		},
 		{
 			name:           "server_without_pkce_support",
 			hasPKCESupport: false,
-			expectError:    true,
-			expectedError:  "does not implement PKCE",
+			wantError:      "does not implement PKCE",
 		},
 	}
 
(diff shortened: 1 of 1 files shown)


oauthex/oauthex.go
```diff
@@ -5,88 +5,38 @@
...
+}
 
-	// DPOPSigningAlgValuesSupported (dpop_signing_alg_values_supported) is an optional
-	// slice of JWS signing algorithms supported by the resource server for validating
-	// DPoP proof JWTs (RFC 9449).
-	DPOPSigningAlgValuesSupported []string `json:"dpop_signing_alg_values_supported,omitempty"`
+// RequirePKCE checks that the authorization server for issuerURL supports PKCE,
+// by verifying that CodeChallengeMethodsSupported is non-empty.
+// It returns an error if metadata cannot be fetched or PKCE is not advertised.
+func RequirePKCE(ctx context.Context, issuerURL string, c *http.Client) error {
```
samthanawalla: move this to auth_meta after rebasing

oauthex/oauthex.go
```diff
@@ -5,88 +5,38 @@
...
+	asm, err := oauthex.GetAuthServerMeta(ctx, issuerURL, c)
+	if err != nil {
+		return err
+	}
 
-	// DPOPBoundAccessTokensRequired (dpop_bound_access_tokens_required) is an optional boolean
-	// specifying whether the protected resource always requires the use of DPoP-bound
-	// access tokens (RFC 9449). Defaults to false if omitted.
-	DPOPBoundAccessTokensRequired bool `json:"dpop_bound_access_tokens_required,omitempty"`
+	if len(asm.CodeChallengeMethodsSupported) == 0 {
```
samthanawalla: Don't we also have to check code_verifier? https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1-13#section-7.5.2

jba: The draft spec says that only code challenge methods needs to be checked:
https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/docs/specification/draft/basic/authorization.md, the section entitled Authorization Code Protection.

samthanawalla: Ah got it, thanks for the clarification

RyoKusnadi:
Hi @samthanawalla updated, kindly help to review again. thanks!

oauthex/auth_meta.go
```diff
@@ -145,3 +145,19 @@ func GetAuthServerMeta(ctx context.Context, issuerURL string, c *http.Client) (*
 	}
 	return nil, fmt.Errorf("failed to get auth server metadata from %q: %w", issuerURL, errors.Join(errs...))
 }
+
+// RequirePKCE checks that the authorization server for issuerURL supports PKCE,
```
jba: I don't think we should have an exported function for this. Why can't we do it inside GetAuthServerMeta?

oauthex/oauthex.go
RyoKusnadi: revised

oauthex/auth_meta.go
RyoKusnadi: revised

oauthex/auth_meta_test.go
```diff
@@ -28,3 +32,33 @@ func TestAuthMetaParse(t *testing.T) {
 		t.Errorf("got %q, want %q", g, w)
 	}
 }
+
+func TestGetAuthServerMetaRequirePKCE(t *testing.T) {
```
jba: This doesn't actually test that we do the check. There needs to be a test for an auth server that doesn't do PKCE.

RyoKusnadi: revised

oauthex/auth_meta_test.go
```diff
@@ -33,32 +34,90 @@ func TestAuthMetaParse(t *testing.T) {
 	}
 }
 
-func TestGetAuthServerMetaRequirePKCE(t *testing.T) {
+func TestGetAuthServerMetaPKCESupport(t *testing.T) {
 	ctx := context.Background()
+	tests := []struct {
+		name           string
+		hasPKCESupport bool
+		expectError    bool
```
jba: s/expect/want/

jba: Also wantError can be a string, if empty, then no error is expected, otherwise the error should contain that string.

RyoKusnadi: revised

PR #564, opened 2025-10-07
oauthex: update fake oauth server

samthanawalla:
This PR adds /mcp, /.well-known/oauth-protected-resource, and /register to the fake server setup. This will be useful for future tests.

For: #19

### DIFF
--- a/oauthex/oauthex_test.go
+++ b/oauthex/oauthex_test.go
@@ -118,7 +118,6 @@ func (s *state) handleServerMetadata(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	w.Header().Set("Content-Type", "application/json")
-	w.WriteHeader(http.StatusOK)
 	w.Write(buf.Bytes())
 }
 
@@ -210,7 +209,6 @@ func (s *state) handleToken(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	w.Header().Set("Content-Type", "application/json")
-	w.WriteHeader(http.StatusOK)
 	w.Write(buf.Bytes())
 }
 


oauthex/oauthex_test.go
```diff
@@ -0,0 +1,236 @@
...
+	"net/http"
+	"strings"
+	"time"
+
+	"github.com/golang-jwt/jwt/v5"
+)
+
+const (
+	authServerPort = ":8080"
+	issuer         = "http://localhost" + authServerPort
```
findleyr: I feel like we should probably close over this information.

Does the server have to run on :8080 in order to function? If so, that's a problem, because two instances of this server can't run concurrently. Better to use the httptest server URL, if possible.

oauthex/oauthex_test.go
```diff
@@ -0,0 +1,236 @@
...
+}
+
+// handleMcp is the protected resource endpoint. It requires a valid Bearer token.
+// If the token is missing or invalid, it returns a 401 Unauthorized response
+// with a WWW-Authenticate header pointing to the resource metadata.
+func (s *state) handleMcp(w http.ResponseWriter, r *http.Request) {
+	authHeader := r.Header.Get("Authorization")
+	tokenString := strings.TrimPrefix(authHeader, "Bearer ")
+
+	token, err := jwt.Parse(tokenString, func(token *jwt.Token) (interface{}, error) {
```
findleyr: any

oauthex/oauthex_test.go
```diff
@@ -0,0 +1,236 @@
...
+	mux.HandleFunc("/register", s.handleDynamicClientRegistration)
+	mux.HandleFunc("/authorize", s.handleAuthorize)
+	mux.HandleFunc("/token", s.handleToken)
+	return mux
+}
+
+// handleMcp is the protected resource endpoint. It requires a valid Bearer token.
+// If the token is missing or invalid, it returns a 401 Unauthorized response
+// with a WWW-Authenticate header pointing to the resource metadata.
+func (s *state) handleMcp(w http.ResponseWriter, r *http.Request) {
```
findleyr: Nit: handleMCP

findleyr (commented):
Thanks. @jba it would be good if you could review, as otherwise I'd have to do a lot of research to verify the behavior here.

oauthex/oauthex_test.go
```diff
@@ -0,0 +1,236 @@
...
+	"fmt"
+	"net/http"
+	"strings"
+	"time"
+
+	"github.com/golang-jwt/jwt/v5"
+)
+
+const (
+	authServerPort = ":8080"
```
jba: This only shows up in the issue or field of the JWT. Maybe it's not essential, since this is a test server and I think we can validate the JWT however we want.

oauthex/oauthex_test.go
```diff
@@ -0,0 +1,236 @@
+// Copyright 2025 The Go MCP SDK Authors. All rights reserved.
+// Use of this source code is governed by an MIT-style
+// license that can be found in the LICENSE file.
+
+//go:build mcp_go_client_oauth
+
+package oauthex
```
jba, line 7: By moving the fake auth server into the test package of oauthex, it's no longer possible to use it in tests from the mcp package. We can leave it here for now, but let's be prepared to move it back into internal/tests.

oauthex/oauthex_test.go
```diff
@@ -0,0 +1,236 @@
...
+	redirectURI   string
+}
+
+type state struct {
+	authCodes map[string]authCodeInfo
+}
+
+// NewFakeAuthMux constructs a ServeMux that implements an OAuth 2.1 authentication
+// server. It should be used with [httptest.NewTLSServer].
+func NewFakeAuthMux() *http.ServeMux {
```
jba: If we're going to combine all the endpoints into one ServeMux, then it's no longer simply an authentication server, so let's use a different name. My original thought was to separate the authentication server from the MCP server, but maybe this is fine.

oauthex/oauthex_test.go
```diff
@@ -0,0 +1,236 @@
...
+
+	response := ClientRegistrationResponse{
+		ClientRegistrationMetadata: clientRegistrationMetadata,
+		ClientID:                   clientID,
+		ClientIDIssuedAt:           time.Now(),
+		ClientSecret:               "fake-registration-access-token",
+	}
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusCreated)
+	json.NewEncoder(w).Encode(response)
```
jba: As a general rule, in code to a bytes.Buffer, check for an error, then write to w.

oauthex/oauthex_test.go
```diff
@@ -0,0 +1,236 @@
...
+
+// handleDynamicClientRegistration handles dynamic client registration requests,
+// as defined in RFC 7591.
+func (s *state) handleDynamicClientRegistration(w http.ResponseWriter, r *http.Request) {
+	if r.Method != http.MethodPost {
+		http.Error(w, "Method Not Allowed", http.StatusMethodNotAllowed)
+		return
+	}
+
+	var clientRegistrationMetadata ClientRegistrationMetadata
```
jba: crm as a variable name

oauthex/oauthex_test.go
```diff
@@ -0,0 +1,236 @@
...
+		ClientID:                   clientID,
+		ClientIDIssuedAt:           time.Now(),
+		ClientSecret:               "fake-registration-access-token",
+	}
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusCreated)
+	json.NewEncoder(w).Encode(response)
+}
+
+func (s *state) handleToken(w http.ResponseWriter, r *http.Request) {
```
jba, line 155: I assume this is copied from the original code?

oauthex/oauthex_test.go
```diff
@@ -0,0 +1,236 @@
...
+	tokenResponse := map[string]any{
+		"access_token": accessToken,
+		"token_type":   "Bearer",
+		"expires_in":   int(tokenExpiry.Seconds()),
+	}
+	w.Header().Set("Content-Type", "application/json")
+	json.NewEncoder(w).Encode(tokenResponse)
+}
+
+func (s *state) handleAuthorize(w http.ResponseWriter, r *http.Request) {
```
jba, line 215: Also copied?

oauthex/oauthex_test.go
samthanawalla: Ack

oauthex/oauthex_test.go
samthanawalla: Done. Renamed to NewFakeMCPServerMux

oauthex/oauthex_test.go
samthanawalla: Done

oauthex/oauthex_test.go
samthanawalla: Done

oauthex/oauthex_test.go
samthanawalla: Done

oauthex/oauthex_test.go
samthanawalla: Done

oauthex/oauthex_test.go
samthanawalla: Yes. The diff did not catch that for some reason.

oauthex/oauthex_test.go
samthanawalla: Yes

oauthex/oauthex_test.go
samthanawalla: Removed

oauthex/oauthex_test.go
samthanawalla: Makes sense, updated to use the server url

oauthex/oauthex_test.go
```diff
@@ -117,8 +112,14 @@ func (s *state) handleServerMetadata(w http.ResponseWriter, r *http.Request) {
...
 		CodeChallengeMethodsSupported:     []string{"S256"},
 	}
+	var buf bytes.Buffer
+	if err := json.NewEncoder(&buf).Encode(metadata); err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
 	w.Header().Set("Content-Type", "application/json")
-	json.NewEncoder(w).Encode(metadata)
+	w.WriteHeader(http.StatusOK)
```
jba: you should be able to omit this line.

oauthex/oauthex_test.go
```diff
@@ -197,8 +204,14 @@ func (s *state) handleToken(w http.ResponseWriter, r *http.Request) {
...
 		"expires_in":   int(tokenExpiry.Seconds()),
 	}
+	var buf bytes.Buffer
+	if err := json.NewEncoder(&buf).Encode(tokenResponse); err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
 	w.Header().Set("Content-Type", "application/json")
-	json.NewEncoder(w).Encode(tokenResponse)
+	w.WriteHeader(http.StatusOK)
```
jba: Ditto.

samthanawalla: Done

oauthex/oauthex_test.go
samthanawalla: Done

PR #569, opened 2025-10-10
feat: add ModifyRequest hook to client transports

### LINKED ISSUE #533: Proposal: add `ModifyRequest func(*http.Request)` to HTTP client transports

n33levo:
## Summary
- add optional ModifyRequest callback to StreamableClientTransport and SSEClientTransport so callers can inject auth headers
- cover new hooks with transport tests to ensure requests carry custom headers across POST/GET/DELETE

## Testing
- go test ./...

Fixes #533

### DIFF
--- a/mcp/sse.go
+++ b/mcp/sse.go
@@ -329,6 +329,10 @@ type SSEClientTransport struct {
 	// HTTPClient is the client to use for making HTTP requests. If nil,
 	// http.DefaultClient is used.
 	HTTPClient *http.Client
+	// ModifyRequest, if non-nil, is called before each outbound HTTP request.
+	// It can be used to set headers (for example, auth headers) prior to sending
+	// the request.
+	ModifyRequest func(*http.Request)
 }
 
 // Connect connects through the client endpoint.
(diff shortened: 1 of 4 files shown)


manuelibar:
@n33levo I think this PR duplicates [this one](https://github.com/modelcontextprotocol/go-sdk/pull/546)

jba:
@n33levo does this add anything new to? #546? If so, what? If not, please close.

n33levo:
I implemented the ModifyRequest feature in PR #[569](https://github.com/modelcontextprotocol/go-sdk/pull/569), but I see @manuelibar has a more comprehensive implementation in PR #[546](https://github.com/modelcontextprotocol/go-sdk/pull/546) that covers the same functionality. I'm closing my PR to avoid duplication. The feature works great and addresses the original request perfectly!

PR #570, opened 2025-10-11
Implement SEP-973

### LINKED ISSUE #552: SEP-973: Icons and metadata support
### Description
This issue tracks the implementation of https://github.com/modelcontextprotocol/modelcontextprotocol/pull/955 which adds support for icons and additional metadata for Implementations, Resources, Tools, and Prompts.

Can we add this issue to project: https://github.com/orgs/modelcontextprotocol/projects/16?

IAmSurajBobade:
mcp/protocol: Implement [SEP-973](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/955) specification

- Define Icon structure, which includes source, mimeType and sizes.
- sizes is any array of strings. refer https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1531  
- Support setting websiteUrl, icons for mcp.Implementation
- Support setting icons for mcp.Prompt
- Support setting icons for mcp.Tool
- Support setting icons for mcp.Resource

Fixes #552

### DIFF
--- a/mcp/protocol.go
+++ b/mcp/protocol.go
@@ -660,7 +660,7 @@ func (*ProgressNotificationParams) isParams() {}
 
 // Icon provides visual identifiers for their resources, tools, prompts, and implementations
 // See [/specification/draft/basic/index#icons] for notes on icons
-
+//
 // TODO(iamsurajbobade): update specification url from draft.
 type Icon struct {
 	// Source is A URI pointing to the icon resource (required). This can be:
--- a/mcp/testdata/conformance/server/spec-sep-973-additional-metadata.txtar
+++ b/mcp/testdata/conformance/server/spec-sep-973-additional-metadata.txtar
@@ -3,7 +3,6 @@ Check behavior of server with Icons and websiteUrl metadata added as part of SEP
 check modelcontextprotocol/go-sdk/issues/552 for more details.
 
 Checks following:
-- If client sends protocolVersion as "draft", server responds with same
 - Test setting websiteUrl, icons for mcp.Implementation
 - Test setting icons for mcp.Prompt
 - Test setting icons for mcp.Tool


IAmSurajBobade:
@findleyr, please note - this changes currently do not work with MCP Inspector 0.17.0, as https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1531 is not implemented yet. I have created issue https://github.com/modelcontextprotocol/inspector/issues/861 to track this.

findleyr:
> @findleyr, please note - this changes currently do not work with MCP Inspector 0.17.0, as [modelcontextprotocol/modelcontextprotocol#1531](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1531) is not implemented yet. I have created issue [modelcontextprotocol/inspector#861](https://github.com/modelcontextprotocol/inspector/issues/861) to track this.

Ah, this is a really important point. We need to not return this data based on the client's protocol version!
This definitely warrants a test!

IAmSurajBobade:
> Thanks for this change!
> 
> Can you add a basic smoke test (perhaps just adding to TestEndToEnd) that verifies the client receives the new metadata correctly?
> 
> (I'm sure it works, but good to have at least one end-to-end test).

Yes, was about to ask, if we are good with this changes. I will add tests accordingly.

IAmSurajBobade:
I have updated E2E tests. And updated streamable tests - which were using shared object.


> We need to not return this data based on the client's protocol version!

As of now, Inspector integration fails, as specification draft has changed after original changes were incorporated in inspector. I am not sure, if this would still be a braking change for clients implementing '2025-06-18' or earlier protocol.

@findleyr  if we have to omit this info from server, could you please suggest, how we should go about it? Considering we may have to implement similar checks for future specs as well.

mcp/streamable_test.go
```diff
@@ -1192,7 +1192,7 @@ func TestStreamableStateless(t *testing.T) {
 			Tools:   &ToolCapabilities{ListChanged: true},
 		},
 		ProtocolVersion: latestProtocolVersion,
-		ServerInfo:      &Implementation{Name: "test", Version: "v1.0.0"},
+		ServerInfo:      testImpl,
```
findleyr: I don't think we need to change this.

Can you instead update conformance tests as follows:

- Update the default server in `runServerTest` to have icons and website url.
- Verify that when the client is on an older protocol version, these fields are not sent (existing conformance tests should verify that).
- Update the logic to accept the protocol version "draft", for testing purposes. See `negotiatedVersion`. In other words: if the client sends protocol version "draft", we enable draft features. I think this is both necessary and useful for testing.
- Add a new conformance test that sends protocol version "draft", and verify that it gets icons and website url.

IAmSurajBobade:
https://github.com/modelcontextprotocol/inspector/issues/861 seems to be resolved now. I have tested server with multiple versions (0.16.0, 0.15.0, 0.17.1) of inspector - excluding 0.17.0 - this changes work fine without any errors.

> Verify that when the client is on an older protocol version, these fields are not sent (existing conformance tests should verify that).

@findleyr , I actually wanted your input on this. For example, how do we handle older/newer protocol specific behavior throughout the SDK. For instance, if we were to determine client protocol is older, we have to do string compare, will this be a problem, as we need to do this for each initialize and list call. Or should we change the way we are listing protocols from string to may be a costomType int, so the comparison will be straightforward. Additionally, I analyzed the code for other changes required - and it will have considerable impact. For example, currently listTools, listPrompts, listResources methods, do not have direct access to protocol version information.

For now, to avoid updating existing conformance tests - I have added conditional logic to only add icon and websiteUrl info, if test name is version-draft.txtar.

findleyr:
@IAmSurajBobade you are right, this may be trickier to implement than first thought--I didn't account for the fact that we'd need to suppress these fields from all server features.

Given that clients should ignore extra fields, perhaps it is OK to do nothing. Can you look into how this is done in other SDKs? Do they have special handling for the protocol version? If not, then this CL becomes much simpler.

IAmSurajBobade:
(Just realised, i did not mention this earlier. 🥲) I have checked CSharp, Python, Typescript, Rust SDKs - all of them have implemented draft spec, and there is no special handling basis client version.

Refer https://github.com/orgs/modelcontextprotocol/projects/16/views/1

findleyr:
Ack, thanks @IAmSurajBobade, that's useful. Then I think we can keep it simple. Reviewing now.

mcp/shared.go
```diff
@@ -33,8 +33,10 @@ const (
 	//
 	// It is the version that the client sends in the initialization request, and
 	// the default version used by the server.
-	latestProtocolVersion   = protocolVersion20250618
-	protocolVersion20250618 = "2025-06-18"
+	latestProtocolVersion = protocolVersion20250618
+
+	protocolVersionDraft    = "draft"      // draft protocol version with experimental features for testing
```
findleyr: I guess we can remove this now, if we're not attempting to gate this functionality on the latest spec.

mcp/conformance_test.go
```diff
@@ -135,19 +135,36 @@ func incTool(_ context.Context, _ *CallToolRequest, args incInput) (*CallToolRes
...
...srs9xkm4ho27atgrG7y+BxrPrtAM8Qr/v37ysXnUdoDPnbaCxVqpTKmFAuRRvdZJuA7t27t1HiCGyxiiGT41u1aqVgzZs3d1uPbf9zwKPpzstly5ZJjaq2Y0m21XJEIbOtyDZWYqvsCxl/zM6ljLR1QB6yCGsZY1hvQ7Y5niuz3GSb3R2ipmDEktrdpQ+2DsiIZKkYjW/evFFMgUcbst2wYUNZXMEmcnT9+vWV9axjFavSemRbB2RCkFyKg7k8LViS7aZNmxo25hjE8nPv7uRQtbsr6/1wYOzYsWZssNjUz1iXhifCsdShm4C2Idvs7jVr1jT6HQG+oHZ3t/V+OLB9+3bZgbyMYIpZBtBvVbFijBlp2QrZTbYJQm77FKxv374ezXW/tA0hisYpUqQw3SBzNnerM2+Ib6Lc4B0Bsq0CmloLl04KNnDgQKPHp2DrAIoYFdlTiRIlJOOVPWEWGUaCkclCKqAZFHddcejQoVKVT9kPBzBXTgI2ccJcsWIFMWC6IZNQnS5YsKCynh1ABTRXG1xnKNjIkSONKkvBDwfQSIVCnRawgNsNij+cmDjiqGO7Yx+7r7Ke1cJerqwnT1gaLWH+OUDLmTNnun1QpshHdkDZHzJVHcJPYpApsiuY5aPfDqCXWouKJWWN8wg1wltlB/uGKonCkCdOnKhg9o+BOIB2Dpmkf+rSHk2nCMmRnGKEsgOmqEqiHACnTp2qYH49hgft0Qiblyxrqu2UYznyw8mgqxSQixQpwj2AurREG0dTlsrp06eNZkKRz1bcHMkArAS/3A0Y7P7+gLPRnDlzAlZoGgYYQqa9jeCxJEr13KatT0zIHeD7AziSDAaKh4sXL/ZpmSUgtA5Qigzg+wNL0x1YCB1gt1KVDkrFq1at8ss+n+AQOkDNQkYOOcr9/YFP+3wCQuUArFPeI3E16C6J+jTOBuDhMlAOW8AyhwGqVKY5h3f3F07m17AIoXKAcJcVro0bN8JSw2Kot7ahcoBdVp6zoP5cJuCGNzsCf28TZ4FhKDJzqSMtC8U6DtUidnzmQk2RoqBn0tA6gBsev6Vzf40W2CTTKuQO0IebTfCpDkfKgI2WDf+GA/Sn+Bz3ReraStrkl/yXHMAmGLU57HOtryoUfhktwWE60MgMYyNTv+AOge/qqKb4dbD+g/J/AVf65lqU7WK5AAAAAElFTkSuQmCC",
```
findleyr: Since the actual data does not matter, can we please make this "foobar".

I don't want arbitrary, unreviewable data to be checked into the repo (I don't know what this data represents).

mcp/conformance_test.go
```diff
@@ -135,19 +135,36 @@ func incTool(_ context.Context, _ *CallToolRequest, args incInput) (*CallToolRes
...
+
 // runServerTest runs the server conformance test.
 // It must be executed in a synctest bubble.
 func runServerTest(t *testing.T, test *conformanceTest) {
 	ctx := t.Context()
 	// Construct the server based on features listed in the test.
-	s := NewServer(&Implementation{Name: "testServer", Version: "v1.0.0"}, nil)
+	impl := &Implementation{Name: "testServer", Version: "v1.0.0"}
+
+	// TODO(IAmSurajBobade): Remove this hack once we have a client protocol specific handling.
```
findleyr: You can remove your self assignment here. We'll probably clean up by adding a more general "settings" file to the test archive.

IAmSurajBobade: removed comment

mcp/conformance_test.go
IAmSurajBobade: removed it to foobar. Its a base64 version of official MCP logo. You can check this at https://jaredwinick.github.io/base64-image-viewer/

mcp/shared.go
IAmSurajBobade: removed

IAmSurajBobade:
@findleyr addressed review comments. Also updated everything example to use icons and websiteUrl wherever applicable.

findleyr:
Thanks again -- this looks great.

I was just going through it once more before merging, and noticed that we missed adding icons to ResourceTemplate and ResourceLink. Could you please include those as well?

mcp/protocol.go
```diff
@@ -658,6 +658,21 @@ type ProgressNotificationParams struct {
...
+type Icon struct {
+	// Source is A URI pointing to the icon resource (required). This can be:
+	// - An HTTP/HTTPS URL pointing to an image file
+	// - A data URI with base64-encoded image data
+	Source string `json:"src"`
+	// Optional MIME type if the server's type is missing or generic
+	MIMEType string `json:"mimeType,omitempty"`
+	// Optional size specification (e.g., ["48x48"], ["any"] for scalable formats like SVG, or ["48x48", "96x96"] for multiple sizes)
+	Sizes []string `json:"sizes,omitempty"`
+}
```
jba, line 676: There is also theme: https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/schema/draft/schema.ts#L355

jba (changes requested):

mcp/protocol.go
IAmSurajBobade: Added option to set theme.

IAmSurajBobade:
Thanks @findleyr, @jba for review - I have incorporated your latest comments. Also updated conformance test and Everything server example.

While adding tests, noticed we have missed adding json tag for structs implementing Content interface, created https://github.com/modelcontextprotocol/go-sdk/issues/631 for same.

mcp/testdata/conformance/server/spec-sep-973-additional-metadata.txtar
```diff
@@ -0,0 +1,217 @@
+Check behavior of server with Icons and websiteUrl metadata added as part of SEP-973 specification.
+
+check modelcontextprotocol/go-sdk/issues/552 for more details.
+
+Checks following:
+- If client sends protocolVersion as "draft", server responds with same
```
findleyr: This is no longer accurate.

mcp/protocol.go
```diff
@@ -658,6 +658,23 @@ type ProgressNotificationParams struct {
 
 func (*ProgressNotificationParams) isParams() {}
 
+// Icon provides visual identifiers for their resources, tools, prompts, and implementations
```
findleyr, line 661: This should be a single doc comment group (the newline breaks documentation).

mcp/content_test.go
```diff
@@ -192,3 +193,53 @@ func TestEmbeddedResource(t *testing.T) {
...
+			name:    "ResourceLink",
+			json:    `{"type":"resource_link","mimeType":"text/plain","uri":"https://example.com/resource","name":"Example Resource","title":"A comprehensive example resource","description":"This resource demonstrates all fields","_meta":{"custom":"metadata"},"icons":[{"src":"foobar","mimeType":"image/png","sizes":["48x48"],"theme":"light"}], "size":24,"annotations":{"audience":["user","assistant"],"lastModified":"2025-01-12T15:00:58Z","priority":0.5}}`,
+			content: &mcp.ResourceLink{},
+			expectContent: &mcp.ResourceLink{
+				URI:         "https://example.com/resource",
+				Name:        "Example Resource",
+				Title:       "A comprehensive example resource",
+				Description: "This resource demonstrates all fields",
+				MIMEType:    "text/plain",
+				// Meta:        mcp.Meta{"custom": "metadata"},
```
findleyr, line 217: This should work. I'll investigate why it doesn't.

IAmSurajBobade: For this specific issue only, I have raised https://github.com/modelcontextprotocol/go-sdk/issues/631.

mcp/protocol.go
IAmSurajBobade: updated

mcp/testdata/conformance/server/spec-sep-973-additional-metadata.txtar
IAmSurajBobade: correct. removed.

findleyr (approved):
Thanks for this change!

Issue #580, opened 2025-10-14
Proposal: disable streamable replayability by default


findleyr:
This issue proposes a change in default behavior for the `StreamableServerTransport`. Specifically, I propose changing the the default behavior documented here:
https://github.com/modelcontextprotocol/go-sdk/blob/31e97ad69edad679c726449c4584353502965cf5/mcp/streamable.go#L366

To the following: `// If nil, stream replayability is disabled.`.

While this is technically a breaking change in behavior, I believe the current behavior is unlikely to be useful, and is problematic in a number of ways:

1. As reported in #571, replayability may be insecure if the authorization middleware does not validate the `Mcp-Session-Id` header against the authorized user, meaning the server author must be explicitly aware of the session hijacking attack pattern. This would require the attacker to have access to the cryptographically random session ID and have the ability to terminate a client request, so I think this is unlikely, but still possible.
2. Even with bounded buffer size (as currently implemented), the `MemoryEventStore` incurs memory cost on the server, and the accounting metadata (the stream mapping) is a slow memory leak. This is a barrier to scalability.
3. With replayability enabled, message delivery is necessarily asynchronous. When a client cancels their HTTP request, we can't automatically cancel the corresponding server operation, because we can never know if a subsequent request will resume the stream. This is yet another performance concern.
4. With replayability enabled, we can't know if server->client calls might eventually succeed, so server operations that interact with the client must hang until they timeout (assuming the server is using a reasonable timeout, which is not guaranteed!).
5. Servers that care about replayability are likely to be hosted in ways where an in-memory event store is not sufficient. For example, the current in-memory event store provides no benefit to stateless serving.

With all these considerations, I feel strongly that the current default behavior is not worth the marginal benefit it provides, and the best path forward is to disable replayability by default.

Subsequent proposals will provide APIs for enabling secure, robust replayability.

CC @jba @samthanawalla @neild

findleyr:
This was done. Since no objection, marking this as approved and including it in the v1.1.0 prerelease.

PR #585, opened 2025-10-15
mcp: allow disabling stream replay; disable by default

### LINKED ISSUE #576: Streamable HTTP: no way to disable support for event replay
We spent some effort implementing a fixed buffer for streamable replay. However, in doing so, we made a simplification (at my suggestion, FWIW) that we consolidate the replay codepath with the ordinary response codepath:
https://github.com/modelcontextprotocol/go-sdk/blob/920e5a39ba9026af14ab459ce856e6ef907de0b9/mcp/streamable.go#L825

This means that the streamable server transport relies on its event store for correct functionality, which therefore means there's no way to disable replayability.

For some (perhaps many) servers, replayability incurs a significant performance overhead for little benefit. We should decouple the event store from synchronous response to an ongoing request.

findleyr:
This CL implements two changes that must be logically connected.

The first is to provide a mechanism for disabling stream replay. Previously, our documentation stated that if
StreamableServerTransport.EventStore was nil, a default in-memory event store would be used. In that case, there would be no way to completely disable stream replay, and as described in #576, there are many cases where replay is undesirable.

But of course changing the behavior of the transport zero value implicitly affects its default behavior, and so this CL also implements the proposal #580: stream replay is disabled by default.

Implementing this change required a significant refactoring, as previously we were relying on the event store for serialized message delivery from the JSON-RPC layer to the MCP layer: the connection would write to the event store, and independently the stream (be it an incoming POST or replay GET) would iterate and serve messages in the stream.

In order to achieve the goals of this CL, it was necessary to decouple storage from delivery. The 'stream' abstraction now tracks a delivery callback that writes straight to the HTTP response. It would have been convenient to store the ongoing http.ResponseWriter directly in the stream (this is how typescript does it), but due to the design of our EventStore API, only the HTTP handler knows the next event index, so a 'deliver' abstraction was an unfortunate requirement (suggestions for how to further simplify this are welcome).

More simplification is possible: in particular, as a result of this refactoring it should be entirely possible to clean up streams once we've received all responses. Any replay would only need access to the EventStore, if at all. This is left to a follow-up CL, to limit this already significant change.

Furthermore, a nice consequence of this refactoring is that, when not using event storage, servers can get synchronous feedback that message delivery failed, which should avoid unnecessary work. We can additionally cancel ongoing requests on early client termination, but that is also left to a follow-up CL.

Throughout, the terminology 'standalone SSE stream' replaced 'hanging GET' when referring to the non-replay GET request issued by the client. This is consistent with other SDKs.

Fixes #576
Updates #580

### DIFF
--- a/mcp/streamable.go
+++ b/mcp/streamable.go
@@ -11,7 +11,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"iter"
 	"log/slog"
 	"math"
 	"math/rand/v2"
@@ -72,6 +71,9 @@ type StreamableHTTPOptions struct {
 	// Logger specifies the logger to use.
 	// If nil, do not log.
 	Logger *slog.Logger
+
+	// TODO(rfindley): file a proposal to export this option, or something equivalent.
+	configureTransport func(req *http.Request, transport *StreamableServerTransport)
 }
 
 // NewStreamableHTTPHandler returns a new [StreamableHTTPHandler].
@@ -238,6 +240,9 @@ func (h *StreamableHTTPHandler) ServeHTTP(w http.ResponseWriter, req *http.Reque
 			jsonResponse: h.opts.JSONResponse,
 			logger:       h.opts.Logger,
 		}
+		if h.opts.configureTransport != nil {
+			h.opts.configureTransport(req, transport)
+		}
 
 		// To support stateless mode, we initialize the session with a default
 		// state, so that it doesn't reject subsequent requests.
@@ -335,16 +340,16 @@ func (h *StreamableHTTPHandler) ServeHTTP(w http.ResponseWriter, req *http.Reque
 //
 // Reads from the streamable server connection receive messages from http POST
 // requests from the client. Writes to the streamable server connection are
-// sent either to the hanging POST response, or to the hanging GET, according
-// to the following rules:
+// sent either to the related stream, or to the standalone SSE stream,
+// according to the following rules:
 //   - JSON-RPC responses to incoming requests are always routed to the
 //     appropriate HTTP response.
 //   - Requests or notifications made with a context.Context value derived from
 //     an incoming request handler, are routed to the HTTP response
 //     corresponding to that request, unless it has already terminated, in
-//     which case they are routed to the hanging GET.
+//     which case they are routed to the standalone SSE stream.
 //   - Requests or notifications made with a detached context.Context value are
-//     routed to the hanging GET.
+//     routed to the standalone SSE stream.
 type StreamableServerTransport struct {
 	// SessionID is the ID of this session.
 	//
@@ -372,15 +377,16 @@ type StreamableServerTransport struct {
 	// Specifically, responses will be application/json whenever incoming POST
 	// request contain only a single message. In this case, notifications or
 	// requests made within the context of a server request will be sent to the
-	// hanging GET request, if any.
+	// standalone SSE stream, if any.
 	//
 	// TODO(rfindley): jsonResponse should be exported, since
-	// StreamableHTTPOptions.JSONResponse is exported.
+	// StreamableHTTPOptions.JSONResponse is exported, and we want to allow users
+	// to write their own streamable HTTP handler.
 	jsonResponse bool
 
 	// optional logger provided through the [StreamableHTTPOptions.Logger].
 	//
-	// TODO(rfindley): logger should be exported, since we want to allow people
+	// TODO(rfindley): logger should be exported, since we want to allow users
 	// to write their own streamable HTTP handler.
 	logger *slog.Logger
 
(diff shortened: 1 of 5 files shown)


mcp/streamable.go
```diff
@@ -72,6 +71,9 @@ type StreamableHTTPOptions struct {
 	// Logger specifies the logger to use.
 	// If nil, do not log.
 	Logger *slog.Logger
+
+	// TODO(rfindley): file a proposal to export this option, or something equivalent.
+	configureTransport func(req *http.Request, transport *StreamableServerTransport)
```
neild, line 76: I'm not sure why `StreamableServerTransport` is an exported type. `StreamableHTTPHandler` creates them and manages their lifecycle.

`StreamableServerTransport` has a couple fields which are configured by `StreamableHTTPOptions`: `Stateless` and `Logger`. Should the event store be configured in the same way, by adding `StreamableHTTPOptions.EventStore`?

Are there any occasions today when a user is expected to touch a `StreamableServerTransport`?

mcp/streamable.go
```diff
@@ -472,58 +480,45 @@ func (c *streamableServerConn) SessionID() string {
...
+	// If non-nil, deliver writes data directly to the HTTP response.
 	//
-	// Lifecycle: each channel value persists for the duration of an HTTP POST or
-	// GET request for the given streamID.
-	signal atomic.Pointer[chan struct{}]
-
-	// The following mutable fields are protected by the mutex of the containing
-	// StreamableServerTransport.
+	// Only one HTTP response may receive messages at a given time. An active
+	// HTTP connection acquires ownership of the stream by setting
```
neild: Incomplete sentence?

mcp/streamable.go
```diff
@@ -472,58 +480,45 @@ func (c *streamableServerConn) SessionID() string {
...
 	requests map[jsonrpc.ID]struct{}
 }
 
-func (c *streamableServerConn) newStream(ctx context.Context, id string, isInitialize, jsonResponse bool) (*stream, error) {
-	if err := c.eventStore.Open(ctx, c.sessionID, id); err != nil {
-		return nil, err
+// done reports whether the stream is logically complete.
+//
+// s.mu must be held while calling this function.
+func (s *stream) done() bool {
```
neild: This is your codebase, so use whatever local style you prefer, but I'd call this `doneLocked` to indicate the stream mutex must be held.

mcp/streamable.go
```diff
@@ -571,37 +566,149 @@ func (t *StreamableServerTransport) ServeHTTP(w http.ResponseWriter, req *http.R
...
+	defer cancel()
+
+	// Acquire the stream.
+	//
+	// Importantly, hold the mutex until we've replayed all messages, since we
+	// don't want to allow delivery (or storage) of new messages until we've
+	// replayed everything that has been stored thus far.
+	stream.mu.Lock()
+	if stream.deliver != nil {
+		stream.mu.Unlock()
```
neild: I hate trying to remember to unlock a mutex on every early return path; it's much too easy to miss a case, and much to easy to not notice if that case only happens in an uncommon return path.

I'd do one of:

1. Factor out the part of this function that needs to lock the stream mutex:

   ```
   donec, err := c.serveGETLocked(...)
   if err != nil {
     return err
   }
   // Release the stream when we're done.
   defer func() {
     stream.mu.Lock()
     stream.deliver = nil
     stream.mu.Unlock()
   }()
   select {
   case <-ctx.Done():
     // request cancelled
   case <-donec:
     // request complete
   case <-c.done:
     // session closed
   }
   ```

2. Set a bool after unlocking:
   ```
   stream.mu.Lock()
   needStreamMuUnlock := true
   defer func() {
     if needStreamMuUnlock {
       stream.mu.Unlock()
     }
   }()
   // ...
   stream.mu.Unlock()
   needStreamMuUnlock = false
   ```

neild (approved):

mcp/streamable.go
jba: It's so users can write their own version of StreamableHTTPHandler.

mcp/streamable.go
jba: +1

mcp/streamable.go
jba: I agree: there are too many places where you unlock before returning. It's definitely a problem for maintenance. I prefer the idea of a separate function, but either of these will work fine.

mcp/streamable.go
```diff
@@ -661,204 +775,145 @@ func (c *streamableServerConn) servePOST(w http.ResponseWriter, req *http.Reques
...
+			}
+			e := Event{
+				Name: "message",
+				Data: data,
+			}
+			if c.eventStore != nil {
+				e.ID = formatEventID(stream.id, lastIndex)
+			}
+			lastIndex++
+			if _, err := writeEvent(w, e); err != nil {
```
jba: Can you do what you did on line 852?

mcp/streamable.go

mcp/streamable.go
findleyr: Funny, I normally do that but thought it was specific to *my* style, so opted not to do it here.

mcp/streamable.go
findleyr: Moot now, since this is factored out.

mcp/streamable.go
findleyr: Let's leave configuring the event store to a separate CL (and API proposal).

For now, users can configure it by writing their own StreamableHTTPHandler.

mcp/streamable.go
findleyr: Thanks. Yeah, I'd considered this and thought the abstraction was too arbitrary, but I think it worked out OK. PTAL.

findleyr:
@neild could you please re-approve when you have a chance? I had to rebase and resolve merge conflicts (I also squashed since I had to force-push anyway).

No changes other than resolving the conflict.

neild (approved):

PR #594, opened 2025-10-20
mcp: add StreamableHTTPOptions.SessionTimeout

### LINKED ISSUE #499: Proposal: Add StreamableHTTPOptions.SessionTimeout (Ungraceful client disconnects leads to zombie sessions)
### Describe the bug

The streamable HTTP transport implementation in `mcp/streamable.go` leaks a goroutine on the server for each client that disconnects ungracefully. Over time, a server can accumulate many "zombie" sessions, consuming resources.

#### Details

When a new MCP session is established via the `connect` function in [transport.go:141](https://github.com/modelcontextprotocol/go-sdk/blob/c2458f9a769b73239c6737cab3c95be5578e9734/mcp/transport.go#L141), the `jsonrpc2` connection layer spawns a long-running goroutine to read incoming client messages. This goroutine, created in [`internal/jsonrpc2/conn.go:279`](https://github.com/modelcontextprotocol/go-sdk/blob/728e0e3ac967a0cbf94eca9bce7a32b7ce0d287d/internal/jsonrpc2/conn.go#L279), eventually blocks on the [`streamableServerConn.Read(...)`](https://github.com/modelcontextprotocol/go-sdk/blob/02f0b2528642b41fa9be80f7f4773cbd3eb28994/mcp/streamable.go#L868) method.

  The Read method waits on one of three conditions:
   1. The session's context is canceled.
   2. A message arrives on the incoming channel.
   3. The session's done channel is closed.

When a client disconnects abruptly (e.g., due to network failure), no more HTTP requests can be made, so no new messages will ever arrive on the incoming channel. The session's context is [intentionally detached](https://github.com/modelcontextprotocol/go-sdk/blob/728e0e3ac967a0cbf94eca9bce7a32b7ce0d287d/internal/jsonrpc2/conn.go#L232) from the underlying HTTP requests, so it does not get canceled. The session is only terminated, and its `done` channel closed, when a keep-alive fails or the client sends an explicit `DELETE` request.

In the case of an ungraceful disconnect, the server is never notified, leaving the reading goroutine blocked indefinitely in the `Read(...)` method, thus leaking the goroutine.

### To Reproduce

Checkout the code from #496 (which adds an opt-in leak detector for unit tests) and run the test with the leak detector enabled:

```shell
❯ git checkout fgrosse:fgrosse/goleaks

❯ go test -run TestClientReplay/default -leak ./mcp
PASS
goleak: Errors on successful test run: found unexpected goroutines:
[Goroutine 66 in state select, with github.com/modelcontextprotocol/go-sdk/mcp.(*streamableServerConn).Read on top of the stack:
github.com/modelcontextprotocol/go-sdk/mcp.(*streamableServerConn).Read(0x140003ae0c0, {0x10553b1f8, 0x1400038a090})
	/Users/fgroe/src/panw/mcp/official-go-sdk/mcp/streamable.go:869 +0x7c
github.com/modelcontextprotocol/go-sdk/internal/jsonrpc2.(*Connection).readIncoming(0x1400038c410, {0x10553b1f8, 0x1400038a090}, {0x10c77c280, 0x140003ae0c0}, {0x105533c80, 0x140003a2010})
	/Users/fgroe/src/panw/mcp/official-go-sdk/internal/jsonrpc2/conn.go:536 +0x54
created by github.com/modelcontextprotocol/go-sdk/internal/jsonrpc2.NewConnection.(*Connection).start.func1 in goroutine 46
	/Users/fgroe/src/panw/mcp/official-go-sdk/internal/jsonrpc2/conn.go:279 +0xd8
]
exit status 1
FAIL	github.com/modelcontextprotocol/go-sdk/mcp	3.386s
```

### Expected behavior

The server should never leak any goroutines.

### Additional context

I found this issue while investigating flaky unit tests via #489

findleyr:
Add a timeout option for the streamable HTTP handler that automatically cleans up idle sessions.

Also, fix a bug in the streamable client, where we hang on a request even though the client can never get a response (because the HTTP request terminated without a response or Last-Event-Id).

Fixes #499

### DIFF
--- a/mcp/streamable.go
+++ b/mcp/streamable.go
@@ -12,6 +12,7 @@ import (
 	"fmt"
 	"io"
 	"log/slog"
+	"maps"
 	"math"
 	"math/rand/v2"
 	"net/http"
@@ -40,12 +41,76 @@ type StreamableHTTPHandler struct {
 	getServer func(*http.Request) *Server
 	opts      StreamableHTTPOptions
 
-	onTransportDeletion func(sessionID string) // for testing only
+	onTransportDeletion func(sessionID string) // for testing
 
-	mu sync.Mutex
-	// TODO: we should store the ServerSession along with the transport, because
-	// we need to cancel keepalive requests when closing the transport.
-	transports map[string]*StreamableServerTransport // keyed by IDs (from Mcp-Session-Id header)
+	mu       sync.Mutex
+	sessions map[string]*sessionInfo // keyed by session ID
+}
+
+type sessionInfo struct {
+	session   *ServerSession
+	transport *StreamableServerTransport
+
+	// If timeout is set, automatically close the session after an idle period.
+	timeout time.Duration
+	timerMu sync.Mutex
+	refs    int // reference count
+	timer   *time.Timer
+}
+
+// startPOST signals that a POST request for this session is starting (which
+// carries a client->server message), pausing the session timeout if it was
+// running.
+//
+// TODO: we may want to also pause the timer when resuming non-standalone SSE
+// streams, but that is tricy to implement. Clients should generally make
+// keepalive pings if they want to keep the session live.
+func (i *sessionInfo) startPOST() {
+	if i.timeout <= 0 {
+		return
+	}
+
+	i.timerMu.Lock()
+	defer i.timerMu.Unlock()
+
+	if i.timer == nil {
+		return // timer stopped permanently
+	}
+	if i.refs == 0 {
+		i.timer.Stop()
+	}
+	i.refs++
+}
+
+// endPOST sigals that a request for this session is ending, starting the
+// timeout if there are no other requests running.
+func (i *sessionInfo) endPOST() {
+	if i.timeout <= 0 {
+		return
+	}
+
+	i.timerMu.Lock()
+	defer i.timerMu.Unlock()
+
+	if i.timer == nil {
+		return // timer stopped permanently
+	}
+
+	i.refs--
+	assert(i.refs >= 0, "negative ref count")
+	if i.refs == 0 {
+		i.timer.Reset(i.timeout)
+	}
+}
+
+// stopTimer stops the inactivity timer permanently.
+func (i *sessionInfo) stopTimer() {
+	i.timerMu.Lock()
+	defer i.timerMu.Unlock()
+	if i.timer != nil {
+		i.timer.Stop()
+		i.timer = nil
+	}
 }
 
 // StreamableHTTPOptions configures the StreamableHTTPHandler.
@@ -77,6 +142,14 @@ type StreamableHTTPOptions struct {
 	// If set, EventStore will be used to persist stream events and replay them
 	// upon stream resumption.
 	EventStore EventStore
+
+	// SessionTimeout configures a timeout for idle sessions.
+	//
+	// When sessions receive no new HTTP requests from the client for this
+	// duration, they are automatically closed.
+	//
+	// If SessionTimeout is the zero value, idle sessions are never closed.
+	SessionTimeout time.Duration
 }
 
 // NewStreamableHTTPHandler returns a new [StreamableHTTPHandler].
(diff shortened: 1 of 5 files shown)


mcp/streamable.go
```diff
@@ -313,23 +365,36 @@ func (h *StreamableHTTPHandler) ServeHTTP(w http.ResponseWriter, req *http.Reque
...
 			h.mu.Lock()
-			h.transports[transport.SessionID] = transport
+			h.sessions[transport.SessionID] = sessInfo
 			h.mu.Unlock()
 		}
 	}
 
-	transport.ServeHTTP(w, req)
+	sessInfo.resetTimeout()
+	sessInfo.transport.ServeHTTP(w, req)
```
neild, line 435: If I'm reading this correctly, a HTTP request will time out after SessionTimeout, even if the handler is still handling a long-running request. Is this intentional?

mcp/streamable.go
```diff
@@ -40,12 +41,46 @@ type StreamableHTTPHandler struct {
...
+		return
+	}
+
+	i.timerMu.Lock()
+	defer i.timerMu.Unlock()
+
+	if i.timer == nil {
+		return
+	}
+	// Reset the timer if we successfully stopped it.
```
neild: "if we successfully stopped it" doesn't appear to be correct; the timer is not stopped. (If it was stopped, it is nil and we don't reset it.)

mcp/streamable_test.go
```diff
@@ -1623,7 +1624,69 @@ func TestStreamableClientContextPropagation(t *testing.T) {
...
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+
+	server := NewServer(testImpl, nil)
+
+	deleted := make(chan string, 1)
+	handler := NewStreamableHTTPHandler(
+		func(req *http.Request) *Server { return server },
+		&StreamableHTTPOptions{
+			SessionTimeout: 50 * time.Millisecond,
```
neild, line 1642: This would be a good use for `testing/synctest`, but it would require limiting the test to only running on Go 1.25+.

mcp/streamable.go
findleyr: Whoops, yes this was a stale comment from a previous iteration. Thanks for catching.

mcp/streamable.go
findleyr: Excellent point. Yes, you were reading this correctly, and I didn't sufficiently consider long-running requests.

I've updated it to use a reference count, and only reset the timer when the reference count drops to zero.
We only instrument this for POST requests, because GETs may be the standalone GET which is never terminated. We could potentially do better in the future, as noted, but I think this is a good enough start.

mcp/streamable_test.go
findleyr: Unfortunately, it doesn't work because the test uses httptest.NewServer, which does (real) I/O. Therefore, synctest.Wait never terminates.

findleyr:
Staticcheck failures are being addressed in https://github.com/modelcontextprotocol/go-sdk/pull/599.

mcp/streamable_test.go
```diff
@@ -1623,7 +1624,69 @@ func TestStreamableClientContextPropagation(t *testing.T) {
...
+	sessionID := session.ID()
+	if sessionID == "" {
+		t.Fatal("client session has empty ID")
+	}
+
+	// Verify the session exists on the server.
+	serverSessions := slices.Collect(server.Sessions())
+	if len(serverSessions) != 1 {
+		t.Fatalf("got %d sessions, want 1", len(serverSessions))
+	}
```
jba, line 1669: Add TODO to test multiple request in flight.

jba (approved):

mcp/streamable_test.go
findleyr: Instead of a TODO, I just tested this.

jba (approved):

PR #595, opened 2025-10-20
examples: Add an example oauth-enabled client

jhrozek:
Despite the PR guidelines asking for a single commit, I thought it would be easier to split the PR into several - I can rework the PR if there's general interest in accepting this work.

The motivation for creating an example client was both a user asking on Discord as well as me exploring [a proposed client credentials flow extension](https://github.com/modelcontextprotocol/ext-auth/pull/3) and finally it seems that the [upcoming SDK tier proposals](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1648) put an emphasis on docs with examples.

For now, there is no example server, that would make the PR much more complex before I'm even sure upstream is interested, but I did test with the example server and AS from the Python SDK.

Provided you have a checkout of the Python SDK available, you'd do:
```
cd examples/servers/simple-auth
uv run mcp-simple-auth-as --port=9000 &
uv run mcp-simple-auth-rs --port=8001 --auth-server=http://localhost:9000  --transport=streamable-http --oauth-strict
```

And then run the client as:
```
MCP_ALLOW_HTTP_LOCALHOST=1 MCP_SERVER_PORT=8001 go run -tags mcp_go_client_oauth examples/client/simple-auth/main.go
```

I'd be happy to work on the PR further or split it if having a full OAuth example test is of interest to the SDK maintainers.

### DIFF
--- a/examples/client/simple-auth/main.go
+++ b/examples/client/simple-auth/main.go
@@ -533,15 +533,29 @@ func main() {
 
 	ctx := context.Background()
 
-	// Create MCP transport
+	// Create OAuth-enabled HTTP transport
+	oauthTransport, err := auth.NewHTTPTransport(performOAuthFlow, nil)
+	if err != nil {
+		log.Fatalf("Failed to create OAuth transport: %v", err)
+	}
+
+	// Create HTTP client with OAuth support
+	httpClient := &http.Client{
+		Transport: oauthTransport,
+		Timeout:   60 * time.Second,
+	}
+
+	// Create MCP transport with OAuth-enabled HTTP client
 	var transport mcp.Transport
 	if transportType == "sse" {
 		transport = &mcp.SSEClientTransport{
-			Endpoint: serverURL,
+			Endpoint:   serverURL,
+			HTTPClient: httpClient,
 		}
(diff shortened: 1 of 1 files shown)


jba:
Thanks for doing this. I think you found a design flaw in our implementation. Please take a look at #600 and comment.

jhrozek:
> Thanks for doing this. I think you found a design flaw in our implementation. Please take a look at #600 and comment.

Thanks for braving a review of this large PR! I will be able to take a look probably during the weekend, maybe on Monday. Other work/life got in the way, sorry for the delay.

maciej-kisiel:
Hi @jhrozek! The most recent `v1.4.0` release includes proper client side OAuth support behind the `mcp_go_client_auth` build tag. It also comes with an example how to use it: https://github.com/modelcontextprotocol/go-sdk/blob/main/examples/auth/client/main.go. It should cover what you wanted to achieve with this PR in a simpler way. I will close this PR, but I would encourage you to try out the new integration if you're interested, as we're still in a phase where we can react to most of the feedback without being tied by stability guarantees. Thanks!

PR #596, opened 2025-10-21
oauthex: add missing build tag

findleyr:
The file oauthex/oauthex.go was missing the mcp_go_client_oauth build tag, making the package importable. I believe that was accidental.


samthanawalla:
It was importable in v1.0.0 so we had to keep the omission otherwise it would be a breaking change

findleyr:
Thanks, did not know that.

PR #597, opened 2025-10-22
auth: clone the client request body before roundtripping

### LINKED ISSUE #590: Replace body for second call and OAuth flow
HTTPTransport.RoundTrip needs to copy the body before it makes a second request. See the comment on line 100 of client.go at https://github.com/modelcontextprotocol/go-sdk/pull/544/files.

Also, add a test that would have found this bug. Look at how https://github.com/docker/cagent/pull/547 does testing.

findleyr:
RoundTrippers may read and close the body, so be careful to clone before roundtripping during client oauth, as the request may be issued multiple times.

Fixes #590

### DIFF
--- a/auth/client.go
+++ b/auth/client.go
@@ -69,9 +69,6 @@ func (t *HTTPTransport) RoundTrip(req *http.Request) (*http.Response, error) {
 	base := t.opts.Base
 	t.mu.Unlock()
 
-	// req1 is our first request in the authorization flow.
-	//
-	// If we mutate its body, we must clone it first.
 	var (
 		// If haveBody is set, the request has a nontrivial body, and we need avoid
 		// reading (or closing) it multiple times. In that case, bodyBytes is its
--- a/auth/client_test.go
+++ b/auth/client_test.go
@@ -122,7 +122,6 @@ func TestHTTPTransport(t *testing.T) {
 		client := &http.Client{Transport: transport}
 
 		resp, err := client.Post(authServer.URL, "application/json", &basicReader{strings.NewReader("{}")})
-		// resp, err := client.Post(authServer.URL, "application/json", strings.NewReader("{}"))
 		if err != nil {
 			t.Fatalf("client.Post() failed: %v", err)
 		}


findleyr:
Staticcheck failures are being addressed in https://github.com/modelcontextprotocol/go-sdk/pull/599.

auth/client.go
```diff
@@ -67,7 +69,32 @@ func (t *HTTPTransport) RoundTrip(req *http.Request) (*http.Response, error) {
 	base := t.opts.Base
 	t.mu.Unlock()
 
-	resp, err := base.RoundTrip(req)
+	// req1 is our first request in the authorization flow.
+	//
+	// If we mutate its body, we must clone it first.
+	req1 := req
+	var (
+		// If haveBody is set, the request has a nontrivial body, and need avoid
```
jba: we need?

auth/client.go
```diff
@@ -67,7 +69,32 @@ func (t *HTTPTransport) RoundTrip(req *http.Request) (*http.Response, error) {
...
+		// reading (or closing) it multiple times. In that case, bodyBytes is its
+		// content.
+		haveBody  bool
```
jba, line 76: Couldn't you just test whether bodyBytes is non-nil?

auth/client.go
```diff
@@ -67,7 +69,32 @@ func (t *HTTPTransport) RoundTrip(req *http.Request) (*http.Response, error) {
 	base := t.opts.Base
 	t.mu.Unlock()
 
-	resp, err := base.RoundTrip(req)
+	// req1 is our first request in the authorization flow.
+	//
+	// If we mutate its body, we must clone it first.
+	req1 := req
```
jba: I don't think you need req1. You can just update req in place.

auth/client.go
```diff
@@ -97,7 +124,16 @@ func (t *HTTPTransport) RoundTrip(req *http.Request) (*http.Response, error) {
 		}
 		t.opts.Base = &oauth2.Transport{Base: t.opts.Base, Source: ts}
 	}
-	return t.opts.Base.RoundTrip(req.Clone(req.Context()))
+
+	// If we don't have a body, the request is reusable, though it will be cloned
+	// by the base. However, if we've had to read the body, we must clone.
+	req2 := req
```
jba: I also think you can get away without req2.

jba (changes requested):
Thanks, this was a bit more complicated than I had thought.

auth/client.go
findleyr: I'm following the documentation for HTTP.RoundTripper precisely. The only thing a round tripper is allowed to do  is read and close the body. Setting a new body is something else: it's mutating the value of the Body field.

Similarly, this is why I make a req2 below. But it's ALSO why I don't even clone if the body is trivial, because the base roundtripper shouldn't mutate.

I also noted that the RoundTripper MUST close the body, which wasn't done in the PR you referenced (I commented as much).

Let's align on this before I make further changes.
https://pkg.go.dev/net/http#RoundTripper

jba: req is a pointer. I'm not saying anything about cloning, I'm just saying you don't need a second name for the pointer. `req = req.Clone(...)` below, for example.

auth/client.go
findleyr: I'd rather just have a variable keep track of whether we mutated, than rely on some convention of nillness vs emptyness.

Strictly speaking, the body could be empty but have a non-trivial Close function, and we should be very precise here.

auth/client.go
findleyr: Right, good point, done.

auth/client.go
findleyr: Done.

auth/client.go
findleyr: Oh, I see. Right. Done.

findleyr (commented):
Sorry, I had to rebase and squash in order to pick up the CI fixes.

auth/client_test.go
```diff
@@ -82,6 +107,32 @@ func TestHTTPTransport(t *testing.T) {
...
+		}
+
+		transport, err := NewHTTPTransport(handler, nil)
+		if err != nil {
+			t.Fatalf("NewHTTPTransport() failed: %v", err)
+		}
+		client := &http.Client{Transport: transport}
+
+		resp, err := client.Post(authServer.URL, "application/json", &basicReader{strings.NewReader("{}")})
+		// resp, err := client.Post(authServer.URL, "application/json", strings.NewReader("{}"))
```
findleyr: Cruft.

PR #599, opened 2025-10-22
.github: fix staticcheck linting

findleyr:
Update the test.yml lint job to run linting with Go 1.25, and use the ambient Go command to run staticcheck. This seems to fix a mysterious bug where staticcheck was failing, presumably in the Go installation step. Using the ambient Go command resulted in staticcheck failures, since some symbols are detected as unused at Go 1.23 due to build tags. Upgrading the Go version for this job avoids those errors.

Also, use a pinned version of staticcheck (as recommended), to avoid future failures when new versions are released.


rickkir686-cloud:
Google

PR #603, opened 2025-10-27
auth: change OAuthHandler to take http Request and Response

### LINKED ISSUE #600: auth.OAuthHandler should take http Request and Response
Currently the [definition of auth.OAuthHandler](https://github.com/modelcontextprotocol/go-sdk/blob/main/auth/client.go#L21) is
```
type OAuthHandler func(context.Context, OAuthHandlerArgs) (oauth2.TokenSource, error)
```
where `OAuthHandlerArgs` contains the resource metadata URL extracted from the WWW-Authenticate header of the HTTP response.

There is an impedance mismatch between this value and the protected resource metadata functions of the oauthex package, as illustrated by [this PR](https://github.com/modelcontextprotocol/go-sdk/pull/595/files#diff-1dc4da244d8738fab1892da4a78b0bfa507750a0578bbca8c92ded19c16a2cd7R112), which must perform additional processing on the resource metadata URL to obtain the actual resource metadata.

Furthermore, the resource meditated URL by itself is not sufficient to validate the protected resource metadata returned by the server. That requires the request URL.

A better definition would be:
```
type OAuthHandler func(req *http.Request, res *http.Response) (oauth2.TokenSource, error)
```
Although more work is required by the handler author, we already provide the function to do it. The  handler implementation should call [oauthex.GetProtectedResourceMetadataFromHeader](https://github.com/modelcontextprotocol/go-sdk/blob/main/oauthex/resource_meta.go#L60), which does all the necessary work and validation. The caller need only extract the URL from the request and the header from the response.

rutu1717:
Change OAuthHandler signature from func(context.Context, OAuthHandlerArgs) to func(req *http.Request, res *http.Response).

- Remove OAuthHandlerArgs struct
- Update HTTPTransport to pass req and resp to handler
- Update tests to use new signature
- Handler can now call oauthex.GetProtectedResourceMetadataFromHeader with proper validation against request URL

This change fixes an impedance mismatch between OAuthHandler and the protected resource metadata functions of the oauthex package. The new signature allows handlers to properly validate resource metadata against the request URL, as required by RFC 9728.

Fixes #600

### DIFF
--- a/auth/client.go
+++ b/auth/client.go
@@ -8,26 +8,19 @@ package auth
 
 import (
 	"bytes"
-	"context"
 	"errors"
 	"io"
 	"net/http"
 	"sync"
 
-	"github.com/modelcontextprotocol/go-sdk/oauthex"
 	"golang.org/x/oauth2"
 )
 
 // An OAuthHandler conducts an OAuth flow and returns a [oauth2.TokenSource] if the authorization
 // is approved, or an error if not.
-type OAuthHandler func(context.Context, OAuthHandlerArgs) (oauth2.TokenSource, error)
-
-// OAuthHandlerArgs are arguments to an [OAuthHandler].
-type OAuthHandlerArgs struct {
-	// The URL to fetch protected resource metadata, extracted from the WWW-Authenticate header.
-	// Empty if not present or there was an error obtaining it.
-	ResourceMetadataURL string
-}
+// The handler receives the HTTP request and response that triggered the authentication flow.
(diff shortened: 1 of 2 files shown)


findleyr:
@jba this LGTM, but I'm new to this code. Unless you have comments, I'll merge this tomorrow morning to include in the v1.1.0 release.

jba:
(better late than never) yes, I'm pretty convinced this is the right choice.

Issue #607, opened 2025-10-27
How to check client capabilities and handle unsupported capability errors?


robbyt:
**Is your feature request related to a problem? Please describe.**
I have two semi-related questions, caused by uncertainty of how to check client capabilities from an MCP server.
1. How can I checking the client capabilities before using them?
2. Can you expos the errors from internal/jsonrpc so I can check for them? (is this a possible solution?)

**re 1: How to check Capabilities**
I'd like to figure out if a MCP client supports a specific MCP server capability (e.g., `ListRoots`). I'm assuming client capabilities by reading the fields in the `mcp.ClientCapabilities` struct- Is this advisable?

Also, the Roots field is a value struct, so I can't check for `nil` on that like I can with the other capabilities fields `Sampling` or `Elicitation`: https://github.com/modelcontextprotocol/go-sdk/blob/d256a9c1094ee4044dc3fa6e55d3559df0047107/mcp/protocol.go#L183-L195

**re 2: Error types from internal/jsonrpc**
When using the elicitation/sampleing/logging capabilities, the examples just return an error to the client. The don't differentiate between the different types, or use any `errors.Is()`. So when a client doesn't support a capability (e.g., roots), the SDK returns a JSON-RPC error code -32601 (Method Not Found). However, `ErrMethodNotFound` is in the internal jsonrpc2 package and cannot be imported by my project, so I can't use `errors.Is()` to check if an error from ListRoots(), CreateMessage(), or Elicit() is due to an unsupported client capability.

**Describe the solution you'd like**

1. Consistent capability structure: Make Roots a pointer type like the other capabilities:

```go
type ClientCapabilities struct {
	Roots       *RootsCapabilities       `json:"roots,omitempty"`  // Changed from inline struct
	Sampling    *SamplingCapabilities    `json:"sampling,omitempty"`
	Elicitation *ElicitationCapabilities `json:"elicitation,omitempty"`
}

type RootsCapabilities struct {
	ListChanged bool `json:"listChanged,omitempty"`
}
```
This would allow checking `Capabilities.Roots != nil` to determine if roots are supported at all.

2. Export JSON-RPC errors: Export common JSON-RPC errors from the public jsonrpc package:

```go
// jsonrpc/jsonrpc.go
var (
	ErrMethodNotFound = internal.ErrMethodNotFound
	ErrInvalidParams  = internal.ErrInvalidParams
	// etc.
)
```

[...]

PR #608, opened 2025-10-27
mcp: add ClientCapabilities.Roots.Supported

findleyr:
Address an API oversight (#607) by exposing a Supported field on the ClientCapabilities.Roots.

Given our promise of compatibility, we unfortunately can't change the Roots field, so this may be the best we can do to work around this oversight.


jba:
I'm not sure it was an oversight. Our client always supports roots, even if the user doesn't add any.

robbyt:
I assume that servers using the Go SDK should support any client that correctly implements the spec (including clients that lack support of roots)

findleyr:
@jba I'm guessing that the origin of the mistake was that this type was internal only, for use by the client. In that case, it was acceptable to have inlined structs. But when capabilities were exported it became an API oversight, because we need to be able to represent whether the roots capability is present on the client.

internal/docs/warts.src.md
robbyt, line 1: The inclusion of this file is a great idea- but is `warts` part of an internal coding style? I'd prefer it called `known_issues.md` but that's just my 2¢.

WRT the content of the file, I'd suggest creating a PR with the inclusion of the file, and then update the content to add each entry in a different PR, instead of adding the file and the 2 entries here.

mcp/protocol.go
```diff
@@ -183,17 +183,71 @@ func (x *CancelledParams) SetProgressToken(t any) { setProgressToken(x, t) }
...
+
+	// Roots reports roots capabilities.
+	//
+	// Due to an API oversight (#607), roots is a non-pointer. Check the
+	// Supported field to see if the roots capability is present
 	Roots struct {
+		// Suppported reports whether roots/list is supported. It works around an
+		// API oversight: roots capabilities should have been a distinguished type,
+		// similar to other capabilities.
+		Supported bool `json:"-"`
```
robbyt, line 195: If this is added, then I prefer that the other capabilities structs have a similar `Supported bool` field added, to keep the API consistent. (or just remove this `Supported` workaround field, and break backwards compatibility, because it's a bug)

Right now, my server must assume the client supports Sampling if `ClientCapabilities.Sampling != nil`, which is fine, but for simplicity I would like all capability fields to behave similarly.

internal/docs/warts.src.md
findleyr: Thanks for the feedback! Yes, I had intended to change the name of this file -- "warts" is a term I've hear before, but isn't nice.

"known issues" is good, but makes me think of "bugs we haven't yet fixed" rather than "API decisions we can't revert". In https://github.com/modelcontextprotocol/go-sdk/pull/609 I went with "rough edges", but let me know what you think.

jba:
How about this option? I'm not totally sure it would work.

Add a Roots2 field, and move the json:roots tag to that field. Its type is a pointer. Change our client to populate Roots2.

At least this removes the asymmetry between Roots and the other capabilities. Servers have to know to check Roots2, but is this so different from them having to know to check Supported?

findleyr:
Technically speaking any change to struct tags is a breaking API change, no? I realized that I removed omitempty in this CL, and probably can't even do that.

jba:
> Technically speaking any change to struct tags is a breaking API change, no?

Excellent point. The spec says (Italics mine):

> Two struct types are identical if they have the same sequence of fields, and if corresponding pairs of fields have the same names, identical types, _and identical tags_...

richard-rance:
> Technically speaking any change to struct tags is a breaking API change, no? I realized that I removed omitempty in this CL, and probably can't even do that.

Adding the `MarshalJSON` and `UnmarshalJSON` methods is just as likely to break other people's usage of the struct as changing the json tags. Both change the behavior of the json package.

I don't see any public functions where the `ClientCapabilities` struct is used as an input or output so the odds of someone outside the SDK being affected are very low. The only possibility that I see is if client middleware is processing the params of the `initialize` method.

Adding a new field to a struct is almost always safe. ([ref](https://go.dev/blog/module-compatibility#maintaining-struct-compatibility:~:text=If%20you%20have%20an%20exported%20struct%20type%2C%20you%20can%20almost%20always%20add%20a%20field%20or%20remove%20an%20unexported%20field%20without%20breaking%20compatibility.)) 

Adding RootsV2 and deprecating the old field is technically the right way to go here but I don't think anyone would notice if you made the current field a pointer.

findleyr:
@richard-rance thanks for the feedback. You're right that adding new encoding fields is at least a gray area, if not an outright breaking change. The [standard library](https://pkg.go.dev/encoding) defines this as a breaking change, but leaves a bit of wiggle room for libraries where encoding is not strictly part of their compatibility promise.

But nevertheless, I think we should be very conservative about any breaking changes, and so I've come up with what I think is a better solution in #698: we do indeed add RootsV2, and fix the encoding only within the SDK JSON-RPC layer.

At some point in the next 6 months, I'm guessing we will go to v2, and so it's better to be extremely conservative for now, and fix it in the next major version.

Abandoning this PR.

PR #611, opened 2025-10-30
mcp: relax SSE connection response handling in non-strict mode

### LINKED ISSUE #610: Relax SSE connection response handling
This issue is similar to #393.

As described in the previous issue, a MCP server may return an error code instead of 405 Method Not Allowed when the client initializes an SSE connection for incoming events.

Previously, We have addressed this by allowing 404 Not Found response code. However, I have encountered an MCP server that responds with 400 Bad Request on the SSE connection.

When testing with MCP Inspector, it also reports a Bad Request error in thsis case, but the session proceeds normally afterward. I also checked other SDK implementations:

- Python: ignores the SSE connection when an error occurs (https://github.com/modelcontextprotocol/python-sdk/blob/db9e4515514255e7378c0e120b042dd5981ec120/src/mcp/client/streamable_http.py#L218)
- Typescript: retries several times and then ignores the error (https://github.com/modelcontextprotocol/typescript-sdk/blob/7387c44bf548b23c5cd8f1b7ae88020bfa7058ad/src/client/streamableHttp.ts#L291)
- Go: terminates the MCP session

Would it make sense to further relax the requirement to accept non-200 responses in non-strict mode? I can submit a patch if sounds reasonable.

CC @findleyr

zhxie:
Some MCP servers do not support SSE, but when receiving a request, they return 400 Bad Request or 404 Not Found instead of the standard 405 Method Not Allowed. The TypeScript and Python SDK ignore this error, while the Go SDK terminates the session. This patch addresses the issue in the non-strict mode.

Fixed #610

### DIFF
--- a/mcp/streamable_client_test.go
+++ b/mcp/streamable_client_test.go
@@ -239,7 +239,11 @@ func TestStreamableClientGETHandling(t *testing.T) {
 	}{
 		{http.StatusOK, ""},
 		{http.StatusMethodNotAllowed, ""},
-		{http.StatusBadRequest, "standalone SSE"},
+		// The client error status code is not treated as an error in non-strict
+		// mode.
+		{http.StatusNotFound, ""},
+		{http.StatusBadRequest, ""},
+		{http.StatusInternalServerError, "standalone SSE"},
 	}
 
 	for _, test := range tests {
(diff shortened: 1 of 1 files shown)


zhxie:
CC @findleyr

findleyr (dismissed):

findleyr:
Test failure looks real (albeit the error message is missing from the logs?)

I think there's another test that asserts on behavior given a 400. PTAL and let me know if it's not straightforward to fix.

zhxie:
@findleyr Fixed. It looks like that approval is required to run the CI.

findleyr (approved):

PR #614, opened 2025-11-05
examples: Fix SSE streaming issue by moving logging from HTTP to MCP middlewarre

SpringMT:
In `examples/http`,  previous HTTP-level logging middleware blocked SSE streaming because `handler.ServeHTTP()` does not return until the connection closes for GET requests (SSE streams).
This prevented response headers from being sent to the client.

Changes:
- Replace HTTP-level logging handler with MCP middleware pattern
- Implement logging at the MCP protocol layer using mcp.Middleware
- Add middleware to server using AddReceivingMiddleware()
- Simplify main.go by removing HTTP handler wrapping

This allows SSE streams to work properly while still providing detailed logging for all MCP protocol-level method calls.

### DIFF
--- a/examples/http/logging_middleware.go
+++ b/examples/http/logging_middleware.go
@@ -24,8 +24,7 @@ func createLoggingMiddleware() mcp.Middleware {
 			sessionID := req.GetSession().ID()
 
 			// Log request details.
-			log.Printf("[REQUEST] %s | Session: %s | Method: %s",
-				start.Format(time.RFC3339),
+			log.Printf("[REQUEST] Session: %s | Method: %s",
 				sessionID,
 				method)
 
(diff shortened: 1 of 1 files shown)


findleyr:
Thank you. I'll see if @jba wants to review, since he reviewed the original. Otherwise, I will review on Monday.

examples/http/logging_middleware.go
```diff
@@ -5,47 +5,52 @@
...
+			ctx context.Context,
+			method string,
+			req mcp.Request,
+		) (mcp.Result, error) {
+			start := time.Now()
+			sessionID := req.GetSession().ID()
+
+			// Log request details.
+			log.Printf("[REQUEST] %s | Session: %s | Method: %s",
+				start.Format(time.RFC3339),
```
jba: The log package already prints the time by default. I know the previous code had this, but I think it's redundant.

jba (changes requested):

examples/http/logging_middleware.go
SpringMT: fixed https://github.com/modelcontextprotocol/go-sdk/pull/614/commits/e8b1d4beb3482da41ef7bc209a6d2f9d2fc314c5

Issue #633, opened 2025-11-11
Streamable HTTP clients can get stuck connecting


richard-rance:
**Describe the bug**
When connecting to some MCP servers that support making SSE GET requests, client.Connect gets hung up waiting for the first notification from the server. This is particularly a problem if they don't manually flush the response buffer after writing the headers. 

**To Reproduce**
I found the problem when running the following to connect to Zapier's MCP server. I then adjusted the unit tests to reproduce it locally in https://github.com/modelcontextprotocol/go-sdk/pull/634/files#diff-bc2db7f8b168cfd247ed95d1ac34698f70d9ad808c40102361253f4beda91058R110-R121

```golang
func main() {
	ctx := context.Background()

	client := mcp.NewClient(&mcp.Implementation{}, &mcp.ClientOptions{})
	logging.Infof(ctx, "Initializing...")

	cs, err := client.Connect(ctx,
		&mcp.StreamableClientTransport{
			Endpoint: "https://mcp.zapier.com/api/mcp/s/my-streamable-server-key-here/mcp",
		},
		&mcp.ClientSessionOptions{})
	if err != nil {
		logging.Errorf(ctx, "Failed to connect to MCP: %v", err)
		return
	}
	defer cs.Close()

	logging.Infof(ctx, "This doesn't run until the GET request times out 5 times")
}
```

**Expected behavior**
I expect creating a client connection to complete in under 1 second. (I tested this same server URL in other MCP clients.)
If the MCP server has configured their server correctly they will be able to send the client notifications. 
If they haven't the client should still be able to use the server's tools, prompts and resources.

**Additional context**
This works as expected using the SDK version prior to #604 / #583. That PR changed initialization of the standalone SSE connection to be synchronous. That allows some server errors from starting the stand alone SSE connection be returned as part of client.Connect.

As a client developer I have no control over the server's behaviour. The server misbehaving only affects the server's ability to send notifications. If the POST initialization request is successful but the GET SSE stream is not it is probably because the server does not send messages so didn't test it. I don't think that creating the connection needs to wait for the stand alone SSE connection.

In my particular use case I am building a stateless client so would like to disable creating the standalone SSE connections entirely.

findleyr:
Hi, I may have been hasty to make this block in #583, but let's figure it out once and for all.

The rationale was that connection should provide some guarantee that the bidirectional stream is established. For the streamable transport, that's not guaranteed until the standalone GET exists.

[...]

Issue #583, opened 2025-10-15
Establishing standalone SSE stream should be synchronous with streamable client connection


findleyr:
Right now, the streamable client transport establishes the standalone SSE stream as soon as it receives its initialized result, but does so asynchronously:
https://github.com/modelcontextprotocol/go-sdk/blob/31e97ad69edad679c726449c4584353502965cf5/mcp/streamable.go#L1134

This didn't matter much to us before (at least not in our tests), because we always guaranteed message delivery via our in-memory replay. However, as we look into disabling replay by default #580, some of our tests fail because the client can't establish a happens-before relationship such that background messages are guaranteed to be delivered.

Looking at the typescript SDK, it starts the request synchronously, immediately after sending 'initialized'.

PR #634, opened 2025-11-11
streamable HTTP client: Don't block on standalone SSE Get

richard-rance:
# Problem
When connecting to some MCP servers that supports making SSE GET requests, client.Connect gets hung up waiting for the first notification from the server. 

More details in issue #633

# Solution

This change modifies the tests to simulate a server that keeps the stand alone SSE connection open. It then adjusts the connection setup to run the standalone SSE connection in the background.

# Open question
Work was done 2 weeks ago in #604 / #583 to make some errors from starting the stand alone SSE connection be returned as part of client.Connect. I'm not clear on why that is necessary. Clients are not required to accept SSE messages from servers. Stateless clients will want to opt out of even attempting to create the standalone SSE connection.
This PR undoes some of it so may not be the correct solution. 

I'll clean up the tests once I know more.


richard-rance:
Closing in favor of https://github.com/modelcontextprotocol/go-sdk/pull/729

PR #640, opened 2025-11-14
mcp: validate tool names

### LINKED ISSUE #621: SEP-986: Enforce Format for Tool Names
In [SEP-986](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/986), requirements for tool names are added.

We should consider ways that we can enforce these requirements. The SEP has some guidance on a compatibility transition period.

findleyr:
Implement SEP 986, enforcing restrictions on tool names. Since we're already at v1 we couldn't turn this into a panic, so use an error log instead. This is consistent with the typescript SDK. A note is left in our 'rough edges' doc to panic in v2.

Fixes #621

### DIFF
--- a/mcp/tool.go
+++ b/mcp/tool.go
@@ -103,6 +103,8 @@ func applySchema(data json.RawMessage, resolved *jsonschema.Resolved) (json.RawM
 	return data, nil
 }
 
+// validateToolName checks whether name is a valid tool name, reporting a
+// non-nil error if not.
 func validateToolName(name string) error {
 	if name == "" {
 		return fmt.Errorf("tool name cannot be empty")
@@ -110,10 +112,12 @@ func validateToolName(name string) error {
 	if len(name) > 128 {
 		return fmt.Errorf("tool name exceeds maximum length of 128 characters (current: %d)", len(name))
 	}
+	// For consistency with other SDKs, report characters in the order the appear
+	// in the name.
 	var invalidChars []string
 	seen := make(map[rune]bool)
 	for _, r := range name {
-		if !((r >= 'a' && r <= 'z') || (r >= 'A' && r <= 'Z') || (r >= '0' && r <= '9') || r == '_' || r == '-' || r == '.') {
+		if !validToolNameRune(r) {
 			if !seen[r] {
 				invalidChars = append(invalidChars, fmt.Sprintf("%q", string(r)))
 				seen[r] = true
@@ -124,4 +128,12 @@ func validateToolName(name string) error {
 		return fmt.Errorf("tool name contains invalid characters: %s", strings.Join(invalidChars, ", "))
 	}
 	return nil
+}
+
+// validToolNameRune reports whether r is valid within tool names.
+func validToolNameRune(r rune) bool {
+	return (r >= 'a' && r <= 'z') ||
+		(r >= 'A' && r <= 'Z') ||
+		(r >= '0' && r <= '9') ||
+		r == '_' || r == '-' || r == '.'
 }


mcp/tool.go
```diff
@@ -101,3 +102,26 @@ func applySchema(data json.RawMessage, resolved *jsonschema.Resolved) (json.RawM
...
+	if name == "" {
+		return fmt.Errorf("tool name cannot be empty")
+	}
+	if len(name) > 128 {
+		return fmt.Errorf("tool name exceeds maximum length of 128 characters (current: %d)", len(name))
+	}
+	var invalidChars []string
+	seen := make(map[rune]bool)
+	for _, r := range name {
+		if !((r >= 'a' && r <= 'z') || (r >= 'A' && r <= 'Z') || (r >= '0' && r <= '9') || r == '_' || r == '-' || r == '.') {
```
samthanawalla: nit: a switch would be clearer

mcp/tool.go
```diff
@@ -101,3 +102,26 @@ func applySchema(data json.RawMessage, resolved *jsonschema.Resolved) (json.RawM
...
+			if !seen[r] {
+				invalidChars = append(invalidChars, fmt.Sprintf("%q", string(r)))
```
jba, line 122: Use maps.Keys(seen) at the end.

findleyr: Can't: need to report characters in the order they occur. Want to be consistent with other SDKs.

mcp/tool.go
findleyr: Hmm, checked the standard library and we don't use switches for this sort of check anywhere: we either use a conditional or boolean array.

Factored out to an (inlined) helper function, which I think makes it clearer. WDYT?

findleyr:
@markus-kusano or @samthanawalla could you review?

mcp/tool.go
```diff
@@ -101,3 +102,38 @@ func applySchema(data json.RawMessage, resolved *jsonschema.Resolved) (json.RawM
...
+// validateToolName checks whether name is a valid tool name, reporting a
+// non-nil error if not.
+func validateToolName(name string) error {
+	if name == "" {
+		return fmt.Errorf("tool name cannot be empty")
+	}
+	if len(name) > 128 {
+		return fmt.Errorf("tool name exceeds maximum length of 128 characters (current: %d)", len(name))
+	}
+	// For consistency with other SDKs, report characters in the order the appear
```
markus-kusano, line 115: "order the appear" ==> "order *they* appear"

mcp/tool.go
```diff
@@ -101,3 +102,38 @@ func applySchema(data json.RawMessage, resolved *jsonschema.Resolved) (json.RawM
...
 }
+
+// validateToolName checks whether name is a valid tool name, reporting a
+// non-nil error if not.
+func validateToolName(name string) error {
+	if name == "" {
+		return fmt.Errorf("tool name cannot be empty")
+	}
+	if len(name) > 128 {
+		return fmt.Errorf("tool name exceeds maximum length of 128 characters (current: %d)", len(name))
```
markus-kusano, line 113: feel free to ignore: should we include the tool name here to make it easier for debugging? i can also see this being a problem since it is too long. i could see this error message being annoying if the user could have multiple tool names and they don't know which one is too long

findleyr: The tool name is included at the call site.
An arguable style choice, but I mildly prefer having validators only describe the specific error, because this then leave it up to the caller for how to present that error. For example the caller could do "Adding tool %q: %v"

mcp/tool.go
findleyr: Done.

PR #642, opened 2025-11-16
mcp:audio-content-reference fix

### LINKED ISSUE #641: Issue with AudioContent Struct MarshalJSON
AudioContent.MarshalJSON uses a value receiver instead of a pointer receiver
The MarshalJSON method for the AudioContent struct in content.go currently uses a value receiver:
`func (c AudioContent) MarshalJSON() ([]byte, error)`

This is problematic because:
**Misleading Behavior:** The method modifies the Data field (e.g., setting it to an empty slice if nil), but since the method operates on a copy of the struct, these changes are not applied to the original instance. This can lead to unexpected behavior.
**Inconsistent Design**: Other methods for AudioContent, such as fromWire, use a pointer receiver (*AudioContent), which is the correct approach when modifying the struct or ensuring efficient memory usage.
**Performance Overhead**: Using a value receiver creates a copy of the struct, which is unnecessary and inefficient, especially for large structs.

**Proposed Solution:**
Change the receiver of the MarshalJSON method to a pointer receiver:

Request for Clarification:
Correct me if I'm wrong here, but if there is a specific reason for this code to exist in its current form (e.g., intentional use of a value receiver for immutability or other design considerations), I would appreciate clarification. Otherwise, this change would improve consistency, performance, and maintainability.

hemanthfr:
audio-content-reference fix 
#641

### DIFF
--- a/mcp/content.go
+++ b/mcp/content.go
@@ -88,7 +88,7 @@ type AudioContent struct {
 	Annotations *Annotations
 }
 
-func (c AudioContent) MarshalJSON() ([]byte, error) {
+func (c *AudioContent) MarshalJSON() ([]byte, error) {
 	// Custom wire format to ensure required fields are always included, even when empty.
 	data := c.Data
 	if data == nil {


findleyr:
Unfortunately we can't accept breaking changes until the next major version of the SDK.

This mistake has been noted in docs/rough_edges.md

Issue #617, opened 2025-11-07
SEP-1034: elicitation default values.


findleyr:
Our elicitation schemas are intentionally decoupled from their JSON-RPC implementation. However, when validating the schema, we should apply defaults.

Unlike tool handling, I'm not sure we should inject these defaults into the Content map.

This implements [SEP-1034](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1034).

findleyr:
If someone wanted to help move this along this week, here are TODOs:

- Investigate how other SDKs treat this.
- Use `jsonschema.Resolved.ApplyDefaults` when validating the schema
- Add tests.

markus-kusano:
Open question: Do we need to handle elicitation defaults on the server side, or just on the client, or both? I think we need to just do it in 1 location (either on the client or the server). But! If we are expected to talk to clients/server that are not using the Go SDK then we might want to make sure to use the defaults in both, rather than assuming we are talking to a Go MCP SDK client and server

Why? Elicitation requests are a request for input from the server to the client. The schema on the elicitation request from the server to the client contains the expected format of the response, as well as the defaults. If the client chooses to not include anything in the response for fields with defaults then the default value should be used. The default can be applied eon the client side (before sending the response back), or on the server side (before the values are used).

The TL;DR flow for this is:

1. Server sends elicitation request with schema with optional defaults
2. Client decides how to respond to the request, potentially not including values for some parts of the schema
3. [Option A] The client sets the fields in the response to their default value, sends the response back to the server
4. [Option B] The server sets fields on the response to their default value after getting the response

We only need one of [Option A] or [Option B] since they are redundant. OTOH, we might want to harden the server to also use the defaults, e.g., this would make the system work if we are talking to a client that is not using the Go implementation

findleyr:
@markus-kusano I'm hesitant to do it on the server side on the basis of typescript _not_ doing this, and preserving information: the handler should see exactly which information the client submitted.

A middle ground might be to simply apply defaults on the server side before validation. That way, if the default is "6", with min=4 and max=8, and the client doesn't submit the field, the content is still considered valid. WDYT?

findleyr:
BTW, I actually noticed that we weren't doing any validation server side, and wondered if this is just the practice. It's not: the typescript SDK validates on both caller and callee.

I filed https://github.com/modelcontextprotocol/go-sdk/issues/648. Fixing this for eliciation will be one half of that.

PR #644, opened 2025-11-19
Elicit defaults

markus-kusano:
https://github.com/modelcontextprotocol/go-sdk/issues/617

### DIFF
--- a/mcp/server.go
+++ b/mcp/server.go
@@ -1024,20 +1024,14 @@ func (ss *ServerSession) Elicit(ctx context.Context, params *ElicitParams) (*Eli
 	if params.RequestedSchema == nil {
 		return res, nil
 	}
-	val := reflect.ValueOf(params.RequestedSchema)
-	switch val.Kind() {
-	case reflect.Chan, reflect.Func, reflect.Map, reflect.Pointer, reflect.UnsafePointer, reflect.Interface, reflect.Slice:
-		if val.IsNil() {
-			return res, nil
-		}
-	default:
-		return res, nil
-	}
-
 	schema, err := validateElicitSchema(params.RequestedSchema)
 	if err != nil {
 		return nil, err
 	}
+	if schema == nil {
+		return res, nil
+	}
+
 	resolved, err := schema.Resolve(nil)
 	if err != nil {
 		fmt.Printf("  resolve err: %s", err)


mcp/mcp_test.go
```diff
@@ -1459,6 +1461,106 @@ func TestElicitationCapabilityDeclaration(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer cs.Close()
+
+	schemas := []struct {
+		name          string
+		schema        *jsonschema.Schema
+		expectedKey   string
+		expectedValue any
```
findleyr: Maybe just have `want map[string]any`, and use cmp.Diff to compare with the content we got?

That allows for more complex tests, and (I think) simplifies the test body.

mcp/mcp_test.go
```diff
@@ -1459,6 +1461,106 @@ func TestElicitationCapabilityDeclaration(t *testing.T) {
...
+			schema: &jsonschema.Schema{
+				Type: "object",
+				Properties: map[string]*jsonschema.Schema{
+					"key": {Type: "string", Enum: []any{"one", "two"}, Default: json.RawMessage("\"one\"")},
+				},
+			},
+			expectedKey:   "key",
+			expectedValue: "one",
+		},
+		// TODO: should we have a test with number (float) types?
```
findleyr: Yes please!

mcp/mcp_test.go
```diff
@@ -1459,6 +1461,106 @@ func TestElicitationCapabilityDeclaration(t *testing.T) {
...
+			return &ElicitResult{Action: "accept", Content: map[string]any{"enabled": "value"}}, nil
+		},
+	})
+	cs, err := c.Connect(ctx, ct, nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer cs.Close()
+
+	schemas := []struct {
```
findleyr: 'tests'? More than just schemas.

mcp/mcp_test.go
```diff
@@ -1459,6 +1461,106 @@ func TestElicitationCapabilityDeclaration(t *testing.T) {
...
+				Type: "object",
+				Properties: map[string]*jsonschema.Schema{
+					"key": {Type: "string", Enum: []any{"one", "two"}, Default: json.RawMessage("\"one\"")},
+				},
+			},
+			expectedKey:   "key",
+			expectedValue: "one",
+		},
+		// TODO: should we have a test with number (float) types?
+	}
```
findleyr, line 1569: Can we have tests with malformed schemas? 'wantErr', and all that?

markus-kusano: Invalid schemas  are tested on line 1137 above, were there some other testcases you had in mind?

mcp/mcp_test.go
markus-kusano: good point, renamed to testcases

mcp/mcp_test.go
markus-kusano: Done

mcp/mcp_test.go
markus-kusano: Great idea! I liked that this simplified things and even handled the float comparison out of the box

markus-kusano:
PTAL

mcp/mcp_test.go
findleyr: Ah, that was largely beneath the fold.

Maybe add a case or two more related to the new default values? How about an invalid number and enum?

markus-kusano: nice catch, i meant to add those. thanks

markus-kusano:
PTAL, I updated the server side to both apply defaults and do validation. I also removed the use of reflect as discussed offline (thanks for the help!)

mcp/client.go
```diff
@@ -341,6 +344,9 @@ func validateElicitSchema(wireSchema any) (*jsonschema.Schema, error) {
 	if err := remarshal(wireSchema, &schema); err != nil {
 		return nil, err
 	}
+	if schema == nil {
```
findleyr, line 347: This is possible? wireShema is non-nil, so wouldn't a nil schema be an error here?

findleyr (approved):

mcp/client.go
markus-kusano: TL;DR: I think we either need a reflect.IsNil() call on `wireSchema` to determine if it is really nil, or we need to check if the result of remarshall is nil when err is != nil (which feels gross).

In more detail:

In the server, as I understand, the wireschema is `**jsonschema`. we pass in a `nil` value for this which gets wrapped in another pointer. so, in the server, `wireschema` is non-nil but `*wireschema` is nil. This means we end up passing a non-nil wireschema here, but it is unmarshalling a `nil` jsonschema.

This means that when we get back here since we passed `&schema` into remarshall now `schema` is nil.

It might make remarshall more robust to have a reflect isNil call and return an err, but then we'd need to add in the reflect call.

findleyr: On the server the schema should be a *jsonschema.Schema, not a **jsonschema.Schema. Where are you seeing this?

Let's merge this as-is, and I will take a look.

findleyr: BTW, I looked into it: the test uses a typed nil (one of the downsides of having `any` for the schema).

I think the code, as written, is fine.

mcp/server.go
```diff
@@ -1015,7 +1015,37 @@ func (ss *ServerSession) Elicit(ctx context.Context, params *ElicitParams) (*Eli
...
+	if err != nil {
+		return nil, err
+	}
+	if schema == nil {
+		return res, nil
+	}
+
+	resolved, err := schema.Resolve(nil)
+	if err != nil {
+		fmt.Printf("  resolve err: %s", err)
```
liut, line 1037: Is this okay?

Issue #623, opened 2025-11-07
SEP-1036: support URL Mode Elicitation


findleyr:
Implement support for the pending [SEP-1036](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1036), adding a URL mode for secure out-of-band interactions.

PR #646, opened 2025-11-20
mcp: support url mode elicitation

findleyr:
Add initial support for URL mode elicitation, by introducing the new
types, as well as both server and client-side validation.

A follow-up CL will add support for ErrURLElicitationRequired.

For #623

### DIFF
--- a/mcp/client.go
+++ b/mcp/client.go
@@ -66,6 +66,11 @@ type ClientOptions struct {
 	// Setting ElicitationHandler to a non-nil value causes the client to
 	// advertise the elicitation capability.
 	ElicitationHandler func(context.Context, *ElicitRequest) (*ElicitResult, error)
+	// ElicitationModes specifies the elicitation modes supported by the client.
+	// If ElicitationHandler is set and ElicitationModes is empty, it defaults to ["form"].
+	ElicitationModes []string
+	// ElicitationCompleteHandler handles incoming notifications for notifications/elicitation/complete.
+	ElicitationCompleteHandler func(context.Context, *ElicitationCompleteNotificationRequest)
 	// Handlers for notifications from the server.
 	ToolListChangedHandler      func(context.Context, *ToolListChangedRequest)
 	PromptListChangedHandler    func(context.Context, *PromptListChangedRequest)
@@ -123,6 +128,15 @@ func (c *Client) capabilities() *ClientCapabilities {
 	}
 	if c.opts.ElicitationHandler != nil {
 		caps.Elicitation = &ElicitationCapabilities{}
+		modes := c.opts.ElicitationModes
+		if len(modes) == 0 || slices.Contains(modes, "form") {
+			// Technically, the empty ElicitationCapabilities value is equivalent to
+			// {"form":{}} for backward compatibility, but we explicitly set the form
+			// capability.
+			caps.Elicitation.Form = &FormElicitationCapabilities{}
+		} else if slices.Contains(modes, "url") {
+			caps.Elicitation.URL = &URLElicitationCapabilities{}
+		}
 	}
 	return caps
 }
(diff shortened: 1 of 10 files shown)


mcp/elicitation_test.go
```diff
@@ -0,0 +1,193 @@
...
+			if err != nil {
+				t.Fatal(err)
+			}
+			defer cs.Close()
+
+			result, err := ss.Elicit(ctx, tc.params)
+
+			if tc.wantErrMsg != "" {
+				if err == nil {
+					t.Fatalf("expected an error containing %q, got nil", tc.wantErrMsg)
```
markus-kusano: should we add `tc.name` into the test failure messages to make it obvious which test is failing (here and below)?

mcp/elicitation_test.go
```diff
@@ -0,0 +1,193 @@
...
+			name: "decline",
+			handler: func(ctx context.Context, req *ElicitRequest) (*ElicitResult, error) {
+				return &ElicitResult{Action: "decline"}, nil
+			},
+			params: &ElicitParams{
+				Mode:    "url",
+				Message: "Please provide information via URL",
+				URL:     "https://example.com/form",
+			},
+			wantResultAction: "decline",
```
markus-kusano, line 53: I think I'm missing something, what's the difference between the "success" testcase above and the "decline" testcase here? I see in the spec that the Action field has specific meanings (accept, decline, cancel), but does our code ever interpret these? I see this ``json:"action"`` magic, maybe that's why we need both of these.

mcp/client.go
```diff
@@ -297,37 +311,51 @@ func (c *Client) createMessage(ctx context.Context, req *CreateMessageRequest) (
...
+			if err != nil {
+				return nil, jsonrpc2.NewError(codeInvalidParams, fmt.Sprintf("failed to resolve requested schema: %v", err))
+			}
+			if err := resolved.Validate(res.Content); err != nil {
+				return nil, jsonrpc2.NewError(codeInvalidParams, fmt.Sprintf("elicitation result content does not match requested schema: %v", err))
+			}
+		}
+		return res, nil
+	case "url":
+		if req.Params.RequestedSchema != nil {
```
markus-kusano, line 352: double checking: do we need to add any documentation about what fields are valid for certain requests? (i wasn't really sure where this documentation should go, though, which is another sign this might not be needed. i wanted to double check)

markus-kusano (approved):
mostly just nits

mcp/elicitation_test.go
findleyr: I'm not sure I follow. The MCP server explicitly calls `ServerSession.Elicit`, and gets the result, so the server author can choose to handle this action however they choose. The SDK doesn't care: declining an elicitation is still a successful RPC.

If you're asking what this test is exercising, you're right that it's essentially the same as the "success" case. (AI wrote the test, so the friction for _writing_ it was trivial. The problem with that is that it puts burden on the review process).

mcp/elicitation_test.go
findleyr: This is not necessary for subtests: the name will be in the test name (this is one of the reasons to use subtests!)

However, some of the test errors were unidiomatic, which I've fixed.

mcp/client.go
findleyr: I'd documented this in the field documentation for the type. I think that's a good place for it; do you think that documentation should be improved?

mcp/elicitation_test.go
markus-kusano: Got it! I was wondering what the difference was in the test, thanks for clarifying. I was worried in the future I'd look at these 2 testcases and be afraid to delete one :)

mcp/client.go
markus-kusano: LGTM, thakns, I missed that

Issue #624, opened 2025-11-07
SEP-1613: Establish JSON Schema 2020-12 as default


findleyr:
We already use 2020-12 as the default schema, so this issue may be a no-op, or may require some additional documentation.

This is [SEP-1613](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1613), which is accepted but not yet merged.

markus-kusano:
Summarizing the discussion on https://github.com/modelcontextprotocol/go-sdk/pull/647

We support JSON-2020-12, once https://github.com/google/jsonschema-go/pull/50 is finished we will also support the commonly used 07 schema.

findleyr:
I noticed there is a conformance test for this SEP:
https://github.com/modelcontextprotocol/conformance/blob/main/src/scenarios/server/json-schema-2020-12.ts

I think we'll pass, but I'll confirm once I get to #650.

PR #647, opened 2025-11-20
Enforce JSON-2020-12 as the default

markus-kusano:
Fixes https://github.com/modelcontextprotocol/go-sdk/issues/624 and implements the spec update referenced in that issue.

Please see the commit message for more details. We already were only supporting JSON-2020-12 (or, pedantically, anything that could unmarshall into the jsonschema.Schema object). 

There could be a chance this changes adds in errors in the user's code. This would be for the case where they are setting the schema object to something other than a JSON-2020-12 but we were able to unmarshall it successfully.  We could update the code to handle this case but we would see worse error messages.

### DIFF
--- a/mcp/client.go
+++ b/mcp/client.go
@@ -42,16 +42,16 @@ func NewClient(impl *Implementation, opts *ClientOptions) *Client {
 	if impl == nil {
 		panic("nil Implementation")
 	}
-	cc := &Client{
+	c := &Client{
 		impl:                    impl,
 		roots:                   newFeatureSet(func(r *Root) string { return r.URI }),
 		sendingMethodHandler_:   defaultSendingMethodHandler[*ClientSession],
 		receivingMethodHandler_: defaultReceivingMethodHandler[*ClientSession],
 	}
 	if opts != nil {
-		cc.opts = *opts
+		c.opts = *opts
 	}
-	return cc
+	return c
 }
 
 // ClientOptions configures the behavior of the client.


mcp/client.go
```diff
@@ -363,6 +366,31 @@ func validateElicitSchema(wireSchema any) (*jsonschema.Schema, error) {
 	return schema, nil
 }
 
+func checkSchemaVersion(wireSchema any) error {
```
findleyr, line 369: I'm not sure this function is actually necessary, since our schema validation already assumes 2020-12.

Also, I believe the SEP only applied to tools. Elicitation uses an even more restricted subset of json schema.

markus-kusano: Whoops I missed that the tools are also using schemas. I think [the SEP](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1613) refers to all schemas used anywhere in the MCP, not just tools.

Agreed, this function might not be necessesary. 

I'm struggling with this change since we might not need to do anything :) 

We might be able to add a test that sees the expected failure if the schema is something else other than JSON 2020-12 (jsonschema.Schema). This test doesn't add much value, though, since we only use jsonschema.Schema in our codebase, though.

I'm more in favor of leaving this as-is and calling the issue as fixed. Please let me know what you think!

findleyr: I think "do nothing" is actually a valid action here. Maybe we just add tests based on the tests of the typescript change?

markus-kusano: There's a lot of commits attached to the still-in-progress typescipt change. From what I can see, they are planning to use a library to handle the JSON schemas, similar to us. I don't see a ton of tests with relevant testcases we could borrow, probably because they are deferring to the library to handle the schema (similar as us).

I poked around to see what happens to our code when we use a non-`jsonschema.Schema` value.  It crashes when it attempts to unmarshall it into the actual `jsonschema.Schema` value. E.g., addint a new tool with an invalid InputSchema causes an error to be returned here: https://github.com/modelcontextprotocol/go-sdk/blob/1943780255521f98ef57f083ae95b92fe727ee96/mcp/server.go#L250

We could add some tests to send invalid values in fields like InputSchema and catch failures. This would clearly document that we are not supporting anything other than `Schema`, but doesn't add a ton of value other than that. Let me know what you think, I can start adding tests along these lines.

It could be useful to have very descriptive error messages here. We are violating the spec by crashing on non-JSON-2020-12 schemas. My initial commit in this PR attempted to make it more obvious when we are crashing due to invalid schema versions.

But, overall it seems like we can just mark this as closed

findleyr: Ack, closing SGTM.

Once https://github.com/google/jsonschema-go/pull/50 lands, we will also support the draft-07 schema.

markus-kusano:
See discussion above and on the issue, closing this PR and marking the associated bug as fixed

PR #659, opened 2025-11-26
mcp: introduce options for AddTool

### LINKED ISSUE #658: Feature Request: Support custom jsonschema type mappings
Hello! Thanks very much for this library 🙌 

I'm building an MCP server where I have 4 Go structs like this:

https://github.com/sumo-mcp/sumoapi-go/blob/e15596d6feabadc0d805064ddd8789d4dcb6a441/match.go#L31-L82

This Go struct represents a string ID with a well-defined structure, where the struct fields are extracted from specific parts of the string format. By implementing `MarshalJSON` and `UnmarshalJSON` this ID can be represented in Go as a struct, and as a string in the serialized JSON format. From the jsonschema perspective, the type has to be string, but the default behavior of jsonschema does not detect this. But jsonschema allows, however, to pass a mapping from `reflect.Type` to `jsonschema.Schema` to customize the jsonschema type for a Go type, like they do for Go's standard library types here:

https://github.com/google/jsonschema-go/blob/98a387e3b9757466bedf13f41d8504784c335018/jsonschema/infer.go#L345-L359

I'd like to request the ability to pass an option to `mcp.AddTool()` with a custom `jsonschema.ForOptions` where the `TypeSchemas map[reflect.Type]*Schema` field can be specified.

I have already implemented and tested the change in a fork and am opening a PR right now 🫡

matheuscscp:
This change introduces a non-breaking API for passing options to `mcp.AddTool()`. The first option introduced is the `jsonschema.ForOptions`, which allows jsonschema to use a specified mapping of Go types to schemas defined by the caller, which are used instead of inference via reflection. This is very useful for example for Go structs implementing the `MarshalJSON` and `UnmarshalJSON` interfaces in order to represent the struct as a string. For example, imagine a MonthID struct containing Year and Month as integers, and whose serialization format implemented by `MarshalJSON` is `YYYYMM`. From the jsonschema perspective the type must be string, but the representation in Go can be rich through a struct.

Fixes #658

### DIFF
--- a/mcp/server.go
+++ b/mcp/server.go
@@ -239,7 +239,7 @@ func (s *Server) AddTool(t *Tool, h ToolHandler) {
 		func() bool { s.tools.add(st); return true })
 }
 
-func toolForErr[In, Out any](t *Tool, h ToolHandlerFor[In, Out]) (*Tool, ToolHandler, error) {
+func toolForErr[In, Out any](t *Tool, h ToolHandlerFor[In, Out], schemaOpts *jsonschema.ForOptions) (*Tool, ToolHandler, error) {
 	tt := *t
 
 	// Special handling for an "any" input: treat as an empty object.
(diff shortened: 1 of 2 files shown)


findleyr:
As I mentioned on the issue, this is a breaking change (a minor one, admittedly, but still a breaking change). Therefore, we'd have to only consider it for v2, which _will_ happen eventually.

findleyr:
Have to decline this for now. The issue tracks reconsidering tool options in v2.

PR #663, opened 2025-11-27
mcp: support SSE polling and server disconnects (SEP-1699)

### LINKED ISSUE #630: SEP-1699: support SSE polling
If [SEP-1699](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1699) lands for the 2025-11-25 version of the spec, we must implement it.

(This one should fall to me).

findleyr:
This CL implements support for SEP-1699, adding the following new API:

- Event.Retry corresponds to the 'retry' field of a server sent event. This is used to signal client retry behavior.
- RequestExtra.CloseStream is set by the streamable transport to allow closing a stream.

The streamable server transport is updated to set CloseStream for calls, allowing server operations or middleware to close the underlying stream. If the configured 'reconnectAfter' delay is set, a 'retry' message is sent prior to terminating the stream.

The streamable client transport is updated to read the configured delay.

Fixes #630

### DIFF
--- a/mcp/streamable_test.go
+++ b/mcp/streamable_test.go
@@ -568,11 +568,9 @@ func TestStreamableServerDisconnect(t *testing.T) {
 
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
-			// memory event store to support replayability.
-			// Then implement the new SEP, and assert that the tool call succeeds.
 			notifications := make(chan string, 2)
 			handler := NewStreamableHTTPHandler(func(*http.Request) *Server { return server }, &StreamableHTTPOptions{
-				EventStore: NewMemoryEventStore(nil),
+				EventStore: NewMemoryEventStore(nil), // support replayability
 			})
 			httpServer := httptest.NewServer(mustNotPanic(t, handler))
 			defer httpServer.Close()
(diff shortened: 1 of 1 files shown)


mcp/streamable_test.go
```diff
@@ -517,6 +517,94 @@ func testClientReplay(t *testing.T, test clientReplayTest) {
...
+				_, err := cs.GetPrompt(ctx, &GetPromptParams{Name: "disconnect"})
+				return err
+			},
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			// memory event store to support replayability.
+			// Then implement the new SEP, and assert that the tool call succeeds.
```
markus-kusano: it might be confusing for the reader to see "the new SEP" in the code-comment without knowing which SEP it is. if we want to add more context about what we're testing maybe we could link to our github issue or something similar? this could go on line 522 with comments about the high-level structure of the test

mcp/streamable_test.go
```diff
@@ -517,6 +517,94 @@ func testClientReplay(t *testing.T, test clientReplayTest) {
 	}
 }
 
+func TestStreamableServerDisconnect(t *testing.T) {
+	server := NewServer(testImpl, nil)
+
+	// Test that client replayability allows the server to terminate incoming
+	// requests immediately, and have the client replay them.
+
+	// testStream exercises stream resumption by interleaving stream termination
```
markus-kusano, line 526: do we need to check/is it possible to check that the stream was closed and re-opened? from my naive understanding, the test checks that we get `msg1` and `msg2` as progress notifications. wouldn't we see the same progress notifications if the close events were just ignored, and the stream was connected the entire time?

mcp/streamable_test.go
findleyr: Whoops, sorry this comment was actually a partial artifact from a rebase--sorry!

mcp/streamable_test.go
findleyr: Good point. I think as an integration test, we should be too invasive here. However, I added an assertion to TestStreamableServerTransport that the client request _doesn't_ receive the "result" event. In combination, perhaps these two tests provide enough coverage? WDYT?

markus-kusano: that's perfect! thanks

PR #671, opened 2025-12-01
mcp: debounce server change notifications

### LINKED ISSUE #649: batch notifications
The server should batch notifications to the client, to avoid a flurry of messages when many tools, etc. are added at once.

This actually happens in practice: see https://github.com/modelcontextprotocol/go-sdk/issues/73#issuecomment-3558960658.

All AddTool notifications are the same (the params are empty), so we can count them and send one when the count reaches a threshold.
The same is true for all the other add and remove notifications.

This should happen in Server.changeAndNotify.

Also, we should asynchronously and periodically send notifications for any non-zero counts. The period should be fairly short (50 ms?) So that single calls to AddTool, etc. are notified promptly.

jba:
Debounce notifications for changes in feature sets (adding or removing tools, prompts, resources, or resource templates).

Delay sending the notifications until some time has passed.

This PR handles server-side change notifications only. We could do the same for client-side notifications (e.g. roots changed), but at the moment we don't need to.

We get errors about undelivered messages, because the delayed notifications are sent before the standalone GET happens. See #684. We'll deal with them in another PR (if at all).

Fixes #649.

### DIFF
--- a/mcp/mcp_test.go
+++ b/mcp/mcp_test.go
@@ -1673,14 +1673,15 @@ func TestAddTool_DuplicateNoPanicAndNoDuplicate(t *testing.T) {
 }
 
 func TestSynchronousNotifications(t *testing.T) {
-	var toolsChanged atomic.Bool
+	var toolsChanged atomic.Int32
 	clientOpts := &ClientOptions{
 		ToolListChangedHandler: func(ctx context.Context, req *ToolListChangedRequest) {
-			toolsChanged.Store(true)
+			toolsChanged.Add(1)
 		},
 		CreateMessageHandler: func(ctx context.Context, req *CreateMessageRequest) (*CreateMessageResult, error) {
-			if !toolsChanged.Load() {
-				return nil, fmt.Errorf("didn't get a tools changed notification")
+			// See the comment after "from server" below.
+			if n := toolsChanged.Load(); n != 1 {
+				return nil, fmt.Errorf("got %d tools-changed notification, wanted 1", n)
 			}
 			// TODO(rfindley): investigate the error returned from this test if
 			// CreateMessageResult is new(CreateMessageResult): it's a mysterious
@@ -1720,7 +1721,12 @@ func TestSynchronousNotifications(t *testing.T) {
 	})
 
 	t.Run("from server", func(t *testing.T) {
-		// TODO: test that multiple changes result in a single notification.
+		// Despite all this tool-changed activity, we expect only one notification.
+		for range 10 {
+			server.RemoveTools("tool")
+			addTool(server)
+		}
+
 		time.Sleep(notificationDelay * 2) // Wait for delayed notification.
 		if _, err := ss.CreateMessage(context.Background(), new(CreateMessageParams)); err != nil {
 			t.Errorf("CreateMessage failed: %v", err)


mcp/server.go
```diff
@@ -497,18 +493,57 @@ func (s *Server) complete(ctx context.Context, req *CompleteRequest) (*CompleteR
...
+// so a single struct can be reused.
+var changeNotificationParams = map[string]Params{
+	notificationToolListChanged:     &ToolListChangedParams{},
+	notificationPromptListChanged:   &PromptListChangedParams{},
+	notificationResourceListChanged: &ResourceListChangedParams{},
+}
+
+// The maximum number of change notifications of a particular type (e.g. tools-changed)
+// that can be pending.
+const maxPendingNotifications = 10
```
findleyr: Why include this? Shouldn't a debouncing delay be sufficient? Put differently, isn't the point of this CL to reduce notification noise for huge sets of features? Why are we capping it at a 10x reduction?

mcp/server.go
```diff
@@ -497,18 +493,57 @@ func (s *Server) complete(ctx context.Context, req *CompleteRequest) (*CompleteR
...
+
+// How long to wait before sending a change notification.
+var notificationDelay = 50 * time.Millisecond
```
findleyr: const, unless it has to be a var

mcp/server.go
```diff
@@ -497,18 +493,57 @@ func (s *Server) complete(ctx context.Context, req *CompleteRequest) (*CompleteR
...
-		sessions = slices.Clone(s.sessions)
+		pending := s.pendingNotifications[notification]
+		if pending >= maxPendingNotifications {
+			send = true
+			pending = 0
+			// Make a local copy of the session list so we can use it without holding the lock.
+			sessions = slices.Clone(s.sessions)
+		} else {
+			pending++
+			if pending == 1 {
```
findleyr: Typically this sort of thing would be achieved with debouncing, i.e. reset the timer after each new notification, so that the actual notification is sent only after a period of quiescence.

findleyr:
> BUG: notifications sent after a connection closes result in an error message printed out. I'm not sure how to avoid that. It would be racy to check for a closed connection just before sending, though it would reduce the number of messages considerably.

I think there will always be a race, but shouldn't we have deleted the session after it's closed, making this race very unlikely? I.e. it should no longer be present in the sessions set?

mcp/server.go
```diff
@@ -497,18 +493,57 @@ func (s *Server) complete(ctx context.Context, req *CompleteRequest) (*CompleteR
...
+				time.AfterFunc(notificationDelay, func() { s.sendNotification(notification) })
```
markus-kusano: is the code more readable if we re-named `s.sendNotification` something closer to `notifySessions`, or vice versa,  e.g., `getAndNotifySessions`?

context: reading the code from ~lines 519 to 535 in isolation (without knowing what `sendNotification` is), i was initially wondering why the time-delay-reached (the else branch) and the max-notifications-reached (the if branch) seemed to have different logic: the time-delay branch was sending notifications, but the max-pending branch was notifying sessions. when i load in the implementation of `s.sendNoticication` into my brain i can see that it is re-reading the `session`s in case they've changed after the time-delay.

jba: Changed to notifySessions.

mcp/server.go

jba:
> > BUG: notifications sent after a connection closes result in an error message printed out. I'm not sure how to avoid that. It would be racy to check for a closed connection just before sending, though it would reduce the number of messages considerably.
> 
> I think there will always be a race, but shouldn't we have deleted the session after it's closed, making this race very unlikely? I.e. it should no longer be present in the sessions set?

Yes, it's not sessions closing.
Looks like this is related: https://github.com/modelcontextprotocol/go-sdk/discussions/669.

jba:
> > > BUG: notifications sent after a connection closes result in an error message printed out. I'm not sure how to avoid that. It would be racy to check for a closed connection just before sending, though it would reduce the number of messages considerably.
> > 
> > 
> > I think there will always be a race, but shouldn't we have deleted the session after it's closed, making this race very unlikely? I.e. it should no longer be present in the sessions set?
> 
> Yes, it's not sessions closing. Looks like this is related: #669.

The problem is that when a POST is done, the deliver function is nilled out (https://github.com/modelcontextprotocol/go-sdk/blob/main/mcp/streamable.go#L1092), so there is nowhere to send the notification. Perhaps changed notifications should always go to the hanging GET.

As #669 points out, this can happen for progress notifications too, but in that case dropping the notification makes sense, because that notification is useless.

findleyr:
> > > > BUG: notifications sent after a connection closes result in an error message printed out. I'm not sure how to avoid that. It would be racy to check for a closed connection just before sending, though it would reduce the number of messages considerably.
> > > 
> > > 
> > > I think there will always be a race, but shouldn't we have deleted the session after it's closed, making this race very unlikely? I.e. it should no longer be present in the sessions set?
> > 
> > 
> > Yes, it's not sessions closing. Looks like this is related: #669.
> 
> The problem is that when a POST is done, the deliver function is nilled out (https://github.com/modelcontextprotocol/go-sdk/blob/main/mcp/streamable.go#L1092), so there is nowhere to send the notification. Perhaps changed notifications should always go to the hanging GET.
> 
> As #669 points out, this can happen for progress notifications too, but in that case dropping the notification makes sense, because that notification is useless.

Ack, thanks for investigating. I think this is working as intended, then. If the message is made with the request context, then it's invalid after the request completes. To send a background notification, use `context.Background()` (if this isn't apparent from our documentation, we should fix that).

mcp/server.go
```diff
@@ -497,18 +493,43 @@ func (s *Server) complete(ctx context.Context, req *CompleteRequest) (*CompleteR
...
+func (s *Server) changeAndNotify(notification string, change func() bool) {
 	s.mu.Lock()
+	defer s.mu.Unlock()
 	if change() {
-		sessions = slices.Clone(s.sessions)
+		// Stop the outstanding delayed call, if any.
+		if t := s.pendingNotifications[notification]; t != nil {
+			t.Stop()
+		}
+		//
```
findleyr, line 520: Cruft?

mcp/shared.go
```diff
@@ -349,6 +349,8 @@ func notifySessions[S Session, P Params](sessions []S, method string, params P)
 	if sessions == nil {
 		return
 	}
+	// Notify with the background context, so the messages are sent on the
+	// standalone stream.
 	// TODO: make this timeout configurable, or call handleNotify asynchronously.
```
findleyr, line 354: Is this TODO now obsolete? Are all notifications asynchronous?

mcp/server.go
```diff
@@ -497,18 +493,43 @@ func (s *Server) complete(ctx context.Context, req *CompleteRequest) (*CompleteR
...
+// This debounces change notifications: a single notification is sent after
+// multiple changes occur in close proximity.
+func (s *Server) changeAndNotify(notification string, change func() bool) {
 	s.mu.Lock()
+	defer s.mu.Unlock()
 	if change() {
-		sessions = slices.Clone(s.sessions)
+		// Stop the outstanding delayed call, if any.
+		if t := s.pendingNotifications[notification]; t != nil {
+			t.Stop()
```
findleyr, line 518: I think we can just call t.Reset() and probably avoid some allocation, no?

findleyr:
@markus-kusano is going to take this over.

mcp/server.go
findleyr: Done in https://github.com/modelcontextprotocol/go-sdk/pull/717

mcp/server.go
findleyr: Done in https://github.com/modelcontextprotocol/go-sdk/pull/717.

mcp/shared.go
findleyr: It's still relevant for the client.

findleyr (commented):
Addressed comments in https://github.com/modelcontextprotocol/go-sdk/pull/717. Still your commit.

findleyr:
Done in #717

PR #676, opened 2025-12-02
mcp/mcp_test: tests for elicit content validation

### LINKED ISSUE #625: SEP-1330: Elicitation Enum Schema Improvements
Implement support for [SEP-1330](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1330), when it lands.

I actually think we were erroneously not supporting `enumNames` in schema validation, so this may be a no-op. We should consider supporting `enumNames` in a backward-compatible way.

markus-kusano:
This adds a simple schema and checks that the content returned by an elicitation handler is validated against the schema. Previously, we had no test coverage checking if the content was validated (i.e., we would see no test failures if we removed content validation against the schema).

The bulk of the actual validation logic is handled by the jsonschema library so adding comprehensive tests covering different schema-features is redundant, we simply need to see that some validation is happening.

See https://github.com/modelcontextprotocol/go-sdk/issues/625 for more details on the motivation of this change.

Fixes #625

### DIFF
--- a/mcp/mcp_test.go
+++ b/mcp/mcp_test.go
@@ -1424,6 +1424,9 @@ func TestElicitContentValidation(t *testing.T) {
 	}
 	defer ss.Close()
 
+	// Set up a client that exercises valid/invalid elicitation: the returned
+	// Content from the handler ("potato") is validated against the schemas
+	// defined in the testcases below.
 	c := NewClient(testImpl, &ClientOptions{
 		ElicitationHandler: func(context.Context, *ElicitRequest) (*ElicitResult, error) {
 			return &ElicitResult{Action: "accept", Content: map[string]any{"test": "potato"}}, nil


mcp/mcp_test.go
```diff
@@ -1377,6 +1413,93 @@ func TestElicitationSchemaValidation(t *testing.T) {
...
+	ct, st := NewInMemoryTransports()
+
+	s := NewServer(testImpl, nil)
+	ss, err := s.Connect(ctx, st, nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer ss.Close()
+
+	c := NewClient(testImpl, &ClientOptions{
```
findleyr, line 1430: Some doc comments explaining how this test works could be helpful. For example, 

```
// Set up a client that exercises valid/invalid elicitation:
// "potato" may or may not be a valid elicitation result, depend on the test schemas
// below.
```

findleyr (approved):
LGTM, with one suggestion (I'll reapprove if needed).

mcp/mcp_test.go
markus-kusano: Great idea! I agree this connection in the test was vague, adding the comment is the right fix

findleyr (approved):

PR #677, opened 2025-12-03
mcp: better handling for streamable context cancellation

### LINKED ISSUE #662: Use call context when reconnecting; remove streamableClientConn.ctx
I just ran into a lifecycle issue in tests, which I finally traced back to our reuse of the connection context for streamable reconnection (in #133).

We should use the same context for reconnection that we use for connection, and should not need to store a context on the streamableClientConn.

Filing as a separate issue for tracking purposes.

findleyr:
After walking through our handling of streamable client context cancellation (due to encountering a shutdown deadlock), I think I've settled on a more coherent strategy for handling call cancellation:

- In our call handler, retire the request if the call exits due to cancellation: the caller will never see the actual result anyway.
- In connectSSE, use the actual request context (the same context used in Write) for the client request, so that it terminates when the context is cancelled. Thread through the initialization context for the standalone SSE request.

Also, a couple minor improvements:
- Use a detached context for the background context of the client connection. We want to preserve context values (see #513), but it is not right to cancel the connection after Connect has already returned, if the context times out.
- Don't use Last-Event-ID != "" as the signal for whether the connectSSE call is initial: if the standalone SSE stream disconnects without an event ID, we'll still reconnect it, and don't want to do so without a delay.

+ tests, updating the streamable client connection test harness to accomodate the new aspects being exercised.

Fixes #662

### DIFF
--- a/mcp/streamable.go
+++ b/mcp/streamable.go
@@ -1371,8 +1371,18 @@ func (t *StreamableClientTransport) Connect(ctx context.Context) (Connection, er
 	// This is crucial for cleanly shutting down the background SSE listener by
 	// cancelling its blocking network operations, which prevents hangs on exit.
 	//
-	// This context should be detached, to decouple the standalone SSE from the
-	// call to Connect.
+	// This context should be detached from the incoming context: the standalone
+	// SSE request should not break when the connection context is done.
+	//
+	// For example, consider that the user may want to wait at most 5s to connect
+	// to the server, and therefore uses a context with a 5s timeout when calling
+	// client.Connect. Let's suppose that Connect returns after 1s, and the user
+	// starts using the resulting session. If we didn't detach here, the session
+	// would break after 4s, when the background SSE stream is terminated.
+	//
+	// Instead, creating a cancellable context detached from the incoming context
+	// allows us to preserve context values (which may be necessary for auth
+	// middleware), yet only cancel the standalone stream when the connection is closed.
 	connCtx, cancel := context.WithCancel(xcontext.Detach(ctx))
 	conn := &streamableClientConn{
 		url:        t.Endpoint,
@@ -1684,8 +1694,10 @@ func (c *streamableClientConn) handleJSON(requestSummary string, resp *http.Resp
 func (c *streamableClientConn) handleSSE(ctx context.Context, requestSummary string, resp *http.Response, persistent bool, forCall *jsonrpc2.Request) {
 	for {
 		// Connection was successful. Continue the loop with the new response.
-		// TODO: we should set a reasonable limit on the number of times we'll try
-		// getting a response for a given request.
+		//
+		// TODO(#679): we should set a reasonable limit on the number of times
+		// we'll try getting a response for a given request, or enforce that we
+		// actually make progress.
 		//
 		// Eventually, if we don't get the response, we should stop trying and
 		// fail the request.
--- a/mcp/streamable_client_test.go
+++ b/mcp/streamable_client_test.go
@@ -429,6 +429,8 @@ func TestStreamableClientResumption_Cancelled(t *testing.T) {
 	// between the initial attempt and first reconnection.
 	//
 	// For easier tuning (and debugging), factor out the tick size.
+	//
+	// TODO(#680): experiment with instead using synctest.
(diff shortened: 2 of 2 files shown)


mcp/streamable.go
```diff
@@ -1364,7 +1370,10 @@ func (t *StreamableClientTransport) Connect(ctx context.Context) (Connection, er
 	// Create a new cancellable context that will manage the connection's lifecycle.
 	// This is crucial for cleanly shutting down the background SSE listener by
 	// cancelling its blocking network operations, which prevents hangs on exit.
-	connCtx, cancel := context.WithCancel(ctx)
+	//
+	// This context should be detached, to decouple the standalone SSE from the
+	// call to Connect.
+	connCtx, cancel := context.WithCancel(xcontext.Detach(ctx))
```
markus-kusano, line 1386: Why do we need to `xcontext.Detach()` in addition to `context.WithCancel()`? My naive understanding is that `xcontxt.Detach` forgets `Done` and `Err` on the context, and then `WithCancel` creates a new `Done`. From this perspective, forgetting `Done` initially seems redudant since it gets overwritten.

I think I'm missing something subtle though! Is it the `Err` part that can cause problems? My also naive understanding is that `Err` would only be used via `Done` but perhaps this is where my understanding breaks down

mcp/streamable.go
```diff
@@ -1668,15 +1681,15 @@ func (c *streamableClientConn) handleJSON(requestSummary string, resp *http.Resp
 //
 // If forCall is set, it is the call that initiated the stream, and the
 // stream is complete when we receive its response.
-func (c *streamableClientConn) handleSSE(requestSummary string, resp *http.Response, persistent bool, forCall *jsonrpc2.Request) {
+func (c *streamableClientConn) handleSSE(ctx context.Context, requestSummary string, resp *http.Response, persistent bool, forCall *jsonrpc2.Request) {
 	for {
 		// Connection was successful. Continue the loop with the new response.
 		// TODO: we should set a reasonable limit on the number of times we'll try
```
markus-kusano: How serious are the implications of not doing this (setting a reasonable retry limit?)

If this also cause a resource leak then it feels like it makes sense to create an explicit bug so we don't forget to fix this too

Related to comment on line 1839 below too

mcp/streamable_client_test.go
```diff
@@ -406,3 +416,140 @@ func TestStreamableClientUnresumableRequest(t *testing.T) {
 		t.Errorf("Connect: got error %v, want containing %q", err, msg)
 	}
 }
+
+func TestStreamableClientResumption_Cancelled(t *testing.T) {
+	// This test verifies that the resumed requests are closed when their context
+	// is cancelled (issue #662).
+
+	// This test (unfortunately) relies on timing, so may have false positives.
```
markus-kusano, line 424: Is it standard in Go to *not* mock out the clock in tests relying on time? In other languages I typically see this handled by dependency injecting a timer/clock interface. In this case, I think we'd want to inject on `timer.After`, this would allow the test logic below to explicit control the injected clock to move time forward, rather than implicitly synchronizing by `sleep`ing in the test and code-under-test (i.e., injecting the clock makes the test behave like message passing/notification based synchronization)

mcp/streamable.go
findleyr: Added a much larger comment--does that clarify?

mcp/streamable_client_test.go
findleyr: At this point, I wouldn't mock time because the synctest package exists for this purpose. Filed an issue to follow up on integrating synctest.

We unfortunately have several other tests that using timing. For now, if you're OK with it, I'd like to get this in: the failure mode is false positives, as we won't just get the 'cancellatoin' in the right place in the logic.

mcp/streamable.go
markus-kusano: thanks!

mcp/streamable_client_test.go
markus-kusano: Sounds good! Agreed we can get this in without using `synctest`. Most important is to have the bug open so we can keep track of what's left, thank you

PR #685, opened 2025-12-04
perf: add schema caching to avoid repeated reflection

SamMorrowDrums:
## Summary

This PR adds schema caching that dramatically reduces the cost of tool registration in stateless server patterns.

**Contributors:** Thanks to @omgitsads and @almaleksia for collaborating on and researching this solution!

## Problem

In stateless deployments like `github/github-mcp-server`, a new `*mcp.Server` is created for each incoming request. This means `AddTool` is called repeatedly for the same tools, causing:

1. **For typed handlers**: `jsonschema.ForType()` reflection is called every time
2. **For pre-defined schemas**: `schema.Resolve()` is called every time

## Solution

Add a `schemaCache` that stores:
- Schemas by `reflect.Type` (for auto-generated schemas from typed handlers)
- Resolved schemas by schema pointer (for pre-defined schemas)

The cache is:
- Concurrent-safe using `sync.Map`
- Unbounded (typical MCP servers have &lt;100 tools)
- Global (lives across server instances)

## Benchmark Results

```
BenchmarkAddToolTypedHandler-22          977080    1223 ns/op     1208 B/op    21 allocs/op
BenchmarkAddToolTypedHandlerNoCache-22    6764  161463 ns/op    39262 B/op  1072 allocs/op
```

| Metric | With Cache | Without Cache | Improvement |
|--------|------------|---------------|-------------|
| Time | 1,223 ns/op | 161,463 ns/op | **132x faster** |
| Allocations | 21 allocs | 1,072 allocs | **51x fewer** |
| Memory | 1,208 B/op | 39,262 B/op | **32x less** |

## Files Changed

- `mcp/schema_cache.go` - New cache implementation
- `mcp/server.go` - Modified `setSchema` to use cache
- `mcp/schema_cache_test.go` - Unit tests for caching behavior
- `mcp/schema_cache_benchmark_test.go` - Benchmarks

## Impact for Integrators

**Automatic** - no code changes required. Integrators using:
- Typed handlers (`AddTool[In, Out]`) → cache by type
- Pre-defined schemas (`Tool{InputSchema: schema}`) → cache by pointer

Both patterns benefit from caching after the first call.

---

## Real-World Performance Validation

The following benchmarks were conducted using `github/github-mcp-server` (a production MCP server with ~130 tools) to validate the performance impact in a real-world scenario.

### Test Environment
- **Server:** github-mcp-server (stateless HTTP deployment)
- **Tools registered per request:** ~130 tools
- **Test methodology:** Python benchmark client, 30 iterations (latency), 100 iterations (stress)
- **Date:** December 4, 2025

### Configurations Tested

| Configuration | Description |
|--------------|-------------|
| **main (mcp-go)** | Original implementation using mcp-go library |
| **go-sdk (no cache)** | go-sdk WITHOUT schema caching (broken) |
| **go-sdk (with cache)** | go-sdk WITH this PR&#39;s schema caching fix |

#### Operation: `initialize`

| Configuration | P50 | P99 | Status |
|--------------|-----|-----|--------|
| main (mcp-go) | 11.48ms | 14.00ms | ✅ Baseline |
| go-sdk (no cache) | 20.47ms | 25.30ms | 🔴 **+78% REGRESSION** |
| go-sdk (with cache) | 10.36ms | 14.33ms | ✅ **FIXED (-10%)** |

#### Operation: `tools/list`

| Configuration | P50 | P99 | Status |
|--------------|-----|-----|--------|
| main (mcp-go) | 13.46ms | 22.47ms | ✅ Baseline |
| go-sdk (no cache) | 22.98ms | 29.11ms | 🔴 **+71% REGRESSION** |
| go-sdk (with cache) | 14.15ms | 15.91ms | ✅ **FIXED (+5%)** |

#### Operation: `prompts/list`

| Configuration | P50 | P99 | Status |
|--------------|-----|-----|--------|
| main (mcp-go) | 11.56ms | 13.81ms | ✅ Baseline |
| go-sdk (no cache) | 21.00ms | 26.42ms | 🔴 **+82% REGRESSION** |
| go-sdk (with cache) | 10.77ms | 15.25ms | ✅ **FIXED (-7%)** |

### Stress Test Results (n=100)

| Configuration | P50 (initialize) | P50 (tools/list) | P50 (prompts/list) |
|--------------|------------------|------------------|-------------------|
| main (mcp-go) | 12.06ms | 14.30ms | 11.34ms |
| go-sdk (no cache) | 20.06ms | 23.44ms | 19.59ms |
| go-sdk (with cache) | **11.83ms** | **14.58ms** | **11.05ms** |

### Memory/Allocation Comparison (from pprof)

| Configuration | Total Allocations | Comparison |
|--------------|-------------------|------------|
| main (mcp-go) | 355.91 MB | Baseline |
| go-sdk (no cache) | **1208.70 MB** | 🔴 **3.4x MORE allocations** |

#### Top Allocation Sources - go-sdk WITHOUT cache (broken)

| Function | Size | % | Issue |
|----------|------|---|-------|
| `jsonschema.UnmarshalJSON` | 324.60 MB | 27% | 🚨 Schema re-parsing |
| `encoding/json.Unmarshal` | 341.10 MB | 28% | JSON deserialization |
| `jsonschema.resolve` | 219.51 MB | 18% | 🚨 Schema re-resolution |
| `jsonschema.MarshalJSON` | 92.52 MB | 8% | Schema JSON encoding |

### Root Cause Analysis

The `google/jsonschema-go` library regenerates JSON schemas on every request instead of caching them. In a server with ~130 tools, this causes:
- **70-80% latency regression** on all MCP operations
- **3.4x more memory allocations** per request
- **~70% of all allocations** are schema-related operations

### Key Findings

1. ✅ **REGRESSION CONFIRMED:** go-sdk without schema caching is 70-80% slower than mcp-go
2. ✅ **FIX VERIFIED:** Schema caching restores performance to baseline (or better)
3. ✅ **MEMORY IMPACT:** 3.4x reduction in allocations with caching
4. ✅ **PRODUCTION READY:** Fixed version performs equivalently to mcp-go baseline


findleyr:
Thank you! This makes sense in principle, and we hadn't considered the effect that lack of a cache would have on the stateless hosting model.

Will review expediently.

SamMorrowDrums:
@findleyr also don’t feel attacked by the AI generated performance test results report and PR description 😅.

I believe the principle of this change is exactly right but happy for you to take over implementation or re-implement if you like the idea too.

I have enabled maintainer edits.

mcp/schema_cache.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+// setBySchema caches a resolved schema by the original schema pointer.
+func (c *schemaCache) setBySchema(schema *jsonschema.Schema, resolved *jsonschema.Resolved) {
+	c.bySchema.Store(schema, resolved)
+}
+
+// resetForTesting clears the cache. Only for use in tests.
+func (c *schemaCache) resetForTesting() {
+	c.byType = sync.Map{}
```
jba: Call sync.Map.Clear?

mcp/schema_cache_benchmark_test.go
```diff
@@ -0,0 +1,161 @@
...
+import (
+	"context"
+	"testing"
+
+	"github.com/google/jsonschema-go/jsonschema"
+)
+
+// BenchmarkAddTool_TypedHandler measures performance of AddTool with typed handlers.
+// This simulates the stateless server pattern where new servers are created per request.
+func BenchmarkAddTool_TypedHandler(b *testing.B) {
```
jba: nit: remove underscores in benchmark and test names

mcp/server.go
```diff
@@ -274,22 +274,32 @@ func toolForErr[In, Out any](t *Tool, h ToolHandlerFor[In, Out]) (*Tool, ToolHan
...
 		if input != nil {
 			if err := json.Unmarshal(input, &in); err != nil {
 				return nil, fmt.Errorf("%w: %v", jsonrpc2.ErrInvalidParams, err)
 			}
 		}
 
+		// Apply defaults directly to the struct.
+		// This is more efficient than re-marshaling the map with defaults applied.
+		if inputResolved != nil {
+			if err := inputResolved.ApplyDefaults(&in); err != nil {
```
jba: The jsonschema ApplyDefaults function is just wrong when applied to structs. I'm going to remove that behavior, as I did for validating against structs.
Consider where the default is 1 and a zero is explicitly passed in the JSON. ApplyDefaults will overwrite that zero with a 1. That's wrong.

jba (changes requested):

mcp/server.go
jba: See https://github.com/google/jsonschema-go/pull/57.

jba:
As I detail in my comments, I don't believe the tool call implementation will work as it stands.

But you also haven't motivated it, as you have for the schema cache. Maybe it isn't necessary, especially given the latency of current LLMs.

It should have been in a second PR anyway, since it's an independent change.

SamMorrowDrums:
Thanks for the review @jba I'll definitely be happy to split. I agree.

Regarding:

> Maybe it isn't necessary, especially given the latency of current LLMs.

We do have sufficient motivation beyond user perception of latency to keep our fixed startup costs down, consider that this adds additional fixed cost latency to every single agent that runs our MCP by default (like say Copilot Coding Agent that runs on Actions runners), which adds up, and it's not per request as all clients need to get all the initial routes at startup and I don't believe all client SDKs make those requests in parallel so the effect is multiplicative to the number of server features offered.

I'll redo benchmarking with just the cache changes and see, but I suspect we will leave the shim in place and avoid the problem, gradually moving away from the tool for type method more permanently.

Appreciate the quick response.

SamMorrowDrums:
Thanks for the thorough review @jba! I've addressed all feedback:

1. ✅ **Removed the input validation optimization entirely** - Per your comment about ApplyDefaults being wrong for structs (zero vs missing values), I've dropped that commit. The PR now contains only the schema caching changes.

2. ✅ **Changed to `sync.Map.Clear()`** - Using the proper Clear() method instead of reassigning a new sync.Map.

3. ✅ **Removed underscores from benchmark names** - Now using `BenchmarkAddToolTypedHandler` instead of `BenchmarkAddTool_TypedHandler`.

The PR is now squashed to a single commit with just the schema caching optimization. Force pushed the cleaned branch.

SamMorrowDrums:
# Benchmark Results v2 - Schema Caching Performance Impact

I've run comprehensive benchmarks comparing 4 configurations to validate this fix. Here are the results:

## Configurations Tested

| # | Name | Description |
|---|------|-------------|
| 1 | **main** | Original MCP implementation using mcp-go library |
| 2 | **gosdk** | Go SDK WITHOUT schema caching (broken) |
| 3 | **gosdk_shimmed** | Go SDK with tool handler shim workaround |
| 4 | **gosdk_optimized** | Go SDK with this PR's schema caching fix |

---

## 🏆 Winners Summary

| Category | Winner | Value |
|----------|--------|-------|
| **initialize latency** | gosdk_optimized | 11.73ms |
| **tools/list latency** | gosdk_optimized | 14.79ms |
| **prompts/list latency** | gosdk_shimmed | 11.08ms |
| **Memory Efficiency** | gosdk_optimized | **149.43 MB** |

## Key Findings

1. ✅ **REGRESSION FIXED:** Both `gosdk_shimmed` and `gosdk_optimized` restore performance to baseline levels
2. 🔴 **REGRESSION CONFIRMED:** `gosdk` (without fix) is **58-71% slower** than main
3. 🏆 **BEST OVERALL:** `gosdk_optimized` has best latency AND memory usage
4. 💾 **MEMORY REDUCTION:** This PR reduces memory from **1186.82 MB → 149.43 MB** (87% reduction!)

---

## 💾 Memory Allocation Comparison (pprof)

| Configuration | Total Allocations | vs main | vs broken gosdk |
|--------------|-------------------|---------|-----------------|
| **main** | 343 MB | baseline | — |
| **gosdk (broken)** | 1186.82 MB | **+246% (3.46x)** | baseline |
| **gosdk_optimized** | **149.43 MB** | **-56%** 🏆 | **-87%** 🏆 |

### Root Cause - Why gosdk (broken) Uses So Much Memory

Top allocation sources in broken gosdk:
| Function | Size | % | Issue |
|----------|------|---|-------|
| **reflect.New** | 105.57 MB | 8.9% | 🚨 Schema reflection on every request |
| **slices.AppendSeq** | 85.50 MB | 7.2% | String collection for schema |
| **encoding/json.Unmarshal** | 66.01 MB | 5.6% | Schema parsing |
| **jsonschema.checkStructure** | 24.50 MB | 2.1% | 🚨 Schema validation |

### Why gosdk_optimized Is Best

Top allocation sources in optimized build:
| Function | Size | % | Notes |
|----------|------|---|-------|
| bytes.growSlice | 7.82 MB | 5.2% | Normal JSON operations |
| runtime.allocm | 5.51 MB | 3.7% | Goroutine stacks |
| encoding/json.Marshal | 5.10 MB | 3.4% | Normal serialization |

**No more `reflect.New`, `jsonschema.checkStructure`, or repeated schema generation!**

---

## Latency Test Results (n=30)

### Operation: `initialize`

| Configuration | P50 | P99 | vs main | Status |
|--------------|-----|-----|---------|--------|
| main | 13.42ms | 16.38ms | baseline | ✅ |
| gosdk | 21.20ms | 25.61ms | **+58%** | 🔴 REGRESSION |
| gosdk_shimmed | 11.89ms | 14.45ms | **-11%** | ✅ FIXED |
| **gosdk_optimized** | **11.73ms** | **15.22ms** | **-13%** | 🏆 WINNER |

### Operation: `tools/list`

| Configuration | P50 | P99 | vs main | Status |
|--------------|-----|-----|---------|--------|
| main | 15.44ms | 24.84ms | baseline | ✅ |
| gosdk | 24.74ms | 31.06ms | **+60%** | 🔴 REGRESSION |
| gosdk_shimmed | 15.13ms | 21.46ms | **-2%** | ✅ FIXED |
| **gosdk_optimized** | **15.07ms** | **19.40ms** | **-2%** | 🏆 WINNER |

### Operation: `prompts/list`

| Configuration | P50 | P99 | vs main | Status |
|--------------|-----|-----|---------|--------|
| main | 13.01ms | 15.56ms | baseline | ✅ |
| gosdk | 22.33ms | 29.35ms | **+72%** | 🔴 REGRESSION |
| **gosdk_shimmed** | **11.72ms** | 17.78ms | **-10%** | 🏆 WINNER |
| gosdk_optimized | 11.41ms | 16.30ms | **-12%** | ✅ FIXED |

---

## Stress Test Results (n=100)

### Operation: `initialize`

| Configuration | P50 | P99 | vs main | Status |
|--------------|-----|-----|---------|--------|
| main | 13.12ms | 18.95ms | baseline | ✅ |
| gosdk | 21.96ms | 29.70ms | **+67%** | 🔴 REGRESSION |
| gosdk_shimmed | 11.63ms | 15.39ms | **-11%** | ✅ FIXED |
| **gosdk_optimized** | **11.73ms** | **18.05ms** | **-11%** | 🏆 WINNER |

---

## Performance Summary

```
Latency (P50 initialize) - Lower is Better
================================================================================
main (mcp-go)           █████████████████████████████████ 13.42ms (baseline)
gosdk (broken)          ██████████████████████████████████████████████████████ 21.20ms (+58%)
gosdk_shimmed           █████████████████████████████ 11.89ms (-11%) ✅
gosdk_optimized         █████████████████████████████ 11.73ms (-13%) 🏆
================================================================================

Memory Allocations - Lower is Better  
================================================================================
main (mcp-go)           ████████████████████████████████████████████████████████████ 343 MB
gosdk (broken)          ████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████ 1187 MB (+246%)
gosdk_optimized         ██████████████████████████ 149 MB (-56% vs main) 🏆
================================================================================
```

---

## Conclusion

This PR's schema caching approach **fully resolves the performance regression** and actually provides:
- **13% faster latency** than the original mcp-go baseline
- **87% less memory** than the broken go-sdk implementation  
- **56% less memory** than the original mcp-go baseline

The fix eliminates repeated JSON schema generation via reflection by caching schemas after first generation.

**Strongly recommend merging this fix.** 🚀

SamMorrowDrums:
# Comprehensive Benchmark Report: Schema Caching Performance Impact

## Executive Summary

This PR's schema caching fix **fully resolves** the performance regression introduced when migrating from mcp-go to go-sdk, while also providing significant memory improvements.

| Metric | Before Fix (gosdk) | After Fix (gosdk_optimized) | Improvement |
|--------|-------------------|----------------------------|-------------|
| initialize latency | 21.20ms | 11.73ms | **-45%** |
| tools/list latency | 24.74ms | 15.07ms | **-39%** |
| Memory allocations | 1186.82 MB | 149.43 MB | **-87%** |

---

## Test Configurations

| # | Name | Description |
|---|------|-------------|
| 1 | **main** | Original MCP implementation using mcp-go library |
| 2 | **gosdk** | Go SDK WITHOUT schema caching (broken) |
| 3 | **gosdk_shimmed** | Go SDK with tool handler shim workaround |
| 4 | **gosdk_optimized** | Go SDK with this PR's schema caching fix |

---

## 1. MCP Protocol Latency Tests (n=30)

### Operation: `initialize`

| Configuration | P50 | P99 | vs main | Status |
|--------------|-----|-----|---------|--------|
| main | 13.42ms | 16.38ms | baseline | ✅ |
| gosdk | 21.20ms | 25.61ms | **+58%** | 🔴 REGRESSION |
| gosdk_shimmed | 11.89ms | 14.45ms | -11% | ✅ FIXED |
| **gosdk_optimized** | **11.73ms** | **15.22ms** | **-13%** | 🏆 WINNER |

### Operation: `tools/list`

| Configuration | P50 | P99 | vs main | Status |
|--------------|-----|-----|---------|--------|
| main | 15.44ms | 24.84ms | baseline | ✅ |
| gosdk | 24.74ms | 31.06ms | **+60%** | 🔴 REGRESSION |
| gosdk_shimmed | 15.13ms | 21.46ms | -2% | ✅ FIXED |
| **gosdk_optimized** | **15.07ms** | **19.40ms** | **-2%** | 🏆 WINNER |

### Operation: `prompts/list`

| Configuration | P50 | P99 | vs main | Status |
|--------------|-----|-----|---------|--------|
| main | 13.01ms | 15.56ms | baseline | ✅ |
| gosdk | 22.33ms | 29.35ms | **+72%** | 🔴 REGRESSION |
| gosdk_shimmed | 11.72ms | 17.78ms | -10% | ✅ FIXED |
| **gosdk_optimized** | **11.41ms** | **16.30ms** | **-12%** | 🏆 WINNER |

---

## 2. Stress Tests (n=100)

### Operation: `initialize`

| Configuration | P50 | P99 | vs main | Status |
|--------------|-----|-----|---------|--------|
| main | 13.12ms | 18.95ms | baseline | ✅ |
| gosdk | 21.96ms | 29.70ms | **+67%** | 🔴 REGRESSION |
| gosdk_shimmed | 11.63ms | 15.39ms | -11% | ✅ FIXED |
| **gosdk_optimized** | **11.73ms** | **18.05ms** | **-11%** | 🏆 WINNER |

---

## 3. Memory Allocation Comparison (pprof)

| Configuration | Total Allocations | vs main | vs broken gosdk |
|--------------|-------------------|---------|-----------------|
| **main** | 343 MB | baseline | — |
| **gosdk (broken)** | 1186.82 MB | **+246%** | baseline |
| **gosdk_optimized** | **149.43 MB** | **-56%** 🏆 | **-87%** 🏆 |

### Root Cause - Top Allocations in Broken gosdk

| Function | Size | % | Issue |
|----------|------|---|-------|
| **reflect.New** | 105.57 MB | 8.9% | 🚨 Schema reflection on every request |
| **slices.AppendSeq** | 85.50 MB | 7.2% | String collection for schema |
| **encoding/json.Unmarshal** | 66.01 MB | 5.6% | Schema parsing |
| **jsonschema.checkStructure** | 24.50 MB | 2.1% | 🚨 Schema validation |

### Top Allocations in gosdk_optimized (After Fix)

| Function | Size | % | Notes |
|----------|------|---|-------|
| bytes.growSlice | 7.82 MB | 5.2% | Normal JSON operations |
| runtime.allocm | 5.51 MB | 3.7% | Goroutine stacks |
| encoding/json.Marshal | 5.10 MB | 3.4% | Normal serialization |

**No more `reflect.New`, `jsonschema.checkStructure`, or repeated schema generation!**

---

## 4. Tool Call Tests (n=25, includes GitHub API latency)

> ⚠️ Times include GitHub API round-trip (~200-750ms)

### Tool: `get_me`

| Configuration | P50 | P99 | Avg | Errors |
|--------------|-----|-----|-----|--------|
| **main** | **215.33ms** 🏆 | 411.90ms | 232.58ms | 0 |
| gosdk | 230.62ms | 414.06ms | 241.64ms | 0 |
| gosdk_shimmed | 226.53ms | **282.34ms** | 228.62ms | 0 |
| gosdk_optimized | 219.92ms | 412.27ms | 235.64ms | 0 |

### Tool: `list_issues`

| Configuration | P50 | P99 | Avg | Errors |
|--------------|-----|-----|-----|--------|
| main | 758.72ms | 1114.68ms | 798.82ms | 0 |
| gosdk | 754.56ms | 1093.81ms | 758.48ms | 0 |
| gosdk_shimmed | **710.04ms** 🏆 | 1725.47ms | 781.53ms | 0 |
| gosdk_optimized | 745.10ms | **928.19ms** 🏆 | **744.38ms** 🏆 | 0 |

**Tool call performance is equivalent across all configurations** - variance is primarily from GitHub API response times.

---

## 5. Visual Summary

```
Latency (P50 initialize) - Lower is Better
================================================================================
main (mcp-go)           █████████████████████████████████ 13.42ms (baseline)
gosdk (broken)          ██████████████████████████████████████████████████████ 21.20ms (+58%)
gosdk_shimmed           █████████████████████████████ 11.89ms (-11%) ✅
gosdk_optimized         █████████████████████████████ 11.73ms (-13%) 🏆
================================================================================

Memory Allocations - Lower is Better  
================================================================================
main (mcp-go)           ██████████████████ 343 MB
gosdk (broken)          ██████████████████████████████████████████████████████████████ 1187 MB (+246%)
gosdk_optimized         █████████ 149 MB (-56% vs main, -87% vs broken) 🏆
================================================================================
```

---

## Winners Summary

| Category | Winner | Value |
|----------|--------|-------|
| **initialize latency** | gosdk_optimized | 11.73ms |
| **tools/list latency** | gosdk_optimized | 15.07ms |
| **prompts/list latency** | gosdk_optimized | 11.41ms |
| **Memory efficiency** | gosdk_optimized | **149.43 MB** |
| **get_me tool call** | main | 215.33ms |
| **list_issues P99** | gosdk_optimized | 928.19ms |

---

## Conclusion

This PR's schema caching approach **fully resolves the performance regression** and provides:

- ✅ **13% faster latency** than mcp-go baseline for MCP protocol operations
- ✅ **87% less memory** than broken go-sdk (1187 MB → 149 MB)
- ✅ **56% less memory** than mcp-go baseline (343 MB → 149 MB)
- ✅ **Equivalent tool call performance** (within 2-5%)
- ✅ **Lowest P99 variance** for complex operations

The fix eliminates repeated JSON schema generation via reflection by caching schemas after first generation.

**Strongly recommend merging this fix.** 🚀

findleyr:
Awesome analysis, thank you.

mcp/server.go
findleyr, line 368: This very subtle logic is changed in this CL, so at the very least we need to update or delete this comment.

I think we were overly cautious when changing the behavior with respect to pointers, and only changed our handling of null if we were responsible for deriving the schemal.

But looking at it now, I actually think this change is correct: the only point of NOT returning a zero if the user supplied their own schema would be to allow returning JSON "null", but that's not valid for structured content according to the MCP spec. Therefore, I think we SHOULD change this, but not without a bug report and a test demonstrating the incorrect behavior. I will do that right now.

mcp/server.go
```diff
@@ -367,26 +367,68 @@ func toolForErr[In, Out any](t *Tool, h ToolHandlerFor[In, Out]) (*Tool, ToolHan
...
+		}
+		*rfield = resolved
+		globalSchemaCache.setByType(rt, internalSchema, resolved)
+		return zero, nil
 	}
+
+	// Case 2: Schema was provided
+	// Check if it's a *jsonschema.Schema we can cache by pointer
+	if providedSchema, ok := (*sfield).(*jsonschema.Schema); ok {
+		if resolved, ok := globalSchemaCache.getBySchema(providedSchema); ok {
```
findleyr: I'll just point out that this is a rather subtle change in behavior: previously, if the user re-used a schema, we'd always validate against the version of the schema that was provided when they called AddTool. Now, we may get a cache hit and use a stale resolved schema.

I know this is unlikely, but given its subtlety this is a rather scary change in behavior.

I wonder if we should therefore put this optimization behind a ServerOption. For example, imagine a `ServerOptions.SchemaCache` field that accepted a `schemacache.Cache`.

SamMorrowDrums: It's certainly not a problem to put it behind a server option from my perspective. I'll do that.

jba: When you make that change, push it as a separate commit so we can look at it in isolation.

mcp/server.go
findleyr: (as you noticed, this is #692)

mcp/server.go
SamMorrowDrums: OK I have done so: https://github.com/modelcontextprotocol/go-sdk/pull/685/commits/f476e90b08e7b74c1a421bab5f2efdf4aaf6533f

It only changes the public API by adding a server option. It does have to add a bunch of nil checks, I couldn't find a better way to reduce that, but I think it's still pretty clear.

SamMorrowDrums: I also removed the underscores per your nit @jba https://github.com/modelcontextprotocol/go-sdk/pull/685/commits/900051a0f1a5ffbef6eed19e39342d483fd25724

findleyr:
@SamMorrowDrums do you need us to land this for v1.2.0? I think it's a good change, but just trying to triage. We're trying to get a prerelease out this week.

SamMorrowDrums:
@findleyr thanks for checking. Our workaround is working well so there's no urgent fire to put out.

findleyr:
Just checking in to say that I'm back from break, and we cut v1.2.0, so we should try to land this PR!

Quick check: Sam, is this ready for re-review?

SamMorrowDrums:
@findleyr awesome! Let me give it a self-review and then I'll ping you once I'm happy,

mcp/server.go
```diff
@@ -239,7 +244,7 @@ func (s *Server) AddTool(t *Tool, h ToolHandler) {
 		func() bool { s.tools.add(st); return true })
 }
 
-func toolForErr[In, Out any](t *Tool, h ToolHandlerFor[In, Out]) (*Tool, ToolHandler, error) {
+func toolForErr[In, Out any](t *Tool, h ToolHandlerFor[In, Out], cache *SchemaCache) (*Tool, ToolHandler, error) {
```
SamMorrowDrums, line 247: Could be nicer to pass a struct through with this sort of stuff to avoid signature churn or large numbers of future arguments, but I think as it's internal it can be refactored should it get unwieldy later.

SamMorrowDrums:
OK @findleyr I think it's clean now, I did a few edits, added short documentation and tried to keep it to the point. LMK what you think. :bow:

SamMorrowDrums:
@jba were you perhaps looking at a previous commit?

The `SchemaCache` type is now exported in `schema_cache.go`. The field `ServerOptions.SchemaCache` has type `*SchemaCache` so both the field name and type are exported.

Changing to `CacheSchemas bool` would force a global/package-level cache, removing user control over the cache lifecycle. For stateless deployments (the primary use case), that would only complicate things I think.

Perhaps I'm missing something, let me know.

SamMorrowDrums:
@jba re-reading, perhaps you meant you don't want to expose the SchemaCache type?

The end user needs to create the cache at application startup time, and pass it to servers during server creation per request for this to make sense. I can't think of a good way to do this unless we do the package global cache. Let me know your preference.

jba:
I must have been looking at an old commit. This is fine.

Are there any other concerns, @findleyr?

mcp/server.go
findleyr: Agreed. Let's keep it simple for now.

findleyr:
@jba indeed this looks good. I just need to page back in why I decided not to proceed with https://github.com/modelcontextprotocol/go-sdk/pull/692... I think I got concerned about the silent behavior change. Need to convince myself its OK. If I can't, we can just keep the current (subtle) behavior, as it's orthogonal to the rest of this PR.

Will get to this today.

SamMorrowDrums:
@findleyr is this still planned for merge or are there any outstanding issues?

PR #688, opened 2025-12-05
mcp: improve documentation of Add*Middleware

jboolean:
This updates the documentation comments for `AddReceivingMiddleware` and `AddSendingMiddleware` to clarify how middlewares are ordered. The existing documentation is confusing and could lead one to believe middleware is applied from right to left _at the time of execution_ (e.g. the `mcp-go` sdk does it this way). 
In fact, because the wrapping is done at the time of adding and not at the time of execution, middleware is only applied in reverse order (and therefore run _in order_) within the set of middlewares being added together, while middleware added successively is applied on top of existing middleware and thus run in _reverse order_.

### DIFF
--- a/mcp/server.go
+++ b/mcp/server.go
@@ -1101,12 +1101,15 @@ func (ss *ServerSession) Log(ctx context.Context, params *LoggingMessageParams) 
 	return handleNotify(ctx, notificationLoggingMessage, newServerRequest(ss, orZero[Params](params)))
 }
 
-// AddSendingMiddleware wraps the current sending method handler using the provided
-// middleware. Middleware is applied from right to left, so that the first one is
-// executed first.
+// AddSendingMiddleware wraps the current sending method handler using
+// the provided middleware.
+// Middleware supplied via arguments is applied from right to left at the time it is added,
+// so that the first one among these is executed first.
 //
 // For example, AddSendingMiddleware(m1, m2, m3) augments the method handler as
 // m1(m2(m3(handler))).
+// Then adding another set of middleware with AddSendingMiddleware(m4, m5, m6) yields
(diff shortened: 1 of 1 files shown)


mcp/server.go
```diff
@@ -1101,12 +1101,15 @@ func (ss *ServerSession) Log(ctx context.Context, params *LoggingMessageParams)
 	return handleNotify(ctx, notificationLoggingMessage, newServerRequest(ss, orZero[Params](params)))
 }
 
-// AddSendingMiddleware wraps the current sending method handler using the provided
-// middleware. Middleware is applied from right to left, so that the first one is
-// executed first.
+// AddSendingMiddleware wraps the current sending method handler using
+// the provided middleware.
+// Middleware supplied via arguments is applied from right to left at the time it is added,
```
findleyr, line 1106: When I first read this, I wondered "how else can middleware be applied other than as arguments" -- to me at least, it's a bit hard to parse.

How about "Arguments to AddSendingMiddleware are applied from right to left, so that the first middleware is executed first."

jboolean: I see your point. I'd like to keep the phrase "among these" though because the key is that not all middleware is applied right to left, it's each set that was added together.
"Middleware in each call is applied from right to left, so that the first middleware among these is executed first." Or something like that.

findleyr: Sorry for not replying: that sounds like a good middle ground.

maciej-kisiel:
Hi @jboolean, would you like to finalize the PR with the latest feedback from from Rob?

maciej-kisiel:
I'm closing the PR for house-keeping. Feel free to reopen, if you'd like to continue working on it.

Issue #199, opened 2025-07-29
No tools are advertised when assigning a pointer type in the generic for `CallToolParamsFor`


sethvargo:
**Describe the bug**
When the generic type for `CallToolParamsFor` is a pointer in the `ToolHandlerFor`, the tool is not included in the response to `tools/list`.

**To Reproduce**

Prove that tools are included in the response with non-pointer types:

1. Create a small server:

    ```go
    package main
    
    import (
    	"context"
    	"net/http"
    
    	"github.com/modelcontextprotocol/go-sdk/mcp"
    )
    
    func main() {
    	server := mcp.NewServer(&mcp.Implementation{Name: "hello"}, nil)
    
    	mcp.AddTool(server, &mcp.Tool{
    		Name:        "greet",
    		Description: "Say hello",
    	}, func(ctx context.Context, ss *mcp.ServerSession, params *mcp.CallToolParamsFor[struct{}]) (*mcp.CallToolResultFor[struct{}], error) {
    		return &mcp.CallToolResultFor[struct{}]{
    			Content: []mcp.Content{
    				&mcp.TextContent{Text: "hello"},
    			},
    		}, nil
    	})
    
    	http.ListenAndServe(":8080", mcp.NewStreamableHTTPHandler(func(*http.Request) *mcp.Server {
    		return server
    	}, nil))
    }
    ```

2. Start the server:

    ```shell
    go run main.go
    ```

3. List the tools:

    ```shell
    SESSION_ID=$(curl -sfo /dev/null -w '%header{mcp-session-id}' localhost:8080 \
      -H 'Accept: application/json,text/event-stream' \
      -H 'Mcp-Protocol-Version: 2025-06-18' \
      -d '{
      "jsonrpc":"2.0",
      "id":1,
      "method":"initialize",
      "params":{
        "protocolVersion":"2025-06-18",
        "capabilities":{},
        "clientInfo":{"name":"Seth", "title":"Seth Test Client", "version":"1.0.0"}
      }
    }'
    )
    ```

    ```shell
    curl -v localhost:8080 \
      -H "Accept: application/json,text/event-stream" \
      -H "Mcp-Protocol-Version: 2025-06-18" \
      -H "Mcp-Session-Id: ${SESSION_ID}" \
      -d '{
      "jsonrpc":"2.0",
      "method":"notifications/initialized"
    }'
    ```

    ```shell
    curl -v localhost:8080 \
[...]

PR #692, opened 2025-12-05
mcp: don't treat pointers differently given an OutputSchema override

### LINKED ISSUE #691: Fix handling for nil pointers given user-provided output schema
Following up on some discussion in #685: When fixing #199, I was overly cautious about handling typed nils: if the user provided their own output schema, we wouldn't fix typed nils for them, and so returning 'nil' would result in a missing structured output, rather than its zero value.

This was wrong: switching from an inferred schema to a user-provided schema shouldn't change the handling of nil, and structured output, if provided, must be an object.

PR incoming.

findleyr:
In the fix for https://github.com/modelcontextprotocol/go-sdk/issues/199, I was overly cautious: I assumed that if the user
provided their own schema, we shouldn't fix up typed nils for them.

But this was wrong: consider that:
1. If the user provides a schema, it must be an "object" schema.
2. One way that we recommend customizing schemas is by calling
   jsonschema.For and modifying. It's very confusing that doing
   this would change the treatment of the zero value.

The good news is that any tool that was returning a typed nil is
provably wrong, so this is just a bug that we can and should fix.

Fixes #691

### DIFF
--- a/mcp/mcp_test.go
+++ b/mcp/mcp_test.go
@@ -1902,16 +1902,20 @@ func TestPointerArgEquivalence(t *testing.T) {
 	type input struct {
 		In string `json:",omitempty"`
 	}
+	inputSchema := json.RawMessage(`{"type":"object","properties":{"In":{"type":"string"}},"additionalProperties":false}`)
 	type output struct {
 		Out string
 	}
+	outputSchema := json.RawMessage(`{"type":"object","required":["Out"],"properties":{"Out":{"type":"string"}},"additionalProperties":false}`)
 	cs, _, cleanup := basicConnection(t, func(s *Server) {
-		// Add two equivalent tools, one of which operates in the 'pointer' realm,
-		// the other of which does not.
+		// Add three equivalent tools:
+		//  - one operates on pointers, with inferred schemas
+		//  - one operates on pointers, with user-provided schemas
+		//  - one operates on non-pointers
 		//
 		// We handle a few different types of results, to assert they behave the
 		// same in all cases.
-		AddTool(s, &Tool{Name: "pointer"}, func(_ context.Context, req *CallToolRequest, in *input) (*CallToolResult, *output, error) {
+		handlePointers := func(_ context.Context, req *CallToolRequest, in *input) (*CallToolResult, *output, error) {
 			switch in.In {
 			case "":
 				return nil, nil, fmt.Errorf("must provide input")
@@ -1924,7 +1928,13 @@ func TestPointerArgEquivalence(t *testing.T) {
 			default:
 				panic("unreachable")
 			}
-		})
+		}
+		AddTool(s, &Tool{Name: "pointer-inferred"}, handlePointers)
+		AddTool(s, &Tool{
+			Name:         "pointer-provided",
+			InputSchema:  inputSchema,
+			OutputSchema: outputSchema,
+		}, handlePointers)
 		AddTool(s, &Tool{Name: "nonpointer"}, func(_ context.Context, req *CallToolRequest, in input) (*CallToolResult, output, error) {
 			switch in.In {
 			case "":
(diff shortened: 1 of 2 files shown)


fahimfazledev-glitch:
@findleyr
[mcp: don't treat pointers differently given an OutputSchema override](https://github.com/modelcontextprotocol/go-sdk/pull/692/commits/7dc5160249253672f69fcc10b203f1895527a671)

mcp/server.go
```diff
@@ -359,21 +359,14 @@ func toolForErr[In, Out any](t *Tool, h ToolHandlerFor[In, Out]) (*Tool, ToolHan
...
 // If an indirection occurred to derive the schema, a non-nil zero value is
 // returned to be used in place of the typed nil zero value.
-//
-// Note that if sfield already holds a schema, zero will be nil even if T is a
-// pointer: if the user provided the schema, they may have intentionally
-// derived it from the pointer type, and handling of zero values is up to them.
-//
-// TODO(rfindley): we really shouldn't ever return 'null' results. Maybe we
-// should have a jsonschema.Zero(schema) helper?
 func setSchema[T any](sfield *any, rfield **jsonschema.Resolved) (zero any, err error) {
```
markus-kusano, line 362: Maybe add some more documentation around `sfield` and `rfield`, and what exactly this function is doing? E.g., it seems like `sfield` is an inpput, and `rfield` is an output, in addition to the 2 return variables. I think it might be beneficial to document the type of `sfield` since it is `any` but I think this is an `any` expected to unmarshall into a valid jsonschema object (cf Tool.OutputSchema).

I am also not familiar with the `sfield` and `rfield` names but I could be missing some conventions around this.

I could be missing some context around how this function is used, or conventions around these names.

mcp/mcp_test.go
```diff
@@ -1947,50 +1957,51 @@ func TestPointerArgEquivalence(t *testing.T) {
...
-	for _, args := range []any{nil, struct{}{}} {
-		r0, err := cs.CallTool(ctx, &CallToolParams{Name: t0.Name, Arguments: args})
-		if err != nil {
-			t.Fatal(err)
-		}
-		r1, err := cs.CallTool(ctx, &CallToolParams{Name: t1.Name, Arguments: args})
-		if err != nil {
-			t.Fatal(err)
+	t0 := tools.Tools[0]
+	for _, t1 := range tools.Tools[1:] {
```
markus-kusano, line 1965: `t1` is a confusing name here since it is index `1` or `2`.

`tn` might be slightly better.

maybe it would be even clearer if we iterated over adjacent pairs in this array, then we could use names like `i` and `j` and avoid giving explicit indices.

mcp/server.go
```diff
@@ -359,21 +359,14 @@ func toolForErr[In, Out any](t *Tool, h ToolHandlerFor[In, Out]) (*Tool, ToolHan
...
+	rt := reflect.TypeFor[T]()
+	if rt.Kind() == reflect.Pointer {
```
markus-kusano, line 364: I'm on the fence, should we put a comment here explaining the importance of setting `rt` and `zero` when `sfield != nil`?  As in, the change in this PR is (I think) making it so we never return a typed `nil` for pointer elements an rather return a element-typed zero. Is there a good reference as to why we should never return typed nils? From the comment in the PR it sounds like this is always a bug to return typed nils, but why?

mcp/server.go
```diff
@@ -359,21 +359,14 @@ func toolForErr[In, Out any](t *Tool, h ToolHandlerFor[In, Out]) (*Tool, ToolHan
...
+		rt = rt.Elem()
+		zero = reflect.Zero(rt).Interface()
+	}
```
markus-kusano, line 367: Somewhat related to my previous comment, should we check if our handling of zero/nil for pointer types is "interfering" with the user provided schema? It sounds like actually the user should never be including the ability to return typed nils in their schema, so maybe this is something we can check at validation time?

its also OK if the answer is "no" and this is too much added complexity, I just wanted to double check if there's any corner cases we could easily identify and flag for the user to make their UX better

PR #696, opened 2025-12-08
mcp: gate prime and close events on protocol version 2025-11-25

### LINKED ISSUE #686: SEP-1699: Don't send priming message if client protocol version is <2025-11-25
Reported by @KKonstantinov in discord: several SDKs implemented SEP-1699 by sending a priming message with empty `data` field, which breaks the client if it tries to interpret this field as a JSON-RPC message.

We should investigate this for the Go SDK implementation. IIRC, I thought about this and chose to use a "prime" event type. That may or may not be an acceptable fix, depending on how clients treat it in the wild. Probably we should:

1. Only issue priming/retry events if the client protocol version is >= 2025-11-25.
2. Be consistent with other SDKs in the structure of our events: if they don't use "prime", then neither should we (but shouldn't they use something other than the default "message" event type?)

findleyr:
The prime and close SSE events (SEP-1699) were added in protocol version 2025-11-25. Only send these events when the client negotiates that version or later.

- In servePOST, extract protocol version from InitializeParams for initialize requests, otherwise use the Mcp-Protocol-Version header
- In serveGET/acquireStream, read protocol version from header
- Only set closeLocked callback when prime/close is supported
- Add unexported protocolVersion field to ClientSessionOptions for testing

Fixes #686

### DIFF
--- a/mcp/streamable.go
+++ b/mcp/streamable.go
@@ -710,7 +710,15 @@ func (c *streamableServerConn) serveGET(w http.ResponseWriter, req *http.Request
 	ctx, cancel := context.WithCancel(req.Context())
 	defer cancel()
 
-	stream, done := c.acquireStream(ctx, w, streamID, &lastIdx)
+	// Read the protocol version from the header. For GET requests, this should
+	// always be present since GET only happens after initialization.
+	protocolVersion := req.Header.Get(protocolVersionHeader)
+	if protocolVersion == "" {
+		protocolVersion = protocolVersion20250326
+	}
+	supportsPrimeClose := protocolVersion >= protocolVersion20251125
+
+	stream, done := c.acquireStream(ctx, w, streamID, &lastIdx, supportsPrimeClose)
 	if stream == nil {
 		return
 	}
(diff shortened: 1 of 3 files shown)


mcp/streamable.go
```diff
@@ -710,7 +710,15 @@ func (c *streamableServerConn) serveGET(w http.ResponseWriter, req *http.Request
...
-	stream, done := c.acquireStream(ctx, w, streamID, &lastIdx)
+	// Read the protocol version from the header. For GET requests, this should
+	// always be present since GET only happens after initialization.
+	protocolVersion := req.Header.Get(protocolVersionHeader)
+	if protocolVersion == "" {
+		protocolVersion = protocolVersion20250326
+	}
+	supportsPrimeClose := protocolVersion >= protocolVersion20251125
+
+	stream, done := c.acquireStream(ctx, w, streamID, &lastIdx, supportsPrimeClose)
```
markus-kusano, line 721: nit, feel free to ignore: would it be a cleaner separation if `acquireStream` took the `protocolVersion` instead of `supportsPrimeClose`? as in, we could move the logic on lines 715--719 into `acquireStream`, and then the caller only needs to pass in the version.

given that this logic isn't duplicated anywhere right now i'm also OK leaving it as is

mcp/streamable.go
```diff
@@ -878,16 +889,18 @@ func (c *streamableServerConn) acquireStream(ctx context.Context, w http.Respons
 		}
 		return c.writeEvent(w, s.id, Event{Name: "message", Data: data}, lastIdx)
 	}
-	s.closeLocked = func(reconnectAfter time.Duration) {
```
markus-kusano, line 881: Is it more readable to define and set `s.supportsPrimeClose` here instead of defining or implicitly not defining `s.closeLocked`? I think this makes the intent clearer: I only need to set `closeLocked` when we have to support reconnect after, and not the other way around.

To put it another way, when first reviewing this CL it isn't obvious why it is safe to *not* set `s.closeLocked` in some cases. At first glance it looks like we don't do anything on `close` in that case, and it isn't obvious why its OK for us to do nothing.

So, in more detail, I'd propose the following if it helps readability:

1. Replace `closeLocked` from `streamableServerConn` with either `supportsPrimeClose`, or maybe just `protocolVersion` (cf my comment above)
2. Encapsulate the definition of `closeLocked` within stream.close() (i think this is the only place it is used). We can call `closeLocked` in `close` is `supportsPrimeClose`

I can also understand keeping the ability to customize `closeLocked` depending on different use cases, however the code isn't using this feature right now so this might be [yagni](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it)

This might be a nitpick, so I'd also be OK with some comments if that's easier

PR #698, opened 2025-12-09
mcp: fix broken client root capabilities

### LINKED ISSUE #607: How to check client capabilities and handle unsupported capability errors?

findleyr:
To address #607, add ClientCapabilities.RootsV2, and populate it when constructing sending and receiving InitializeParams.

Fixes #607

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -51,7 +51,7 @@ func ExampleLoggingTransport() {
 
 	// Output:
 	// read: {"jsonrpc":"2.0","id":1,"result":{"capabilities":{"logging":{}},"protocolVersion":"2025-06-18","serverInfo":{"name":"server","version":"v0.0.1"}}}
-	// write: {"jsonrpc":"2.0","id":1,"method":"initialize","params":{"capabilities":{"roots":{"listChanged":true}},"clientInfo":{"name":"client","version":"v0.0.1"},"protocolVersion":"2025-06-18"}}
+	// write: {"jsonrpc":"2.0","id":1,"method":"initialize","params":{"clientInfo":{"name":"client","version":"v0.0.1"},"protocolVersion":"2025-06-18","capabilities":{"roots":{"listChanged":true}}}}
 	// write: {"jsonrpc":"2.0","method":"notifications/initialized","params":{}}
 }
 ```
--- a/mcp/protocol.go
+++ b/mcp/protocol.go
@@ -226,6 +226,10 @@ type clientCapabilitiesV2 struct {
 func (c *clientCapabilitiesV2) toV1() *ClientCapabilities {
 	caps := c.ClientCapabilities
 	caps.RootsV2 = c.Roots
+	// Sync Roots from RootsV2 for backward compatibility (#607).
+	if caps.RootsV2 != nil {
+		caps.Roots = *caps.RootsV2
+	}
 	return &caps
 }
 
--- a/mcp/shared.go
+++ b/mcp/shared.go
@@ -104,12 +104,12 @@ func defaultSendingMethodHandler(ctx context.Context, method string, req Request
 	}
 	// Notifications don't have results.
 	if strings.HasPrefix(method, "notifications/") {
-		return nil, req.GetSession().getConn().Notify(ctx, method, req.GetParams())
+		return nil, req.GetSession().getConn().Notify(ctx, method, params)
 	}
 	// Create the result to unmarshal into.
 	// The concrete type of the result is the return type of the receiving function.
 	res := info.newResult()
-	if err := call(ctx, req.GetSession().getConn(), method, req.GetParams(), res); err != nil {
+	if err := call(ctx, req.GetSession().getConn(), method, params, res); err != nil {
 		return nil, err
 	}
 	return res, nil
--- a/mcp/transport_example_test.go
+++ b/mcp/transport_example_test.go
@@ -46,7 +46,7 @@ func ExampleLoggingTransport() {
 
 	// Output:
 	// read: {"jsonrpc":"2.0","id":1,"result":{"capabilities":{"logging":{}},"protocolVersion":"2025-06-18","serverInfo":{"name":"server","version":"v0.0.1"}}}
-	// write: {"jsonrpc":"2.0","id":1,"method":"initialize","params":{"capabilities":{"roots":{"listChanged":true}},"clientInfo":{"name":"client","version":"v0.0.1"},"protocolVersion":"2025-06-18"}}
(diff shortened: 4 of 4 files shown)


mcp/client.go
```diff
@@ -58,19 +58,23 @@ func NewClient(impl *Implementation, opts *ClientOptions) *Client {
...
+	//
+	// For advertising capabilities to the server, Capabilities.Roots is ignored:
+	// use Capabilities.RootsV2.
+	Capabilities *ClientCapabilities
+
 	// CreateMessageHandler handles incoming requests for sampling/createMessage.
 	//
 	// Setting CreateMessageHandler to a non-nil value causes the client to
-	// advertise the sampling capability.
+	// advertise the sampling capability (if Capabilities is nil).
```
markus-kusano: nit, potential wording bike shedding, feel free to ignore: maybe `(see comment on Capabilities above)` instead of mentioning if Capabilities is nil? i like the description on the `Capabilities` comment that describes the overall behavior.

mcp/client.go
```diff
@@ -127,6 +131,12 @@ type ClientSessionOptions struct {
 }
 
 func (c *Client) capabilities() *ClientCapabilities {
+	// If Capabilities is set, use it as the override.
+	if c.opts.Capabilities != nil {
+		return c.opts.Capabilities
+	}
+
+	// Otherwise, infer capabilities from handlers.
```
markus-kusano: it feels a little foot gunny for the `Capabilities` struct to, e.g., disable a feature when the user also has a handler registered for it. should we do something like emit warning messages for cases like this? on the other hand, i could see people doing this so maybe its fine

mcp/client.go
```diff
@@ -137,16 +147,8 @@ func (c *Client) capabilities() *ClientCapabilities {
...
 	if c.opts.ElicitationHandler != nil {
-		caps.Elicitation = &ElicitationCapabilities{}
-		modes := c.opts.ElicitationModes
-		if len(modes) == 0 || slices.Contains(modes, "form") {
-			// Technically, the empty ElicitationCapabilities value is equivalent to
-			// {"form":{}} for backward compatibility, but we explicitly set the form
-			// capability.
-			caps.Elicitation.Form = &FormElicitationCapabilities{}
-		}
-		if slices.Contains(modes, "url") {
```
markus-kusano, line 147: double checking: does this change remove inferring URL elicitation capabilities, meaning the user needs to explicitly set them on the capabilities object? if so, this seems counter intuitive because it means in some cases the inference is incomplete. perhaps i am missing something, maybe there's no way to infer URL elicitation from simply the handler being set?

if the inference is often incomplete and the user needs to explicitly set Capabilities then it might be better to remove inference, but this is a larger change and out of scope here probably.

mcp/client.go
findleyr: Ack, let's review in a separate PR.

mcp/client.go
findleyr: This change means that instead of a bespoke ClientOption, they just set the capability explicitly.

Let's finalize in a separate PR.

mcp/client_test.go
```diff
@@ -202,9 +203,8 @@ func TestClientCapabilities(t *testing.T) {
 			name:            "With initial capabilities",
 			configureClient: func(s *Client) {},
 			wantCapabilities: &ClientCapabilities{
-				Roots: struct {
-					ListChanged bool "json:\"listChanged,omitempty\""
-				}{ListChanged: true},
+				Roots:   canListRoots,
```
markus-kusano: nit: there is mixed use of `Roots: canListRoots` (here) and in-lining its definition (line 254 below). imo, inlining it seems simpler, since the type name isn't too long, instead of introducing another variable for the reader to remember

findleyr: Good point. I was just excited to not have to repeat the type, but we can reuse `RootCapabilities`.
Done.

findleyr:
PTAL: I renamed `RootsCapabilities` to `RootCapabilities`, to be consistent with `ToolCapabilities`, `PromptCapabilities`, etc.

(Almost another mistake!)

findleyr:
Gentle nudge: this needs reapproval as I address comments and made the nontrivial change to rename to `RootCapabilities`.

markus-kusano (approved):

PR #699, opened 2025-12-09
mcp: implement sampling with tools

findleyr:
Summary

  Add support for tool use within sampling requests, as described in the MCP 2025-11-25 spec's sampling.tools capability.
  New types

  Content types:
  - ToolUseContent — represents a tool invocation request from the assistant
  - ToolResultContent — represents the result of a tool invocation, with nested []Content for the result blocks

  Capability types:
  - SamplingCapabilities gains Tools and Context sub-fields
  - SamplingToolsCapabilities, SamplingContextCapabilities

  Tool-enabled sampling (parallel tool calls):
  - CreateMessageWithToolsParams with SamplingMessageV2 (supports array content)
  - CreateMessageWithToolsResult (supports array content)
  - ToolChoice — controls tool invocation mode (auto, required, none)
  - ServerSession.CreateMessageWithTools
  - ClientOptions.CreateMessageWithToolsHandler

  Design

  Following the TypeScript SDK's pattern, tool-enabled sampling uses separate types from basic sampling to avoid breaking the existing API. The basic CreateMessage/CreateMessageResult path
   is unchanged.

  Both paths share the same wire method (sampling/createMessage) and go through the method info table. The table uses the broader CreateMessageWithToolsResult type internally;
  CreateMessage downconverts (erroring if multiple content blocks are returned). This is documented as a rough edge to unify in v2.

  Setting CreateMessageWithToolsHandler automatically infers the sampling.tools capability. It is a panic to set both CreateMessageHandler and CreateMessageWithToolsHandler.

  References

  - https://modelcontextprotocol.io/specification/2025-11-25/client/sampling
  - TypeScript SDK: CreateMessageResultWithToolsSchema / CreateMessageRequestParamsWithTools


EronWright:
Good stuff @findleyr.

findleyr:
If you skim over the tests, this change isn't actually that large.

Unfortunately, we needed to add new APIs to work around the spec change, but what we've chosen is consistent with typescript.

Notably, the new `ToolUseContent` has `Input` as `map[string]any`, rather than `any` or `json.RawMessage` as we've used in `CallToolParams` and `CallToolParamsRaw`. Per my note in `rough_edges.md`, I actually think it was a mistake to not have `CallToolParams.Arguments` just be a `map[string]any`, since we always need to translate to this in order to validate.

mcp/client.go
```diff
@@ -198,11 +222,14 @@ func (c *Client) capabilities(protocolVersion string) *ClientCapabilities {
...
 	}
 
-	// Augment with sampling capability if handler is set.
-	if c.opts.CreateMessageHandler != nil {
+	// Augment with sampling capability if a handler is set.
+	if c.opts.CreateMessageHandler != nil || c.opts.CreateMessageWithToolsHandler != nil {
 		if caps.Sampling == nil {
 			caps.Sampling = &SamplingCapabilities{}
 		}
+		if c.opts.CreateMessageWithToolsHandler != nil && caps.Sampling.Tools == nil {
```
maciej-kisiel: Shouldn't this be inside the `caps.Sampling == nil` if body? Otherwise, it may override manually set sampling capabilities, which would be slightly contrary to the doc comment above.

mcp/content.go
```diff
@@ -183,6 +186,104 @@ func (c *EmbeddedResource) fromWire(wire *wireContent) {
...
+
+// ToolResultContent represents the result of a tool invocation.
+// This content type is only valid in sampling messages with role "user".
+type ToolResultContent struct {
+	// ToolUseID references the ID from the corresponding ToolUseContent.
+	ToolUseID string
+	// Content holds the unstructured result of the tool call.
+	Content []Content
+	// StructuredContent holds an optional structured result as a JSON object.
+	StructuredContent any
```
maciej-kisiel, line 236: For my own education: why the same logic as to `ToolUseContent.Input` doesn't apply here? They are both defined the same way in the specification.

mcp/content.go
```diff
@@ -242,10 +342,40 @@ type wireContent struct {
 	Meta        Meta              `json:"_meta,omitempty"`
 	Annotations *Annotations      `json:"annotations,omitempty"`
 	Icons       []Icon            `json:"icons,omitempty"`
+	// Fields for ToolUseContent (type: "tool_use")
```
maciej-kisiel: Should we add similar comments for other fields, for completeness? I assume that the above fields are not universal based on the comment you removed from line 230.

mcp/content.go
```diff
@@ -242,10 +342,40 @@ type wireContent struct {
...
+	ID    string         `json:"id,omitempty"`
+	Input map[string]any `json:"input,omitempty"`
+	// Fields for ToolResultContent (type: "tool_result")
+	ToolUseID         string         `json:"toolUseId,omitempty"`
+	NestedContent     []*wireContent `json:"content,omitempty"` // nested content for tool_result
```
maciej-kisiel: Unnecessary comment at the end of the line.

mcp/client.go
```diff
@@ -453,12 +480,23 @@ func (c *Client) listRoots(_ context.Context, req *ListRootsRequest) (*ListRoots
...
-	if c.opts.CreateMessageHandler == nil {
-		// TODO: wrap or annotate this error? Pick a standard code?
-		return nil, &jsonrpc.Error{Code: codeUnsupportedMethod, Message: "client does not support CreateMessage"}
+func (c *Client) createMessage(ctx context.Context, req *CreateMessageWithToolsRequest) (*CreateMessageWithToolsResult, error) {
+	if c.opts.CreateMessageWithToolsHandler != nil {
+		return c.opts.CreateMessageWithToolsHandler(ctx, req)
+	}
+	if c.opts.CreateMessageHandler != nil {
+		// Downconvert the request for the basic handler.
+		baseReq := &CreateMessageRequest{
```
maciej-kisiel: Any reason not to implement `toBase` on the request level, as opposed to params? It would mirror the response functionality.

mcp/protocol.go
```diff
@@ -379,6 +388,105 @@ func (x *CreateMessageParams) isParams()              {}
...
+// from each message. Tools, ToolChoice, and any additional content blocks
+// (e.g. parallel tool calls) are dropped. The first block may be a
+// ToolUseContent or ToolResultContent, which the basic handler should
+// tolerate since SamplingMessage accepts tool content types.
+func (p *CreateMessageWithToolsParams) toBase() *CreateMessageParams {
+	var msgs []*SamplingMessage
+	for _, m := range p.Messages {
+		var content Content
+		if len(m.Content) > 0 {
+			content = m.Content[0]
```
maciej-kisiel, line 428: Wouldn't this behavior be a bit confusing that the remaining content blocks are ignored? Shouldn't we return an error?

mcp/server.go
```diff
@@ -1174,7 +1178,44 @@ func (ss *ServerSession) CreateMessage(ctx context.Context, params *CreateMessag
...
+		Role:       res.Role,
+		StopReason: res.StopReason,
+	}, nil
+}
+
+// CreateMessageWithTools sends a sampling request with tools to the client,
+// returning a [CreateMessageWithToolsResult] that supports array content
+// (for parallel tool calls). Use this instead of [ServerSession.CreateMessage]
+// when the request includes tools.
+func (ss *ServerSession) CreateMessageWithTools(ctx context.Context, params *CreateMessageWithToolsParams) (*CreateMessageWithToolsResult, error) {
```
maciej-kisiel, line 1207: Should these check if the client has the capability?

maciej-kisiel (commented):
Sharing the comments without looking at the test file to expedite the process. I will look at the remaining file soon.

mcp/sampling_tools_test.go
```diff
@@ -0,0 +1,1253 @@
+// Copyright 2025 The Go MCP SDK Authors. All rights reserved.
```
maciej-kisiel: I would suggest to rename this to `sampling_test.go` and leave a TODO to move other sampling related tests to this file.

mcp/sampling_tools_test.go
```diff
@@ -0,0 +1,1253 @@
...
+	if err != nil {
+		t.Fatalf("contentFromWire() error = %v", err)
+	}
+
+	toolUse, ok := content.(*ToolUseContent)
+	if !ok {
+		t.Fatalf("expected *ToolUseContent, got %T", content)
+	}
+
+	if toolUse.ID != "tool_123" {
```
maciej-kisiel: Nit: wouldn't it be a bit more concise to compare with a whole "want" `ToolUseContent`? In the subsequent test functions as well.

mcp/sampling_tools_test.go
```diff
@@ -0,0 +1,1253 @@
...
+	}
+	if !toolResult.IsError {
+		t.Error("IsError should be true")
+	}
+	if toolResult.ToolUseID != "tool_1" {
+		t.Errorf("ToolUseID = %v, want tool_1", toolResult.ToolUseID)
+	}
+}
+
+func TestToolResultContent_ImageNestedContent(t *testing.T) {
```
maciej-kisiel: This could be a table driven test case of `TestToolResultContent_UnmarshalJSON`.

mcp/sampling_tools_test.go
```diff
@@ -0,0 +1,1253 @@
...
+	textContent, ok := toolResult.Content[0].(*TextContent)
+	if !ok {
+		t.Fatalf("expected *TextContent, got %T", toolResult.Content[0])
+	}
+	if textContent.Text != "42" {
+		t.Errorf("Text = %v, want %v", textContent.Text, "42")
+	}
+}
+
+func TestCreateMessageWithToolsResult_ToolUseContent(t *testing.T) {
```
maciej-kisiel: Rename to `TestCreateMessageWithToolsResult_UnmarshalJSON` and gather all interesting cases here (e.g. array of contents as well)?

mcp/sampling_tools_test.go
```diff
@@ -0,0 +1,1253 @@
...
+
+import (
+	"context"
+	"encoding/json"
+	"reflect"
+	"strings"
+	"testing"
+)
+
+func TestToolUseContent_MarshalJSON(t *testing.T) {
```
maciej-kisiel: Should marshaling/unmarshaling tests for `ToolUseContent`/`ToolResultContent` be moved to `content_test.go`?

mcp/sampling_tools_test.go
```diff
@@ -0,0 +1,1253 @@
...
+	}
+	if toolUse.ID != "tool_1" {
+		t.Errorf("Content.ID = %v, want %v", toolUse.ID, "tool_1")
+	}
+	if toolUse.Name != "calculator" {
+		t.Errorf("Content.Name = %v, want %v", toolUse.Name, "calculator")
+	}
+}
+
+func TestSamplingMessage_ToolUseContent(t *testing.T) {
```
maciej-kisiel: This should probably become an additional test case of `TestSamplingMessage_UnmarshalJSON` in `protocol_test.go`. I think it's not there yet, but we can start it and leave a TODO to fill more cases. Similarly `TestSamplingMessage_ToolResultContent` below.

mcp/sampling_tools_test.go
```diff
@@ -0,0 +1,1253 @@
...
+	}
+	if toolResult.ToolUseID != "tool_1" {
+		t.Errorf("Content.ToolUseID = %v, want %v", toolResult.ToolUseID, "tool_1")
+	}
+	if len(toolResult.Content) != 1 {
+		t.Fatalf("len(Content.Content) = %v, want 1", len(toolResult.Content))
+	}
+}
+
+func TestSamplingCapabilities_WithTools(t *testing.T) {
```
maciej-kisiel: This should probably become an additional test case of `TestSamplingCapabilities_MarshalJSON` in `protocol_test.go`. I think it's not there yet, but we can start it and leave a TODO to fill more cases. Similarly `TestSamplingCapabilities_Empty` below.

I would suggest to separate the unmarshaling part into `TestSamplingCapabilities_UnmarshalJSON`.

mcp/sampling_tools_test.go
```diff
@@ -0,0 +1,1253 @@
...
+
+	if caps2.Tools != nil {
+		t.Error("Tools capability should be nil for empty capabilities")
+	}
+	if caps2.Context != nil {
+		t.Error("Context capability should be nil for empty capabilities")
+	}
+}
+
+func TestCreateMessageWithToolsParams(t *testing.T) {
```
maciej-kisiel: Split into `_MarshalJSON` and `_UnmarshalJSON` and move to `protocol_test.go`?

mcp/sampling_tools_test.go
```diff
@@ -0,0 +1,1253 @@
...
+	}
+	if params2.Tools[0].Name != "calculator" {
+		t.Errorf("Tools[0].Name = %v, want %v", params2.Tools[0].Name, "calculator")
+	}
+	if params2.ToolChoice == nil || params2.ToolChoice.Mode != "auto" {
+		t.Errorf("ToolChoice.Mode = %v, want %v", params2.ToolChoice, &ToolChoice{Mode: "auto"})
+	}
+}
+
+func TestToolChoice_Modes(t *testing.T) {
```
maciej-kisiel: This is simple enough to test both Marshal and Unmarshal at once. I would suggest to rename to `TestToolChoice_MarshalUnmarshalJSON` and move to `protocol_test.go`.

mcp/sampling_tools_test.go
```diff
@@ -0,0 +1,1253 @@
...
+	tests := []struct {
+		name string
+		mode string
+	}{
+		{"auto", "auto"},
```
maciej-kisiel: Just prepare a slice of `ToolChoice` structs and do `DeepEqual(tc, Unmarshal(Marshal(tc))`?

mcp/sampling_tools_test.go
```diff
@@ -0,0 +1,1253 @@
...
+		CreateMessageHandler: func(_ context.Context, req *CreateMessageRequest) (*CreateMessageResult, error) {
+			receivedMessages = req.Params.Messages
+			return &CreateMessageResult{
+				Model:   "test-model",
+				Role:    "assistant",
+				Content: &TextContent{Text: "The result is 3"},
+			}, nil
+		},
+		Capabilities: &ClientCapabilities{
+			Sampling: &SamplingCapabilities{Tools: &SamplingToolsCapabilities{}},
```
maciej-kisiel: Should specifying tool capability without actually providing the `CreateMessageWithToolsHandler` be allowed? More generally, do we ever validate the consistency of provided capabilities and the handlers provided? Based on my quick research we don't, maybe we should?

mcp/sampling_tools_test.go
```diff
@@ -0,0 +1,1253 @@
...
+	}
+	if tc, ok := toolResult.Content[0].(*TextContent); !ok || tc.Text != "3" {
+		t.Errorf("toolResult.Content[0] = %v, want TextContent with '3'", toolResult.Content[0])
+	}
+}
+
+func TestSamplingToolsCapability_Integration(t *testing.T) {
+	ctx := context.Background()
+
+	t.Run("client advertises tools capability", func(t *testing.T) {
```
maciej-kisiel: Nit: `client with explicit tools capability`.

mcp/sampling_tools_test.go
```diff
@@ -0,0 +1,1253 @@
...
+		}
+		if caps.Sampling.Tools != nil {
+			t.Error("client should NOT advertise sampling.tools capability")
+		}
+		if caps.Sampling.Context != nil {
+			t.Error("client should NOT advertise sampling.context capability")
+		}
+	})
+
+	t.Run("CreateMessageWithToolsHandler infers tools capability", func(t *testing.T) {
```
maciej-kisiel: Nit: `client with inferred tools capability`.

mcp/sampling_tools_test.go
```diff
@@ -0,0 +1,1253 @@
...
+	}
+	if len(toolResult.Content) != 1 {
+		t.Fatalf("toolResult.Content len = %d, want 1", len(toolResult.Content))
+	}
+	if tc, ok := toolResult.Content[0].(*TextContent); !ok || tc.Text != "3" {
+		t.Errorf("toolResult.Content[0] = %v, want TextContent with '3'", toolResult.Content[0])
+	}
+}
+
+func TestSamplingToolsCapability_Integration(t *testing.T) {
```
maciej-kisiel: Nit: `TestSamplingToolsCapabilities`.

mcp/sampling_tools_test.go
```diff
@@ -0,0 +1,1253 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer cs.Close()
+
+	// Server sends CreateMessage with error tool result
+	_, err = ss.CreateMessage(ctx, &CreateMessageParams{
+		MaxTokens: 1000,
+		Messages: []*SamplingMessage{
+			{Role: "user", Content: &ToolResultContent{
```
maciej-kisiel: Just for more realistic scenario: can we add the preceding messages like in `TestSamplingWithToolResult_Integration`?

mcp/sampling_tools_test.go
```diff
@@ -0,0 +1,1253 @@
...
+	img, ok := toolResult.Content[0].(*ImageContent)
+	if !ok {
+		t.Fatalf("nested content type = %T, want *ImageContent", toolResult.Content[0])
+	}
+	if img.MIMEType != "image/png" {
+		t.Errorf("MIMEType = %v, want image/png", img.MIMEType)
+	}
+}
+
+func TestToolUseContent_MetaRoundTrip(t *testing.T) {
```
maciej-kisiel: This could be covered in `TestToolUseContent_MarshalJSON` and `TestToolUseContent_UnmarshalJSON`.

mcp/sampling_tools_test.go
```diff
@@ -0,0 +1,1253 @@
...
+	}
+	if result.Content[0].(*ToolUseContent).ID != "call_1" {
+		t.Errorf("Content[0].ID = %v, want call_1", result.Content[0].(*ToolUseContent).ID)
+	}
+	if result.Content[1].(*ToolUseContent).ID != "call_2" {
+		t.Errorf("Content[1].ID = %v, want call_2", result.Content[1].(*ToolUseContent).ID)
+	}
+}
+
+func TestCreateMessageWithToolsResult_ArrayRoundTrip(t *testing.T) {
```
maciej-kisiel: This could be covered in `TestCreateMessageWithToolsResult_MarshalJSON` and `TestCreateMessageWithToolsResult_UnmarshalJSON`. The former doesn't exist at this point and probably should be created.

mcp/sampling_tools_test.go
```diff
@@ -0,0 +1,1253 @@
...
+			t.Fatalf("Content[%d] type = %T, want *ToolUseContent", i, c)
+		}
+		origTU := orig.Content[i].(*ToolUseContent)
+		if tu.ID != origTU.ID || tu.Name != origTU.Name {
+			t.Errorf("Content[%d] = %+v, want %+v", i, tu, origTU)
+		}
+	}
+}
+
+func TestCreateMessageWithToolsResult_SingleContentBackwardCompat(t *testing.T) {
```
maciej-kisiel: This could be covered by `TestCreateMessageWithToolsResult_MarshalJSON`.

mcp/sampling_tools_test.go
```diff
@@ -0,0 +1,1253 @@
...
+	})
+	if err == nil {
+		t.Fatal("expected error for multiple content blocks")
+	}
+	if !strings.Contains(err.Error(), "CreateMessageWithTools") {
+		t.Errorf("error should mention CreateMessageWithTools, got: %v", err)
+	}
+}
+
+func TestUnmarshalContent_NullJSON(t *testing.T) {
```
maciej-kisiel: This could be covered by `TestCreateMessageWithToolsResult_UnmarshalJSON`. Similarly `TestUnmarshalContent_EmptyArray` below.

mcp/sampling_tools_test.go
```diff
@@ -0,0 +1,1253 @@
...
+	}
+	if result.Content == nil {
+		t.Error("Content should be non-nil empty slice, got nil")
+	}
+	if len(result.Content) != 0 {
+		t.Errorf("len(Content) = %d, want 0", len(result.Content))
+	}
+}
+
+func TestSamplingMessageV2_EmptyContent(t *testing.T) {
```
maciej-kisiel: Rewrite the two subsequent test cases to `TestSamplingMessageV2_MarshalJSON` and `TestSamplingMessageV2_UnmarshalJSON` and move to `protocol_test.go`?

mcp/sampling_tools_test.go
```diff
@@ -0,0 +1,1253 @@
...
+	}
+	if _, ok := got.Content[0].(*TextContent); !ok {
+		t.Errorf("Content[0] type = %T, want *TextContent", got.Content[0])
+	}
+	if _, ok := got.Content[1].(*ToolUseContent); !ok {
+		t.Errorf("Content[1] type = %T, want *ToolUseContent", got.Content[1])
+	}
+}
+
+func TestCreateMessageWithToolsResult_RejectsToolResult(t *testing.T) {
```
maciej-kisiel: This could be covered by `TestCreateMessageWithToolsResult_UnmarshalJSON`.

mcp/sampling_tools_test.go
```diff
@@ -0,0 +1,1253 @@
...
+	if tc, ok := base.Messages[0].Content.(*TextContent); !ok || tc.Text != "hello" {
+		t.Errorf("Messages[0].Content = %v, want TextContent{hello}", base.Messages[0].Content)
+	}
+	// Second message: only first content block kept
+	if tu, ok := base.Messages[1].Content.(*ToolUseContent); !ok || tu.ID != "c1" {
+		t.Errorf("Messages[1].Content = %v, want ToolUseContent{c1}", base.Messages[1].Content)
+	}
+}
+
+func TestToWithTools_Conversion(t *testing.T) {
```
maciej-kisiel: Maybe create a table driven test with the two subsequent test methods?

mcp/sampling_tools_test.go
```diff
@@ -0,0 +1,1253 @@
...
+		"model": "m",
+		"role": "assistant"
+	}`
+	var result CreateMessageWithToolsResult
+	if err := json.Unmarshal([]byte(jsonData), &result); err == nil {
+		t.Error("expected error for tool_result in CreateMessageWithToolsResult")
+	}
+}
+
+func TestToBase_Conversion(t *testing.T) {
```
maciej-kisiel: Move the remaining test functions in this file to `protocol_test.go`?

mcp/client.go
findleyr: Done — moved the `Tools` inference inside the `caps.Sampling == nil` block so it only applies when capabilities are auto-inferred.

mcp/content.go
findleyr: They serve different purposes: `ToolUseContent.Input` is a flat JSON object (tool arguments), so `map[string]any` is the natural representation — and we normalize `nil` to `{}` on marshal to satisfy the spec's required field. `ToolResultContent.Content` is a list of typed content blocks (`[]Content`), which goes through `contentsFromWire` for proper type dispatch. The spec defines them the same way structurally (both are required), but their Go representations differ because of the nested content semantics.

mcp/content.go
findleyr: Done — annotated every field in `wireContent` with the content type(s) that use it.

mcp/protocol.go
findleyr: Done — `toBase()` now returns an error if any message has multiple content blocks, rather than silently dropping them.

mcp/server.go
findleyr: Good question. Neither `CreateMessage` nor `CreateMessageWithTools` currently validates capabilities, and `checkInitialized` is itself still a TODO. I think adding capability validation is worth doing, but it's broader scope — it should probably cover all capability-gated server→client requests consistently. Want me to file an issue for this?

mcp/sampling_tools_test.go
findleyr: Done. Restructured all tests per your suggestions:
- Moved `ToolUseContent`/`ToolResultContent` marshal/unmarshal tests to `protocol_test.go` (needs internal access to `wireContent`/`contentFromWire`, so `content_test.go` with `package mcp_test` wouldn't work)
- Moved `SamplingMessage`, `SamplingCapabilities`, `SamplingMessageV2`, `CreateMessageWithToolsParams`, `CreateMessageWithToolsResult`, `ToolChoice`, `toBase`, `toWithTools` tests to `protocol_test.go`
- Used table-driven tests throughout
- `ToolChoice` test uses `DeepEqual` on whole structs via marshal→unmarshal round-trip
- Split `SamplingCapabilities` into `_MarshalJSON`/`_UnmarshalJSON`
- Consolidated result tests into `TestCreateMessageWithToolsResult_MarshalJSON`/`_UnmarshalJSON` (5 table cases including null, empty array, rejects tool_result)
- Fixed naming: `TestSamplingToolsCapabilities`, `"client with explicit tools capability"`, `"client with inferred tools capability"`
- Added preceding messages to `TestSamplingToolResultWithError_Integration`

anowardear062-svg:
Thanks

mcp/server.go
maciej-kisiel: Please do. I just wonder, given this is a new handler that it would be easier to validate than to introduce a behavior change later on, but as you say, if we want to do it consistently we will need to face this challenge anyways.

mcp/content.go
maciej-kisiel: My comment was more about `ToolResultContent.StructuredContent` and the differentiation between `any` and `map[string]any`. But after some thought, I think this is because we want to be able let the user put any result object here, that is marshalable to a JSON object, and not an intermediate data structure that would enforce correct schema. I just wonder if we ever enforce that this needs to be a JSON object vs. any other JSON value.

mcp/protocol_test.go
```diff
@@ -488,6 +489,587 @@ func TestCompleteResult(t *testing.T) {
 	}
 }
 
+func TestToolUseContent_MarshalJSON(t *testing.T) {
+	tests := []struct {
+		name    string
+		content *ToolUseContent
+		want    *ToolUseContent
```
maciej-kisiel: Maybe it would be simpler to use the expected JSON string as `want`, similarly to the `TestToolResult` case?

maciej-kisiel (approved):

PR #702, opened 2025-12-10
docs: add GitHub Pages redirect to latest release documentation

jonathanhefner:
**Context:** I am working on getting documentation up for each MCP SDK at `https://modelcontextprotocol.github.io/*-sdk/`.  I realize the Go ecosystem has its own idioms for generating and hosting documentation, so I am proposing we make https://modelcontextprotocol.github.io/go-sdk/ redirect to the latest tag of https://github.com/modelcontextprotocol/go-sdk/blob/v1.1.0/docs/README.md.

---

Adds a redirect page at https://modelcontextprotocol.github.io/go-sdk/ that automatically redirects visitors to the `docs/README.md` of the latest release tag. Uses Jekyll templating with `site.github.latest_release` to dynamically resolve the current release version at build time.

### DIFF
--- a/.github/workflows/publish-docs.yml
+++ b/.github/workflows/publish-docs.yml
@@ -1,0 +1,24 @@
+name: Publish Docs to GitHub Pages
+
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
+
+permissions:
+  contents: write
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@08eba0b27e820071cde6df949e0beb9ba4906955 # v4
+
+      - name: Deploy to GitHub Pages
+        uses: peaceiris/actions-gh-pages@4f9cc6602d3f66b9c108549d475ec49e8ef4d45e # v4
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          publish_dir: ./docs
+          publish_branch: gh-pages
+          enable_jekyll: true
--- a/docs/index.html
+++ b/docs/index.html
@@ -1,0 +1,15 @@
+---
+---
+<!DOCTYPE html>
+<html>
+<head>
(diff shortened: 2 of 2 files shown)


findleyr:
Thanks! Will we be the only SDK with this solution? I think we'd actually wanted to contribute official docs, but are slim on bandwidth at the moment.

Either way, this is a fine solution, even if only temporary.

Will ask a colleague to review the PR, or I can get to it tomorrow.

jonathanhefner:
> Thanks! Will we be the only SDK with this solution? I think we'd actually wanted to contribute official docs, but are slim on bandwidth at the moment.

Yes, but I've ended up having to take a unique approach for each SDK (because of ecosystem idioms, as well as maintainer buy-in), so I think that's fine.

My naive understanding was that hosting markdown files directly in GitHub's UI was typical for Go packages, but I am not a regular Go developer.  If you would prefer to have the `docs/` markdown files fully rendered at https://modelcontextprotocol.github.io/go-sdk/ instead of a redirect, I can pursue that approach.

findleyr:
> My naive understanding was that hosting markdown files directly in GitHub's UI was typical for Go packages, but I am not a regular Go developer.

I think Go projects also want to use nicely formatted HTML when they can, but markdown works in a pinch. It's only typical in the sense that Go developers often prefer the simplest solution, even if it's not pretty :)

Many projects rely solely on their package documentation, but that will not suffice for correlating MCP features with API. This is why we added the markdown docs. Your solution is fine for now, and thanks again for contributing.

Took a quick skim, and this all looks very reasonable, but I'm a GitHub noob, so need to make sure I understand the implications of permissions, etc (or just verify that this is consistent with other SDKs). Will review tomorrow if @jba or @markus-kusano don't get to it.

findleyr:
LGTM, will merge later / tomorrow (want to run and test the action after merging, possibly tuning it, and can't now).

PR #703, opened 2025-12-10
mcp: add documentation for the streamable implementation

findleyr:
Two commits to be rebased:
- add a constant for lastEventID
- add significant documentation

See individual commits for details.

### DIFF
--- a/mcp/streamable.go
+++ b/mcp/streamable.go
@@ -2,6 +2,10 @@
 // Use of this source code is governed by an MIT-style
 // license that can be found in the LICENSE file.
 
+// NOTE: see streamable_server.go and streamable_client.go for detailed
+// documentation of the streamable server design.
+// TODO(cleanup): move the client and server logic into those files.
+
 package mcp
 
 import (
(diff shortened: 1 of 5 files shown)


mcp/streamable.go
```diff
@@ -895,20 +910,23 @@ func (c *streamableServerConn) acquireStream(ctx context.Context, w http.Respons
...
-		if err := ctx.Err(); err != nil {
-			return err
-		}
 		if final {
 			defer close(done)
 		}
 		return c.writeEvent(w, s.id, Event{Name: "message", Data: data}, lastIdx)
 	}
+	// TODO(cleanup): consolidate closeLocked functions, and be consistent about
+	// checking ctx.Err().
```
markus-kusano, line 929: nit, feel free to ignore: should be `ctx.Done()` I think since we changed the .Err() check to a .Done() check on line 919.

markus-kusano (approved):

mcp/streamable.go
findleyr: Ack, will consider this when I get to this cleanup item.

PR #704, opened 2025-12-10
mcp: simplify stream state management in streamable transport

findleyr:
Replace closure-based deliver and closeLocked fields with explicit state:
- pendingJSONMessages: signals JSON mode and collects messages
- w: indicates stream is claimed by HTTP request
- done: signals stream completion (non-nil implies w is non-nil)
- lastIdx, supportsPrimeClose, logger: moved from closures to fields

Add stream methods to consolidate logic:
- close: sends close event and closes done channel
- release: releases stream from HTTP request for resumption
- deliverLocked: writes/stores messages, handles request accounting

Factor out hangResponse to document HTTP response lifecycle.

Inline writeEvent to eliminate confusing pointer-based index management.

### DIFF
--- a/mcp/streamable.go
+++ b/mcp/streamable.go
@@ -618,22 +618,39 @@ type stream struct {
 	// The standalone SSE stream has id "".
 	id string
 
+	// logger is used for logging errors during stream operations.
+	logger *slog.Logger
+
 	// mu guards the fields below, as well as storage of new messages in the
 	// connection's event store (if any).
 	mu sync.Mutex
 
-	// If non-nil, deliver writes data directly to the HTTP response.
+	// If pendingJSONMessages is non-nil, this is a JSON stream and messages are
+	// collected here until the stream is complete, at which point they are
+	// flushed as a single JSON response.
+	pendingJSONMessages []json.RawMessage
+
+	// w is the HTTP response writer for this stream. A non-nil w indicates
+	// that the stream is claimed by an HTTP request; it is set to nil when
+	// the request completes.
+	w http.ResponseWriter
+
+	// done is closed to release the hanging HTTP request.
 	//
-	// Only one HTTP response may receive messages at a given time. An active
-	// HTTP connection acquires ownership of the stream by setting this field.
-	deliver func(data []byte, final bool) error
+	// Invariant: a non-nil done implies w is also non-nil, though the converse
+	// is not necessarily true: done is set to nil when it is closed, to avoid
+	// duplicate closure.
+	done chan struct{}
 
-	// closeLocked sends a 'close' event to the client with a configurable retry
-	// delay, if there is a delivery channel available. The stream must be
-	// locked.
-	closeLocked func(reconnectAfter time.Duration)
+	// lastIdx is the index of the last written SSE event, for event ID generation.
+	// It starts at -1 since indices start at 0.
+	lastIdx int
 
-	// streamRequests is the set of unanswered incoming requests for the stream.
+	// supportsPrimeClose indicates whether the client supports prime and close
+	// events (protocol version 2025-11-25 or later).
+	supportsPrimeClose bool
+
+	// requests is the set of unanswered incoming requests for the stream.
 	//
 	// Requests are removed when their response has been received.
 	// In practice, there is only one request, but in the 2025-03-26 version of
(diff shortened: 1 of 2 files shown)


mcp/streamable.go
```diff
@@ -642,12 +659,92 @@ type stream struct {
...
+	s.done = nil
+}
+
+// release releases the stream from its HTTP request, allowing it to be
+// claimed by another request (e.g., for resumption).
+func (s *stream) release() {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	s.w = nil
+	s.done = nil // may already be nil, if the stream is done or closed
```
markus-kusano, line 692: why is it OK to not `close(s.done)` here? i think it is because release() is only used below after hangresponse() returns, so the stream is finished, but first glance looking at this function in isolation from how its used it feels like the close() is missing

findleyr: s.done just signals to terminate the hanging response. Since we're terminating anyway, it doesn't matter! Ok to garbage collect a non-closed channel.

mcp/streamable.go
```diff
@@ -618,22 +618,39 @@ type stream struct {
...
+
 	// mu guards the fields below, as well as storage of new messages in the
 	// connection's event store (if any).
 	mu sync.Mutex
 
-	// If non-nil, deliver writes data directly to the HTTP response.
+	// If pendingJSONMessages is non-nil, this is a JSON stream and messages are
+	// collected here until the stream is complete, at which point they are
+	// flushed as a single JSON response.
+	pendingJSONMessages []json.RawMessage
```
jba, line 631: Is this a slice only because of batching?

mcp/streamable.go
```diff
@@ -618,22 +618,39 @@ type stream struct {
...
+
+	// w is the HTTP response writer for this stream. A non-nil w indicates
+	// that the stream is claimed by an HTTP request; it is set to nil when
```
jba, line 634: What are the alternatives? I believe only the standalone GET.

jba: By which I mean that the replay GETs are associated with a request.

mcp/streamable.go
```diff
@@ -618,22 +618,39 @@ type stream struct {
...
-	// locked.
-	closeLocked func(reconnectAfter time.Duration)
+	// lastIdx is the index of the last written SSE event, for event ID generation.
+	// It starts at -1 since indices start at 0.
+	lastIdx int
 
-	// streamRequests is the set of unanswered incoming requests for the stream.
+	// supportsPrimeClose indicates whether the client supports prime and close
+	// events (protocol version 2025-11-25 or later).
+	supportsPrimeClose bool
```
jba, line 651: wouldn't it be more general to have a version field?

mcp/streamable.go
```diff
@@ -642,12 +659,92 @@ type stream struct {
...
+// The done value is always accurate, even if an error is returned.
+//
+// s.mu must be held when calling this method.
+func (s *stream) deliverLocked(data []byte, eventID string, responseTo jsonrpc.ID) (done bool, err error) {
+	if responseTo.IsValid() {
+		delete(s.requests, responseTo)
+	}
+	done = len(s.requests) == 0 && s.id != ""
+	if s.done == nil {
+		return done, fmt.Errorf("stream not connected or already closed")
```
jba, line 712: I would normally put this at the top of the function, to fail as early as possible. Is there a reason to remove the request ID from the map before failing here?

mcp/streamable.go
```diff
@@ -1092,44 +1142,16 @@ func (c *streamableServerConn) servePOST(w http.ResponseWriter, req *http.Reques
...
-				toWrite, err = json.Marshal(msgs)
-				if err != nil {
-					return err
-				}
-			}
-			_, err = w.Write(toWrite)
-			return err
-		}
+		// JSON mode: collect messages in pendingJSONMessages until done.
+		stream.pendingJSONMessages = []json.RawMessage{}
```
jba, line 1152: why is this the empty slice instead of nil? Is it so that it will render as the empty JSON array?

mcp/streamable.go
findleyr: Yes, only because of batching. Otherwise, it could be a bool. Adding a note.

mcp/streamable.go
findleyr: It could be the initial hanging POST. If the server thinks it's still handing a POST, we don't want to allow a GET to replay the stream. (or maybe we do, but let's not decide that now). Adding a comment.

mcp/streamable.go
findleyr: It's a trick (but I think an elegant one): the non-nilness of the field signals whether this is a JSON response.

I came up with that, not Claude which was much more verbose. Do you like it?

PR #705, opened 2025-12-10
mcp: finish cleanup notes in streamable.go

findleyr:
Address remaining TODO(cleanup) comments in streamable.go: eliminate a pointless context cancellation, and simplify forCall logic.

### DIFF
--- a/mcp/streamable.go
+++ b/mcp/streamable.go
@@ -4,7 +4,7 @@
 
 // NOTE: see streamable_server.go and streamable_client.go for detailed
 // documentation of the streamable server design.
-// TODO(cleanup): move the client and server logic into those files.
+// TODO: move the client and server logic into those files.
 
 package mcp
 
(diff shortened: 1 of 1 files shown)


mcp/streamable.go
```diff
@@ -1674,22 +1673,26 @@ func (c *streamableClientConn) Write(ctx context.Context, msg jsonrpc.Message) e
...
 		//
 		// [§2.1.4]: https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#listening-for-messages-from-the-server
-		resp.Body.Close()
-		return nil
-	} else if !isCall && !c.strict {
-		// Some servers return 200, even with an empty json body.
-		// Ignore this response in non-strict mode.
-		if c.logger != nil {
-			c.logger.Warn(fmt.Sprintf("unexpected status code %d from non-call", resp.StatusCode))
+		if resp.StatusCode != http.StatusNoContent && resp.StatusCode != http.StatusAccepted {
```
markus-kusano, line 1684: Do we want a test for this, or maybe there's one that already covers it? Fine to not put it in this CL if we don't have it, we can leave a TODO

markus-kusano (approved):

mcp/streamable.go
findleyr: We do have a test for this (I initially broke that test!).

TestStreamableClientStrictness.

PR #712, opened 2025-12-10
oauthex: validate URL schemes in auth server metadata and DCR

### LINKED ISSUE #526: protect against attacks from MCP server URLs

findleyr:
Use URL scheme validation for remaining URL fields in auth datastructures (AuthServerMeta and ClientRegistrationMetadata), to prevent XSS attacks.

By comparison, the typescript SDK doesn't validate introspection_endpoint, and I don't know why. Otherwise the implementations align.

Fixes #526


oauthex/auth_meta.go
```diff
@@ -144,9 +144,40 @@ func GetAuthServerMeta(ctx context.Context, issuerURL string, c *http.Client) (*
...
+// validateAuthServerMetaURLs validates all URL fields in AuthServerMeta
+// to ensure they don't use dangerous schemes that could enable XSS attacks.
+func validateAuthServerMetaURLs(asm *AuthServerMeta) error {
+	urls := []struct {
+		name  string
+		value string
+	}{
+		{"authorization_endpoint", asm.AuthorizationEndpoint},
+		{"token_endpoint", asm.TokenEndpoint},
+		{"jwks_uri", asm.JWKSURI},
```
markus-kusano, line 172: do we need to be worried about adding new `asm.*Endpoint` fields and forgetting to update the function here to validate them? that feels like an easy way to accidentally add a XSS bug (add a new endpoint and forget to validate it). seems like this is a standard data structure, though, so perhaps it wont change often unless the oauth spec changes

markus-kusano (dismissed):

oauthex/auth_meta.go
findleyr: Thanks, good point. Added comments as a reminder.

Issue #706, opened 2025-12-10
Proposal: Add {Client,Server}Options.Capabilities, and deprecate Has{Tools,Prompts,Resources}


findleyr:
In #135 we addressed a simple problem--that there was no way to force advertising unused capabilities--by adding the simplest possible API: `ServerOptions.Has{Tools,Prompts,Resources}` allowed server authors to advertise the relevant capability even if there were no features (yet) registered on the server.

However, as I work on the 2025-11-25 version of the spec, particularly #629, I am finding problems with this design.

### Problem statement

- What if the user wants to *customize* those capabilities, for example by disabling the `listChanged` feature?
- Lack of discoverability, and confusing semantics: if I set `HasTools: false`, that doesn't actually mean that my server has no tools, it just means that it has no tools *yet*.
- Explosion of API: as we require further customization of capabilities, we have to add more and more arbitrary fields to the `{Client,Server}Options` structs.

This was already a problem, but is becoming more noticeable with 2025-11-25 version of the spec, which added more capability configuration (URL elicitation and Tool sampling, to name two). For example, in #623 I added `ClientOptions.ElicitationModes`, but didn't like it. Fortunately, we've not yet had a release and so can change that API.

### Proposal

I propose the following solution, which I believe is similar to what the typescript SDK does (CC @felixweinberger for any advice from typescript).

1. Add `{Client.Server}Options.Capabilities`, an optional field that may be set by the API user to configure default capabilities. **This overrides the current default capabilities, which are not currently empty** (we always advertise `"roots"` from the client, and `"logging"` from the server).
2. Whenever the user adds a new capability, for example by adding a tool or prompt, or by setting an `ElicitationHandler`, augment the advertised capabilities to reflect that new capability. So if the user sets `ServerOptions.Capabilities` to `new(ServerCapabilities)`, but then adds a tool, we WILL advertise tool capabilities.
3. Check `listChanged` capabilities before sending notifications, so that server behavior is consistent with its advertised capabilities.
4. **Deprecate** `Has{Tools,Prompts,Resources}`, and **remove** `ClientOptions.ElicitationModes` (which has not yet been released).

That's it.

### Notes

- This means there's no way to e.g. add a tool but not advertise the tool capability (but why would you want to do that?).
[...]

PR #713, opened 2025-12-11
mcp: add Capabilities fields to ServerOptions and ClientOptions

findleyr:
This allows users to configure default capabilities, disable defaults,
or suppress listChanged notifications by setting ListChanged: false.

Key changes:
- Add ServerOptions.Capabilities and ClientOptions.Capabilities fields.
- Merge user-provided capabilities with dynamically-added ones (tools,
  prompts, resources, sampling, elicitation).
- Check ListChanged before sending list-changed notifications.
- Sync Roots from RootsV2 for backward compatibility (https://github.com/modelcontextprotocol/go-sdk/issues/607).
- Deprecate HasPrompts, HasResources, HasTools on ServerOptions.
- Remove unreleased ElicitationModes field from ClientOptions.
- Add clone methods to capability structs using shallowClone helper.
- Add synctest-based tests for notification behavior (go1.25).
- Add TestClientCapabilitiesOverWire and TestServerCapabilitiesOverWire
- Document the new Capabilities API in docs/

Fixes https://github.com/modelcontextprotocol/go-sdk/issues/706

### DIFF
--- a/docs/client.md
+++ b/docs/client.md
@@ -4,6 +4,9 @@
 1. [Roots](#roots)
 1. [Sampling](#sampling)
 1. [Elicitation](#elicitation)
+1. [Capabilities](#capabilities)
+	1. [Capability inference](#capability-inference)
+	1. [Explicit capabilities](#explicit-capabilities)
 
 ## Roots
 
@@ -130,7 +133,9 @@ allows servers to request user inputs. It is implemented in the SDK as follows:
 **Client-side**: To add the `elicitation` capability to a client, set
 [`ClientOptions.ElicitationHandler`](https://pkg.go.dev/github.com/modelcontextprotocol/go-sdk/mcp#ClientOptions.ElicitationHandler).
 The elicitation handler must return a result that matches the requested schema;
-otherwise, elicitation returns an error.
+otherwise, elicitation returns an error. If your handler supports [URL mode
+elicitation](https://modelcontextprotocol.io/specification/2025-11-25/client/elicitation#url-mode-elicitation-requests),
+you must declare that capability explicitly (see [Capabilities](#capabilities))
 
 **Server-side**: To use elicitation from the server, call
 [`ServerSession.Elicit`](https://pkg.go.dev/github.com/modelcontextprotocol/go-sdk/mcp#ServerSession.Elicit).
@@ -171,3 +176,60 @@ func Example_elicitation() {
 	// Output: value
 }
 ```
+
+## Capabilities
+
+Client capabilities are advertised to servers during the initialization
+handshake. [By default](rough_edges.md), the SDK advertises the `logging`
+capability. Additional capabilities are automatically added when server
+features are added (e.g. via `AddTool`), or when handlers are set in the
+`ServerOptions` struct (e.g., setting `CompletionHandler` adds the
+`completions` capability), or may be configured explicitly.
+
+### Capability inference
+
+When handlers are set on `ClientOptions` (e.g., `CreateMessageHandler` or
+`ElicitationHandler`), the corresponding capability is automatically added if
+not already present, with a default configuration.
+
+For elicitation, if the handler is set but no `Capabilities.Elicitation` is
+specified, the client defaults to form elicitation. To enable URL elicitation
+or both modes, [configure `Capabilities.Elicitation`
+explicitly](#explicit-capabilities).
+
+See the [`ClientCapabilities`
+documentation](https://pkg.go.dev/github.com/modelcontextprotocol/go-sdk/mcp#ClientCapabilities)
+for further details on inference.
+
+### Explicit capabilities
+
+To explicitly declare capabilities, or to override the [default inferred
+capability](#capability-inference), set
+[`ClientOptions.Capabilities`](https://pkg.go.dev/github.com/modelcontextprotocol/go-sdk/mcp#ClientOptions.Capabilities).
+This sets the initial client capabilities, before any capabilities are added
+based on configured handlers. If a capability is already present in
+`Capabilities`, adding a handler will not change its configuration.
+
+This allows you to:
+
+- **Disable default capabilities**: Pass an empty `&ClientCapabilities{}` to
+  disable all defaults, including roots.
+- **Disable listChanged notifications**: Set `ListChanged: false` on a
+  capability to prevent the client from sending list-changed notifications
+  when roots are added or removed.
+- **Configure elicitation modes**: Specify which elicitation modes (form, URL)
+  the client supports.
+
+```go
+// Configure elicitation modes and disable roots.
+client := mcp.NewClient(impl, &mcp.ClientOptions{
+    Capabilities: &mcp.ClientCapabilities{
+        Elicitation: &mcp.ElicitationCapabilities{
+            Form: &mcp.FormElicitationCapabilities{},
+            URL:  &mcp.URLElicitationCapabilities{},
+        },
+    },
+    ElicitationHandler: handler,
+})
+```
+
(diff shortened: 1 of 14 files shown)


mcp/client.go
```diff
@@ -130,26 +151,40 @@ type ClientSessionOptions struct {
...
-			// {"form":{}} for backward compatibility, but we explicitly set the form
-			// capability.
-			caps.Elicitation.Form = &FormElicitationCapabilities{}
-		}
-		if slices.Contains(modes, "url") {
-			caps.Elicitation.URL = &URLElicitationCapabilities{}
+		if caps.Elicitation == nil {
+			// Default to form elicitation for backward compatibility.
+			caps.Elicitation = &ElicitationCapabilities{
+				Form: &FormElicitationCapabilities{},
```
findleyr: TODO: should we leave '"form"' blank here? The default is interpreted as form capabilities.

Maybe we should pass in the protocol version? Look at what other SDKs do

mcp/client.go
```diff
@@ -345,12 +380,38 @@ func changeAndNotify[P Params](c *Client, notification string, params P, change
...
 	}
 	c.mu.Unlock()
 	notifySessions(sessions, notification, params, c.logger)
 }
 
+// shouldSendListChangedNotification checks if the client's capabilities allow
+// sending the given list-changed notification.
+func (c *Client) shouldSendListChangedNotification(notification string) bool {
+	// Get effective capabilities (considering user-provided defaults).
+	caps := c.opts.Capabilities
```
findleyr, line 417: Note that this is effectively duplicating the capability inference logic. Is that OK?

mcp/server.go
```diff
@@ -80,14 +80,28 @@ type ServerOptions struct {
...
+	//
+	// Set to an empty &ServerCapabilities{} to disable all default capabilities.
+	// For example, to disable listChanged notifications:
+	//
+	//	Capabilities: &ServerCapabilities{
+	//	    Tools: &ToolCapabilities{ListChanged: false},
+	//	}
+	Capabilities *ServerCapabilities
+
+	// Deprecated: Use Capabilities instead. If true, advertises the prompts
```
findleyr: We should probably leave the original doc comment, not just the deprecation notice. The user still needs to know how to interpret these fields.

docs/client.md
```diff
@@ -171,3 +172,45 @@ func Example_elicitation() {
 	// Output: value
 }
 ```
+
+## Capabilities
+
+Client capabilities are advertised to servers during the initialization
+handshake. By default, the SDK advertises the `roots` capability with
```
markus-kusano: Should we add a note why the SDK advertises `roots` by default? i.e., should we mention here that this was somewhat of a mistake, maybe just link to an issue that explains it, or the text below ("**workaround**"? I'm asking since we might change the `roots` by default on a new version of the SDK so maybe we want to not lock in the roots default in the docs here

docs/rough_edges.md
```diff
@@ -36,3 +36,11 @@ v2.
 
   **Workaround**: use `ClientCapabilities.RootsV2`, which aligns with the
   semantics of other capability fields.
+
+- Default capabilities should have been empty. Instead, servers default to
+  advertising `logging`, and clients default to advertising `roots` with
+  `listChanged: true`. This is confusing because a nil `Capabilities` field
+  does not mean "no capabilities".
+
+  **Workaround**: set `ServerOptions.Capabilities` or `ClientOptions.Capabilities`
```
markus-kusano: IMO it would be slightly clearer to be very explicit here about the difference between an empty Capabilities and a `nil` capabilities.

I'd reword here to: "**Workaround**: to advertise no capabilities set `ServerOptions`... to an empty ..."

docs/server.md
```diff
@@ -535,6 +536,43 @@ func Example_logging() {
 }
 ```
 
+### Capabilities
+
+Server capabilities are advertised to clients during the initialization
+handshake. By default, the SDK advertises only the `logging` capability.
```
markus-kusano, line 544: ditto about why logging is advertised by default

docs/client.md
```diff
@@ -171,3 +172,45 @@ func Example_elicitation() {
...
+`listChanged: true`. Additional capabilities are automatically added when
```
markus-kusano: think one confusion is that there's some subtleties when capabilities *must* be set, i.e., automatic inference doesn't handle these cases.  This would be the implicit defaults (i.e., when I add no handlers it doesn't make sense that automatic inference infers roots capabilities), and lack of URL elicitation advertisement (if I add a handler supporting URL elicitation then automatic inference will fail to notice that I added support for this).

My reasoning for this is a busy dev might have the following thought process reading these docs quickly:

1. *starts reading Cappabilities section here*
2. oh shoot, I need to setup capabilities to make my server work
3. *finishes first paragraph here*
4. nevermind, go's SDK is so great it automatically infers capabilities, I don't need to think about capabilities anymore

this could put them in a state where later down the line they're struggling with the SDK because they don't understand Capabilities, but its because our automatic inference abstraction is sort of leaking and only half-way taking care of the job for them.

mcp/client.go
markus-kusano: duplicating my comment from our 1/1:  I'm not sure if its safer to use both (include the old fields and new fields), or be strict (if we run on an older protocol use the older fields only). I would guess in general people would want both: we might want to be strict and throw errors when the protocol does something unexpected, or we might want to handle as many different features at the same time (be loose with the spec) in order to get more traffic through

mcp/protocol.go
```diff
@@ -208,6 +208,29 @@ type ClientCapabilities struct {
 	Elicitation *ElicitationCapabilities `json:"elicitation,omitempty"`
 }
 
+// clone returns a deep copy of the ClientCapabilities.
+func (c *ClientCapabilities) clone() *ClientCapabilities {
```
markus-kusano, line 216: Is this another function we need to remember to keep in sync with the `ClientCapabilities` definition? As in, if we add more fields to `ClientCapabilities` then we'll need to update the clone logic here. Maybe this deserves a comment?

mcp/server.go
```diff
@@ -80,14 +80,28 @@ type ServerOptions struct {
...
 	UnsubscribeHandler func(context.Context, *UnsubscribeRequest) error
-	// If true, advertises the prompts capability during initialization,
-	// even if no prompts have been registered.
+
+	// Capabilities configures the server's default capabilities.
+	// If non-nil, this overrides the default capabilities (which includes
+	// only Logging). Features like tools, prompts, and resources are
+	// automatically added to the capabilities when registered.
+	//
+	// Set to an empty &ServerCapabilities{} to disable all default capabilities.
```
markus-kusano: Is it more precise to say that setting to an empty ServerCapabilities will disable *all* capabilities, not just the defaults?

findleyr: Rewrote this entire section, PTAL.

mcp/server.go
findleyr: Done.

mcp/protocol.go
findleyr: Added a note in the corresponding struct.

mcp/client.go
findleyr: Passed in protocol version.

docs/client.md
findleyr: Done. Rewrote this by hand.

docs/client.md
findleyr: I tried to clarify this significantly. Rewrote. PTAL.

docs/rough_edges.md
findleyr: Done.

docs/server.md
findleyr: Done.

jba (dismissed):

Issue #715, opened 2025-12-11
Proposal: expose ErrSessionMissing


findleyr:
The MCP spec states that the streamable server may terminate the session at any time, and when the client encounters a 404 it should initialize a new session:
https://modelcontextprotocol.io/specification/2025-11-25/basic/transports#session-management

The SDK doesn't currently expose an easy way to achieve this: for consistency with other transports, we do not transparently reinitialize the session for the user: a `ClientSession` corresponds to a logical session on the server, and if it is broken a new `ClientSession` must be created.

However, there's no easy way to detect if the session is broken, because our internal sentinel error value for this is not exported.

I propose that we export `ErrSessionMissing`, a sentinel error value that is wrapped by Write errors from the `streamableClientConn` when an HTTP request returns `NotFound`.

An alternative would be to expose the HTTP errors directly (see also #579), but that is a rather large API surface, and tightly couples the client logic to the transport. I would prefer to just expose this one error value. We can add more as needed.

CC @dpasiukevich @jba @markus-kusano

findleyr:
In https://github.com/modelcontextprotocol/go-sdk/pull/719#issuecomment-3646681481, [dpasiukevich](https://github.com/dpasiukevich) points out that the pre-existing `ErrConnectionClosed` might be sufficient.

So for that reason, let's hold off on this proposal. Moving out of the release milestone.

findleyr:
I think we can close this and reopen if there's a clear need.

dpasiukevich:
@findleyr I did more experiments, and it's possible to hit this. With this new `mcp.ErrSessionMissing` and already existing `mcp.ErrConnectionClosed` it should be sufficient to have a proper automatic session restart. Right now we'd use `strings.Contains(err.Error(), "Go")` temporarily.

It would be great to expose this error type.

Apologies for the false negative earlier.

findleyr:
CC @maciej-kisiel 

OK! I think we can expose this.

PR #719, opened 2025-12-11
mcp: export ErrSessionMissing for session termination detection

findleyr:
Export ErrSessionMissing as a sentinel error that clients can use to detect when a server has terminated or lost their session.

This allows users to implement session recovery logic, creating a new ClientSession when needed.

Updates #715

### DIFF
--- a/mcp/streamable.go
+++ b/mcp/streamable.go
@@ -1376,6 +1376,10 @@ func (c *streamableServerConn) Close() error {
 // A StreamableClientTransport is a [Transport] that can communicate with an MCP
 // endpoint serving the streamable HTTP transport defined by the 2025-03-26
 // version of the spec.
+//
+// If the server terminates the session (returning 404 Not Found), subsequent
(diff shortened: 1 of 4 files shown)


findleyr:
@dpasiukevich, please let me know if you have any opinion on this PR or API.

markus-kusano (approved):

dpasiukevich:
I've did some experiments. It seems like if client times out (e.g. by cancelled context) or server terminated a session (e.g. via by setting StreamableHTTPOptions.SessionTimeout).

In all these cases `errors.Is(err, mcp.ErrConnectionClosed)` returns true. So maybe exposing `mcp.ErrSessionMissing` may be excessive?

findleyr:
Thanks for testing, @dpasiukevich!

In that case, let's hold off on this until there's a clearer need.

findleyr:
Let's close this since there isn't a clear need.

PR #721, opened 2025-12-12
oauthex: fix content type check in getJSON

nikolavp:
The content type header contains full mimetype string which might contain parameters. The getJSON function doesn't account for this and so some responses fail to parse although they are perfectly acceptable. For example 
```
curl -vv 'https://github.com/.well-known/oauth-authorization-server/login/oauth'
```
returns a content type header with the value
```
content-type: application/json; charset=utf-8
```
which is perfectly valid but fails in the current logic.

### DIFF
--- a/oauthex/oauth2.go
+++ b/oauthex/oauth2.go
@@ -13,6 +13,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"mime"
 	"net/http"
 	"net/url"
 	"strings"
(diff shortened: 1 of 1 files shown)


oauthex/oauth2.go
```diff
@@ -57,7 +57,7 @@ func getJSON[T any](ctx context.Context, c *http.Client, url string, limit int64
 		return nil, fmt.Errorf("bad status %s", res.Status)
 	}
 	// Specs require application/json.
-	if ct := res.Header.Get("Content-Type"); ct != "application/json" {
+	if ct := res.Header.Get("Content-Type"); !strings.Contains(ct, "application/json") {
```
jba: The problem is that this will also match "x-application/jsonfoo" or anything else of that form.

Can we cut off everything from ";" to the end and compare the prefix?

nikolavp: yeah, sounds good. Thanks for pointing that out. Decided to go with the mime parser from the standard library, tell me if you still prefer the easier/simpler split and then checking the first part.

yousifani16-source: I think the easiest is the best

PR #729, opened 2025-12-18
mcp: add DisableListening option to StreamableClientTransport

### LINKED ISSUE #728: Proposal: Add option to disable continuous listening in StreamableClientTransport
## Is your feature request related to a problem? Please describe.

Yes, I'm experiencing issues with the current implementation where `StreamableClientTransport` always establishes a standalone SSE (Server-Sent Events) connection after initialization, which causes problems in several scenarios:

1. **Unnecessary resource usage**: In many use cases, especially in scientific computing scenarios, I only need simple request-response communication. I don't need to receive server-initiated notifications, so maintaining a persistent SSE connection is wasteful.

2. **Server compatibility issues**: Some third-party MCP servers don't properly handle GET requests for SSE streams. When the client automatically tries to establish the SSE connection, it fails or hangs, blocking the entire connection process. This is particularly problematic when using third-party servers that I cannot modify. (Related to issue #634)

3. **Lack of user control**: Currently, there's no way to opt out of the standalone SSE stream. The SDK always attempts to establish it, which doesn't align with the MCP specification that states the standalone SSE stream is optional.

## Describe the solution you'd like

I would like to add a `DisableListening` boolean field to `StreamableClientTransport` that allows users to control whether the client establishes the standalone SSE stream. 

- Default value should be `false` (maintains current behavior for backward compatibility)
- When set to `true`, the client should skip establishing the standalone SSE connection
- Add a convenience function `WithDisableListening()` similar to other transport options

This would give users explicit control over SSE connection establishment, similar to how `github.com/mark3labs/mcp-go` implements `getListeningEnabled` (though with inverse semantics to maintain backward compatibility).

## Describe alternatives you've considered

1. **Always disable by default**: This would break backward compatibility and isn't desirable.

2. **Use a pointer type**: Could use `*bool` to distinguish between "not set" (default) and "explicitly false", but a simple `bool` with default `false` is cleaner and maintains backward compatibility.

3. **Separate transport type**: Creating a new transport type would be overkill for this simple option.

4. **Environment variable or config file**: Less idiomatic in Go and doesn't provide programmatic control.

The proposed solution (simple `bool` field with default `false`) is the most straightforward and maintains backward compatibility while providing the needed flexibility.

## Additional context

- This feature is already implemented in `github.com/mark3labs/mcp-go` as `getListeningEnabled` (defaults to false, requires explicit enablement)
- Related issue: #634 discusses similar server compatibility problems
- The MCP specification states that the standalone SSE stream is optional: "The client MAY issue an HTTP GET to the MCP endpoint"
- Reference: https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#listening-for-messages-from-the-server

zxxf18:
mcp: add DisableListening option to control standalone SSE stream

Currently, StreamableClientTransport always establishes a standalone SSE connection after initialization to receive server-initiated messages. While this is useful for receiving notifications like ToolListChangedNotification, it causes problems in several real-world scenarios:

1. **Unnecessary resource usage**: Many use cases (for me), especially in scientific computing scenarios, only require simple request-response communication. Clients don't need server-initiated notifications, making the persistent SSE connection wasteful.

2. **Server compatibility issues**: Some third-party MCP servers don't properly handle GET requests for SSE streams. When the client automatically tries to establish the SSE connection, it fails or hangs, blocking the entire connection process. This is particularly problematic when using third-party servers that cannot be modified. This issue is similar to what was discussed in issue #634.

3. **Lack of user control**: The MCP specification states that the standalone SSE stream is optional ("The client MAY issue an HTTP GET"), but the current SDK implementation doesn't provide a way to opt out. This implementation is also inconsistent with other MCP SDKs like github.com/mark3labs/mcp-go, which provides getListeningEnabled control (defaults to false, requires explicit enablement).

This change adds a DisableListening boolean field to StreamableClientTransport that allows users to control whether the client establishes the standalone SSE stream. The default value is false, maintaining backward compatibility with current behavior. When set to true, the client skips establishing the standalone SSE connection.

The implementation includes DisableListening bool field in StreamableClientTransport, WithDisableListening() convenience function, conditional logic in sessionUpdated() to respect the option, and comprehensive tests in TestStreamableClientDisableListening.

Fixes #728

### DIFF
--- a/mcp/streamable.go
+++ b/mcp/streamable.go
@@ -1389,19 +1389,25 @@ type StreamableClientTransport struct {
 	// It defaults to 5. To disable retries, use a negative number.
 	MaxRetries int
 
-	// DisableListening disables receiving server-to-client notifications when no request is in flight.
+	// DisableStandaloneSSE controls whether the client establishes a standalone SSE stream
+	// for receiving server-initiated messages.
 	//
-	// By default, the client establishes a standalone long-live GET HTTP connection to the server
-	// to receive server-initiated messages (like ToolListChangedNotification).
-	// https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#listening-for-messages-from-the-server
-	// NOTICE: Even if continuous listening is enabled, the server may not support this feature.
+	// When false (the default), after initialization the client sends an HTTP GET request
+	// to establish a persistent server-sent events (SSE) connection. This allows the server
+	// to send messages to the client at any time, such as ToolListChangedNotification or
+	// other server-initiated requests and notifications. The connection persists for the
+	// lifetime of the session and automatically reconnects if interrupted.
 	//
-	// If false (default), the client will establish the standalone SSE stream.
-	// If true, the client will not establish the standalone SSE stream and will only receive
-	// responses to its own requests.
+	// When true, the client does not establish the standalone SSE stream. The client will
+	// only receive responses to its own POST requests. Server-initiated messages will not
+	// be received.
 	//
-	// Defaults to false to maintain backward compatibility with existing behavior.
-	DisableListening bool
+	// According to the MCP specification, the standalone SSE stream is optional.
+	// Setting DisableStandaloneSSE to true is useful when:
+	//   - You only need request-response communication and don't need server-initiated notifications
+	//   - The server doesn't properly handle GET requests for SSE streams
+	//   - You want to avoid maintaining a persistent connection
+	DisableStandaloneSSE bool
 
 	// TODO(rfindley): propose exporting these.
 	// If strict is set, the transport is in 'strict mode', where any violation
@@ -1429,28 +1435,6 @@ var (
 	// Mutable for testing.
 	reconnectInitialDelay = 1 * time.Second
 )
-
-// WithDisableListening disables receiving server-to-client notifications when no request is in flight.
-//
-// By default, the client establishes a standalone long-live GET HTTP connection to the server
-// to receive server-initiated messages. This function disables that behavior.
-//
-// If you want to disable continuous listening, you can either:
-//
-//	transport := &mcp.StreamableClientTransport{
-//		Endpoint: "http://localhost:8080/mcp",
-//		DisableListening: true,
-//	}
-//
-// Or use this convenience function:
-//
-//	transport := &mcp.StreamableClientTransport{
-//		Endpoint: "http://localhost:8080/mcp",
-//	}
-//	mcp.WithDisableListening(transport)
-func WithDisableListening(transport *StreamableClientTransport) {
-	transport.DisableListening = true
-}
 
 // Connect implements the [Transport] interface.
 //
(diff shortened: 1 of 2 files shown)


mcp/streamable.go
```diff
@@ -1389,6 +1389,20 @@ type StreamableClientTransport struct {
 	// It defaults to 5. To disable retries, use a negative number.
 	MaxRetries int
 
+	// DisableListening disables receiving server-to-client notifications when no request is in flight.
+	//
+	// By default, the client establishes a standalone long-live GET HTTP connection to the server
```
findleyr: long-lived

mcp/streamable.go
```diff
@@ -1389,6 +1389,20 @@ type StreamableClientTransport struct {
 	// It defaults to 5. To disable retries, use a negative number.
 	MaxRetries int
 
+	// DisableListening disables receiving server-to-client notifications when no request is in flight.
```
findleyr: "disables the standalone GET request that receives server-initiated messages and notifications."

* it's not just notifications: it can also be requests
* it's not related to whether a request is in flight: a request could be in flight, and yet the server sends messages unassociated with that request.

findleyr: Should we just call this "DisableStandaloneSSE" to be clear? What do other SDKs call it, if they have such a setting?

mcp/streamable.go
```diff
@@ -1389,6 +1389,20 @@ type StreamableClientTransport struct {
...
+	// By default, the client establishes a standalone long-live GET HTTP connection to the server
+	// to receive server-initiated messages (like ToolListChangedNotification).
+	// https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#listening-for-messages-from-the-server
+	// NOTICE: Even if continuous listening is enabled, the server may not support this feature.
+	//
+	// If false (default), the client will establish the standalone SSE stream.
+	// If true, the client will not establish the standalone SSE stream and will only receive
+	// responses to its own requests.
+	//
+	// Defaults to false to maintain backward compatibility with existing behavior.
```
findleyr: Let's omit this last sentence: it's documenting the change (adding a new field), not the steady state.

It defaults to false because we thought the SSE stream should be established by default (which IIRC is consistent with other SDKs). I don't think we need to caveat it.

mcp/streamable.go
```diff
@@ -1416,6 +1430,28 @@ var (
...
+//		DisableListening: true,
+//	}
+//
+// Or use this convenience function:
+//
+//	transport := &mcp.StreamableClientTransport{
+//		Endpoint: "http://localhost:8080/mcp",
+//	}
+//	mcp.WithDisableListening(transport)
+func WithDisableListening(transport *StreamableClientTransport) {
```
findleyr: Why would we have this function when there's already a field on the StreamableClientTransport?

mcp/streamable.go
zxxf18: Updated the description of DisableListening to clarify that it disables the standalone SSE connection for receiving server-initiated messages, including both notifications and requests, independent of any in-flight requests.

mcp/streamable.go
zxxf18: On the WithDisableListening() function: You're right about the redundancy. I've removed it to simplify the API.

mcp/streamable.go
zxxf18: I've renamed the field accordingly and updated all references, including tests.

jba:
Is the same name used in other clients, typescript in particular?

zxxf18:
> Is the same name used in other clients, typescript in particular?

After a comprehensive search, I confirm that in the repositories of each programming language under the modelcontextprotocol organization (including the TypeScript version), the variable names DisableStandaloneSSE or disableStandaloneSSE are not used.
Further examination of StandaloneSSE related terms: In the Go SDK's mcp/streamable.go file at line 1536, there is a function named connectStandaloneSSE; in the TypeScript SDK's packages/server/src/server/webStandardStreamableHttp.ts and other files, there are functions named closeStandaloneSSEStream, as well as some variables named receivedCloseStandaloneSSEStream. These terms are primarily used for internal functions and variables, not for exposed configuration options.
Additionally, in third-party MCP SDKs (such as mark3labs/mcp-go), the similar functionality is implemented through a chained call function named WithContinuousListening to support continuous listening mode.

PR #730, opened 2025-12-18
go.mod: upgrade to jsonschema v0.4.1

jba:
DO NOT SUBMIT

Visible changes:

- Inference on structs populates PropertyOrder

- An empty map for Schema.Properties marshals as "{}"

### PR Guideline

You **must** follow the form:

```
net/http: handle foo when bar

Fixes #12345
```

- there is no trailing period

Additionally:

- Markdown is allowed.


findleyr:
Hi, is this ready for review? Do we want to land this for v1.2.0-pre.2? It seems like it should land after the v1.2.0 release, given its sensitivity. WDYT?

jba:
This one was broken. I'll send another one tomorrow or Monday.

I'd be happy to delay it after 1.2.

Issue #726, opened 2025-12-15
Proposal: make maxTokenSize of client transport SSE event handling configurable


dingliang88526-dev:
mcp/event.go

func scanEvents(r io.Reader) iter.Seq2[Event, error] {
	scanner := bufio.NewScanner(r)
	const maxTokenSize = 1 * 1024 * 1024 // 1 MiB max line size
	scanner.Buffer(nil, maxTokenSize)
        ...
        if err := scanner.Err(); err != nil {
			if errors.Is(err, bufio.ErrTooLong) {
				err = fmt.Errorf("event exceeded max line length of %d", maxTokenSize)
			}
			if !yield(Event{}, err) {
				return
			}
		}

}

why  set a small and no-configurable limit?

findleyr:
Is this actually causing you problems? Our assumption is that the in-memory event store is not really practical for production applications.

Of course, we could make this configurable, but need to understand the use-case first.

zxxf18:
I have also been affected by this 1MB limit. Here is my business scenario:

I registered a tool used to list a resource containing a list of server endpoint information descriptions on the server side. Each resource entry includes descriptive information of varying lengths (some long, some short).

When I enter a relatively large amount of server information (currently around 0.87MB), constructing the JSON-RPC response body—according to the protocol specification, where structured data is placed in StructuredContent and the JSON-serialized data is placed in Content—exceeds the 1MB limit.

It is expected that my list will continue to grow, so this 1MB restriction does have a practical impact on me.

Of course, I could implement alternative strategies at the business level, but I similarly believe that this limit should be exposed as a configurable option.

zxxf18:
My work involves life science research, particularly in the area of complex molecular analysis. These analyses often include sequences and corresponding analytical content that can easily exceed the 1 MB limit.

As a result, I am currently forced to fork a copy of the code and modify the restriction myself in order to continue using it.

findleyr:
@jba @markus-kusano any objection to making this configurable?

We did design the memory event store to be somewhat production grade (that's why we set a maximum cache size in the first place), so making this configurable seems OK to me.

findleyr:
In my haste, I misunderstood the report, FWIW: the issue is not about MemoryEventStore, but rather our event scanner.

Agree that we should fix this. Probably 1MB is too small a default, as well.

findleyr:
[...]

Issue #683, opened 2025-12-04
StreamableClientTransport: Connection poisoned by transient errors


Agustin-Jerusalinsky:
## Problem

The `StreamableClientTransport` permanently breaks after any transient error, making the connection unusable for all subsequent requests. **This affects both stateful and stateless server modes.**

Transient errors include:
- Network timeouts (context deadline exceeded)
- HTTP errors (503 Service Unavailable, 500 Internal Server Error, 502 Bad Gateway)
- Network interruptions (connection refused, connection reset)

After a single transient error occurs, **all subsequent requests fail** with "client is closing", requiring a full reconnection. This affects production systems where temporary issues (server restarts, network glitches, load spikes) should be recoverable.

---

## 1. What did you do?

Created an MCP client using `StreamableClientTransport` and made three sequential tool calls, where the second call encounters a transient error (timeout, 503, network interruption, etc.):

```go
httpClient := &http.Client{Timeout: /* some timeout */}
transport := &mcp.StreamableClientTransport{
    Endpoint:   serverURL,
    HTTPClient: httpClient,
}

session, _ := client.Connect(ctx, transport, nil)

// Call 1: Fast tool - should succeed
result1, err1 := session.CallTool(ctx, &mcp.CallToolParams{Name: "delay_tool"})

// Call 2: Encounters transient error (timeout, 503, etc.)
result2, err2 := session.CallTool(ctx, &mcp.CallToolParams{Name: "delay_tool"})

// Call 3: Fast tool - should succeed
result3, err3 := session.CallTool(ctx, &mcp.CallToolParams{Name: "delay_tool"})
```

## 2. What did you see?

```
Call #1: SUCCESS
Call #2: FAILED - calling "tools/call": sending "tools/call": Post "http://...": context deadline exceeded (Client.Timeout exceeded while awaiting headers)
Call #3: FAILED - connection closed: calling "tools/call": client is closing: EOF
Call #4+: FAILED - connection closed: calling "tools/call": client is closing: EOF
```

After `Call #2` encounters a transient error, all subsequent calls fail permanently with "client is closing: EOF" errors. The connection becomes unusable and must be recreated.

## 3. What did you expect to see?

```
Call #1: SUCCESS ✅
Call #2: FAILED - transient error (expected)
Call #3: SUCCESS ✅ (connection should survive the transient error)
Call #4+: SUCCESS ✅
```

[...]

PR #734, opened 2025-12-19
mcp: improve http transports error handling and make transport work with any size message

alexbumbacea:
We have a use case for messages larger than 1MB, hence making this configurable would help.
In order to discover the source of our problems we had to spend some time investigating, as errors were silently discarded. So we also included changes to improving surfacing errors .

### DIFF
--- a/mcp/event.go
+++ b/mcp/event.go
@@ -140,11 +140,10 @@ func scanEvents(r io.Reader) iter.Seq2[Event, error] {
 				data := bytes.TrimSpace(after)
 				if dataBuf == nil {
 					dataBuf = new(bytes.Buffer)
-					dataBuf.Write(data)
 				} else {
 					dataBuf.WriteByte('\n')
-					dataBuf.Write(data)
 				}
+				dataBuf.Write(data)
 			}
 
 			if isEOF {
--- a/mcp/streamable.go
+++ b/mcp/streamable.go
@@ -1867,10 +1867,7 @@ func (c *streamableClientConn) processStream(ctx context.Context, requestSummary
 				return "", 0, true
 			}
 
-			// Network/I/O errors during reading should trigger reconnection, not permanent failure.
-			// Return from processStream so handleSSE can attempt to reconnect.
-			c.logger.Debug(fmt.Sprintf("%s: stream read error (will attempt reconnect): %v", requestSummary, err))
-			return lastEventID, reconnectDelay, false
+			break
 		}
 
 		if evt.ID != "" {


mcp/sse.go
```diff
@@ -329,6 +329,9 @@ type SSEClientTransport struct {
 	// HTTPClient is the client to use for making HTTP requests. If nil,
 	// http.DefaultClient is used.
 	HTTPClient *http.Client
+
+	// MaxLineSize is the maximum buffer size used when reading a message. It defaults to 1MB
```
findleyr: "MaxLineSize is the maximum buffer size (in bytes) used when reading SSE events. If unset, defaults to 1MiB."

Here, and in the streamable client.

mcp/event.go
```diff
@@ -66,10 +66,12 @@ func writeEvent(w io.Writer, evt Event) (int, error) {
 //
 // TODO(rfindley): consider a different API here that makes failure modes more
 // apparent.
-func scanEvents(r io.Reader) iter.Seq2[Event, error] {
+func scanEvents(r io.Reader, maxLineSize int) iter.Seq2[Event, error] {
 	scanner := bufio.NewScanner(r)
-	const maxTokenSize = 1 * 1024 * 1024 // 1 MiB max line size
-	scanner.Buffer(nil, maxTokenSize)
+	if maxLineSize == 0 {
+		maxLineSize = 1 * 1024 * 1024 // defaults to 1MB
```
findleyr: MiB was correct.

findleyr (commented):
Let's discuss this change in #726. I'm not sure we shouldn't just remove the buffering.

alexbumbacea:
@findleyr current proposal (no fixed buffer size) approach is inline with most sdks. Can we get this merged?

findleyr:
@alexbumbacea sorry for the delay, will review tomorrow.

mcp/event.go
```diff
@@ -100,15 +98,40 @@ func scanEvents(r io.Reader) iter.Seq2[Event, error] {
 				dataBuf = nil
 			}
 		}
-		for scanner.Scan() {
-			line := scanner.Bytes()
+		emitEvent := func() bool {
```
findleyr: We probably should have two so similar closures. I think flushdata can be inlined here.

mcp/event.go
```diff
@@ -100,15 +98,40 @@ func scanEvents(r io.Reader) iter.Seq2[Event, error] {
...
+		for {
+			line, err := scanner.ReadBytes('\n')
+			if err != nil {
+				if errors.Is(err, io.EOF) {
+					// Handle EOF below
+				} else if ctx.Err() != nil {
+					yield(Event{}, fmt.Errorf("context done: %w", ctx.Err()))
+					return
+				} else {
+					yield(Event{}, fmt.Errorf("error reading event: %w", err))
```
findleyr: I don't think we want to expose these errors in the API: we can just format them with %v.

mcp/event.go
```diff
@@ -66,10 +66,8 @@ func writeEvent(w io.Writer, evt Event) (int, error) {
 //
 // TODO(rfindley): consider a different API here that makes failure modes more
 // apparent.
-func scanEvents(r io.Reader) iter.Seq2[Event, error] {
-	scanner := bufio.NewScanner(r)
-	const maxTokenSize = 1 * 1024 * 1024 // 1 MiB max line size
-	scanner.Buffer(nil, maxTokenSize)
+func scanEvents(ctx context.Context, r io.Reader) iter.Seq2[Event, error] {
+	scanner := bufio.NewReader(r)
```
findleyr: 'reader'

mcp/event.go
```diff
@@ -66,10 +66,8 @@ func writeEvent(w io.Writer, evt Event) (int, error) {
 //
 // TODO(rfindley): consider a different API here that makes failure modes more
 // apparent.
-func scanEvents(r io.Reader) iter.Seq2[Event, error] {
-	scanner := bufio.NewScanner(r)
-	const maxTokenSize = 1 * 1024 * 1024 // 1 MiB max line size
-	scanner.Buffer(nil, maxTokenSize)
+func scanEvents(ctx context.Context, r io.Reader) iter.Seq2[Event, error] {
```
findleyr: Why are we adding a ctx parameter? Does not appear to be used?

mcp/sse.go
```diff
@@ -438,12 +448,15 @@ func (c *sseClientConn) Read(ctx context.Context) (jsonrpc.Message, error) {
 	case <-c.done:
 		return nil, io.EOF
 
-	case data := <-c.incoming:
+	case m := <-c.incoming:
+		if m.err != nil {
```
findleyr: Returning an error from Read breaks the connection. I don't think that's actually what we want to happen here, is it?

mcp/event.go
alexbumbacea: Not clear why we don't wat to wrap it and instead we just want to print it.

jba: If you wrap an error, then callers can extract the error. That means the error type becomes part of your effective API. Which means you can't change it.

If you just include the error text, you don't have that problem. If later, someone needs to distinguish particular errors, we can design the proper API for them.

mcp/event.go
```diff
@@ -100,15 +98,40 @@ func scanEvents(r io.Reader) iter.Seq2[Event, error] {
...
+				return true
+			}
+			if !yield(evt, nil) {
+				return false
+			}
+			evt = Event{}
+			return true
+		}
+		for {
+			line, err := scanner.ReadBytes('\n')
```
jba: Do we think this is a potential DOS attack vector? True, the client is reading from the server, which it presumably trusts in many other ways. But a malicious server could make the client crash by feeding it a large event.

What about a high and configurable upper limit?

mcp/event.go
jba: Used on L117 to stop the iteration on cancellation.

mcp/sse.go
alexbumbacea: Given that this method return either an error or a jsonrpc.Message, I think this is an acceptable behaviour.
Considering previous behaviour when there was an error during scan events, parsing the body would stop as the loop over scanEvents would break in case there was an error, but without bubbling up the error.

mcp/event.go
alexbumbacea: inlined flush data, but I don't get it why we would need 2 similar clousures.

mcp/event.go
alexbumbacea: it was already discussed in #726 that having this kind of limit would not protect against this in the event the server would present that as a multi "data" event.

Looking into other implementations of the MCP client (other languages and even other written in go) they do not impose such a limitation.

alexbumbacea:
@findleyr all comments have been either replied to or addressed. are there any other concerns?

mcp/event.go
```diff
@@ -94,30 +92,55 @@ func scanEvents(r io.Reader) iter.Seq2[Event, error] {
...
 			}
 			before, after, found := bytes.Cut(line, []byte{':'})
 			if !found {
 				yield(Event{}, fmt.Errorf("malformed line in SSE stream: %q", string(line)))
 				return
 			}
-			if !bytes.Equal(before, dataKey) {
-				flushData()
+			// If we're starting a new event field and we already have an event, emit it first
+			if bytes.Equal(before, eventKey) && !evt.Empty() {
```
findleyr: This doesn't look right: the `event:` field does not need to be the first field in the event. Maybe add a failing test for this.

mcp/event.go
```diff
@@ -94,30 +92,55 @@ func scanEvents(r io.Reader) iter.Seq2[Event, error] {
 			evt     Event
 			dataBuf *bytes.Buffer // if non-nil, preceding field was also data
 		)
-		flushData := func() {
+		emitEvent := func() bool {
```
findleyr: s/emitEvent/yieldEvent

mcp/sse.go
findleyr: Sorry, I don't think I sufficiently conveyed the consequences.
The JSON-RPC library expects to operate on a logical stream. Any failure to read from that stream indicates that the logical connection is broken.

So as a consequence of this change, a context cancellation calling a tool will break the entire MCP session.

I think I know the desired behavior: you want to get an error from the `CallTool` request, but this change doesn't achieve that result.

The jsonrpc2 library doesn't really support this: we'd need to add something like `jsonrpc2.Connection.Fail(jsonrpc2.ID, error)` to allow this type of layering traversal.

If you'd like to land this PR, I suggest not returning an error here, and I can make the change to achieve what you want.

mcp/event.go
findleyr: Ack, ok... but is that necessary? Wouldn't a context cancellation close the response body?

mcp/event.go
findleyr: Sorry, I mean't "shouldn't"!

mcp/event.go
alexbumbacea: removed and added tests the scenario you mentioned

mcp/sse.go
alexbumbacea: How is this error different from the error returned a few lines bellow by `jsonrpc2.DecodeMessage`?

Your assumption is correct, I want to bubble up any errors from lower levels, so that for the caller the problem would be clear enough, but this implementation doesn't seem to be using a jsonrpc2 connection, just the encode/decode part. (this is sse implementation, not jsonrpc).

mcp/event.go
```diff
@@ -94,31 +92,43 @@ func scanEvents(r io.Reader) iter.Seq2[Event, error] {
...
 				}
-				evt = Event{}
 				continue
 			}
 			before, after, found := bytes.Cut(line, []byte{':'})
 			if !found {
 				yield(Event{}, fmt.Errorf("malformed line in SSE stream: %q", string(line)))
 				return
 			}
-			if !bytes.Equal(before, dataKey) {
```
findleyr, line 119: Removing this means that we'd concatenate data for the following malformed event:

event: foo
data: 123
retry: 5
data 456

https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events#data

Only consecutive data fields should be concatenated.

mcp/sse.go
findleyr: The jsonrpc2.Connection calls Read, which must have stream semantics: an error from Read breaks the stream.

I think there's an argument to be made that a malformed payload _should_ break the stream: if the server sends something that we can't even parse as a jsonrpc2.Message, then the stream is corrupt.

But if the error is due to a network error or client cancellation, we don't want to terminate the session.

See also #683.

Let's focus on fixing the size limit. We can leave the bubbling up of errors, but should not return an error here. I'll make the change to surface the error to the application layer:it's a subtle change to the jsonrpc2 connection.

mcp/event.go
alexbumbacea: good catch! but adding this back as it was would trigger yield for an incomplete event if there are other keys after data (other new tests would fail).
I handled it in a different way now.

alexbumbacea:
@findleyr is there anything else?

alexbumbacea:
@findleyr can we get this merged?

findleyr:
@alexbumbacea very sorry, I thought I had merged this. Total oversight on my part.

@maciej-kisiel we may want to cherry-pick this into the release, or do a second prerelease.

mcp/sse.go
findleyr: FWIW, I also didn't realize that this comment thread was resolved. I think this is why I hadn't reviewed earlier.

mcp/sse.go
```diff
@@ -438,12 +448,16 @@ func (c *sseClientConn) Read(ctx context.Context) (jsonrpc.Message, error) {
...
```
findleyr: Sorry, but returning (nil, nil) here is not correct. Previously the code would simply not receive the message: now it's reading a nil message, which is almost certain to cause a nil pointer exception.

I think we should simply not surface this error to the JSON-RPC layer.

mcp/streamable.go
```diff
@@ -1894,7 +1859,11 @@ func (c *streamableClientConn) processStream(ctx context.Context, requestSummary
 			if ctx.Err() != nil {
 				return "", 0, true // don't reconnect: client cancelled
 			}
-			break
+
+			// Network errors during reading should trigger reconnection, not permanent failure.
+			// Return from processStream so handleSSE can attempt to reconnect.
+			c.logger.Debug(fmt.Sprintf("%s: stream read error (will attempt reconnect): %v", requestSummary, err))
+			return lastEventID, reconnectDelay, false
```
findleyr: FWIW, I think we actually need to be more subtle than this: a malformed event is a hard error, and different from a reader error.

mcp/event.go
```diff
@@ -128,27 +140,25 @@ func scanEvents(r io.Reader) iter.Seq2[Event, error] {
...
 			case bytes.Equal(before, dataKey):
 				data := bytes.TrimSpace(after)
-				if dataBuf != nil {
-					dataBuf.WriteByte('\n')
+				previousLineEmptyOrData := previousLineType == nil || bytes.Equal(previousLineType, dataKey)
+				if dataBuf == nil {
+					dataBuf = new(bytes.Buffer)
 					dataBuf.Write(data)
+				} else if !previousLineEmptyOrData {
+					yield(Event{}, fmt.Errorf("non-continuous data items in the event"))
```
findleyr: I'm very sorry for reverting this feedback: but despite the MDN article suggesting otherwise, I did a bunch of research that indicates almost every implementation just concatenates data fields regardless. So I think you can simplify here as you'd done before.

findleyr:
Apologies again for this getting lost in the weeds: I think there are a few things that need to be addressed before merging.

mcp/sse.go
alexbumbacea: removed the return here, just a TODO comment to have track of the idea that it needs to be handled

mcp/streamable.go
alexbumbacea: did some adaptaptation. let me know WDYT

mcp/event.go
```diff
@@ -310,6 +313,11 @@ func (s *MemoryEventStore) Append(_ context.Context, sessionID, streamID string,
 // index is no longer available.
 var ErrEventsPurged = errors.New("data purged")
 
+// ErrMalformedEvent is returned when an SSE event cannot be parsed due to format violations.
+// This is a hard error indicating corrupted data or protocol violations, as opposed to
+// transient I/O errors which may be retryable.
+var ErrMalformedEvent = errors.New("malformed event")
```
findleyr: Please unexport this. It does not need to be part of the public API.

mcp/sse.go
findleyr: This is still not right: we'd proceed to process the nil message.

alexbumbacea: not surfacing the error at all, would require even more extensive refactoring. the best I can think of it's to wrap the error in a jsonrpc.Response

findleyr: This is why it was factored the way it was before: Read really only needs to know about EOF errors or things that break the connection: all other errors are an aspect of handling the HTTP request, in the HTTP layer.

Returning a jsonrpc.Response with corrupt the stream. I think we should revert incoming to be a queue of data.

mcp/streamable.go
```diff
@@ -1909,7 +1859,17 @@ func (c *streamableClientConn) processStream(ctx context.Context, requestSummary
 			if ctx.Err() != nil {
 				return "", 0, true // don't reconnect: client cancelled
 			}
-			break
+
+			// Malformed events are hard errors that indicate corrupted data or protocol			// violations. These should fail the connection permanently.
```
maciej-kisiel:
```suggestion
			// Malformed events are hard errors that indicate corrupted data or protocol
			// violations. These should fail the connection permanently.
```

mcp/event.go
```diff
@@ -128,27 +138,20 @@ func scanEvents(r io.Reader) iter.Seq2[Event, error] {
 				evt.Retry = strings.TrimSpace(string(after))
 			case bytes.Equal(before, dataKey):
 				data := bytes.TrimSpace(after)
-				if dataBuf != nil {
-					dataBuf.WriteByte('\n')
+				if dataBuf == nil {
+					dataBuf = new(bytes.Buffer)
 					dataBuf.Write(data)
```
findleyr: Optional nit (preexisting): you could move this redundant `dataBuf.Write` out of the `if` block.

mcp/streamable.go
```diff
@@ -1909,7 +1859,18 @@ func (c *streamableClientConn) processStream(ctx context.Context, requestSummary
...
+			// violations. These should fail the connection permanently.
+			if errors.Is(err, errMalformedEvent) {
+				c.fail(fmt.Errorf("%s: %v", requestSummary, err))
+				return "", 0, true
+			}
+
+			// Network/I/O errors during reading should trigger reconnection, not permanent failure.
+			// Return from processStream so handleSSE can attempt to reconnect.
+			c.logger.Debug(fmt.Sprintf("%s: stream read error (will attempt reconnect): %v", requestSummary, err))
+			return lastEventID, reconnectDelay, false
```
maciej-kisiel: I'm reading it one last time and I started wondering if we don't need to handle the "no lastEventID for a non-standalone stream" scenario, as explained on line 1913. Maybe `break` would still be a better alternative?

Sorry for a late comment.

alexbumbacea:
@findleyr @maciej-kisiel can we get this merged and included in a prerelease? Thanks!

maciej-kisiel:
It looks good to me, I wanted to give @findleyr some time to take a look since there was a small change from the version he approved. But I think we can merge it and adjust as a follow-up if needed.

With regard to the release, I think I would prefer to only include fixes to bugs that were introduced with a given pre-release. `1.3.0-pre.1` is already soaking some time and I wouldn't like to prolong the release more than necessary. This change will then be included in `1.4.0` which is planned ~in the last week of February.

PR #736, opened 2025-12-25
fix: connectStandaloneSSE checking Content-Type header

liushuangls:
```
mcp/streamable: checking Content-Type header
```

Issue encountered: the MCP server request was unexpectedly redirected to an HTML page. Because the client didn’t validate the response Content-Type, it still entered handleSSE, which then kept treating the HTML as an SSE stream and repeatedly re-requesting the same page—creating an infinite request loop.

### DIFF
--- a/mcp/streamable.go
+++ b/mcp/streamable.go
@@ -1550,8 +1550,16 @@ func (c *streamableClientConn) connectStandaloneSSE() {
 	// SSE stream at this endpoint."
 	//
 	// [§2.2.3]: https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#listening-for-messages-from-the-server
-	if resp.StatusCode == http.StatusMethodNotAllowed || resp.Header.Get("Content-Type") != "text/event-stream" {
+	if resp.StatusCode == http.StatusMethodNotAllowed {
 		// The server doesn't support the standalone SSE stream.
+		resp.Body.Close()
+		return
+	}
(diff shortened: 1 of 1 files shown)


findleyr:
Thanks for this CL. We're on break now, but I'll review next week.

mcp/streamable.go
```diff
@@ -1550,7 +1550,7 @@ func (c *streamableClientConn) connectStandaloneSSE() {
 	// SSE stream at this endpoint."
 	//
 	// [§2.2.3]: https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#listening-for-messages-from-the-server
-	if resp.StatusCode == http.StatusMethodNotAllowed {
+	if resp.StatusCode == http.StatusMethodNotAllowed || resp.Header.Get("Content-Type") != "text/event-stream" {
```
findleyr: Can we add a test for this case? You should be able to modify TestStreamableClientGETHandling to set content type (see `streamableResponse.header`).

liushuangls: Added

findleyr: Sorry for not catching this before: I think this is a non-conformant server implementation, so we should at the very least log.

Let's pull this out into a separate if block, and similar to the block on line 1558 below, we should log a warning.

findleyr (commented):
Thanks for adding the test! One more minor comment.

mcp/streamable.go
liushuangls: Done.

PR #738, opened 2025-12-27
feat: deprecated logger in client & add Logger in ClientOption

### LINKED ISSUE #708: Proposal: add ClientOptions.Logger
For symmetry with `ServerOptions.Logger`, we should have `ClientOptions.Logger`

(I'm fixing a hardcoded log in `notifySession`, which is reached from both client and server).

Not a high priority.

CocaineCong:
Fix #708

1. deprecated logger in client
2. add Logger in ClientOption
3. add unit test for Logger

### DIFF
--- a/mcp/client.go
+++ b/mcp/client.go
@@ -70,7 +70,7 @@ func NewClient(impl *Implementation, options *ClientOptions) *Client {
 
 // ClientOptions configures the behavior of the client.
 type ClientOptions struct {
-	// If non-nil, log client activity.
+	// Logger may be set to a non-nil value to enable logging of client activity.
 	Logger *slog.Logger
 	// CreateMessageHandler handles incoming requests for sampling/createMessage.
 	//
--- a/mcp/server.go
+++ b/mcp/server.go
@@ -58,7 +58,7 @@ type Server struct {
 type ServerOptions struct {
 	// Optional instructions for connected clients.
 	Instructions string
-	// If non-nil, log server activity.
+	// Logger may be set to a non-nil value to enable logging of server activity.
 	Logger *slog.Logger
 	// If non-nil, called when "notifications/initialized" is received.
 	InitializedHandler func(context.Context, *InitializedRequest)


mcp/client.go
```diff
@@ -42,25 +44,34 @@ type Client struct {
...
-	if opts != nil {
-		c.opts = *opts
-	}
-	return c
 }
 
 // ClientOptions configures the behavior of the client.
 type ClientOptions struct {
+	// If non-nil, log client activity.
+	Logger *slog.Logger
```
findleyr, line 74: To be consistent with Go idioms, and other field documentation, this should start with "Logger".

For example:
"Logger may be set to a non-nil value to enable logging of client activity."

CocaineCong:
> LGTM following one minor nit.

I've already corrected it. PTAL.

findleyr (approved):

mcp/client.go
```diff
@@ -18,15 +18,17 @@ import (
...
 
 // A Client is an MCP client, which may be connected to an MCP server
 // using the [Client.Connect] method.
 type Client struct {
-	impl                    *Implementation
-	opts                    ClientOptions
+	impl *Implementation
+	opts ClientOptions
+	// Deprecated
 	logger                  *slog.Logger // TODO: file proposal to export this
```
findleyr, line 32: Can you update the CL to use the client's logger instead of this one? Otherwise, the behavior is counterintuitive. If you'd prefer to do this in a separate CL, or if you'd prefer I do it, no problem.

(if I don't hear from you in a couple days, I'll just merge this as-is and then send a follow-up CL)

CocaineCong: yep, I just prefer to do this in a separate CL, and I will send a follow-up CL in the next few days.

mcp/client.go
```diff
@@ -42,25 +44,34 @@ type Client struct {
...
+func NewClient(impl *Implementation, options *ClientOptions) *Client {
 	if impl == nil {
 		panic("nil Implementation")
 	}
-	c := &Client{
+	var opts ClientOptions
+	if options != nil {
+		opts = *options
+	}
+	options = nil // prevent reuse
```
jba, line 55: How does this prevent reuse?
As far as I can tell, it does so only for the current function, which is short.

mcp/client.go
```diff
@@ -42,25 +44,34 @@ type Client struct {
 // The first argument must not be nil.
 //
 // If non-nil, the provided options configure the Client.
-func NewClient(impl *Implementation, opts *ClientOptions) *Client {
+func NewClient(impl *Implementation, options *ClientOptions) *Client {
```
jba, line 47: Why rename to `options`? The caller knows the type, which already says `Options`?

mcp/client.go
```diff
@@ -18,15 +18,17 @@ import (
...
 )
 
 // A Client is an MCP client, which may be connected to an MCP server
 // using the [Client.Connect] method.
 type Client struct {
-	impl                    *Implementation
-	opts                    ClientOptions
+	impl *Implementation
+	opts ClientOptions
+	// Deprecated
```
jba, line 31: You can just remove it and all references. It's unexported.

mcp/client.go
CocaineCong: Similar above, I just keep the code style consistent. 

image

If you think this will cause confusion, I will rename back.

mcp/client.go
CocaineCong: cause i see the func `NewServer` (which in mcp/server.go) do this, i thought this is the code style of this program.😂 
image

If you think it's unnecessary, I will remove it.

mcp/client.go
CocaineCong: yep, I have removed it.

PR #742, opened 2025-12-30
fix: enforce retry limit when SSE stream makes no progress

### LINKED ISSUE #679: Set a total retry limit on streams, or enforce progress
Right now, the streamable client connection retries broken resumable connections up to `StreamableClientTransport.MaxRetries`, but once a connection has been re-established, the counter resets.

This can lead to a failure mode where a server keeps terminating the connection with status 200. We should either:
1. Set a global limit on the number of retries across a single logical stream.
2. Only reset the retry counter when the Last-Event-ID actually makes progress.

I'm leaning toward 2: if the stream is active and long-running, a global limit could be problematic. On the other hand, (2) doesn't have that problem, and still avoids the degenerate failure mode.

CC @jba @markus-kusano

majiayu000:
Fixes #679

## Changes
- Track last event ID to detect progress in SSE streams
- Only reset retry counter when Last-Event-ID advances
- Fail connection after exceeding maxRetries without progress

This implements Option 2 from the issue: progress-based retry limiting that prevents infinite retry loops while allowing long-running streams to continue as long as progress is made.

### DIFF
--- a/mcp/streamable_client_test.go
+++ b/mcp/streamable_client_test.go
@@ -789,7 +789,7 @@ data: {"jsonrpc":"2.0","method":"notifications/message","params":{"level":"info"
 	// We expect maxRetries+1 attempts because we increment before checking the limit.
 	if got := retryCount.Load(); got != int32(maxRetries+1) {
 		t.Errorf("retry count = %d, want exactly %d", got, maxRetries+1)
-  }
+	}
 }
 
 func TestStreamableClientDisableStandaloneSSE(t *testing.T) {


findleyr:
Thanks for this contribution, and sorry for the delay. I'll be catching up on PRs this weekend / next week.

mcp/streamable_client_test.go
```diff
@@ -693,3 +693,98 @@ func TestStreamableClientTransientErrors(t *testing.T) {
...
+	}
+
+	// Check that the error mentions exceeding retries without progress.
+	wantErr := "exceeded"
+	if !strings.Contains(err.Error(), wantErr) {
+		t.Errorf("CallTool error = %q, want containing %q", err.Error(), wantErr)
+	}
+
+	// Verify that we actually retried the expected number of times.
+	// We expect maxRetries+1 attempts because we increment before checking the limit.
```
findleyr, line 789: In this case, make this more assertive: assert that got == maxRetries+1

mcp/streamable_client_test.go
```diff
@@ -693,3 +693,98 @@ func TestStreamableClientTransientErrors(t *testing.T) {
...
+	// Speed up reconnection delays for testing.
+	const tick = 10 * time.Millisecond
+	defer func(delay time.Duration) {
+		reconnectInitialDelay = delay
+	}(reconnectInitialDelay)
+	reconnectInitialDelay = tick
+
+	// Use the fakeStreamableServer pattern like other tests to avoid race conditions.
+	ctx := context.Background()
+	maxRetries := 2
```
findleyr: const maxRetries = 2

(more readable, and avoids some conversions below)

findleyr (approved):
Thanks! @maciej-kisiel feel free to merge.

Issue #626, opened 2025-11-07
SEP-1686: implement Tasks


findleyr:
Add support for [SEP-1686](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1686), when it lands.

jba:
Tasks pose a challenge for backwards compatibility. A tool call can now return its usual result, or a CreateTaskResult.

One way to deal with this would be to have separate functions and types on the server for registering a tool that might be a task. The handler's return type would be some union of CallToolResult and CreateTaskResult. We'd also need a separate function on the client.

A better idea, although it's a bit ugly, is to add a CreateTaskResult field to CallToolResult. When a tool is called as a task, it sets this field and leaves all the others empty. Callers of ClientSession.CallTool would have to recognize that state. Maybe we could have ClientSession.CallToolAsTask that returns a CreateTaskResult – no union type is needed on the client side. But that method would not be able to mark the call as being a task call: that bit is set in the call's params (arguments), not its metadata (see https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1732/files#r2524912400 for my failed pushback). In our SDK, CallToolParams.Arguments is an `any`, so we can't easily look at it. (One hideous possibility: we require that the arguments marshal to JSON, so we could marshal, unmarshal into a `map[string]any`, and set the task info.) We may be able to do better on the server side, recognizing that a call is a task call, because we already unmarshal into a `map[string]any`.

LucaButBoring:
Relaying here from `#general-sdk-dev`:

> There doesn't seem to be a central ticket for all SDK implementations of a particular change, so sharing this here for SEP-1686 -- the following revisions have been merged on top of the base specification PR for tasks:
> - Clarify request lifetimes in task diagrams: [PR](<https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1826>)
> - Fix capability phrasing in spec: [PR](<https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1830>)
> - Remove an incorrectly-added field from `notifications/cancelled`: [PR](<https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1833>)
> - Adjust schema to better-communicate where tasks are valid: [PR](<https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1845>)
> - Rename/move `annotations.taskHint` to `execution.taskSupport`, tweak its enum values, and change a SHOULD to a MUST for the behavior when this is declared:
[...]

PR #755, opened 2026-01-16
mcp: add tasks extension capability negotiation

JoannaaKL:
Adds capability negotiation for the MCP Tasks extension (`io.modelcontextprotocol/tasks`). First step of #626; task execution follows in later PRs.

### Why

Tasks moved out of the core protocol and into an official extension in `2026-07-28` ([SEP-2663](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/seps/2663-tasks-extension.md)), now specified in [modelcontextprotocol/ext-tasks](https://github.com/modelcontextprotocol/ext-tasks/blob/main/specification/draft/tasks.md). The in-core design this PR originally implemented ([SEP-1686](https://modelcontextprotocol.io/specification/2025-11-25/basic/utilities/tasks), `2025-11-25`) is gone from the current schema. The branch is rewritten against the extension, and scoped to capabilities only.

This also fixes a live bug. A `CreateTaskResult` decoded into an empty `CallToolResult` **without error**, so a tool call against any server that returns a task silently looked like it succeeded and returned nothing.

### What

- `ExtensionTasks`: the extension identifier.
- `HasExtension` on `ClientCapabilities` and `ServerCapabilities` — the read-side counterpart to the existing `AddExtension`, which had no equivalent for testing whether a peer declared an extension. Generic, not tasks-specific.
- Reserve the [`"task"` `resultType` discriminator](https://github.com/modelcontextprotocol/ext-tasks/blob/main/specification/draft/tasks.md#polymorphic-results) and reject it while decoding, returning `UnsupportedTaskResultError` with the task ID.

Negotiation is then one line on each side, over plumbing that already exists:

```go
req.ClientCapabilities().HasExtension(mcp.ExtensionTasks)              // server
cs.InitializeResult().Capabilities.HasExtension(mcp.ExtensionTasks)    // client
```

The guard lives in `UnmarshalJSON` rather than in `CallTool` or a middleware: a middleware can be disabled via `MultiRoundTrip.Disabled`, which is the wrong shape for a correctness check. [`tools/call` is the only task-eligible method today](https://github.com/modelcontextprotocol/ext-tasks/blob/main/specification/draft/tasks.md#supported-methods), but the spec asks implementations to accommodate more in future revisions, so the guard also covers `prompts/get` and `resources/read`. It adds no extra parsing, since all three result types already decode `resultType`.

**Nothing is declared by default**, on either side. Declaring the extension obliges a client to poll task handles and a server to serve `tasks/*`, neither of which the SDK implements yet; users with their own implementation can still declare it via `AddExtension`.

Note for whoever implements `tasks/get`: the error code for "Missing Required Client Capability" is `-32021`, so this SDK's existing `CodeMissingRequiredClientCapabilities` is already correct. The ext-tasks prose briefly said `-32003` — a slip introduced when SEP-2663 was ported into that repo, where the SEP itself, the core `2026-07-28` schema and the TypeScript SDK all use `-32021`. Fixed upstream in modelcontextprotocol/ext-tasks#15, so the spec and this SDK now agree.

### DIFF
--- a/docs/client.md
+++ b/docs/client.md
@@ -334,3 +334,43 @@ that optional capabilities outside the core protocol can be declared on the
 wire. Keys are namespaced as `"{vendor-prefix}/{extension-name}"`; values
 are per-extension settings objects.
 
+Use `AddExtension` to declare one and `HasExtension` to test for one:
+
+```go
+caps := &mcp.ClientCapabilities{}
+caps.AddExtension("io.example/my-extension", nil)
+client := mcp.NewClient(impl, &mcp.ClientOptions{Capabilities: caps})
+
+cs, err := client.Connect(ctx, transport, nil)
+...
+if cs.InitializeResult().Capabilities.HasExtension("io.example/my-extension") {
+    // The server declared it too.
+}
+```
+
+#### Tasks
+
+[SEP-2663](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2663)
+moved tasks out of the core protocol and into the
+[tasks extension](https://github.com/modelcontextprotocol/ext-tasks),
+identified by the `mcp.ExtensionTasks` constant. A server that has negotiated
+it may answer a request with a durable task handle instead of the result that
+was asked for, which the client then polls to completion.
+
+**The SDK does not implement task execution**, and does not declare the
+extension by default. Declaring it is a promise to the peer: a client that
+declares it must be prepared for any eligible request to return a task handle
+instead of a result. Only declare it if you implement that polling flow
+yourself.
+
+If a server returns a task handle anyway, decoding fails with
+`*mcp.UnsupportedTaskResultError`, which carries the task ID:
+
+```go
+res, err := cs.CallTool(ctx, params)
+var terr *mcp.UnsupportedTaskResultError
+if errors.As(err, &terr) {
+    log.Printf("server created task %s, which this SDK cannot resolve", terr.TaskID)
+}
+```
+
--- a/docs/client.md
+++ b/docs/client.md
@@ -350,9 +350,9 @@ if cs.InitializeResult().Capabilities.HasExtension("io.example/my-extension") {
 
 #### Tasks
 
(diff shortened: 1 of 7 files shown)


jba:
Please delete the guidelines and write a proper commit message. Link to the part of the spec that defines tasks.

jba:
Also link to our issue about tasks.

jba:
I just skimmed the [spec](https://modelcontextprotocol.io/specification/2025-11-25/basic/utilities/tasks). It is huge. I think we should implement it in multiple PRs. Let's start with the relatively easy stuff. How about a PR for capabilities first? And maybe one for the task data structure and related code, like generating the task ID?

jba:
I don't quite understand how calling a tool with a task works. It looks like the tool itself doesn't have to know about tasks. Where is the code that returns the task result immediately and then handles task lifetime management?

I think everything else is pretty straightforward, but this part of the design needs to be discussed on our tasks issue.

JoannaaKL:
Hey @jba pr is a draft and I assumed it's clear that it's not ready for a review yet. Body would be filled once the implementation is done.
Regarding the the tool calling - happy to move to the main issue and discuss there.

JoannaaKL:
Hey @jba, this is the capabilities-only pr you suggested, out of draft now.

Scope is just negotiation: `ExtensionTasks`, `HasExtension` on both capability types, and reserving the `"task"` resultType.

On your question about where task lifetime is handled - it isn't. Nothing is declared by default on either side, so we never promise a peer we can serve tasks. Happy to discuss the execution design in #626.

Two things changed since you last looked. Tasks moved out of core into an extension ([SEP-2663](https://github.com/modelcontextprotocol/ext-tasks/blob/main/specification/draft/tasks.md)) so I rewrote the branch against ext-tasks. And this fixes a bug: a `CreateTaskResult` decodes into an empty `CallToolResult` with no error today, so a tool call against a task-returning server looks like it succeeded and returned nothing.

Also ci has never run here, I think it needs your approval.

PR #756, opened 2026-01-17
Testing: use synctest for timing-dependent tests

### LINKED ISSUE #680: Testing: use synctest throughout
We have several tests that rely on timing, and have been meaning to replace this with uses of synctest. This issue tracks that work.

(see any usage of the `time` package in test files)

La002:
Testing: use synctest for timing-dependent tests

Converted timing-based tests to use Go 1.25's synctest for instant, deterministic execution with simulated time.

Moved converted tests to *_go125_test.go files. Tests using real I/O (exec.Command, httptest) remain timing-based as they require actual system operations.

Fixes #680

### DIFF
--- a/mcp/mcp_go125_test.go
+++ b/mcp/mcp_go125_test.go
@@ -602,6 +602,9 @@ func TestKeepAliveFailure_Synctest(t *testing.T) {
 	synctest.Test(t, func(t *testing.T) {
 		ctx := context.Background()
 
+		// This test doesn't pinpoint keepalive detected failures well due to the fact that in-memory transports
+		// propagate `io.EOF` synchronously, causing the transport level connection to be closed immediately.
+		// TODO: propose better transports that would allow testing this scenario precisely.
 		ct, st := NewInMemoryTransports()
 
 		// Server without keepalive (to test one-sided keepalive)
@@ -624,24 +627,20 @@ func TestKeepAliveFailure_Synctest(t *testing.T) {
 		defer cs.Close()
 
 		// Let the connection establish properly first
-		time.Sleep(30 * time.Millisecond)
 		synctest.Wait()
 
 		// simulate ping failure
 		ss.Close()
 
-		// Wait for keepalive to detect the failure and close the client
-		// Check periodically with simulated time advancement
-		for i := 0; i < 40; i++ { // 40 iterations * 25ms = 1 second total
-			time.Sleep(25 * time.Millisecond)
-			synctest.Wait()
-			_, err = cs.CallTool(ctx, &CallToolParams{
-				Name:      "greet",
-				Arguments: map[string]any{"Name": "user"},
-			})
-			if err != nil && (errors.Is(err, ErrConnectionClosed) || strings.Contains(err.Error(), "connection closed")) {
-				return // Test passed
-			}
+		time.Sleep(100 * time.Millisecond)
+		synctest.Wait()
+
+		_, err = cs.CallTool(ctx, &CallToolParams{
+			Name:      "greet",
+			Arguments: map[string]any{"Name": "user"},
+		})
+		if err != nil && (errors.Is(err, ErrConnectionClosed) || strings.Contains(err.Error(), "connection closed")) {
+			return // Test passed
 		}
 
 		t.Errorf("expected connection to be closed by keepalive, but it wasn't. Last error: %v", err)


La002:
@findleyr @jba following up, is this still relevant ? Are any changes required?

La002:
@maciej-kisiel  Pushed with the changes discussed [here](https://github.com/modelcontextprotocol/go-sdk/issues/680#issuecomment-3809722984)

mcp/cmd_go125_test.go
```diff
@@ -0,0 +1,57 @@
...
+		if err := session.Ping(context.Background(), nil); err != nil {
+			t.Fatal(err)
+		}
+
+		// cancel the context to stop the server
+		cancel()
+
+		// wait for the server to exit
+		// synctest will detect deadlock if server doesn't exit
+		synctest.Wait()
```
maciej-kisiel: Is `synctest.Wait()` required here? If I understand `synctest`'s documentation correctly, I believe blocking on `onServerExit` should be sufficient.

mcp/elicitation_go125_test.go
```diff
@@ -0,0 +1,71 @@
...
+			ElicitationHandler: func(context.Context, *ElicitRequest) (*ElicitResult, error) {
+				return &ElicitResult{Action: "accept"}, nil
+			},
+			ElicitationCompleteHandler: func(_ context.Context, req *ElicitationCompleteNotificationRequest) {
+				elicitationCompleteCh <- req.Params
+			},
+		})
+
+		cs, ss, cleanup := basicClientServerConnection(t, c, nil, nil)
+		_ = cs // Dummy usage to avoid "declared and not used" error.
```
maciej-kisiel: I know it is pre-existing, but maybe we can leave the campsite better:

Wouldn't putting `_` in place of `cs` one line above work as well?

mcp/elicitation_go125_test.go
```diff
@@ -0,0 +1,71 @@
...
+		// 2. Server sends elicitation complete notification (simulating out-of-band completion)
+		err = handleNotify(ctx, notificationElicitationComplete, newServerRequest(ss, &ElicitationCompleteParams{
+			ElicitationID: elicitID,
+		}))
+		if err != nil {
+			t.Fatalf("failed to send elicitation complete notification: %v", err)
+		}
+
+		// 3. Client should receive the notification
+		synctest.Wait()
```
maciej-kisiel: Similarly here, we're blocking on a channel, so I don't think `synctest.Wait()` call is required.

mcp/mcp_go125_test.go
```diff
@@ -0,0 +1,805 @@
...
+		// Server.sendNotification).
+		notificationChans := map[string]chan int{}
+		for _, name := range []string{"initialized", "roots", "tools", "prompts", "resources", "progress_server", "progress_client", "resource_updated", "subscribe", "unsubscribe", "elicitation_complete"} {
+			notificationChans[name] = make(chan int, 1)
+		}
+
+		waitForNotification := func(t *testing.T, name string) {
+			t.Helper()
+			time.Sleep(notificationDelay * 2)
+			synctest.Wait()
```
maciej-kisiel: Similarly here, I believe it should work without `synctest.Wait()`.

mcp/mcp_go125_test.go
```diff
@@ -0,0 +1,805 @@
...
+			// Check tools-changed notifications.
+			s.AddTool(&Tool{Name: "T", InputSchema: &jsonschema.Schema{Type: "object"}}, nopHandler)
+			waitForNotification(t, "tools")
+			s.RemoveTools("T")
+			waitForNotification(t, "tools")
+		}
+
+		// ===== resources =====
+		t.Log("Testing resources")
+		if runtime.GOOS != "windows" {
```
maciej-kisiel, line 257: Can we leave the TODO as a comment?

mcp/mcp_go125_test.go
```diff
@@ -0,0 +1,805 @@
...
+					},
+				},
+			}
+
+			check := func(t *testing.T) {
+				t.Helper()
+				var got []*LoggingMessageParams
+				// Read messages from this test until we've seen all we expect.
+				for len(got) < len(want) {
+					synctest.Wait()
```
maciej-kisiel: Do we need the `synctest.Wait()`? If we remove it, we should probably also remove the select with default and leave a plain channel read.

mcp/mcp_go125_test.go
```diff
@@ -0,0 +1,805 @@
...
+		// ===== progress =====
+		t.Log("Testing progress")
+		{
+			ss.NotifyProgress(ctx, &ProgressNotificationParams{
+				ProgressToken: "token-xyz",
+				Message:       "progress update",
+				Progress:      0.5,
+				Total:         2,
+			})
+			synctest.Wait()
```
maciej-kisiel: Here and in line 458 as well.

Also, why not use `waitForNotification`?

mcp/mcp_go125_test.go
```diff
@@ -0,0 +1,805 @@
...
+		// ===== resource_subscriptions =====
+		t.Log("Testing resource_subscriptions")
+		{
+			err := cs.Subscribe(ctx, &SubscribeParams{
+				URI: "test",
+			})
+			if err != nil {
+				t.Fatal(err)
+			}
+			synctest.Wait()
```
maciej-kisiel, line 465: Here, line 477 and 486 as well.

Also, why not use `waitForNotification`?

mcp/mcp_go125_test.go
```diff
@@ -0,0 +1,805 @@
...
+			<-notificationChans["unsubscribe"]
+
+			// Verify the client does not receive the update after unsubscribing.
+			s.ResourceUpdated(ctx, &ResourceUpdatedNotificationParams{
+				URI: "test",
+			})
+			time.Sleep(notificationDelay * 2)
```
maciej-kisiel: Could this sleep be replaced with `synctest.Wait`?

This seems to actually be a good place for this call, since we want to make sure that whatever server wants to do, it gets done before proceeding.

mcp/mcp_go125_test.go
```diff
@@ -0,0 +1,805 @@
...
+	"sync/atomic"
+	"testing"
+	"testing/synctest"
+	"time"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/jsonschema-go/jsonschema"
+)
+
+func TestEndToEnd_Synctest(t *testing.T) {
```
maciej-kisiel, line 28: Could you leave the following TODO for me above the function signature?

```
// TODO: split this test into multiple test cases that target specific functionality
```

It was already big before your modifications and will be even harder to debug now that we don't have `t.Run` calls that would separate the results.

mcp/mcp_go125_test.go
```diff
@@ -0,0 +1,805 @@
...
+		defer cs.Close()
+
+		errs := make(chan error, batchSize)
+		for i := range batchSize {
+			go func() {
+				_, err := cs.ListTools(ctx, nil)
+				errs <- err
+			}()
+			time.Sleep(2 * time.Millisecond)
+			if i < batchSize-1 {
```
maciej-kisiel: I'm confused, if `batchSize` is 1, would this `if` ever execute? It seems the only possible value for `i` is `0`, so this will be `0 < 1-1 = 0` which is false. If you can confirm this is never executed, let's just remove this test.

mcp/mcp_go125_test.go
```diff
@@ -0,0 +1,805 @@
...
+		}
+		c := NewClient(testImpl, clientOpts)
+		cs, err := c.Connect(ctx, ct, nil)
+		if err != nil {
+			t.Fatal(err)
+		}
+		defer cs.Close()
+
+		// Let the connection establish properly first
+		time.Sleep(30 * time.Millisecond)
```
maciej-kisiel: Could this be replaced with `synctest.Wait()`?

mcp/mcp_go125_test.go
```diff
@@ -0,0 +1,805 @@
...
+
+		// simulate ping failure
+		ss.Close()
+
+		// Wait for keepalive to detect the failure and close the client
+		// Check periodically with simulated time advancement
+		for i := 0; i < 40; i++ { // 40 iterations * 25ms = 1 second total
+			time.Sleep(25 * time.Millisecond)
```
maciej-kisiel: I don't have the background why this loop was introduced, but wouldn't a single sleep longer than the keepalive time be sufficient?

mcp/mcp_go125_test.go
```diff
@@ -0,0 +1,805 @@
...
+		})
+		defer cleanup()
+
+		ctx, cancel := context.WithCancel(context.Background())
+		go cs.CallTool(ctx, &CallToolParams{Name: "slow"})
+		<-start
+		cancel()
+
+		// Wait for all goroutines to be blocked (cancellation delivered).
+		synctest.Wait()
```
maciej-kisiel: I believe it should work without `synctest.Wait()`.

mcp/cmd_go125_test.go
La002: Yes, I did misunderstand the behaviour of `synctest.Wait()`. Thanks for all the review comments 
Removed all the places it wasn't needed [bb74dc2](https://github.com/modelcontextprotocol/go-sdk/pull/756/commits/bb74dc2bced88fc2d1091e58e79ad03089e0718a)

mcp/elicitation_go125_test.go
La002: done [bb74dc2](https://github.com/modelcontextprotocol/go-sdk/pull/756/commits/bb74dc2bced88fc2d1091e58e79ad03089e0718a)

mcp/mcp_go125_test.go
La002: done [bb74dc2](https://github.com/modelcontextprotocol/go-sdk/pull/756/commits/bb74dc2bced88fc2d1091e58e79ad03089e0718a)

mcp/mcp_go125_test.go
La002: done [bb74dc2](https://github.com/modelcontextprotocol/go-sdk/pull/756/commits/bb74dc2bced88fc2d1091e58e79ad03089e0718a)

mcp/mcp_go125_test.go
La002: done [bb74dc2](https://github.com/modelcontextprotocol/go-sdk/pull/756/commits/bb74dc2bced88fc2d1091e58e79ad03089e0718a)

mcp/mcp_go125_test.go
La002: done [bb74dc2](https://github.com/modelcontextprotocol/go-sdk/pull/756/commits/bb74dc2bced88fc2d1091e58e79ad03089e0718a)

mcp/mcp_go125_test.go
La002: done [bb74dc2](https://github.com/modelcontextprotocol/go-sdk/pull/756/commits/bb74dc2bced88fc2d1091e58e79ad03089e0718a)

mcp/mcp_go125_test.go
La002: Added the todo [fb973b1](https://github.com/modelcontextprotocol/go-sdk/pull/756/commits/fb973b19849989db0e232c02b84960c260adcca1)

mcp/mcp_go125_test.go
La002: Done [fb973b1](https://github.com/modelcontextprotocol/go-sdk/pull/756/commits/fb973b19849989db0e232c02b84960c260adcca1)

mcp/elicitation_go125_test.go
La002: yes, done [fb973b1](https://github.com/modelcontextprotocol/go-sdk/pull/756/commits/fb973b19849989db0e232c02b84960c260adcca1)

mcp/mcp_go125_test.go
La002: Added [246ae8b](https://github.com/modelcontextprotocol/go-sdk/pull/756/commits/246ae8bd7b3444914d757f868a366c7f89ec33be)

mcp/mcp_go125_test.go
La002: Yes its not executed. The comment above also mentions the buggy situation

Removed as part of [246ae8b](https://github.com/modelcontextprotocol/go-sdk/pull/756/commits/246ae8bd7b3444914d757f868a366c7f89ec33be)

mcp/mcp_go125_test.go
La002: I think it might have been useful for earlier test, to exit early and not have to wait the whole duration of whatever longer sleep time would have been specified.  But now with synctest, time is not an issue, so a sufficient sleep time of 1s should be good enough

La002:
replied to comments @maciej-kisiel

mcp/mcp_go125_test.go
maciej-kisiel: Does the test still pass if you remove the `Sleep`? If so, I would propose to remove it altogether (the comment may be left to explain the `synctest.Wait` call).

mcp/mcp_go125_test.go
maciej-kisiel: Let me try to be more precise: could this loop be replaced with:

```
time.Sleep(51 * time.Millisecond) // One millisecond more than the keepalive duration.
if _, err := cs.CallTool(...); !errors.IsError(err, ErrConnectionClosed) {
  t.Errorf(...)
}
```

La002: I think it should not be ideally. When the connection is closed it should need some buffer time to detect the failure. (2-3x atleast should have been given) 
But I am unable to test that here because of this potential issue,  when we do `CallTool() `- its also capable of detecting error , whether `KeepAlive` mechanism detected it or not. 
I tried this below test for example (without synctest), and this still passes. 

```
func TestKeepAliveFailure(t *testing.T) {
	// same csetup.....
	defer cs.Close()

	// Let the connection establish properly first
	//time.Sleep(30 * time.Millisecond)

	// simulate ping failure
	ss.Close()
	
	time.Sleep(10 * time.Nanosecond) // net wait from session closure is 10 ns
	_, err = cs.CallTool(ctx, &CallToolParams{
		Name:      "greet",
		Arguments: map[string]any{"Name": "user"},
	})
	if errors.Is(err, ErrConnectionClosed) {
		return // Test passed
	}
	t.Errorf("expected connection to be closed by keepalive, but it wasn't. Last error: %v", err)
}
```
So a bigger wait of `51ms` is also passing. 
Is this expected?

mcp/mcp_go125_test.go
La002: Done

La002:
I replied one comment I had missed earlier. For `keepAliveFailure` one, please let me know what you think.

mcp/mcp_go125_test.go
maciej-kisiel: This one is still open.

mcp/mcp_go125_test.go
maciej-kisiel: Good question. My understanding would be that indeed `CallTool` should fail if the server closed the session and it's difficult to distinguish from the keepalive failure. After some LLM-assisted research, it looks like we would need a transport that does not immediately propagate `io.EOF` on closure to detect this appropriately. Fixing this seems out of scope for this PR, especially that we already had several rounds of improvements. Therefore, I would suggest the following:

1) Leave a comment over `NewInMemoryTransports()` line:
```
// This test doesn't pinpoint keepalive detected failures well due to the fact that in-memory transports
// propagate `io.EOF` synchronously, causing the transport level connection to be closed immediately.
// TODO: propose better transports that would allow testing this scenario precisely.
```

2) Remove the loop and just leave a `time.Sleep(100ms)`, as that would be an appropriate value if the test was more precise.

La002: Done [6c590c1](https://github.com/modelcontextprotocol/go-sdk/pull/756/commits/6c590c117629768d3e0c846533d1c6fbfb5360fb)

mcp/mcp_go125_test.go
La002: Done [6c590c1](https://github.com/modelcontextprotocol/go-sdk/pull/756/commits/6c590c117629768d3e0c846533d1c6fbfb5360fb)

maciej-kisiel (approved):
Thanks for your patience and contribution!

PR #757, opened 2026-01-20
mcp: add Allow header to 405 responses per RFC 9110 §15.5.6

SamMorrowDrums:
Hey @findleyr we have an issue where a customer is using Gateways and technically the SDK throws non-compliant method not allowed headers, when they try to inspect server with a HEAD request. It's fine that it fails, but it highlights that it doesn't specify the required allowed methods, and so I think this is a change that should be made. Let me know :pray: 

## Summary

405 Method Not Allowed responses MUST include an Allow header listing supported methods, per RFC 9110 Section 15.5.6. This fixes issues with strict HTTP gateways (like Apigee) that treat 405 responses without an Allow header as malformed, returning 502 Bad Gateway errors.

## Changes

- **SSEHandler**: Add `Allow: GET, POST` header for unsupported methods
- **StreamableHTTPHandler**: Add Allow header for GET-without-session case:
  - Stateless mode: `Allow: POST, DELETE` (GET is never valid)
  - Stateful mode: `Allow: GET, POST, DELETE` (GET is valid once you have a session)
- Add tests to verify Allow header presence in all 405 responses

## RFC 9110 Reference

[Section 15.5.6 (405 Method Not Allowed)](https://httpwg.org/specs/rfc9110.html#status.405):

> The origin server MUST generate an Allow header field in a 405 response containing a list of the target resource's currently supported methods.

## Testing

All existing tests pass, plus new tests added:
- `TestSSE405AllowHeader` - verifies SSE handler compliance
- `TestStreamable405AllowHeader` - verifies Streamable handler compliance in both stateful and stateless modes

### DIFF
--- a/mcp/streamable_test.go
+++ b/mcp/streamable_test.go
@@ -1885,40 +1885,34 @@ func TestStreamable405AllowHeader(t *testing.T) {
 	server := NewServer(testImpl, nil)
 
 	tests := []struct {
-		name      string
-		stateless bool
-		method    string
-		wantAllow string
+		name       string
+		stateless  bool
+		method     string
+		wantStatus int
+		wantAllow  string
 	}{
 		{
-			name:      "unsupported method (PUT) stateful",
-			stateless: false,
-			method:    "PUT",
-			wantAllow: "GET, POST, DELETE",
+			name:       "unsupported method stateful",
+			stateless:  false,
+			method:     "PUT",
+			wantStatus: http.StatusMethodNotAllowed,
+			wantAllow:  "GET, POST, DELETE",
 		},
 		{
-			name:      "unsupported method (PATCH) stateful",
-			stateless: false,
-			method:    "PATCH",
-			wantAllow: "GET, POST, DELETE",
+			name:       "GET in stateless mode",
+			stateless:  true,
+			method:     "GET",
+			wantStatus: http.StatusMethodNotAllowed,
+			wantAllow:  "POST",
 		},
 		{
-			name:      "GET in stateless mode",
-			stateless: true,
-			method:    "GET",
-			wantAllow: "POST",
-		},
-		{
-			name:      "unsupported method (PATCH) stateless",
-			stateless: true,
-			method:    "PATCH",
-			wantAllow: "POST",
-		},
-		{
-			name:      "unsupported method (PUT) stateless",
-			stateless: true,
-			method:    "PUT",
-			wantAllow: "POST",
+			// DELETE without session returns 400 Bad Request (not 405)
+			// because DELETE is a valid method, just requires a session ID.
+			name:       "DELETE without session stateless",
+			stateless:  true,
+			method:     "DELETE",
+			wantStatus: http.StatusBadRequest,
+			wantAllow:  "", // No Allow header for 400 responses
 		},
 	}
 
@@ -1941,8 +1935,8 @@ func TestStreamable405AllowHeader(t *testing.T) {
 			}
 			defer resp.Body.Close()
 
-			if got, want := resp.StatusCode, http.StatusMethodNotAllowed; got != want {
-				t.Errorf("status code: got %d, want %d", got, want)
+			if got := resp.StatusCode; got != tt.wantStatus {
+				t.Errorf("status code: got %d, want %d", got, tt.wantStatus)
 			}
 
 			allow := resp.Header.Get("Allow")


jba (dismissed):

SamMorrowDrums:
Pushed amended commit with gofmt applied properly.

mcp/streamable_test.go
```diff
@@ -1877,6 +1877,80 @@ func TestStreamableGET(t *testing.T) {
...
+func TestStreamable405AllowHeader(t *testing.T) {
+	server := NewServer(testImpl, nil)
+
+	tests := []struct {
+		name        string
+		stateless   bool
+		method      string
+		wantAllow   string
+		wantStatus  int
+		withSession bool
```
maciej-kisiel: `withSession` is not used, let's remove it (I assume GET with session in a stateful setup is covered by other tests). Similarly `wantStatus` is always set to the same value, I would propose to inline it into the test case, similarly as was done in SSE test.

mcp/sse_test.go
```diff
@@ -186,3 +186,38 @@ func TestSSEClientTransport_HTTPErrors(t *testing.T) {
...
+				t.Fatal(err)
+			}
+
+			resp, err := http.DefaultClient.Do(req)
+			if err != nil {
+				t.Fatal(err)
+			}
+			defer resp.Body.Close()
+
+			if got := resp.StatusCode; got != http.StatusMethodNotAllowed {
```
maciej-kisiel: Nit: `http.StatusMethodNotAllowed` could be assigned to `want` to avoid repetition, similarly as was done with `resp.StatusCode`.

SamMorrowDrums:
Thanks for the review @maciej-kisiel, I agree with your comments.

mcp/streamable.go
```diff
@@ -275,6 +275,13 @@ func (h *StreamableHTTPHandler) ServeHTTP(w http.ResponseWriter, req *http.Reque
 	switch req.Method {
 	case http.MethodPost, http.MethodGet:
 		if req.Method == http.MethodGet && (h.opts.Stateless || sessionID == "") {
+			// RFC 9110 §15.5.6: 405 responses MUST include Allow header.
+			// In stateless mode, only POST is allowed (no persistent sessions for GET/DELETE).
+			if h.opts.Stateless {
+				w.Header().Set("Allow", "POST")
+			} else {
+				w.Header().Set("Allow", "GET, POST, DELETE")
```
maciej-kisiel: Wouldn't sending a GET request and receiving a 405 with Allow header containing the GET method be a bit confusing? Maybe when the Handler is not in stateless mode GET with empty session should return BadRequest, just like DELETE does in line 263?

mcp/streamable.go
maciej-kisiel: Should this take the stateless mode into account as well?

SamMorrowDrums:
@maciej-kisiel they are great questions. I was thinking about some of them too when writing. 

I also suggested in MCP discord the idea of adding some HTTP conformance tests to the SDK conformance testing, and so I think as well as trying to get this correct first time, I think we will end up with http conformance being checked and consistent across SDKs with community decision on some of these things, I hope this PR will be the guide for some that work, but I also think that there will likely be follow-ups.

SamMorrowDrums:
Thanks for the review @maciej-kisiel, I agree with your comments.

After researching the MCP spec, I've made the following changes:

**GET without session in stateful mode → 400 Bad Request**
- You're right - returning 405 with GET in the Allow header was confusing
- GET IS a supported method in stateful mode, it just requires a session ID as a precondition
- This now matches the behavior of DELETE without session (line 263)

**Allow header in stateless mode → `POST` only**
- Since stateless mode doesn't support sessions, neither GET (SSE streaming) nor DELETE (session termination) are useful
- The MCP spec says: *'The server MUST either return Content-Type: text/event-stream in response to this HTTP GET, or else return HTTP 405 Method Not Allowed, indicating that the server does not offer an SSE stream at this endpoint.'*

**Summary of behavior now:**
| Mode | Method | Status | Allow Header |
|------|--------|--------|--------------|
| Stateful | GET without session | 400 Bad Request | (none) |
| Stateful | Unsupported method | 405 | GET, POST, DELETE |
| Stateless | GET | 405 | POST |
| Stateless | Unsupported method | 405 | POST |

mcp/streamable_test.go
```diff
@@ -1877,6 +1879,107 @@ func TestStreamableGET(t *testing.T) {
...
+	server := NewServer(testImpl, nil)
+
+	tests := []struct {
+		name      string
+		stateless bool
+		method    string
+		wantAllow string
+	}{
+		{
+			name:      "unsupported method (PUT) stateful",
```
maciej-kisiel: I'm not sure how much value testing both PUT and PATCH brings in both stateful and stateless setups, they are generally not used in the protocol, so one is probably sufficient. Instead, I think I would additionally add "DELETE in stateless mode" case, as it's a one that differs between the modes. WDYT?

maciej-kisiel: Actually, wait, now I realize that DELETE is handled separately here. I'm not sure it's behavior is super correct, but we can probably outscope its adjustment from this PR. Please disregard the above comment.

SamMorrowDrums: I did have a look at adjusting the test.

SamMorrowDrums: I have pushed, but can remove the last commit if it's wrong direction.

findleyr:
LGTM! Thanks @SamMorrowDrums for the contribution, and @maciej-kisiel for the review!

PR #760, opened 2026-01-22
feat: add automatic DNS rebinding protection for localhost servers

pcarleton:
## Summary

Add DNS rebinding protection that is automatically enabled when requests arrive via localhost (127.0.0.1, [::1]). This protects against malicious websites using DNS rebinding to interact with local MCP servers.

(note: this was a claude assisted PR, mostly wanted to see how difficult passing this test would be for this SDK)

## Design Goal: Secure by Default

The primary goal is to make it difficult to run a localhost server without these protections by mistake. There are other approaches that could provide secure defaults (e.g., a helper for `ListenAndServe` that does the localhost check), but using `http.LocalAddrContextKey` for runtime detection seemed like the most backwards compatible approach and least likely to be disabled by accident.

With this implementation:
- **No code changes required** - existing servers get protection automatically
- **No opt-in needed** - protection activates based on the connection's local address
- **Explicit opt-out** - users must deliberately set `DisableLocalhostProtection: true`

## Changes

- Add `DisableLocalhostProtection` option to `StreamableHTTPOptions`
- Add `isLocalhostAddr` and `isLocalhostHost` helper functions  
- Validate Host header at start of `ServeHTTP`, rejecting non-localhost Host headers with 403 Forbidden

## How it works

The protection uses `http.LocalAddrContextKey` to detect the connection's local address at runtime. When a request arrives via localhost (127.0.0.1 or [::1]), the handler validates that the Host header also matches a localhost value. If not, the request is rejected with 403 Forbidden.

This approach means:
- Protection is enabled for requests arriving via localhost, regardless of whether the server listens on `127.0.0.1` or `0.0.0.0`
- Requests arriving via non-localhost IPs (e.g., external network requests) are **not** affected

### Edge case: Reverse proxies

If a reverse proxy (e.g., Envoy, nginx) runs on the same host and forwards requests to the MCP server via localhost while preserving the original Host header, those requests would be rejected. In this case, users should either:
1. Set `DisableLocalhostProtection: true`
2. Configure the proxy to rewrite the Host header to localhost

## Testing

- Added unit tests for `isLocalhostAddr` and `isLocalhostHost` helper functions
- Added integration tests for the full protection flow
- Verified against the MCP conformance test suite:
  - `localhost-host-rebinding-rejected`: PASS
  - `localhost-host-valid-accepted`: PASS

## Related

- Spec: https://modelcontextprotocol.io/specification/2025-11-25/basic/security_best_practices#local-mcp-server-compromise
- TypeScript SDK implementation: `localhostHostValidation()` middleware
- Conformance test: `dns-rebinding-protection` scenario

### DIFF
--- a/mcp/streamable.go
+++ b/mcp/streamable.go
@@ -31,6 +31,7 @@ import (
 
 	"github.com/modelcontextprotocol/go-sdk/auth"
 	"github.com/modelcontextprotocol/go-sdk/internal/jsonrpc2"
+	"github.com/modelcontextprotocol/go-sdk/internal/mcpgodebug"
 	"github.com/modelcontextprotocol/go-sdk/internal/util"
 	"github.com/modelcontextprotocol/go-sdk/internal/xcontext"
 	"github.com/modelcontextprotocol/go-sdk/jsonrpc"
@@ -219,10 +220,16 @@ func (h *StreamableHTTPHandler) closeAll() {
 	}
 }
 
+// disablelocalhostprotection is a compatibility parameter that allows to disable
+// DNS rebinding protection, which was added in the 1.4.0 version of the SDK.
+// See the documentation for the mcpgodebug package for instructions how to enable it.
+// The option will be removed in the 1.6.0 version of the SDK.
+var disablelocalhostprotection = mcpgodebug.Value("disablelocalhostprotection")
+
 func (h *StreamableHTTPHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 	// DNS rebinding protection: auto-enabled for localhost servers.
 	// See: https://modelcontextprotocol.io/specification/2025-11-25/basic/security_best_practices#local-mcp-server-compromise
-	if !h.opts.DisableLocalhostProtection {
+	if !h.opts.DisableLocalhostProtection && disablelocalhostprotection != "1" {
 		if localAddr, ok := req.Context().Value(http.LocalAddrContextKey).(net.Addr); ok && localAddr != nil {
 			if util.IsLoopback(localAddr.String()) && !util.IsLoopback(req.Host) {
 				http.Error(w, fmt.Sprintf("Forbidden: invalid Host header %q", req.Host), http.StatusForbidden)
--- a/internal/util/net.go
+++ b/internal/util/net.go
@@ -1,3 +1,6 @@
+// Copyright 2025 The Go MCP SDK Authors. All rights reserved.
+// Use of this source code is governed by the license
+// that can be found in the LICENSE file.
 package util
 
 import (
--- a/internal/util/net_test.go
+++ b/internal/util/net_test.go
@@ -1,7 +1,6 @@
 // Copyright 2025 The Go MCP SDK Authors. All rights reserved.
-// Use of this source code is governed by an MIT-style
-// license that can be found in the LICENSE file.
-
+// Use of this source code is governed by the license
+// that can be found in the LICENSE file.
 package util
 
 import "testing"


mcp/streamable.go
```diff
@@ -206,7 +217,46 @@ func (h *StreamableHTTPHandler) closeAll() {
...
+	// Remove brackets for IPv6
+	hostname = strings.TrimPrefix(strings.TrimSuffix(hostname, "]"), "[")
+	return hostname == "localhost" || hostname == "127.0.0.1" || hostname == "::1"
+}
+
 func (h *StreamableHTTPHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
+	// DNS rebinding protection: auto-enabled for localhost servers.
+	// See: https://modelcontextprotocol.io/specification/2025-11-25/basic/security_best_practices#local-mcp-server-compromise
+	if !h.opts.DisableLocalhostProtection {
+		if localAddr, ok := req.Context().Value(http.LocalAddrContextKey).(net.Addr); ok {
```
maciej-kisiel: With this change below I think we could only have a single implementation of the `isLocalhost` function, since they are largely the same. It would be called with `localAddr.String()` and `req.Host`.

```suggestion
		if localAddr := req.Context().Value(http.LocalAddrContextKey).(net.Addr); localAddr != nil {
```

mcp/streamable_test.go
```diff
@@ -2280,3 +2280,138 @@ data: {"jsonrpc":"2.0","method":"test2","params":{}}
...
+	server := NewServer(testImpl, nil)
+
+	tests := []struct {
+		name        string
+		listenAddr  string // Address to listen on
+		hostHeader  string // Host header in request
+		disableProt bool   // DisableLocalhostProtection setting
+		wantStatus  int
+	}{
+		// Auto-enabled for localhost listeners (127.0.0.1)
```
maciej-kisiel: Extending the test cases to multiple lines and including field names is more readable. I would suggest to do it.

mcp/streamable_test.go
```diff
@@ -2280,3 +2280,138 @@ data: {"jsonrpc":"2.0","method":"test2","params":{}}
...
+// TestStreamableLocalhostProtection verifies that DNS rebinding protection
+// is automatically enabled for localhost servers.
+func TestStreamableLocalhostProtection(t *testing.T) {
+	server := NewServer(testImpl, nil)
+
+	tests := []struct {
+		name        string
+		listenAddr  string // Address to listen on
+		hostHeader  string // Host header in request
+		disableProt bool   // DisableLocalhostProtection setting
```
maciej-kisiel: `disableProtection` should be fine.

mcp/streamable_test.go
```diff
@@ -2280,3 +2280,138 @@ data: {"jsonrpc":"2.0","method":"test2","params":{}}
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			opts := &StreamableHTTPOptions{
+				Stateless:                  true, // Simpler for testing
+				DisableLocalhostProtection: tt.disableProt,
+			}
+			handler := NewStreamableHTTPHandler(func(req *http.Request) *Server { return server }, opts)
+
+			// Create a listener on the specified address to control LocalAddrContextKey
```
maciej-kisiel:
```suggestion
			// Create a listener on the specified address to control LocalAddrContextKey.
```

mcp/streamable_test.go
```diff
@@ -2280,3 +2280,138 @@ data: {"jsonrpc":"2.0","method":"test2","params":{}}
...
+			listener, err := net.Listen("tcp", tt.listenAddr)
+			if err != nil {
+				t.Fatalf("failed to listen on %s: %v", tt.listenAddr, err)
+			}
+			defer listener.Close()
+
+			// Start server in background
```
maciej-kisiel:
```suggestion
			// Start the server in the background.
```

maciej-kisiel:
Hi @pcarleton, based on the inactivity I assume you just wanted to share your PoC, but have no bandwidth to finalize it. If that's the case, I will take this over and finalize it in the next few days.

herczyn (approved):

pcarleton:
@maciej-kisiel yes thank you! apologies, my github notifications are a mess

PR #764, opened 2026-01-26
conformance: prepare the conformance test suite

maciej-kisiel:
- move the conformance related files to a dedicated directory
- simplify parts of the conformance script
- create a scaffolding for a new GitHub action to run conformance tests
- use the new `--output-dir` flag from the conformance suite for saving results on demand


.github/workflows/conformance.yml
```diff
@@ -0,0 +1,55 @@
...
+        uses: actions/setup-go@d35c59abb061a4a6fb18e82ac0862c26744d6ab5 # v5
+        with:
+          go-version: "^1.25"
+      - name: "Run conformance tests" 
+        uses: modelcontextprotocol/conformance@c2f3fdaf781dcd5a862cb0d2f6454c1c210bf0f0 # v0.1.11
+        with:
+          mode: client 
+          command: go run ./conformance/everything-client/main.go
+          suite: core
+          expected-failures: ./conformance/baseline.yml
```
findleyr: Nit: should have a trailing newline here, otherwise it doesn't work well with tools (e.g. diff tools)

scripts/server-conformance.sh
```diff
@@ -79,28 +77,24 @@ echo "Server pid is $SERVER_PID"
...
         npm --prefix "$CONFORMANCE_REPO" run start -- \
-            server --url "http://localhost:$PORT")
+            server --url "http://localhost:$PORT" \
+            ${RESULT_DIR:+--output-dir "$RESULT_DIR"}) || true
 else
     (cd "$WORKDIR" && \
-        npx @modelcontextprotocol/conformance@latest server --url "http://localhost:$PORT")
+        npx @modelcontextprotocol/conformance@latest \
+        server --url "http://localhost:$PORT" \
+        ${RESULT_DIR:+--output-dir "$RESULT_DIR"}) || true
```
findleyr: Why are we adding || true here and above? Don't we want this to fail?

scripts/server-conformance.sh
```diff
@@ -50,11 +50,9 @@ done
 
 cleanup() {
     if [ -n "$SERVER_PID" ]; then
+        echo "Stopping server..."
         kill "$SERVER_PID" 2>/dev/null || true
-    fi
-    # Clean up the work directory unless --result_dir was specified.
-    if [ -z "$RESULT_DIR" ] && [ -n "$WORKDIR" ]; then
```
findleyr, line 56: Why remove this cleanup?

scripts/client-conformance.sh
```diff
@@ -0,0 +1,82 @@
+#!/bin/bash
+# Copyright 2025 The Go MCP SDK Authors. All rights reserved.
+# Use of this source code is governed by an MIT-style
+# license that can be found in the LICENSE file.
+
+# Run MCP conformance tests against the Go SDK conformance client.
+
+set -e
+
+RESULT_DIR=""
```
findleyr, line 10: Do we need to clean up here, like we had been doing in the conformance script?

scripts/client-conformance.sh
```diff
@@ -0,0 +1,82 @@
...
+go build -o "$WORKDIR/conformance-client" ./conformance/everything-client
+
+# Run conformance tests from the work directory to avoid writing results to the repo.
+echo "Running conformance tests..."
+if [ -n "$CONFORMANCE_REPO" ]; then
+    # Run from local checkout using npm run start.
+    (cd "$WORKDIR" && \
+        npm --prefix "$CONFORMANCE_REPO" run start -- \
+            client --command "$WORKDIR/conformance-client" \
+            --scenario sse-retry \
```
findleyr: Did you mean to leave this as --scenario sse-retry?

conformance/everything-client/main.go
```diff
@@ -0,0 +1,249 @@
...
+	tools, err := session.ListTools(ctx, nil)
+	if err != nil {
+		return fmt.Errorf("session.ListTools(): %v", err)
+	}
+
+	idx := slices.IndexFunc(tools.Tools, func(t *mcp.Tool) bool {
+		return t.Name == "add_numbers"
+	})
+
+	if idx >= 0 {
```
findleyr: Shouldn't it be an error if the tool is not found?

scripts/server-conformance.sh
maciej-kisiel: My intention was to still print the text about the result directory, even when the tests failed, as looking into results may help debugging. But indeed not returning non-zero code from the script will break the CI. I'll handle this more explicitly then.

scripts/server-conformance.sh
maciej-kisiel: I switched to the new `–output-dir` flag implemented recently in the conformance suite. This way there are two scenarios:
- `result-dir` is provided to the script and passed to conformance tests, so results shouldn't be removed.
- `result-dir` is not provided to the script and `–output-dir` is not passed to conformance tests, so no output is produced.

See https://github.com/modelcontextprotocol/conformance/pull/107/changes/b9bc062534a023d7b1e75dde562f8e21a0fe8a49 for more details.

scripts/client-conformance.sh
maciej-kisiel: I think my comment for the server-conformance file should address it. Please re-open otherwise.

conformance/everything-client/main.go
maciej-kisiel: Good question: in general we can think how much we should assume with regard to correctness of the conformance suite. The TS SDK is not consistent in that regard, in some cases it errors, in others it doesn't. We can probably be strict and error.

.github/workflows/conformance.yml
maciej-kisiel: Done

scripts/client-conformance.sh
maciej-kisiel: Good catch! Fixed

conformance/everything-client/main.go
maciej-kisiel: Moved to strict expectations across the scenarios.

scripts/server-conformance.sh
maciej-kisiel: Should be correct now, PTAL

Issue #758, opened 2026-01-21
[MCP Elicitation] Content validation should be done only for accept action


vnalawad-tibco:
**Describe the bug**
When elicitation request is canceled or declined, current implementation still attempts to validate content against schema

**To Reproduce**
1. Register MCP server (with tool sending elicitation request) in GitHub Copilot
2. Enter prompt that would trigger MCP server tool
3. When elicitation request received, cancel it

**Expected behavior**
In such scenario, SDK call(ServerSession.Elicit(...)) expected to return response back with **cancel** action.
Instead, implementation attempts to validate elicitation response.

**Logs**

**Additional context**
I believe validating content against schema only for **accept** action type should solve this issue.

maciej-kisiel:
I can confirm that schema validation happens regardless of the `action` that is returned from the client.

@vnalawad-tibco would you like to contribute a fix and some test cases that verify it?

PR #766, opened 2026-01-28
mcp: validation only for accept action

CocaineCong:
Fix: #758

- validate schema only for accept action
- add enum for action in consts.go
- add unit test


mcp/elicitation_test.go
```diff
@@ -201,3 +201,72 @@ func TestElicitationCompleteNotification(t *testing.T) {
 		t.Fatal("timed out waiting for elicitation complete notification")
 	}
 }
+
+func TestElicitationValidationBug(t *testing.T) {
```
maciej-kisiel:
```suggestion
func TestElicitationNoValidationWithoutAccept(t *testing.T) {
```

mcp/elicitation_test.go
```diff
@@ -201,3 +201,72 @@ func TestElicitationCompleteNotification(t *testing.T) {
...
+	ctx := context.Background()
+
+	// Schema that requires a field "test"
```
maciej-kisiel: I would delete this comment, the code is self-explanatory.

mcp/consts.go
```diff
@@ -0,0 +1,11 @@
+// Copyright 2025 The Go MCP SDK Authors. All rights reserved.
+// Use of this source code is governed by an MIT-style
+// license that can be found in the LICENSE file.
+
+package mcp
+
+const (
+	ElicitActionAccept  = "accept"  // User submitted the form/confirmed the action
```
maciej-kisiel: I think I would just inline these variables and not introduce this file. The const names are longer than the actual strings and given that these are part of the protocol specification, it's unlikely that we will need to change their values across the code base as that would be backwards incompatible.

CocaineCong:
> Thanks a lot for the contribution! I left a few minor comments.

Got it, I've already made all the changes. PTAL.

maciej-kisiel:
Thanks a lot for contributing!

PR #769, opened 2026-01-28
mcp: update JSON marshaling to not HTML-escape messages

### LINKED ISSUE #238: Idea: environment variable for experiments and compatibility
Once this module is at v1, we won't make any breaking API changes. But we may make slight behavior changes. For example, our streaming transport implementation may return more fine-grained HTTP status codes to help clients distinguish different types of errors, or we might turn some code paths that are currently successful into errors if they violate the specification or they inevitably lead to errors further on. We want to give users time to adapt their implementation to these changes.

We also may conduct experiments, adding new features that we're not sure are the right ones, or have the right API. We want to allow users to opt in to these experiments.

We propose to provide for both of these cases with an environment variable, GOMCPSDK. Its value will be a comma-separated list of key=value pairs. Each entry will either restore old behavior or enable an experiment.

The README for the module will document the each entry and, for compatibility-preserving entries, when they will expire. A typical compatibility entry will expire after one minor version has passed, locking in the new behavior. For example, an entry introduced in v1.3 will be removed in v1.5.

### LINKED ISSUE #767: StreamableHTTP transport timeout issue.
**Describe the bug**
The Go MCP SDK client experiences a 60-second timeout (504 Gateway Timeout) when calling tools on the remote MCP server at mcp.on-demand.io, while the Node.js MCP SDK client successfully completes the same request in ~1 second. The issue appears to be a server-side compatibility problem with how the Go SDK's requests are processed.

**To Reproduce**
1. Install the official Go MCP SDK:
```
go get github.com/modelcontextprotocol/go-sdk/mcp
```
2. Create a Go client connecting to the remote MCP server:
```
   client := mcp.NewClient(&mcp.Implementation{
       Name:    "mcp-mermaid-client",
       Version: "1.0.0",
   }, nil)
   transport := &mcp.StreamableClientTransport{
       Endpoint:   "https://mcp.on-demand.io/instances/presight-mermaid-mcp/mcp?apiKey=<REDACTED>",
       HTTPClient: &http.Client{Timeout: 0},
       MaxRetries: 3,
   }
```
3. Call the generate_mermaid_diagram tool:
```
   params := &mcp.CallToolParams{
       Name: "generate_mermaid_diagram",
       Arguments: map[string]any{
           "mermaid": "graph TD\n    A-->B",
           "outputType": "mermaid",
       },
   }
   
   result, err := session.CallTool(ctx, params)
```

**Expected behavior**
The Go client should successfully call the generate_mermaid_diagram tool and receive the mermaid diagram response within a reasonable time (~1-2 seconds), similar to the Node.js client behavior.

**Logs**
```
2026/01/28 12:50:42 Calling generate_mermaid_diagram tool...
2026/01/28 12:50:42 Starting tool call...
2026/01/28 12:50:42 === HTTP REQUEST ===
POST /instances/presight-mermaid-mcp/mcp?apiKey=<REDACTED> HTTP/1.1
Host: mcp.on-demand.io
User-Agent: Go-http-client/1.1
Content-Length: 165
Accept: application/json, text/event-stream
Content-Type: application/json
Mcp-Protocol-Version: 2025-06-18
Accept-Encoding: gzip

{"jsonrpc":"2.0","id":2,"method":"tools/call","params":{"name":"generate_mermaid_diagram","arguments":{"mermaid":"graph TD\n    A--\u003eB","outputType":"mermaid"}}}

2026/01/28 12:51:42 === HTTP RESPONSE (after 1m0.316470083s) ===
HTTP/2.0 504 Gateway Timeout
Content-Length: 102
Access-Control-Allow-Origin: *
Content-Security-Policy: default-src 'none'; frame-ancestors 'none'
Content-Type: application/json; charset=utf-8
Date: Wed, 28 Jan 2026 07:21:42 GMT

2026/01/28 12:51:42 CallTool failed after 1m0.316810209s: calling "tools/call": rejected by transport: sending "tools/call": Gateway Timeout
```

maciej-kisiel:
It's causing some servers to fail: #767. I checked with Typescript implementation and it doesn't do any escaping.

It's still a behavior change.

Fixes #238 
Fixes #767

### DIFF
--- a/internal/jsonrpc2/messages.go
+++ b/internal/jsonrpc2/messages.go
@@ -9,6 +9,8 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+
+	"github.com/modelcontextprotocol/go-sdk/internal/mcpgodebug"
 )
 
 // ID is a Request identifier, which is defined by the spec to be a string, integer, or null.
@@ -159,12 +161,14 @@ func EncodeMessage(msg Message) ([]byte, error) {
 func EncodeIndent(msg Message, prefix, indent string) ([]byte, error) {
 	wire := wireCombined{VersionTag: wireVersion}
 	msg.marshal(&wire)
-	// TODO: do we need to escape HTML here?
-	data, err := json.MarshalIndent(&wire, prefix, indent)
-	if err != nil {
-		return data, fmt.Errorf("marshaling jsonrpc message: %w", err)
+	var buf bytes.Buffer
+	enc := json.NewEncoder(&buf)
+	enc.SetEscapeHTML(false)
+	enc.SetIndent(prefix, indent)
+	if err := enc.Encode(&wire); err != nil {
+		return nil, fmt.Errorf("marshaling jsonrpc message: %w", err)
 	}
-	return data, nil
+	return bytes.TrimRight(buf.Bytes(), "\n"), nil
 }
 
 func DecodeMessage(data []byte) (Message, error) {
@@ -213,8 +217,18 @@ func marshalToRaw(obj any) (json.RawMessage, error) {
 	return json.RawMessage(data), nil
 }
 
+// jsonescaping is a compatibility parameter that allows to restore
+// JSON escaping in the JSON marshaling, which stopped being the default
+// in the 1.4.0 version of the SDK. See the documentation for the
+// mcpgodebug package for instructions how to enable it.
+// The option will be removed in the 1.6.0 version of the SDK.
+var jsonescaping = mcpgodebug.Value("jsonescaping")
+
 // jsonMarshal marshals obj to JSON like json.Marshal but without HTML escaping.
 func jsonMarshal(obj any) ([]byte, error) {
+	if jsonescaping == "1" {
+		return json.Marshal(obj)
+	}
 	var buf bytes.Buffer
 	enc := json.NewEncoder(&buf)
 	enc.SetEscapeHTML(false)
--- a/internal/mcpgodebug/mcpgodebug.go
+++ b/internal/mcpgodebug/mcpgodebug.go
@@ -1,0 +1,52 @@
+// Copyright 2025 The Go MCP SDK Authors. All rights reserved.
+// Use of this source code is governed by the license
+// that can be found in the LICENSE file.
+
+// Package mcpgodebug provides a mechanism to configure compatibility parameters
+// via the MCPGODEBUG environment variable.
+//
+// The value of MCPGODEBUG is a comma-separated list of key=value pairs.
+// For example:
+//
+//	MCPGODEBUG=someoption=1,otheroption=value
+package mcpgodebug
+
+import (
+	"fmt"
+	"os"
+	"strings"
+)
+
+const compatibilityEnvKey = "MCPGODEBUG"
+
+var compatibilityParams map[string]string
+
+func init() {
+	var err error
+	compatibilityParams, err = parseCompatibility(os.Getenv(compatibilityEnvKey))
+	if err != nil {
+		panic(err)
+	}
+}
+
+// Value returns the value of the compatibility parameter with the given key.
+// It returns an empty string if the key is not set.
+func Value(key string) string {
+	return compatibilityParams[key]
+}
+
+func parseCompatibility(envValue string) (map[string]string, error) {
+	if envValue == "" {
+		return nil, nil
+	}
+
+	params := make(map[string]string)
+	for part := range strings.SplitSeq(envValue, ",") {
+		k, v, ok := strings.Cut(part, "=")
+		if !ok {
+			return nil, fmt.Errorf("MCPGODEBUG: invalid format: %q", part)
+		}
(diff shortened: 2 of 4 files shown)


jba:
We should provide a way to preserve the old behavior. We don't know who depends on it.
See https://github.com/modelcontextprotocol/go-sdk/issues/238.
For example code, see https://github.com/google/jsonschema-go/blob/main/jsonschema/infer.go#L138.

internal/jsonrpc2/messages.go
```diff
@@ -158,6 +159,7 @@ func EncodeMessage(msg Message) ([]byte, error) {
 func EncodeIndent(msg Message, prefix, indent string) ([]byte, error) {
 	wire := wireCombined{VersionTag: wireVersion}
 	msg.marshal(&wire)
+	// TODO: do we need to escape HTML here?
```
jba: It's only used in tests, but why should tests behave differently? So you should turn off escaping here too.

internal/jsonrpc2/messages.go
```diff
@@ -204,9 +206,21 @@ func marshalToRaw(obj any) (json.RawMessage, error) {
 	if obj == nil {
 		return nil, nil
 	}
-	data, err := json.Marshal(obj)
+	data, err := jsonMarshal(obj)
```
jba, line 213: If you grep the codebase, there are many other calls to json.Marshal. Should they all go to jsonMarshal? I'm not sure.

findleyr: If this is specifically about messages, this is probably the only location that needs to change.

internal/jsonrpc2/messages.go
maciej-kisiel: Done

maciej-kisiel:
Added a simple package to allow specifying compatibility flag as suggested in https://github.com/modelcontextprotocol/go-sdk/issues/238.

internal/mcpgodebug/mcpgodebug.go
```diff
@@ -0,0 +1,52 @@
...
+
+// Value returns the value of the compatibility parameter with the given key.
+// It returns an empty string if the key is not set.
+func Value(key string) string {
+	return compatibilityParams[key]
+}
+
+func parseCompatibility(envValue string) (map[string]string, error) {
+	if envValue == "" {
+		return make(map[string]string), nil
```
jba: return nil, nil

internal/mcpgodebug/mcpgodebug.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+
+	params := make(map[string]string)
+	for _, part := range strings.Split(envValue, ",") {
+		kv := strings.Split(part, "=")
+		if len(kv) != 2 {
+			return nil, fmt.Errorf("MCPGODEBUG: invalid format: %q", part)
+		}
+		params[kv[0]] = kv[1]
```
jba: strings.TrimSpace for both kv's

internal/mcpgodebug/mcpgodebug.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+func parseCompatibility(envValue string) (map[string]string, error) {
+	if envValue == "" {
+		return make(map[string]string), nil
+	}
+
+	params := make(map[string]string)
+	for _, part := range strings.Split(envValue, ",") {
+		kv := strings.Split(part, "=")
```
jba: Better:
```
k, v, ok := strings.Cut(part, "=")
if !ok {error}
```
A bit easier to write and also allows = in value (why not?)

internal/mcpgodebug/mcpgodebug_test.go
```diff
@@ -0,0 +1,70 @@
...
+
+package mcpgodebug
+
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+func TestParseCompatibility_Success(t *testing.T) {
```
jba, line 13: nit: avoid underscores in test function names. (We seem to have a couple, maybe that's Claude's fault.)

internal/mcpgodebug/mcpgodebug_test.go
```diff
@@ -0,0 +1,70 @@
...
+	got, err := parseCompatibility("foo=bar,baz=qux")
```
jba: test with spaces around some parts

internal/mcpgodebug/mcpgodebug.go
maciej-kisiel: Thanks! I thought it will panic in Value if it's `nil` but TIL that only writes panic, reads do not.

internal/mcpgodebug/mcpgodebug_test.go
maciej-kisiel: They seem to generally be allowed, even in the Google Go style which is pretty opinionated: https://google.github.io/styleguide/go/decisions#underscores. From my perspective they allow easier separation of the function under test from the testing scenario. Do you feel strongly?

jba: You can leave them.

Issue #628, opened 2025-11-07
SEP-990: investigate support for enterprise IdP policy controls extension


findleyr:
Investigate if we can/should implement support for [SEP-990](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/990), and enterprize auth flow. This is an MCP extension.

radar07:
@findleyr I'd like to work on this since my team is working on implementing this in other SDKs

findleyr:
@radar07 that would be great. I'm not sure exactly how we'll approach SDK extensions, but if you want to map out this change, we can help figure out how to integrate it into the SDK.

radar07:
@findleyr [This](https://docs.google.com/document/d/1THr7iu8tsLPIrVSirhNhi_bXCpZu2DeU7QG_S3DxOhQ/edit?tab=t.0) is the plan I'm going far and I've been working on a [test PR](https://github.com/radar07/go-sdk/pull/1). Let me know if you any other ideas for this one.

PR #770, opened 2026-01-30
Enterprise managed authorization

radar07:
`auth`, `oauthex`: implement Enterprise Managed Authorization (SEP-990)
This PR implements Enterprise Managed Authorization (SEP-990) for the Go MCP SDK, enabling MCP Clients and Servers to leverage enterprise Identity Providers for seamless authorization without requiring users to authenticate separately to each MCP Server.
 Overview
Enterprise Managed Authorization follows the Identity Assertion Authorization Grant specification (draft-ietf-oauth-identity-assertion-authz-grant), implementing a three-step flow:
1. **Single Sign-On (SSO)**: User authenticates to the MCP Client via enterprise IdP (Okta, Auth0, Azure AD, etc.)
2. **Token Exchange (RFC 8693)**: Client exchanges ID Token for Identity Assertion JWT Authorization Grant (ID-JAG) at the IdP
3. **JWT Bearer Grant (RFC 7523)**: Client exchanges ID-JAG for Access Token at the MCP Server
This enables:
- **For end users**: Single sign-on across MCP Clients and Servers—no manual connection/authorization per server
- **For enterprise admins**: Centralized visibility and control over which MCP Servers can be used within the organization
- **For MCP clients**: Automatic token acquisition without user interaction for each server

Closes: [#628](https://github.com/modelcontextprotocol/go-sdk/issues/628)

### DIFF
--- a/oauthex/oauth2.go
+++ b/oauthex/oauth2.go
@@ -10,7 +10,6 @@ package oauthex
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"io"
 	"mime"
@@ -18,6 +17,7 @@ import (
 	"net/url"
 	"strings"
 
+	"github.com/modelcontextprotocol/go-sdk/internal/json"
 	"github.com/modelcontextprotocol/go-sdk/internal/util"
 )
 
--- a/internal/json/json.go
+++ b/internal/json/json.go
@@ -8,12 +8,25 @@ package json
 
 import (
 	"bytes"
+	"io"
 
 	"github.com/segmentio/encoding/json"
 )
 
-func Unmarshal(data []byte, v any) error {
-	dec := json.NewDecoder(bytes.NewReader(data))
+type Decoder struct {
+	dec *json.Decoder
+}
+
+func NewDecoder(r io.Reader) *Decoder {
+	dec := json.NewDecoder(r)
 	dec.DontMatchCaseInsensitiveStructFields()
-	return dec.Decode(v)
+	return &Decoder{dec: dec}
+}
+
+func (d *Decoder) Decode(v any) error {
+	return d.dec.Decode(v)
+}
+
+func Unmarshal(data []byte, v any) error {
+	return NewDecoder(bytes.NewReader(data)).Decode(v)
 }
--- a/internal/json/json_test.go
+++ b/internal/json/json_test.go
@@ -8,6 +8,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"strings"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
@@ -60,6 +61,43 @@ func TestUnmarshalCaseSensitivity(t *testing.T) {
 			}
 			if diff := cmp.Diff(tt.want, got); diff != "" {
 				t.Errorf("Unmarshal mismatch (-want +got):\n%s", diff)
+			}
+		})
+	}
+}
+
+func TestNewDecoderCaseSensitivity(t *testing.T) {
+	type Target struct {
+		Field       string `json:"field"`
+		TaggedField string `json:"custom_tag"`
+	}
+
+	tests := []struct {
+		name string
+		json string
+		want Target
+	}{
+		{
+			name: "exact match",
+			json: `{"field": "value", "custom_tag": "tagged"}`,
+			want: Target{Field: "value", TaggedField: "tagged"},
+		},
+		{
+			name: "case mismatch",
+			json: `{"Field": "value", "Custom_tag": "tagged"}`,
+			want: Target{},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			var got Target
+			dec := NewDecoder(strings.NewReader(tt.json))
+			if err := dec.Decode(&got); err != nil {
+				t.Fatalf("Decode failed: %v", err)
+			}
+			if diff := cmp.Diff(tt.want, got); diff != "" {
+				t.Errorf("Decode mismatch (-want +got):\n%s", diff)
 			}
 		})
 	}


maciej-kisiel:
Hi @radar07, thanks for submitting this PR. Could you link the issue that it is addressing?

Also, as a heads-up: it will likely take some time to review your proposal. Both because it's quite large, but more importantly I'm also working on a proposal how to structure the client-side OAuth implementation and this change will need to be aligned with it.

radar07:
Thanks @maciej-kisiel. I updated the description with the SEP that this PR solves.

radar07:
@maciej-kisiel I'd be happy to contribute to OAuth implementation. Let me know if I can help with anything. Just want to know if I should add conformance tests to this because I can see that there are PRs related to conformance tests.

oauthex/jwt_bearer.go
```diff
@@ -0,0 +1,193 @@
...
+	"golang.org/x/oauth2"
+)
+
+// GrantTypeJWTBearer is the grant type for RFC 7523 JWT Bearer authorization grant.
+// This is used in SEP-990 to exchange an ID-JAG for an access token at the MCP Server.
+const GrantTypeJWTBearer = "urn:ietf:params:oauth:grant-type:jwt-bearer"
+
+// JWTBearerResponse represents the response from a JWT Bearer grant request
+// per RFC 7523. This uses the standard OAuth 2.0 token response format.
+type JWTBearerResponse struct {
```
maciej-kisiel: Could https://pkg.go.dev/golang.org/x/oauth2#Token be used for this?

oauthex/jwt_bearer.go
```diff
@@ -0,0 +1,193 @@
...
+	ExpiresIn int `json:"expires_in,omitempty"`
+	// RefreshToken is the refresh token, which can be used to obtain new
+	// access tokens using the same authorization grant.
+	RefreshToken string `json:"refresh_token,omitempty"`
+	// Scope is the scope of the access token as described by RFC 6749 Section 3.3.
+	Scope string `json:"scope,omitempty"`
+}
+
+// JWTBearerError represents an error response from a JWT Bearer grant request.
+type JWTBearerError struct {
```
maciej-kisiel: Could https://pkg.go.dev/golang.org/x/oauth2#RetrieveError be used for this?

oauthex/jwt_bearer.go
```diff
@@ -0,0 +1,193 @@
...
+//	// Then exchange ID-JAG for access token
+//	token, err := ExchangeJWTBearer(
+//		ctx,
+//		"https://auth.mcpserver.example/oauth2/token",
+//		idJAG,
+//		"mcp-client-id",
+//		"mcp-client-secret",
+//		nil,
+//	)
+func ExchangeJWTBearer(
```
maciej-kisiel: It feels like https://pkg.go.dev/golang.org/x/oauth2#Config.Exchange could be used to replicate the behavior of the function. Is there any reason not to use it?

oauthex/oauth2.go
```diff
@@ -58,10 +58,9 @@ func getJSON[T any](ctx context.Context, c *http.Client, url string, limit int64
 		return nil, fmt.Errorf("bad status %s", res.Status)
 	}
 	// Specs require application/json.
-	ct := res.Header.Get("Content-Type")
-	mediaType, _, err := mime.ParseMediaType(ct)
-	if err != nil || mediaType != "application/json" {
-		return nil, fmt.Errorf("bad content type %q", ct)
+	ct := strings.TrimSpace(strings.SplitN(res.Header.Get("Content-Type"), ";", 2)[0])
```
maciej-kisiel: Could you comment why special behavior is needed in place of `mime.ParseMediaType`?

oauthex/token_exchange.go
```diff
@@ -0,0 +1,267 @@
...
+//		RequestedTokenType: TokenTypeIDJAG,
+//		Audience:          "https://auth.mcpserver.example/",
+//		Resource:          "https://mcp.mcpserver.example/",
+//		Scope:             []string{"read", "write"},
+//		SubjectToken:      idToken,
+//		SubjectTokenType:  TokenTypeIDToken,
+//	}
+//
+//	resp, err := ExchangeToken(ctx, idpTokenEndpoint, req, clientID, clientSecret, nil)
+func ExchangeToken(
```
maciej-kisiel, line 83: It looks like https://pkg.go.dev/golang.org/x/oauth2#Config.Exchange can be used for Token Exchange as well? https://github.com/golang/oauth2/issues/409

oauthex/token_exchange.go
```diff
@@ -0,0 +1,267 @@
...
+	SubjectToken string
+
+	// SubjectTokenType is the type of the security token in SubjectToken.
+	// For SEP-990 with OIDC, this MUST be TokenTypeIDToken.
+	SubjectTokenType string
+}
+
+// TokenExchangeResponse represents the response from a token exchange request
+// per RFC 8693 Section 2.2.
+type TokenExchangeResponse struct {
```
maciej-kisiel: Same comments as in oauthex/jwt_bearer.go apply to the request and error types.

oauthex/jwt_bearer.go
radar07: Yes, changed to oauth2.Token.

oauthex/jwt_bearer.go
radar07: Yes, changed to oauth2.RetrieveError.

oauthex/token_exchange.go
radar07: Yes, changed to oauth2.

oauthex/jwt_bearer.go
radar07: [oauth2#Config.Exchange](https://cs.opensource.google/go/x/oauth2/+/refs/tags/v0.35.0:oauth2.go;drc=f28b0b5467dda26d56f1240381158f7c334654d1;l=225) always expect a `code` which we don't have in our Token Ecxhange. IdPs can reject the request if we don't pass `code` or pass it as an empty string.

oauthex/token_exchange.go
radar07: You're right. But it is the same comment as above.

radar07: Also, the [Token](https://cs.opensource.google/go/x/oauth2/+/refs/tags/v0.35.0:token.go;drc=f28b0b5467dda26d56f1240381158f7c334654d1;l=31) struct don't have [`issued_token_type`](https://github.com/modelcontextprotocol/ext-auth/blob/main/specification/draft/enterprise-managed-authorization.mdx#42-token-exchange-response) which is required to indicate the type of the token.

maciej-kisiel:
Hi @radar07! I have recently merged #785 which I wrote about a few comments above. I think it's a good time to see if you could fit your desired auth flow into the `OAuthHandler` interface. I believe the Enterprise Managed authorization, as defined by the `ext-auth` repository should just expose a new implementation of this interface, just as `AuthorizationCodeHandler` implements the main MCP flow.

I would also ask you to implement this handler and put all the related helper files in a new sub-package of the `auth` package called `extauth` to make it clear that the functionality comes from an extension and is subject to different freshness guarantees.

If there is any logic from `authorization_code.go` that you would like to reuse, feel free to do so. We can move it to a more generic file in the future.

Thanks for working on this project!

pranav-okta:
@maciej-kisiel Thanks for the feedback. I made the changes as you suggested. Could you please take a look again?

radar07:
@maciej-kisiel Thanks for the feedback. I made the changes as you suggested. Could you please take a look again?

auth/extauth/enterprise_handler.go
```diff
@@ -0,0 +1,188 @@
...
+	"golang.org/x/oauth2"
+)
+
+// IDTokenFetcher is called to obtain an ID Token from the enterprise IdP.
+// This is typically done via OIDC login flow where the user authenticates
+// with their enterprise identity provider.
+type IDTokenFetcher func(ctx context.Context) (string, error)
+
+// EnterpriseHandlerConfig is the configuration for [EnterpriseHandler].
+type EnterpriseHandlerConfig struct {
```
maciej-kisiel, line 35: Could we document which fields are required and which are optional in their doc comments?

auth/extauth/enterprise_handler.go
```diff
@@ -0,0 +1,188 @@
...
+
+// Authorize performs the Enterprise Managed Authorization flow.
+// It is called when a request fails with 401 or 403.
+func (h *EnterpriseHandler) Authorize(ctx context.Context, req *http.Request, resp *http.Response) error {
+	defer func() {
+		if resp != nil && resp.Body != nil {
+			io.Copy(io.Discard, resp.Body)
+			resp.Body.Close()
+		}
+	}()
```
maciej-kisiel:
```suggestion
	defer resp.Body.Close()
	defer io.Copy(io.Discard, resp.Body)
```

I think we can assume that the response is correctly passed. Good idea with draining `Body`, could you also apply it in `authorization_code.go`?

auth/enterprise_auth.go
```diff
@@ -0,0 +1,196 @@
...
+//	}
+//
+//	// If you already obtained an ID token through your own means
+//	accessToken, err := EnterpriseAuthFlow(ctx, config, myIDToken)
+//	if err != nil {
+//		log.Fatal(err)
+//	}
+//
+//	// Use accessToken to call MCP Server APIs
+func EnterpriseAuthFlow(
```
maciej-kisiel: Do we still need this if we have the `EnterpriseHandler`?

auth/enterprise_auth.go
```diff
@@ -0,0 +1,196 @@
...
+	)
+	if err != nil {
+		return nil, fmt.Errorf("JWT bearer grant failed: %w", err)
+	}
+	return accessToken, nil
+}
+
+// GetAuthServerMetadatForIssuer fetches authorization server metadata for the given issuer URL.
+// It tries standard well-known endpoints (OAuth 2.0 and OIDC) and returns the first successful result.
+func GetAuthServerMetadatForIssuer(ctx context.Context, IssuerURL string, httpClient *httpClient) (*oauthex.AuthServerMeta, error) {
```
maciej-kisiel: It would be good to align it with `authorization_code.go`'s `getAuthServerMetadata`. It can be generalized to not be a method on AuthorizationCodeHandler, but receive the URL as parameter instead of PRM and the client as a parameter instead of using the receiver. It should probably be moved to a different file, maybe something like `shared.go`?

auth/extauth/oidc_login.go
maciej-kisiel, line 1: Please move this file inside `extauth/`, it is Enterprise flow specific.

auth/extauth/oidc_login.go
```diff
@@ -0,0 +1,454 @@
...
+	"time"
+
+	"github.com/modelcontextprotocol/go-sdk/oauthex"
+	"golang.org/x/oauth2"
+)
+
+// OIDCLoginConfig configures the OIDC Authorization Code flow for obtaining
+// an ID Token. This is an OPTIONAL step before calling EnterpriseAuthFlow.
+// Users can alternatively obtain ID tokens through their own methods.
+type OIDCLoginConfig struct {
```
maciej-kisiel, line 30: Please make sure all fields are documented as either required or optional.

auth/extauth/enterprise_handler.go
```diff
@@ -0,0 +1,188 @@
...
+
+	"github.com/modelcontextprotocol/go-sdk/auth"
+	"github.com/modelcontextprotocol/go-sdk/oauthex"
+	"golang.org/x/oauth2"
+)
+
+// IDTokenFetcher is called to obtain an ID Token from the enterprise IdP.
+// This is typically done via OIDC login flow where the user authenticates
+// with their enterprise identity provider.
+type IDTokenFetcher func(ctx context.Context) (string, error)
```
maciej-kisiel: Is the token always guaranteed to be a simple string, without any additional data?

auth/oidc_login.go
```diff
@@ -0,0 +1,454 @@
...
+		RefreshToken: oauth2Token.RefreshToken,
+		TokenType:    oauth2Token.TokenType,
+		ExpiresAt:    oauth2Token.Expiry.Unix(),
+	}, nil
+}
+
+// generateCodeVerifier generates a cryptographically random code verifier
+// for PKCE per RFC 7636 Section 4.1.
+func generateCodeVerifier() (string, error) {
+	// Per RFC 7636: code verifier is 43-128 characters from [A-Z] / [a-z] / [0-9] / "-" / "." / "_" / "~"
```
maciej-kisiel: Why not use https://pkg.go.dev/golang.org/x/oauth2#GenerateVerifier?

auth/oidc_login.go
```diff
@@ -0,0 +1,454 @@
...
+	randomBytes := make([]byte, 32)
+	if _, err := rand.Read(randomBytes); err != nil {
+		return "", fmt.Errorf("failed to generate random bytes: %w", err)
+	}
+	return base64.RawURLEncoding.EncodeToString(randomBytes), nil
+}
+
+// generateCodeChallenge generates the PKCE code challenge from the verifier
+// using SHA256 per RFC 7636 Section 4.2.
+func generateCodeChallenge(verifier string) string {
```
maciej-kisiel: https://pkg.go.dev/golang.org/x/oauth2#S256ChallengeFromVerifier

auth/oidc_login.go
```diff
@@ -0,0 +1,454 @@
...
+	hash := sha256.Sum256([]byte(verifier))
+	return base64.RawURLEncoding.EncodeToString(hash[:])
+}
+
+// generateState generates a cryptographically random state parameter
+// for CSRF protection per RFC 6749 Section 10.12.
+func generateState() (string, error) {
```
maciej-kisiel: Could https://pkg.go.dev/crypto/rand#Text be used here?

auth/extauth/oidc_login.go
```diff
@@ -0,0 +1,454 @@
...
+	AccessToken string
+	// RefreshToken is the OAuth2 refresh token (if issued by IdP).
+	RefreshToken string
+	// TokenType is the token type (typically "Bearer").
+	TokenType string
+	// ExpiresAt is when the ID token expires.
+	ExpiresAt int64
+}
+
+// InitiateOIDCLogin initiates an OIDC Authorization Code flow with PKCE.
```
maciej-kisiel: Do you see a situation occurring of `InitiateOICDLogin` not being used with `CompleteOIDCLogin`? If not, maybe we could be more opinionated and implement the whole flow as a single function that just takes the "direct user to authorization URL and return the authorization result" as a functional argument, similar to how `AuthorizationCodeHandlerConfig` takes `AuthorizationCodeFetcher`?

In general this flow looks very similar to the normal authorization code flow, so maybe we could even reuse some of the data structures? Do you see a risk of the OAuth flow and the OIDC flow diverging in the future?

auth/oidc_login.go
```diff
@@ -0,0 +1,454 @@
...
+	randomBytes := make([]byte, 32)
+	if _, err := rand.Read(randomBytes); err != nil {
+		return "", fmt.Errorf("failed to generate random bytes: %w", err)
+	}
+	return base64.RawURLEncoding.EncodeToString(randomBytes), nil
+}
+
+// buildAuthorizationURL constructs the OIDC authorization URL.
+func buildAuthorizationURL(
```
maciej-kisiel: Have you considered reusing https://godoc.corp.google.com/pkg/google3/third_party/golang/oauth2/oauth2#Config.AuthCodeURL? I believe it supports all that's needed here.

auth/oidc_login.go
```diff
@@ -0,0 +1,454 @@
...
+	// Add login_hint if provided (optional per OIDC spec, but some IdPs may require it)
+	if loginHint != "" {
+		q.Set("login_hint", loginHint)
+	}
+	u.RawQuery = q.Encode()
+	return u.String(), nil
+}
+
+// exchangeAuthorizationCode exchanges the authorization code for tokens.
+func exchangeAuthorizationCode(
```
maciej-kisiel: I think https://godoc.corp.google.com/pkg/google3/third_party/golang/oauth2/oauth2#Config.Exchange could be used here and if you would follow the suggestion to make the whole OIDC a single function, you could even reuse the `oauth2.Config`.

oauthex/id_jag.go
```diff
@@ -0,0 +1,138 @@
...
+//
+// Example:
+//
+//	claims, err := ParseIDJAG(idJAG)
+//	if err != nil {
+//		log.Fatalf("Failed to parse ID-JAG: %v", err)
+//	}
+//	fmt.Printf("Subject: %s\n", claims.Subject)
+//	fmt.Printf("Expires: %v\n", claims.Expiry())
+func ParseIDJAG(jwt string) (*IDJAGClaims, error) {
```
maciej-kisiel: Maybe we can use the `golang-jwt/jwt` library as an implementation detail for parsing to not replicate the logic?

oauthex/jwt_bearer.go
maciej-kisiel: > IdPs can reject the request if we don't pass `code` or pass it as an empty string.

I don't think that should happen. Based on RFC 6749, Section 3.2:

> Parameters sent without a value MUST be treated as if they were
> omitted from the request.  The authorization server MUST ignore
> unrecognized request parameters.

oauthex/token_exchange.go
maciej-kisiel: > Also, the `Token` struct don't have `issued_token_type` which is required to indicate the type of the token.

I believe it could be accessible via `Token.Extra`, no?

oauthex/id_jag.go
maciej-kisiel: Also, is this function used anywhere?

oauthex/oauth2.go
radar07: I was facing an error in the response mime type and later I realized it was on my end. Removed.

auth/extauth/enterprise_handler.go
radar07: Done.

oauthex/jwt_bearer.go
radar07: I did a sample test and you're right. Changed to use `Config.Exchange`

auth/oidc_login.go
radar07: Yes.

auth/oidc_login.go
radar07: Done.

oauthex/id_jag.go
radar07: I added this in case if any one wants to verify the claims of the JWT. But, I believe it is not needed for this. Please let me know what you think.

auth/extauth/oidc_login.go
radar07: Done.

auth/extauth/oidc_login.go
radar07: Yes, done.

auth/oidc_login.go
radar07: Used `oauth2.GenerateVerifier`.

auth/oidc_login.go
radar07: I see what you mean. Good call.

auth/oidc_login.go
radar07: Yes, right! Used `oauth2.Config`

oauthex/token_exchange.go
radar07: Yes. Extracted `issued_token_type` from Token.Extra [here](https://github.com/radar07/go-sdk/blob/a4fd173143ce2a7fd79150b30212aa7df0129f92/oauthex/token_exchange.go#L206).

auth/extauth/enterprise_handler.go
radar07: Yes.

auth/enterprise_auth.go
radar07: Moved the code to **shared.go** (which can be renamed if needed). Also, the `getAuthServerMetadata` method has a different fallback behavior. I want to know your thoughts.

auth/extauth/enterprise_handler.go
radar07: From the [spec](https://github.com/modelcontextprotocol/ext-auth/blob/f29c419151ddf9266432b1b3488265a748794c69/specification/draft/enterprise-managed-authorization.mdx?plain=1#L132), it is understandable that it is a JWT.

auth/enterprise_auth.go
radar07: This uses `OAuthHandler` and allows users to use the Enterprise Auth Flow without the MCP imports. However, if we prefer keeping only the handler, we can remove this.

radar07:
@maciej-kisiel Thank you so much for taking the time to review this PR and your valuable feedback. I have made the changes as you suggested and need your suggestions on the changes.

auth/extauth/oidc_login.go
maciej-kisiel: I see that you have introduced `PerformOIDCLogin` that does it. I would go one step further and not export `InitiateOIDCLogin` and `CompleteOIDCLogin`. It would have the following benefits:

1. You would need to verify the `OIDCLoginConfig` only once.
2. You could reuse the `oauth2.Config` between the functions.
3. You would need only one call for Authorization Metadata of the IdP.
4. You would ensure `state` equality is verified in the flow.
5. You could reuse `auth.AuthorizationResult` and `auth.AuthorizationArgs` instead of introducing new structs.
6. Consistency with `auth/authorization_code.go`.

To achieve those in a readable way, you will likely need to slice the logic into functions slightly differently than currently. I would encourage you to follow `auth/authorization_code.go` example for added consistency.

auth/extauth/enterprise_handler.go
maciej-kisiel: I'm just curious about future extensibility, I'm not a domain expert. As an example, in https://github.com/modelcontextprotocol/go-sdk/blob/a433a831d6e5d5ac3b9e625a8095aa8eaa040dfc/auth/authorization_code.go#L73 we have introduced a struct to contain the `URL`, even though it's the only required data currently. However, if we have set it to string, we would be in a difficult place if there appeared a new use case to carry additional data, since a string is not extensible. This is something to consider here as well.

auth/enterprise_auth.go
maciej-kisiel: I would start with just the handler. It's always possible to add more helpers if there is a need, the other way round (removing from the API) is not possible.

auth/shared.go
```diff
@@ -0,0 +1,69 @@
...
+	"net/http"
+	"net/url"
+	"strings"
+
+	"github.com/modelcontextprotocol/go-sdk/oauthex"
+)
+
+// GetAuthServerMetadataForIssuer fetches authorization server metadata for the given issuer URL.
+// It tries standard well-known endpoints (OAuth 2.0 and OIDC) and returns the first successful result.
+func GetAuthServerMetadataForIssuer(ctx context.Context, issuerURL string, httpClient *http.Client) (*oauthex.AuthServerMeta, error) {
```
maciej-kisiel: Nit: I would keep it at `GetAuthServerMetadata`, the "for issuer" part should be pretty evident from the parameter list and the doc comment.

oauthex/jwt_bearer.go
maciej-kisiel: Great! At this point, I think I would remove this file (and the relevant test file) and just implement it as a private helper function in `auth/extauth/enterprise_handler.go`. It should allow you to skip some validations (relying on validation already present there).

oauthex/oauth2.go
```diff
@@ -94,3 +94,9 @@ func checkHTTPSOrLoopback(addr string) error {
 	}
 	return nil
 }
+
+// CheckURLScheme validates a URL scheme for security.
+// This is exported for use by the auth package.
+func CheckURLScheme(u string) error {
+	return checkURLScheme(u)
```
maciej-kisiel: If it's needed from another package, let's just export it instead of adding a wrapper.

oauthex/token_exchange.go
```diff
@@ -0,0 +1,233 @@
...
+	// Per RFC 6749 Section 3.2, parameters sent without a value (like the empty
+	// "code" parameter) MUST be treated as if they were omitted from the request.
+	// The oauth2 library's Exchange method sends an empty code, but compliant
+	// servers should ignore it.
+	cfg := &oauth2.Config{
+		ClientID:     clientID,
+		ClientSecret: clientSecret,
+		Endpoint: oauth2.Endpoint{
+			TokenURL:  tokenEndpoint,
+			AuthStyle: oauth2.AuthStyleInParams, // Use POST body auth per SEP-990
```
maciej-kisiel: Could you point me where this is defined? I failed to find it.

oauthex/token_exchange.go
```diff
@@ -0,0 +1,233 @@
...
+//		SubjectTokenType:  TokenTypeIDToken,
+//	}
+//
+//	resp, err := ExchangeToken(ctx, idpTokenEndpoint, req, clientID, clientSecret, nil)
+func ExchangeToken(
+	ctx context.Context,
+	tokenEndpoint string,
+	req *TokenExchangeRequest,
+	clientID string,
+	clientSecret string,
```
maciej-kisiel: Are there any other client authentication methods that could be used for token exchange? I think we should be more flexible here and define a struct that could be extended with JWT-based methods in the future.

auth/enterprise_auth.go
maciej-kisiel: If the fallback handling at the end is problematic, you can move it directly to the `Authorize` function, just after the `getAuthServerMetadata` call and modify the latter to use the new function to avoid code repetition.

maciej-kisiel:
Hi @radar07, thanks for applying the suggestions! I think we're pretty close now with the main logic. I have added a few more comments.

One bigger question that I would have is: would it make sense for the token returning methods (OIDC login, Token Exchange) to use `oauth2.Token` instead of custom types? Just like `ExchangeJWTBearer` currently does. I understand that they have some meaningful fields that need to be passed via `oauth2.Token.Extra`, but maybe it's fine? I would be curious to hear your thoughts.

One more idea: instead of adding extensive examples to doc comments of particular functions, what do you think about creating a single, end-to-end example in the `examples/` directory?

radar07:
> Hi @radar07, thanks for applying the suggestions! I think we're pretty close now with the main logic. I have added a few more comments.
> 
> One bigger question that I would have is: would it make sense for the token returning methods (OIDC login, Token Exchange) to use `oauth2.Token` instead of custom types? Just like `ExchangeJWTBearer` currently does. I understand that they have some meaningful fields that need to be passed via `oauth2.Token.Extra`, but maybe it's fine? I would be curious to hear your thoughts.
> 
> One more idea: instead of adding extensive examples to doc comments of particular functions, what do you think about creating a single, end-to-end example in the `examples/` directory?

@maciej-kisiel Thanks again for your time in reviewing this PR!

Moving the examples from doc comments to their own directory makes more sense. I'll make the changes.

For using `oauth2.Token`, I agree with your opinion on making it consistent and use `oauth2` for all tokens in the methods. However, `issued_token_type` is a required field and must be validated as per the spec. So, making this a custom explicit field makes this validation  obvious to the callers. The tradeoff will be these fields will be hidden in the `Token.Extra` and users have to go through the code or comments which is not a huge deal (I think). I'd love to hear your opinion.

auth/extauth/enterprise_handler.go
```diff
@@ -0,0 +1,243 @@
...
+	// REQUIRED.
+	IdPIssuerURL string
+
+	// IdPClientID is the MCP Client's ID registered at the IdP.
+	// OPTIONAL. Required if the IdP requires client authentication for token exchange.
+	IdPClientID string
+
+	// IdPClientSecret is the MCP Client's secret registered at the IdP.
+	// OPTIONAL. Required if the IdP requires client authentication for token exchange.
+	IdPClientSecret string
```
maciej-kisiel: Should these be represented as newly introduced `oauthex.ClientCredentials`?

auth/extauth/enterprise_handler.go
```diff
@@ -0,0 +1,243 @@
...
+	// REQUIRED.
+	MCPResourceURI string
+
+	// MCPClientID is the MCP Client's ID registered at the MCP Server.
+	// OPTIONAL. Required if the MCP Server requires client authentication.
+	MCPClientID string
+
+	// MCPClientSecret is the MCP Client's secret registered at the MCP Server.
+	// OPTIONAL. Required if the MCP Server requires client authentication.
+	MCPClientSecret string
```
maciej-kisiel: Should these be represented as newly introduced `oauthex.ClientCredentials`?

radar07:
@maciej-kisiel Thanks for being patient and taking the time to discuss and come up with the plan. I really appreciate it!

> For using oauth2.Token, I agree with your opinion on making it consistent and use oauth2 for all tokens in the methods. However, issued_token_type is a required field and must be validated as per the spec. So, making this a custom explicit field makes this validation obvious to the callers. The tradeoff will be these fields will be hidden in the Token.Extra and users have to go through the code or comments which is not a huge deal (I think). I'd love to hear your opinion.

I made the changes as suggested and with the separate examples directory for this flow. Please go through and let me what you think.

oauthex/token_exchange.go
radar07: I changed it to ClientCredentials struct (if clientSecret is what you are asking :thinking: )

oauthex/token_exchange.go
radar07: Used a struct for client credentials to support `oauth2.Config.Exchange()`. Should we keep this as it is or implement support for JWT methods?

auth/extauth/oidc_login.go
```diff
@@ -0,0 +1,253 @@
...
+type OIDCLoginConfig struct {
+	// IssuerURL is the IdP's issuer URL (e.g., "https://acme.okta.com").
+	// REQUIRED.
+	IssuerURL string
+	// ClientID is the MCP Client's ID registered at the IdP.
+	// REQUIRED.
+	ClientID string
+	// ClientSecret is the MCP Client's secret at the IdP.
+	// OPTIONAL. Only required if the client is confidential.
+	ClientSecret string
```
maciej-kisiel: Should these be represented as newly introduced `oauthex.ClientCredentials`?

auth/extauth/oidc_login.go
```diff
@@ -0,0 +1,253 @@
...
+//
+// Implementations should:
+//  1. Direct the user to args.URL (e.g., open in browser)
+//  2. Wait for the IdP redirect to the configured RedirectURL
+//  3. Extract the code and state from the redirect query parameters
+//  4. Return them in auth.AuthorizationResult
+//
+// The implementation MUST verify that the returned state matches the state
+// sent in the authorization request (available via parsing args.URL).
+type AuthorizationCodeFetcher func(ctx context.Context, args auth.AuthorizationArgs) (*auth.AuthorizationResult, error)
```
maciej-kisiel: Maybe we can extract this to `auth/authorization_code.go` and use it also in https://github.com/modelcontextprotocol/go-sdk/blob/ea161959d7c87bd3d4bbd63cfa8df01d3ab3b304/auth/authorization_code.go#L103 since they are identical?

auth/extauth/oidc_login.go
```diff
@@ -0,0 +1,253 @@
...
+// to the authorization URL and returning the authorization result.
+//
+// Implementations should:
+//  1. Direct the user to args.URL (e.g., open in browser)
+//  2. Wait for the IdP redirect to the configured RedirectURL
+//  3. Extract the code and state from the redirect query parameters
+//  4. Return them in auth.AuthorizationResult
+//
+// The implementation MUST verify that the returned state matches the state
+// sent in the authorization request (available via parsing args.URL).
```
maciej-kisiel: Nit: I think this comment is obsolete, given we're doing this verification.

auth/extauth/oidc_login.go
```diff
@@ -0,0 +1,253 @@
...
+type AuthorizationCodeFetcher func(ctx context.Context, args auth.AuthorizationArgs) (*auth.AuthorizationResult, error)
+
+// PerformOIDCLogin performs the complete OIDC Authorization Code flow with PKCE
+// in a single function call. This is the recommended approach for obtaining an
+// ID Token for use with [EnterpriseHandler].
+//
+// The authCodeFetcher callback handles the user interaction:
+//   - Directing the user to the IdP login page
+//   - Waiting for the redirect with the authorization code
+//   - Validating CSRF state and returning the result
```
maciej-kisiel: And similarly this comment is also obsolete.

auth/authorization_code.go
```diff
@@ -395,40 +397,6 @@ func (h *AuthorizationCodeHandler) getAuthServerMetadata(ctx context.Context, pr
 	return asm, nil
```
maciej-kisiel: As discussed in the other comment thread, please move this to `Authorize` and call `GetAuthServerMetadata` from `shared.go` instead. This function can then be removed.

examples/auth/enterprise-client/main.go
```diff
@@ -0,0 +1,197 @@
...
+	log.Printf("Connecting to MCP server at %s...", *mcpServerURL)
+	session, err := client.Connect(ctx, transport, nil)
+	if err != nil {
+		log.Fatalf("failed to connect to MCP server: %v", err)
+	}
+	defer session.Close()
+
+	log.Println("Successfully connected to MCP server!")
+
+	// List available tools as a demonstration
```
maciej-kisiel:
```suggestion
	// List available tools as a demonstration.
```

examples/auth/enterprise/main.go
maciej-kisiel, line 1: Nice! The UX is very similar to the already existing flow, which is a big benefit.

One small suggestion I would have is to rename the folder to `enterprise`, given that the other example has standalone `client` and `server`. This has both in a single example, which makes the `-client` in the folder name a bit misleading.

oauthex/token_exchange.go
maciej-kisiel: We don't need to implement the support for JWT yet, as no-one has requested it. But I would like to have a clear idea how it would be implemented if needed. For example, I had similar concerns in https://github.com/modelcontextprotocol/go-sdk/blob/ea161959d7c87bd3d4bbd63cfa8df01d3ab3b304/auth/authorization_code.go#L48

And I ended up creating a struct hierarchy that is not very rich now, but can be extended. Maybe we should follow a similar example here? I assume the secret may not always be needed, e.g. in case of JWT I assume it's the key that matters instead?

oauthex/token_exchange.go
maciej-kisiel: Sorry for not being precise. I was trying to find where in SEP-990 is it defined to use `client_secret_post` and failed. Your comment claims it should be there?

auth/extauth/oidc_login.go
radar07: Done.

auth/extauth/oidc_login.go
radar07: Yes, removed.

auth/extauth/enterprise_handler.go
radar07: Yes!

auth/extauth/enterprise_handler.go
radar07: Yes!

auth/extauth/oidc_login.go
radar07: Yes!

examples/auth/enterprise-client/main.go
radar07: Done.

examples/auth/enterprise/main.go
radar07: Yup. Renamed to `enterprise`.

radar07:
> Sorry for not being precise. I was trying to find where in SEP-990 is it defined to use client_secret_post and failed. Your comment claims it should be there?

You're right! I couldn't find the `client_secret_post` in SEP-990. I've removed the misleading comment.

> We don't need to implement the support for JWT yet, as no-one has requested it. But I would like to have a clear idea how it would be implemented if needed. For example, I had similar concerns in And I ended up creating a struct hierarchy that is not very rich now, but can be extended. Maybe we should follow a similar example here? I assume the secret may not always be needed, e.g. in case of JWT I assume it's the key that matters instead?

Yes, the key matters in JWT. I agree with your point. We could extend the `ClientCredentials` with additional fields or create a new interface with methods for JWT client authentication (similar to `PreregisteredClientConfig`)

auth/authorization_code.go
radar07: Yes, I did that. Please have a look if this aligns with you.

auth/authorization_code.go
```diff
@@ -67,12 +68,21 @@ type AuthorizationResult struct {
...
 	// Authorization URL to be opened in a browser for the user to start the authorization process.
 	URL string
 }
 
+// AuthorizationCodeFetcher is called to initiate the OAuth authorization flow.
+// It is responsible for directing the user to the authorization URL (e.g., opening
+// in a browser) and returning the authorization code and state once the Authorization
+// Server redirects back to the configured RedirectURL.
+//
+// The implementation MUST verify that the returned state matches the state
```
maciej-kisiel: The SDK is doing `state` verification in all current flows. Maybe requiring this from the fetcher is not needed? I would remove these two lines of the comment.

auth/authorization_code.go
```diff
@@ -218,9 +227,17 @@ func (h *AuthorizationCodeHandler) Authorize(ctx context.Context, req *http.Requ
 		return err
 	}
 
-	asm, err := h.getAuthServerMetadata(ctx, prm)
+	asm, err := GetAuthServerMetadata(ctx, prm.AuthorizationServers[0], h.config.Client)
```
maciej-kisiel, line 201: To preserve current behavior I think the helper needs to return `nil, nil` when none of the endpoints responded with a non-4xx error (return statement at the end of the function). In the current version, the fallback will kick in on every error from `GetAuthServerMetadata` and that's not something we want, e.g. if the fetched asm has insecure links we want to propagate the error.

```go
asm, err := h.getAuthServerMetadata(ctx, prm)
if err != nil {
		return err
}
if asm == nil {
		// fallback...
}
```

Note that you will need to add `nil` checks at other call sites as well.

auth/authorization_code_test.go
```diff
@@ -439,15 +439,11 @@ func TestGetProtectedResourceMetadata_Error(t *testing.T) {
 }
 
 func TestGetAuthServerMetadata(t *testing.T) {
```
maciej-kisiel: Can we move this test case to `shared_test.go`? The test should probably not reference "fallback", since it's no longer part of the logic of `GetAuthServerMetadata`.

auth/authorization_code_test.go
```diff
@@ -622,11 +623,15 @@ func TestHandleRegistration(t *testing.T) {
 			if err != nil {
 				t.Fatalf("NewAuthorizationCodeHandler() error = %v, want nil", err)
 			}
-			asm, err := handler.getAuthServerMetadata(t.Context(), &oauthex.ProtectedResourceMetadata{
-				AuthorizationServers: []string{s.URL()},
-			})
+			asm, err := GetAuthServerMetadata(t.Context(), s.URL(), http.DefaultClient)
 			if err != nil {
```
maciej-kisiel, line 544: Does this test fail without the fallback? If not, I would remove it from here, since it shouldn't be relevant to the logic under test.

auth/authorization_code_test.go
```diff
@@ -672,11 +677,15 @@ func TestDynamicRegistration(t *testing.T) {
 	if err != nil {
 		t.Fatalf("NewAuthorizationCodeHandler() error = %v", err)
 	}
-	asm, err := handler.getAuthServerMetadata(t.Context(), &oauthex.ProtectedResourceMetadata{
-		AuthorizationServers: []string{s.URL()},
-	})
+	asm, err := GetAuthServerMetadata(t.Context(), s.URL(), http.DefaultClient)
 	if err != nil {
-		t.Fatalf("getAuthServerMetadata() error = %v, want nil", err)
+		// Fallback to predefined endpoints for testing
```
maciej-kisiel: Does this test fail without the fallback? If not, I would remove it from here, since it shouldn't be relevant to the logic under test.

auth/client.go
```diff
@@ -40,3 +40,10 @@ type OAuthHandler interface {
 	// The function is responsible for closing the response body.
 	Authorize(context.Context, *http.Request, *http.Response) error
 }
+
+// OAuthHandlerBase is an embeddable type that satisfies the private method
+// requirement of [OAuthHandler]. Extension packages should embed this type
+// in their handler structs to implement OAuthHandler.
+type OAuthHandlerBase struct{}
```
maciej-kisiel: I think it may be time to remove `isOAuthHandler` from the `OAuthHandler` interface. We added it temporarily to not encourage the implementations of the interface before it's stable, but in the next release we want to stabilize it. It should allow you to remove this struct and pass lint checks.

docs/protocol.md
```diff
@@ -357,6 +357,70 @@ session, err := client.Connect(ctx, transport, nil)
...
+    Endpoint:     "https://example.com/mcp",
+    OAuthHandler: enterpriseHandler,
+}
+client := mcp.NewClient(&mcp.Implementation{Name: "client", Version: "v0.0.1"}, nil)
+session, err := client.Connect(ctx, transport, nil)
+```
+
+The `EnterpriseHandler` automatically manages the token exchange flow and token refreshing.
+
+For a complete working example, see [examples/auth/enterprise-client](https://github.com/modelcontextprotocol/go-sdk/tree/main/examples/auth/enterprise-client).
```
maciej-kisiel: Update path to `enterprise`.

auth/extauth/enterprise_handler.go
```diff
@@ -0,0 +1,230 @@
...
+	MCPAuthServerURL string
+
+	// MCPResourceURI is the MCP Server's resource identifier (RFC 9728).
+	// Used as the resource parameter in token exchange.
+	// REQUIRED.
+	MCPResourceURI string
+
+	// MCPCredentials contains the MCP Client's credentials registered at the MCP Server.
+	// OPTIONAL. Set to nil if the MCP Server doesn't require client authentication.
+	MCPCredentials *oauthex.ClientCredentials
```
maciej-kisiel, line 65: The credentials are marked as optional, but you dereference them without checking for `nil` in line 206-207.

auth/extauth/enterprise_handler.go
```diff
@@ -0,0 +1,230 @@
...
+
+	// IdPIssuerURL is the enterprise IdP's issuer URL (e.g., "https://acme.okta.com").
+	// Used for OIDC discovery to find the token endpoint.
+	// REQUIRED.
+	IdPIssuerURL string
+
+	// IdPCredentials contains the MCP Client's credentials registered at the IdP.
+	// OPTIONAL. Set to nil for public clients. Required if the IdP requires client
+	// authentication for token exchange.
+	IdPCredentials *oauthex.ClientCredentials
```
maciej-kisiel, line 47: The credentials are marked as optional, but the `ExchangeToken` function validates if they are non-`nil` and throws an error if they are.

auth/extauth/enterprise_handler.go
```diff
@@ -0,0 +1,230 @@
...
+		tokenExchangeResp.AccessToken,
+		h.config.MCPCredentials,
+		httpClient,
+	)
+	if err != nil {
+		return fmt.Errorf("JWT bearer grant failed: %w", err)
+	}
+
+	// Store the token source for subsequent requests
+	h.tokenSource = oauth2.StaticTokenSource(accessToken)
```
maciej-kisiel, line 215: Is it intentional to not support token refreshes in the enterprise flow? The docs in `protocol.md` suggest that it should support the refreshes.

auth/extauth/enterprise_handler.go
```diff
@@ -0,0 +1,230 @@
...
+	// REQUIRED.
+	MCPResourceURI string
+
+	// MCPCredentials contains the MCP Client's credentials registered at the MCP Server.
+	// OPTIONAL. Set to nil if the MCP Server doesn't require client authentication.
+	MCPCredentials *oauthex.ClientCredentials
+
+	// MCPScopes is the list of scopes to request at the MCP Server.
+	// OPTIONAL.
+	MCPScopes []string
```
maciej-kisiel, line 69: Does the enterprise flow support dynamic scopes differing per-tool (e.g. communicated via the `WWW-Authenticate` header or scope step-up flows?

auth/extauth/enterprise_handler.go
maciej-kisiel, line 1: Please create unit tests for this file. We should have at least an e2e test for `Authorize` and validation of `NewEnterpriseHandler`.

For the first one you can check https://github.com/modelcontextprotocol/go-sdk/blob/755b9ed4dfe2da1ce054ddd6b2d236cdff797630/auth/authorization_code_test.go#L24 for an example. I hope we could reuse the fake authorization server with some modifications to support token exchange.

oauthex/token_exchange.go
```diff
@@ -0,0 +1,229 @@
...
+	// TokenTypeIDJAG is the URN for Identity Assertion JWT Authorization Grants.
+	// This is the token type returned by IdP during token exchange for SEP-990.
+	TokenTypeIDJAG = "urn:ietf:params:oauth:token-type:id-jag"
+
+	// GrantTypeTokenExchange is the grant type for RFC 8693 token exchange.
+	GrantTypeTokenExchange = "urn:ietf:params:oauth:grant-type:token-exchange"
+)
+
+// ClientCredentials holds client authentication credentials for OAuth token requests.
+type ClientCredentials struct {
```
maciej-kisiel: I've done some thinking and I believe we should follow the `PreregisteredClientConfig` example here. Consider:

```go
type ClientCredentials struct {
    ClientID string
    ClientSecretAuth *ClientSecretAuth
}
type ClientSecretAuth struct {
    ClientSecret string
}
```
In the future we can add more `*Auth` fields to the struct, e.g. `PrivateKeyJWTAuth`. Let's also move it to a non-token exchange specific file, e.g. `client.go`. I would also propose to introduce a `Validate` method for the struct to check if only one `*Auth` field is set. `client_test.go` should then contain a test that validates the `Validate` method was updated when new field was added.

Once we have it, I would replace `PreregisteredClientConfig` with `oauthex.ClientCredentials` in `AuthorizationCodeConfig`.

auth/extauth/oidc_login.go
```diff
@@ -0,0 +1,234 @@
...
+	// OPTIONAL.
+	LoginHint string
+	// HTTPClient is the HTTP client for making requests.
+	// If nil, http.DefaultClient is used.
+	// OPTIONAL.
+	HTTPClient *http.Client
+}
+
+// OIDCTokenResponse contains the tokens returned from a successful OIDC login.
+type OIDCTokenResponse struct {
```
maciej-kisiel: Regarding the return type discussion we've had: after some thinking I believe we should use `oauth2.Token` to represent the output from OIDC login and token exchange.

If you don't like accessing the important fields via `Extra()`, you can consider embedding the `oauth2.Token` into the existing response types instead of repeating all the fields. The extra fields could be provided with an accessor method, e.g. `oidcResp.IDToken()` accessing `Extra` underneath, but personally I don't think we need it if we document the extra fields explicitly.

auth/extauth/oidc_login_test.go
```diff
@@ -0,0 +1,496 @@
...
+			return
+		}
+		http.NotFound(w, r)
+	}))
+	serverURL = server.URL
+	return server
+}
+
+// createMockOIDCServerWithToken creates a mock OIDC server that also handles token exchange.
+func createMockOIDCServerWithToken(t *testing.T) *httptest.Server {
```
maciej-kisiel, line 348: Could this server be used anywhere `createMockOIDCServer` is used? It seems to be a strict superset, so maybe we can remove the simpler one and the more advanced one everywhere. In that case, I would propose for the advanced server to be renamed to `createMockOIDCServer` since it would be the only one.

auth/extauth/oidc_login_test.go
```diff
@@ -0,0 +1,496 @@
...
+}
+
+// base64EncodeClaims encodes JWT claims for testing.
+func base64EncodeClaims(claims map[string]interface{}) string {
+	claimsJSON, _ := json.Marshal(claims)
+	return base64.RawURLEncoding.EncodeToString(claimsJSON)
+}
+
+// TestPerformOIDCLogin tests the combined OIDC login flow with callback.
+func TestPerformOIDCLogin(t *testing.T) {
```
maciej-kisiel, line 411: Generally it's recommended to test the public API. Do you think it would be possible to move the tests from `TestInitiateOIDCLogin` here?

`TestCompleteOIDCLogin` probably is not needed, since we own the logic that should ensure these values tested there are passed correctly.

auth/extauth/oidc_login_test.go
```diff
@@ -0,0 +1,496 @@
...
+				if !strings.Contains(err.Error(), tt.expectErr) {
+					t.Errorf("expected error containing '%s', got: %v", tt.expectErr, err)
+				}
+			})
+		}
+	})
+}
+
+// TestOIDCLoginE2E tests the complete OIDC login flow end-to-end.
+func TestOIDCLoginE2E(t *testing.T) {
```
maciej-kisiel, line 275: If this is not different from the successful flow tested in `TestPerformOIDCLogin`, then I would remove it.

auth/extauth/oidc_login.go
```diff
@@ -0,0 +1,237 @@
...
+	}
+	if config.RedirectURL == "" {
+		return nil, nil, fmt.Errorf("RedirectURL is required")
+	}
+	if len(config.Scopes) == 0 {
+		return nil, nil, fmt.Errorf("Scopes is required (must include 'openid')")
+	}
+
+	hasOpenID := false
+	for _, scope := range config.Scopes {
```
jba: https://pkg.go.dev/slices#Contains

auth/authorization_code.go
radar07: Yes, done!

docs/protocol.md
radar07: Done.

auth/extauth/oidc_login.go
radar07: I understand. I changed these to return `oauth2.Token`

auth/extauth/enterprise_handler.go
radar07: Oops, my bad. This is required.

auth/extauth/enterprise_handler.go
radar07: This should be required.

auth/extauth/oidc_login.go
radar07: Done.

auth/extauth/oidc_login_test.go
radar07: Yes, removed.

auth/authorization_code.go
radar07: This is removed.

auth/shared.go
```diff
@@ -0,0 +1,79 @@
...
+// Returns (nil, nil) when no metadata endpoints respond (404s), allowing callers to implement
+// fallback logic. Returns an error only for actual failures (network errors, invalid JSON, etc.).
+func GetAuthServerMetadata(ctx context.Context, issuerURL string, httpClient *http.Client) (*oauthex.AuthServerMeta, error) {
+	var lastErr error
+	for _, metadataURL := range authorizationServerMetadataURLs(issuerURL) {
+		asm, err := oauthex.GetAuthServerMeta(ctx, metadataURL, issuerURL, httpClient)
+		if err != nil {
+			// Store the error but continue trying other endpoints
+			lastErr = err
+			continue
```
maciej-kisiel: Just return the error, as was before. `GetAuthServerMeta` only returns an error if it's not a client error. We don't want to proceed (and possibly fallback) if the server has some temporary issues. It will simplify the code.

auth/extauth/enterprise_handler.go
maciej-kisiel: Ok, I see that automatic refresh is not supported, based on the changes in `protocol.md`.

auth/extauth/enterprise_handler.go
maciej-kisiel: Bumping this question.

radar07: Currently, this flow does not support dynamic scopes from the `WWW-Authenticate` header or scope step-up but are configured statically via the `MCPScopes`. The rationale is that in enterprise environments, scopes are typically known ahead of time.

I looked more closely at the `AuthorizeCodeHandler` and it does support dynamic scopes from the header. If you'd like, I can update to parse and use the header or fallback to `MCPScopes` when not.

radar07:
Thanks @maciej-kisiel for taking the time to review! I have been away from my computer for a few days and couldn't answer your questions on time.

I looked at your final comments and replied to them. Please let me know how we can proceed from here. Thanks again.

auth/extauth/enterprise_handler.go
maciej-kisiel: I don't want to add features that nobody uses and I'm not really familiar how the environments where this authorization would be used look like. If you say that scopes are usually static, then I'm willing to trust you. The logic can be extended in the future.

oauthex/oauth2.go
```diff
@@ -63,10 +63,10 @@ func getJSON[T any](ctx context.Context, c *http.Client, url string, limit int64
 	return &t, nil
 }
 
-// checkURLScheme ensures that its argument is a valid URL with a scheme
+// CheckURLScheme ensures that its argument is a valid URL with a scheme
```
maciej-kisiel: One last request: we came to a conclusion that it's not the best approach to export `CheckURLScheme` as it's a low level helper. Could you please unexport it back and just copy the implementation where you need it? Youd can add a sentence to a doc comment to update the other version as well if needed. Sorry for confusion.

radar07: Yes, I made the changes and added a line to doc string comment which I believe is self-explanatory.

PR #773, opened 2026-02-03
chore: use rand.Text and slog.DiscardHandler over intrernal implementation

### LINKED ISSUE #772: Update code for minimum go version 1.24
HI, noticed that we have upgraded go version to 1.24 via https://github.com/modelcontextprotocol/go-sdk/pull/765. I think its good time to address some hacks/tradeoffs done in code earlier.

IAmSurajBobade:
Address TODOs in code requiring go 1.24 as minimum version

mcp/logging.go: use slog.DiscardHandler, over local implementation
mcp/util.go: use rand.Text, over local implementation

Fixes #772

### DIFF
--- a/mcp/server.go
+++ b/mcp/server.go
@@ -176,9 +176,7 @@ func NewServer(impl *Implementation, options *ServerOptions) *Server {
 	}
 
 	if opts.GetSessionID == nil {
-		opts.GetSessionID = func() string {
-			return rand.Text()
-		}
+		opts.GetSessionID = rand.Text
 	}
 
 	if opts.Logger == nil { // ensure we have a logger


mcp/util.go
```diff
@@ -15,18 +15,8 @@ func assert(cond bool, msg string) {
 	}
 }
 
-// Copied from crypto/rand.
-// TODO: once 1.24 is assured, just use crypto/rand.
-const base32alphabet = "ABCDEFGHIJKLMNOPQRSTUVWXYZ234567"
-
 func randText() string {
```
maciej-kisiel: Could we replace calls to this function with `rand.Text()` and remove it altogether?

IAmSurajBobade: Thanks for the review, @maciej-kisiel. Addressed your review comment.

I did not remove randText method earlier, to minimize code change and also as ServerOptions.GetSessionID expects a func, and so anyway need to create anon func. And mcp/streamable.go, uses math/rand/v2, causing a conflict.

IAmSurajBobade:
I did not remove randText method earlier, to minimize code change and ServerOptions.GetSessionID expects a func, and so anyway need to create anon func. Also mcp/streamable.go, uses math/rand/v2, causing a conflict and impacting readability.

mcp/server.go
```diff
@@ -175,7 +176,9 @@ func NewServer(impl *Implementation, options *ServerOptions) *Server {
 	}
 
 	if opts.GetSessionID == nil {
-		opts.GetSessionID = randText
+		opts.GetSessionID = func() string {
```
maciej-kisiel: Just assigning `rand.Text` won't work?

IAmSurajBobade: No, it won't. SessionId expects func()string. I mentioned this in the response to your earlier review comment on why I reused the existing randText method earlier.

maciej-kisiel: `rand.Text` is `func() string`. Please do:

```
opts.GetSessionID = rand.Text
```

IAmSurajBobade: ahaa, my bad. resolved.

maciej-kisiel:
Thanks for the contribution!

PR #774, opened 2026-02-03
fix: add JSON tags to ElicitationCapabilities fields

awschmeder:
## Summary

The `Form` and `URL` fields in `ElicitationCapabilities` were missing JSON struct tags, causing them to serialize as `"Form"` and `"URL"` (capitalized) instead of the lowercase `"form"` and `"url"` required by the MCP protocol.

## Problem

During client initialization, when advertising elicitation capabilities, the JSON payload would contain:

```json
{"elicitation": {"Form": {}, "URL": {}}}
```

Instead of the correct:

```json
{"elicitation": {"form": {}, "url": {}}}
```

This caused servers to not recognize the elicitation capabilities, leading to initialization failures or unexpected behavior.

## Fix

Added `json:"form,omitempty"` and `json:"url,omitempty"` struct tags to the `ElicitationCapabilities` struct fields.

## Related Issue

Discovered via downstream issue: https://github.com/docker/cagent/issues/1574

### DIFF
--- a/mcp/protocol.go
+++ b/mcp/protocol.go
@@ -988,8 +988,8 @@ type SamplingCapabilities struct{}
 //
 // If neither Form nor URL is set, the 'Form' capabilitiy is assumed.
 type ElicitationCapabilities struct {
-	Form *FormElicitationCapabilities
-	URL  *URLElicitationCapabilities
+	Form *FormElicitationCapabilities `json:"form,omitempty"`
+	URL  *URLElicitationCapabilities  `json:"url,omitempty"`
 }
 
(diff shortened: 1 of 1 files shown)


maciej-kisiel:
This looks like a correct change to fix a case where we don't follow the specification, but is still a behavior change. I assume in this case it's acceptable, @findleyr to confirm.

findleyr:
Yes, this is definitely an oversight and bug. Therefore, the behavior change is acceptable / must be accepted: no correct client could have used the old behavior.

PR #783, opened 2026-02-07
mcp: add max body size limit for SSE + streamable HTTP

TheodorNEngoy:
This adds a basic DoS guard for the built-in HTTP transports by limiting POST request body size.

- Adds `MaxBodyBytes` to `SSEOptions` and `StreamableHTTPOptions` (default: `DefaultMaxBodyBytes` = 1_000_000; negative disables).
- Adds `MaxBodyBytes` to `SSEServerTransport` and `StreamableServerTransport` for custom handler usage.
- Returns `413 Request Entity Too Large` when the limit is exceeded.
- Includes tests covering both SSE and streamable handlers/transports.

Rationale: avoid unbounded `io.ReadAll(req.Body)` buffering for network-exposed servers.

### DIFF
--- a/mcp/http_limits.go
+++ b/mcp/http_limits.go
@@ -1,0 +1,47 @@
+// Copyright 2025 The Go MCP SDK Authors. All rights reserved.
+// Use of this source code is governed by an MIT-style
+// license that can be found in the LICENSE file.
+
+package mcp
+
+import (
+	"errors"
+	"net/http"
+)
+
+// DefaultMaxBodyBytes is the default maximum size (in bytes) for HTTP request
+// bodies accepted by the built-in SSE and streamable HTTP handlers.
+//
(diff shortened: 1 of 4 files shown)


TheodorNEngoy:
Local sanity check: `go test ./...` passes on this branch.

maciej-kisiel:
Adding a default for this limit is a backwards incompatible behavior change, which we generally try to avoid. Isn't this already easy enough to achieve by wrapping the MCP HTTP Handler with https://godoc.corp.google.com/pkg/net/http#MaxBytesHandler?

maciej-kisiel:
I will close this PR for house-keeping, I believe this goal is achievable easily with the standard library and doesn't require special support in the SDK. Feel free to re-open if you'd like to continue the discussion.

PR #784, opened 2026-02-08
examples: bind auth-middleware server to localhost by default

TheodorNEngoy:
The auth-middleware example currently defaults to `-http :8080`, which binds to all interfaces (LAN-exposed) by default.

This changes the default to `localhost:8080` and adds a short comment noting `-http 0.0.0.0:8080` if you intentionally want to expose it.

Rationale: this example serves token-generation endpoints and authenticated MCP routes; binding loopback by default avoids accidental exposure while keeping the opt-in remote behavior.

### DIFF
--- a/scripts/server-conformance.sh
+++ b/scripts/server-conformance.sh
@@ -70,8 +70,8 @@ fi
 go build -o "$WORKDIR/conformance-server" ./conformance/everything-server
 
 # Start the server in the background
-echo "Starting conformance server on port $PORT..."
-"$WORKDIR/conformance-server" -http=":$PORT" &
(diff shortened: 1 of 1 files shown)


TheodorNEngoy:
Local sanity check: `go test ./...` passes on this branch.

maciej-kisiel:
Thanks for the change. The same logic probably applies here as well: https://github.com/modelcontextprotocol/go-sdk/blob/main/scripts/server-conformance.sh#L74, could you adjust it as well?

TheodorNEngoy:
Pushed an update per your note: scripts/server-conformance.sh now binds the conformance server to loopback by default (localhost:) instead of : (all interfaces).

PR #785, opened 2026-02-10
all: client side OAuth support

maciej-kisiel:
This PR introduces a proposal for how client-side OAuth support could look like in the SDK. It is open for community feedback.

_The main goal of the PR is to validate the proposed APIs. The code may not be polished yet. In particular:_

- _Please ignore any log lines that were introduced for development._
- _Please restrain from providing code style feedback at this moment._
- _Commentary for the API is not finalized yet and will be improved._
- _Unit tests are largely missing._

# Overall design philosophy

The main new interface introduced to support client-side OAuth is the `OAuthHandler`. Contrary to its counter-parts in some other SDKs, it is intentionally small and generic so that it can support most of the OAuth authorization flow variants. The main `mcp` package purposefully delegates almost all auth handling to `OAuthHandler` implementations. This way, the `mcp` package continues to be largely OAuth agnostic.

The `auth` package comes with a default `OAuthHandler` implementation – `AuthorizationCodeOAuthHandler`, which implements the authorization flow for the `authorization_code` grant type, as defined in the [MCP specification](https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization). It is expected that the `OAuthHandler` implementation granularity will be roughly per grant type, as this parameter influences the authorization flow the most.

# Detailed design

## `OAuthHandler` (auth/client.go)

The main interface for the integration consists of two methods:

```go
type OAuthHandler interface {
	// TokenSource returns a token source to be used for outgoing requests.
	TokenSource(context.Context) (oauth2.TokenSource, error)

	// Authorize is called when an HTTP request results in an error that may
	// be addressed by the authorization flow (currently 401 Unauthorized and 403 Forbidden).
	// It is responsible for initiating the OAuth flow to obtain a token source.
	// The arguments are the request that failed and the response that was received for it.
	// If the returned error is nil, [TokenSource] is expected to return a non-nil token source.
	// After a successful call to [Authorize], the HTTP request should be retried by the transport.
	// The function is responsible for closing the response body.
	Authorize(context.Context, *http.Request, *http.Response) error
}
```

`StreamableClientTransport` accepts this interface as a field and uses the `TokenSource` method to add an `Authorization: Bearer` header to outgoing requests to the MCP server. If the request fails with `401: Unauthorized` or `403: Forbidden`, the transport calls `Authorize` to perform the authorization flow. If `Authorize` returns a non-nil error, the request is retried once. `Authorize` errors do not result in termination of the client session (unless they happen during connection), as some OAuth flows are multi-legged and it should be acceptable to retry for example when the authorization grant was received.

*Note:* There was already an `OAuthHandler` type defined as part of a previous iteration of OAuth support. It was renamed to `OAuthHandlerLegacy`. This is the only backwards incompatible API change in the experimental client `auth` part (protected by the `mcp_go_client_oauth` Go build tag). The cost of replacing previous usages is deemed worthwhile to maintain clear naming, under the assumption that the experimental client `auth` part was not widely used.

## `AuthorizationCodeOAuthHandler` (auth/authorization_code.go)

This is the `OAuthHandler` implementation that fulfills the [MCP specification](https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization). In particular, in supports:

- Protected Resource Metadata discovery and fetching
- Authorization Server Metadata discovery and fetching
- Client registration
  - OpenID Connect metadata document
  - Pre-registration
  - Dynamic Client Registration
- PKCE
- Providing custom `state` field generators
- `none`, `client_secret_post`, `client_secret_basic` token endpoint auth methods
- Automatic refresh of tokens
- Persisting tokens in a provided storage implementation

### Authorization Server redirect

As part of the flow for the `authorization_code` grant type, the client needs to point the user to the authorization URL, so that they confirm the access request. This design proposes that it will be supported via a pluggable dependency that will initiate this process in an application-specific way:

```go
type AuthorizationCodeOAuthHandler struct {
	...

	// AuthorizationURLHandler is called to handle the authorization URL.
	// It is responsible for opening the URL in a browser.
	// It should return once the redirect has been issued.
	// The redirect callback should be handled by the caller and the authorization code
	// should be set by calling [SetAuthorizationCode] before retrying the request.
	AuthorizationURLHandler func(ctx context.Context, authorizationURL string) error

	...
}
```

The function is supposed as soon as the user was pointed to the authorization URL, after which `Authorize` will finish with `auth.ErrRedirected`. The application is expected to wait for the callback and provide the required authorization code and state values to the `AuthorizationCodeOAuthHandler` via `FinalizeAuthorization` call. After that, the client call (or connection) can be retried.

### Token storage

`AuthorizationCodeOAuthHandler` includes a `SetTokenSource` method that allows the application to pre-populate the handler with tokens, for example fetched from a persistent store. The token source may be overridden by the handler if the token from the existing token source does not authorize the user and the automatically initiated `Authorize` flow succeeds. Optionally, a `TokenStore` implementation can be provided to the handler.

```go
// TokenStore is an interface than can be used by OAuthHandler implementations
// to save tokens to a persistent storage.
type TokenStore interface {
	Save(context.Context, *oauth2.Token) error
}
```

When provided, the token source created by the handler on a successful token exchange will save each token it returns. The same effect can be achieved in a pre-populated token source thanks to an exposed `func NewPersistentTokenSource(ctx context.Context, wrapped oauth2.TokenSource, store TokenStore) oauth2.TokenSource` utility function.

## `oauthex` changes

Some additional/adjusted building blocks that may be useful when creating `OAuthHandler` implementations are present in the `oauthex` package:

- `auth_meta.go`: Authorization Server Metadata utilities:
  - `GetAuthServerMeta`: get Authorization Server Metadata from a URL
  - `AuthorizationServerMetadataURLs`: generate ASM URL candidates based on MCP specification
- `dcr.go`: Dynamic Client Registration utilities (unchanged)
- `resource_meta.go`: Protected Resource Metadata utilities:
  - `GetProtectedResourceMetadata`: get Protected Resource Metadata from a URL
  - `ProtectedResourceMetadataURLs`: generate PRM URL candidates based on MCP specification
  - `ResourceMetadataURL`, `Scopes`, `Error` – helpers to retrieve fields from the `WWW-Authenticate` challenges
  - `ParseWWWAuthenticate` – as per name (unchanged)
  - [deprecated] `GetProtectedResourceMetadataFromID` and `GetProtectedResourceMetadataFromHeader` in favor of more generic `GetProtectedResourceMetadata`. They will be kept under the Go build tag and removed two versions of the SDK in the future.

# Input requested

Please provide any feedback/suggestions on the proposed API surface you might have. In particular:

- Share thoughts on the following API areas:
  - Expected handler granularity: is the assumption that they will likely follow grant types correct? 
  - Callback mechanism: returning `ErrRedirected` and retrying the initial call at the application level. Some alternatives could be proposed, for example making the authorization URL handler blocking and returning authorization code and state directly (likely through a channel). It would simplify the control flow (no error returned from `Authorize` and application-level retries) and make the implementation a bit cleaner, but would come at a cost of additional goroutines being created.
  - ~~Persisted token's refreshes: in order to create a refreshing token source based on a stored token it's required to create a `oauth2.Config` object. Deriving the values to populate it is a significant part of the `AuthorizationCodeOAuthHandler` implementation and we can consider exposing some mechanisms to not require the users to replicate this logic.~~
  - ~~Exposed extension points for `AuthorizationCodeOAuthHandler`: `TokenSource` and `StateProvider` – they were added in an anticipation of being useful and because other SDKs provide it, but it would be great to validate if that's really the case.~~
  - Error handling: Generally we assume that `Unauthorized` errors do not close the client session (with the exception of session connection that will fail). Is this the right approach? It's unclear what should happen when getting a handler's token source fails, feedback welcome.
  - Naming: feedback always welcome.
  - Specification (both OAuth and MCP) non-compliance – I have tried to do everything by the book, but might have missed something.
- Share your authorization use cases and whether they fit this design.
- ~~Suggest any real Authorization Server implementations that could be easily deployed in dev mode to facilitate end-to-end testing. So far the implementation was tested only with MCP conformance tests.~~

Further steps:
- [x] Switching to an appropriate alternative to the `mcp_go_client_oauth` build tag to make CI pass
- [x] Create an example OAuth client under `examples/client` (dev AS server would come in handy)
- [x] Adjust documentation
- [x] Add unit tests
- [x] Cleanup

Thank you!

### DIFF
--- a/mcp/streamable.go
+++ b/mcp/streamable.go
@@ -404,7 +404,7 @@ func (h *StreamableHTTPHandler) ServeHTTP(w http.ResponseWriter, req *http.Reque
 				// stateless servers.
 				body, err := io.ReadAll(req.Body)
 				if err != nil {
-					http.Error(w, "failed to read body", http.StatusInternalServerError)
+					http.Error(w, "failed to read body", http.StatusBadRequest)
 					return
 				}
 				req.Body.Close()
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,45 @@
+name: "CodeQL Advanced"
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    branches: [ "main" ]
+  schedule:
+    - cron: '31 9 * * 4'
+
+jobs:
+  analyze:
+    name: Analyze (${{ matrix.language }})
+    runs-on: ubuntu-latest
+    permissions:
+      # required for all workflows
+      security-events: write
+
+      # required to fetch internal or private CodeQL packs
+      packages: read
+
+      # only required for workflows in private repositories
+      actions: read
+      contents: read
+
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+        - language: actions
+          build-mode: none
+        - language: go
+          build-mode: autobuild
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+    - name: Initialize CodeQL
+      uses: github/codeql-action/init@89a39a4e59826350b863aa6b6252a07ad50cf83e # v4.32.4
+      with:
+        languages: ${{ matrix.language }}
+        build-mode: ${{ matrix.build-mode }}
+    - name: Perform CodeQL Analysis
+      uses: github/codeql-action/analyze@89a39a4e59826350b863aa6b6252a07ad50cf83e # v4.32.4
+      with:
+        category: "/language:${{matrix.language}}"


maciej-kisiel:
Hi all, this proposal tries to address a long-standing gap in the SDK, which is client-side OAuth. I'm sure many people have worked around it and thus have an understanding what of the flow that will allow them to assess if this proposal would address their needs. Please review and provide your feedback!

I will go through OAuth related issues and tag people that have been active there. Sorry for the noise if you're not interested.

maciej-kisiel:
cc @findleyr @jba @herczyn

auth/authorization_code.go
```diff
@@ -0,0 +1,336 @@
...
+)
+
+type resolvedClientConfig struct {
+	registrationType registrationType
+	clientID         string
+	clientSecret     string
+	authStyle        oauth2.AuthStyle
+}
+
+type AuthorizationCodeOAuthHandler struct {
```
jba: doc

auth/authorization_code.go
```diff
@@ -0,0 +1,336 @@
...
+	if h.AuthorizationURLHandler != nil {
+		if err := h.AuthorizationURLHandler(ctx, authURL); err != nil {
+			return fmt.Errorf("authorization URL handler failed: %w", err)
+		}
+	}
+
+	return ErrRedirected
+}
+
+func (h *AuthorizationCodeOAuthHandler) SetTokenSource(ts oauth2.TokenSource) {
```
jba: why exported?

auth/authorization_code.go
```diff
@@ -0,0 +1,336 @@
...
+	h.tokenSource = ts
+}
+
+func (h *AuthorizationCodeOAuthHandler) FinalizeAuthorization(code, state string) error {
```
jba: ditto

auth/client.go
```diff
@@ -16,16 +17,74 @@ import (
...
 // To obtain the protected resource metadata, call [oauthex.GetProtectedResourceMetadataFromHeader].
-type OAuthHandler func(req *http.Request, res *http.Response) (oauth2.TokenSource, error)
+// Deprecated: Please use the new OAuthHandler abstraction that is built
+// into the streamable transport.
+type OAuthHandlerLegacy func(req *http.Request, res *http.Response) (oauth2.TokenSource, error)
+
+// Error that will be thrown if the call failed due to authorization.
+var ErrUnauthorized = errors.New("unauthorized")
+
+type OAuthHandler interface {
```
jba, line 24: Add an unexported method, so we can add methods to this interface without breaking clients.

auth/client.go
```diff
@@ -16,16 +17,74 @@ import (
...
+	ctx     context.Context
+}
+
+// NewPersistentTokenSource returns a [oauth2.TokenSource] that
+// persists the token to a given [TokenStore] after every successful
+// [oauth2.TokenSource.Token] call.
+// It is especially useful when wrapping a [oauth2.TokenSource]
+// that automatically refreshes the token when it expires.
+// The passed context is used for [TokenStore.Save] calls.
+func NewPersistentTokenSource(ctx context.Context, wrapped oauth2.TokenSource, store TokenStore) oauth2.TokenSource {
```
jba: This seems nice, but are we sure we need it, and it's the right API? This sounds like an extra we can add later if there is demand for it.

auth/authorization_code.go
maciej-kisiel: Yes, as explained in the preamble of the PR, commentary will be improved. This includes doc comments.

auth/authorization_code.go
maciej-kisiel: So that the developers can set a token source before authorization flow is triggered by an Unauthorized error. This would be helpful if tokens are stored in-between sessions and the OAuth handler should be primed with a token.

auth/authorization_code.go
maciej-kisiel: Developers are required to call this function to finalize the flow after they get the authorization code from the redirect. More details about the flow are present in the PR description.

auth/client.go
maciej-kisiel: I've never used this technique, as interfaces I authored were for local use. Wouldn't it make it difficult for the developers to implement this interface?

auth/client.go
maciej-kisiel: Very good questions.
Is it the right API? → I don't know, this is one of the areas where I would like to get more input
Do we need it? → I don't have any evidence, apart from similar functionality being present in other SDKs. I proposed it, as I believe asking the user to go through the OAuth flow every time a new OAuthHandler is created is poor UX.

findleyr:
The shape of this looks right to me, but I really have to defer to @jba, who has thought about this much more than me.

auth/authorization_code.go
maciej-kisiel: Removed for now, together with other persistence-related APIs.

auth/client.go
maciej-kisiel: Removed for now, together with other persistence-related APIs.

auth/client.go
maciej-kisiel: Added.

maciej-kisiel:
I have added an end-to-end example that uses the new APIs to showcase authorization. It's under `examples/auth`, there's both a client and a server binary. The examples require access to a configured Authorization Server.

I ran the example with a dev instance of the https://www.keycloak.org/ server. I was able to test a pre-registered client and dynamic client registration, as the server doesn't yet support client ID metadata documents.

I would like to highlight my feedback request about the design of the part where an authorization URL is returned to the user. In this flow, the authorization server after getting user's consent, redirects to an URL specified by the client (likely a local URL) and this request contains the authorization code that our handler is able to exchange for a token.

Currently this process happens with a return of control to the application: the SDK request fails with `auth.ErrRedirected` when it was initiated. The SDK request should be retried by the application after calling `FinalizeAuthorization` on the handler with the received authorization code. As you can see in the client example, this logic is not super easy to handle and calling `client.Connect` in a loop (or retrying in an `if`) is not elegant. I also worry about the UX during the "scope step-up" scenario. If application expects it, basically every relevant MCP call would need to be wrapped with such a loop. I don't know how widely this functionality is used though.

As mentioned before, the alternative could be to **not** return control to the caller and adjust `AuthorizationURLHandler` to be something like `func(ctx context.Context, authorizationURL string) (AuthResult, error)` and assume it will block for a longer time.

auth/client.go
jba: Good point. We will eventually want people to implement it. I'm just concerned that we don't have the right API yet.

jba:
Thanks for testing it for real! keycloak.org was a good find.

I believe if you look at the TS or C# examples, they start a small local webserver that the auth server redirects to. That gets the code which can then be exchanged for a token, and then Authorize can return with a token.

There should be a way for both flows to co-exist: one that returns control with an error, and one that does everything in a single call.

jba:
> There should be a way for both flows to co-exist: one that returns control with an error, and one that does everything in a single call.

It looks like you already have this.

maciej-kisiel:
> > There should be a way for both flows to co-exist: one that returns control with an error, and one that does everything in a single call.
> 
> It looks like you already have this.

Do you suggest we should adjust `AuthorizationURLHandler` signature to be able to return the authorization code (and other required data) directly to allow immediate continuation? And if the developers would like to let the control to go back to the application, they could return an error of their choice and check for it in SDK calls? `ErrRedirected` could probably be removed in that case, the developers could define something similar themselves if needed.

neild:
> As mentioned before, the alternative could be to not return control to the caller and adjust AuthorizationURLHandler to be something like func(ctx context.Context, authorizationURL string) (AuthResult, error) and assume it will block for a longer time.

I read over the proposed API and example client before reading this comment, and this was going to be my top suggestion: The current control flow is quite complicated, and could be simplified by making some operations blocking instead of callback-based.

Returning ErrRedirected from client.Connect means that the client user needs to not only retry the error, but must also synchronize with the authorization flow to identify when it has completed. This is necessary because the OAuthHandler interface provides a way to start authorization, but no way to tell when authorization has completed.

I think OAuthHandler.Authorize should block until authorization has completed. Perhaps this leaves an additional goroutine running for the duration of authorization, but goroutines are cheap.

Similarly, AuthorizationCodeOAuthHandler.AuthorizationURLHandler probably should block and return the code rather than returning it asynchronously. (But see the next section below.)

----

AuthorizationCodeOAuthHandler.RedirectURL seems like it might be a problem to me: This must be set before AuthorizationURLHandler is called (because the redirect URL is part of the authorization URL), but the example client does not start the localhost HTTP server until AuthorizationURLHandler is called. This makes it impossible for the server to listen on a non-well-known port (notably, it can't listen on localhost:0) unless the listener is preemptively created even when the authorization flow does not need to execute.

Perhaps there should be another type here, something along the lines of:

```go
type AuthorizationHandler interface {
  // RedirectURL is the URL to redirect to after authorization.
  RedirectURL() string

  // Authorize handles the authorization URL.
  Authorize(ctx context.Context, authorizationURL string) (code, state string, err error)

  // Stop shuts down the localhost server (if one was started).
  Stop()
}

type AuthorizationCodeOAuthHandler struct {
  // NewAuthorizationHandler replaces RedirectURL and AuthorizationURLHandler.
  NewAuthorizationHandler func() (AuthorizationHandler, error)
}
```

----

Are most implementations of AuthorizationURLHandler going to use a localhost HTTP server? If so, it seems like we should provide a reasonable default implementation rather than requiring everyone to reimplement the same rather-subtle behavior.

A few subtleties:

The example client doesn't handle multiple visits to the localhost HTTP server. You can wedge the server by quickly sending it two requests, one of which will block on sending to r.authChan.

If multiple clients copy the same example code, they'll all listen on the same port (localhost:3142) and potentially conflict with each other. I don't know if this is a problem in practice.

The localhost server is started in a new goroutine, and there's a theoretical race condition where it hasn't started by the time the user visits it. (This seems fairly unlikely in practice.) This could be avoided by creating a net.Listener and then starting the server in a new goroutine with server.Serve instead of ListenAndServe.

maciej-kisiel:
> I think OAuthHandler.Authorize should block until authorization has completed. Perhaps this leaves an additional goroutine running for the duration of authorization, but goroutines are cheap.
>
> Similarly, AuthorizationCodeOAuthHandler.AuthorizationURLHandler probably should block and return the code rather than returning it asynchronously.

Thanks, I'll try this approach out and see how it looks.

> AuthorizationCodeOAuthHandler.RedirectURL seems like it might be a problem to me: This must be set before AuthorizationURLHandler is called (because the redirect URL is part of the authorization URL), but the example client does not start the localhost HTTP server until AuthorizationURLHandler is called. This makes it impossible for the server to listen on a non-well-known port (notably, it can't listen on localhost:0) unless the listener is preemptively created even when the authorization flow does not need to execute.

The example is a simplified representation of the flow. My understanding is that in practice, the redirect URLs will either be full URLs like `https://myapp.com/oauth2` for web and mobile apps or custom URI schemes like `my-app://oauth2` in case of native desktop apps. Starting a temporary HTTP server on localhost should be a niche use case at worst, used mostly for development purposes.

The thing with `redirect_uri` is that in most OAuth client registration methods supported will need to be provided to the Authorization Server before the flow even begins, as most ASes will validate that what the `redirect_uri` they have registered for a given client matches the one passed in the authorization request for security reasons.

To sum up, I don't think these quirks of the example should influence the main design here.

> The example client doesn't handle multiple visits to the localhost HTTP server. You can wedge the server by quickly sending it two requests, one of which will block on sending to r.authChan.

Good point, I'll see if I can get around this without complicating the example too much.

> If multiple clients copy the same example code, they'll all listen on the same port (localhost:3142) and potentially conflict with each other. I don't know if this is a problem in practice.

I don't think this is a problem in practice, as explained above I don't think the example provides a very realistic way of handling the URIs, but it's a tradeoff with simplicity.

> The localhost server is started in a new goroutine, and there's a theoretical race condition where it hasn't started by the time the user visits it. (This seems fairly unlikely in practice.) This could be avoided by creating a net.Listener and then starting the server in a new goroutine with server.Serve instead of ListenAndServe.

Good catch, I will fix it.

maciej-kisiel:
I applied the suggestions to make the APIs more direct and blocking and I have to say I'm happy with the result. Especially the user code looks much more elegant.

I also introduce a constructor to be able to validate the inputs and throw errors earlier. Two more changes I'm considering:

- Renaming `AuthorizationCodeOAuthHandler` to `AuthorizationCodeHandler` (that would be consistent with the current name for the config struct, doing it the other way would result in unbearably long name)
- Removing the `StateProvider` argument until users confirm the need and request it.

Let me know if anyone has any thoughts on these.

jba:
> > > There should be a way for both flows to co-exist: one that returns control with an error, and one that does everything in a single call.
> > 
> > 
> > It looks like you already have this.
> 
> Do you suggest we should adjust `AuthorizationURLHandler` signature to be able to return the authorization code (and other required data) directly to allow immediate continuation? And if the developers would like to let the control to go back to the application, they could return an error of their choice and check for it in SDK calls? `ErrRedirected` could probably be removed in that case, the developers could define something similar themselves if needed.

I was referring to this code in streamable.go:

```
if err := c.oauthHandler.Authorize(ctx, req, resp); err != nil {
			// Wrap with ErrRejected so the jsonrpc2 connection doesn't set writeErr
			// Wrap the authorization error as well for client inspection.
			return fmt.Errorf("%s: %w: %w", requestSummary, jsonrpc2.ErrRejected, err)
		}
		// Retry the request after successful authorization.
		resp, err = doRequest()
```

If Authorize returns an error, you return the error all the way up, and the caller has to deal with it and retry. But if it doesn't return an error, you proceed, which calls handler.TokenSource. So Authorize can do the full round-trip.

jba:
> * Renaming `AuthorizationCodeOAuthHandler` to `AuthorizationCodeHandler` (that would be consistent with the current name for the config struct, doing it the other way would result in unbearably long name)
> * Removing the `StateProvider` argument until users confirm the need and request it.
> 
> Let me know if anyone has any thoughts on these.

Both fine. In general, remove anything not absolutely necessary.

oauthex/auth_meta.go
```diff
@@ -113,51 +116,109 @@ type AuthServerMeta struct {
...
-		if err != nil {
-			// issuerURL is bad; no point in continuing.
-			return nil, err
-		}
-		asm, err := getJSON[AuthServerMeta](ctx, c, u, 1<<20)
-		if err == nil {
-			if asm.Issuer != issuerURL { // section 3.3
-				// Security violation; don't keep trying.
-				return nil, fmt.Errorf("metadata issuer %q does not match issuer URL %q", asm.Issuer, issuerURL)
+func GetAuthServerMeta(ctx context.Context, metadataURL AuthorizationServerMetadataURL, c *http.Client) (*AuthServerMeta, error) {
```
jba: Why do we bundle the issuer with the url? They could be separate arguments here.

oauthex/auth_meta.go
```diff
@@ -113,51 +116,109 @@ type AuthServerMeta struct {
...
+	// URL where the Authorization Server Metadata may be retrieved.
+	URL string
+	// Issuer that was used to construct the [URL].
+	Issuer string
+}
+
+// AuthorizationServerMetadataURLs returns a list of URLs to try when looking for
+// authorization server metadata as mandated by the MCP specification:
+// https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#authorization-server-metadata-discovery.
+func AuthorizationServerMetadataURLs(issuerURL string) []AuthorizationServerMetadataURL {
```
jba: All the return values have the argument issuer, which is why I think we may not need the AuthorizationServerMetadataURL type.

examples/auth/client/main.go
```diff
@@ -0,0 +1,131 @@
...
+	flag.Parse()
+	receiver := &codeReceiver{
+		authChan: make(chan *auth.AuthorizationResult),
+		errChan:  make(chan error),
+	}
+	listener, err := net.Listen("tcp", fmt.Sprintf("localhost:%d", *callbackPort))
+	if err != nil {
+		log.Fatalf("failed to listen: %v", err)
+	}
+	go receiver.serveRedirectHandler(listener)
```
jba, line 82: How does this goroutine terminate? (How do we avoid leaking it?)

jba:
> I think OAuthHandler.Authorize should block until authorization has completed. Perhaps this leaves an additional goroutine running for the duration of authorization, but goroutines are cheap.

I'm not sure this is always possible. The MCP client might not be able to handle the redirect URL itself (in fact I think that will be the norm). It must rely on the user's browser, or in general on its caller. We've all seen the CLI interfaces where you have to copy a URL into your browser, then copy the resulting text back to the prompt. That is consistent with Authorize blocking. But what if the MCP client is embedded in a context where it can't (or doesn't know if it can) issue that prompt?

I believe that's why there has to be a way for the MCP client to return control to its caller, so the caller can do whatever it needs to in order to obtain a token, and then restart the client connection from the beginning.

jba:
> Returning ErrRedirected from client.Connect means that the client user needs to not only retry the error, but must also synchronize with the authorization flow to identify when it has completed. This is necessary because the OAuthHandler interface provides a way to start authorization, but no way to tell when authorization has completed.

I'm not sure I understand the problem. It's true that the client user has to know when the auth flow is completed. But that is not the job of anything in the MCP client. By returning ErrRedirected, we're basically giving up. The state of the current OAuthHandler is no longer relevant.

The idea is that the client user handles the rest of the auth flow, handling the redirect URL and obtaining a token, and then calls client.Connect with a different OAuthHandler, a trivial one that just implements TokenSource to hand back the token. 

Or maybe AuthorizationCodeHandler has a SetToken method, so it can be reused. I'm fine with that if it simplifies the coding.

neild:
> I believe that's why there has to be a way for the MCP client to return control to its caller, so the caller can do whatever it needs to in order to obtain a token, and then restart the client connection from the beginning.

In the client example, there are the following layers (at least):

- Top-level user code which calls client.Connect.
- mcp.StreamableClientTransport, which calls StreamableClientTransport.OAuthHandler.Authorize.
- auth.AuthorizationCodeHandler, which calls AuthorizationCodeHandlerConfig.AuthorizationURLHandler.
- The AuthorizationURLHandler, which does something to obtain an authorization code.

In the previous design which I reviewed, the authorization flow happened in both the top and bottom level: The AuthorizationURLHandler starts the authorization process, an ErrRedirected error bubbles up to the user, and the user is now responsible for figuring out when the authorization process completes.

Which parts of this are "the caller"?

If the AuthorizationURLHandler is "the caller", then the updated design (where the authorization flow happens entirely within the AuthorizationURLHandler) still does exactly what you say--control is passed to the caller's AuthorizationURLHandler.

If the AuthorizationURLHandler is not the caller, then this was a layering violation in the prior design (with ErrRedirected), since the top-level user needs to interact with the AuthorizationURLHandler before retrying the connection.

An alternative design might be to drop AuthorizationURLHandler entirely, and return the authorization URL to the caller. Something along the lines of:

```go
for {
  if err == nil {
    break
  }
  e, ok := err.(*auth.ErrRedirected)
  if !ok {
    return err // some fatal error
  }
  authorizeUsing(e.AuthorizationURL)
  // retry
}
```

...but that doesn't seem like an improvement on the AuthorizationURLHandler callback to me.

auth/authorization_code.go
```diff
@@ -0,0 +1,456 @@
...
+	// The caller is responsible for handling the redirect out of band.
+	// If Dynamic Client Registration is used, the RedirectURL must be consistent
+	// with [DynamicClientRegistrationConfig.Metadata.RedirectURIs].
+	RedirectURL string
+
+	// AuthorizationURLHandler is a required function called to handle the authorization request.
+	// It is responsible for opening the URL in a browser for the user to start the authorization process.
+	// It should return the authorization code and state once the Authorization Server
+	// redirects back to the [AuthorizationCodeHandler.RedirectURL].
+	AuthorizationURLHandler func(ctx context.Context, authorizationURL string) (*AuthorizationResult, error)
```
jba: arg should be a struct

auth/authorization_code.go
```diff
@@ -0,0 +1,456 @@
...
+	"github.com/modelcontextprotocol/go-sdk/oauthex"
+	"golang.org/x/oauth2"
+)
+
+// ClientSecretAuthMethod defines "client_secret_*" authentication methods per
+// https://www.iana.org/assignments/oauth-parameters/oauth-parameters.xhtml#token-endpoint-auth-method.
+// "client_secret_jwt" is not currently supported.
+type ClientSecretAuthMethod int
+
+const (
```
jba: remove?

mcp/streamable.go
```diff
@@ -1724,14 +1731,59 @@ func (c *streamableClientConn) Write(ctx context.Context, msg jsonrpc.Message) e
...
+			// Wrap the authorization error as well for client inspection.
+			return fmt.Errorf("%s: %w: %w", requestSummary, jsonrpc2.ErrRejected, err)
+		}
+		// Retry the request after successful authorization.
+		resp, err = doRequest()
+		if err != nil {
+			return err
+		}
+	}
+	if resp.StatusCode == http.StatusForbidden && c.oauthHandler != nil {
```
jba: comment linking to part of spec

jba: delegate all this to authorize

oauthex/resource_meta.go
```diff
@@ -97,22 +109,51 @@ func getPRM(ctx context.Context, purl string, c *http.Client, wantResource strin
...
+	// URL represents a URL where Protected Resource Metadata may be retrieved.
+	URL string
+	// Resource represents the corresponding resource URL for [URL].
+	// It is required to perform validation described in RFC 9728, section 3.3.
+	Resource string
+}
+
+// ProtectedResourceMetadataURLs returns a list of URLs to try when looking for
+// protected resource metadata as mandated by the MCP specification.
+func ProtectedResourceMetadataURLs(metadataURL, resourceURL string) []ProtectedResourceMetadataURL {
```
jba: move to auth package

jba:
In the current design, everything happens in the AuthorizationURLHandler, which blocks. I now believe that can handle any kind of control flow, even the ones I described as non-blocking, by a suitable use of goroutines and channels.

There is no specially defined error in case someone really wants to retry connecting from the top, but also nothing prevents one from defining such an error and returning it from AuthorizationURLHandler.

So I think we're at a place where (1) the API is pretty minimal, (2) it conveniently supports some common control flows, and (3) it's general enough to support other control flows. I think that's as good as we're going to get at this point, before we have real-world users.

auth/authorization_code.go
maciej-kisiel: This was removed in the newest version of the API.

oauthex/auth_meta.go
maciej-kisiel: Done

oauthex/auth_meta.go
maciej-kisiel: Removed.

examples/auth/client/main.go
maciej-kisiel: It runs for the whole lifecycle of the program and is terminated by the defer call in the next line. I thought this will be simpler in an example rather than trying to create a temporary server.

auth/authorization_code.go
maciej-kisiel: Done.

auth/authorization_code.go
maciej-kisiel: Removed.

mcp/streamable.go
maciej-kisiel: Added in Authorize.

oauthex/resource_meta.go
maciej-kisiel: Moved to auth, auth_meta one as well.

mcp/streamable.go
maciej-kisiel: Moved to Authorize.

mcp/streamable.go
```diff
@@ -1718,20 +1724,46 @@ func (c *streamableClientConn) Write(ctx context.Context, msg jsonrpc.Message) e
...
-	req.Header.Set("Accept", "application/json, text/event-stream")
-	c.setMCPHeaders(req)
 
-	resp, err := c.client.Do(req)
-	if err != nil {
-		// Any error from client.Do means the request didn't reach the server.
-		// Wrap with ErrRejected so the jsonrpc2 connection doesn't set writeErr
-		// and permanently break the connection.
-		return fmt.Errorf("%w: %s: %v", jsonrpc2.ErrRejected, requestSummary, err)
+	if slices.Contains([]int{http.StatusUnauthorized, http.StatusForbidden}, resp.StatusCode) && c.oauthHandler != nil {
```
jba: nit: for two things it's simpler to use x==a || x==b

auth/client.go
```diff
@@ -2,122 +2,31 @@
...
+	// TokenSource returns a token source to be used for outgoing requests.
+	// Returned token source might be nil. In that case, the transport will not
+	// add any authorization headers to the request.
+	TokenSource(context.Context) (oauth2.TokenSource, error)
+
+	// Authorize is called when an HTTP request results in an error that may
+	// be addressed by the authorization flow (currently 401 Unauthorized and 403 Forbidden).
+	// It is responsible for performing the OAuth flow to obtain an access token.
+	// The arguments are the request that failed and the response that was received for it.
+	// Currently the body of the passed request is consumed by the transport
```
jba: The doc should just describe behavior, so remove "currently" and the sentence about filing an issue.
Maybe:

The headers of the request are available but the body will have already been consumed when Authorize is called.

auth/client.go
```diff
@@ -2,122 +2,31 @@
...
+	// before [Authorize] is called. Please file an issue if you need the body to be available.
+	// If the returned error is nil, [TokenSource] is expected to return a non-nil token source.
+	// After a successful call to [Authorize], the HTTP request should be retried by the transport.
+	// The function is responsible for closing the response body.
```
jba, line 40: Why does it have to close the response body if it can't read it?

auth/client.go
```diff
@@ -2,122 +2,31 @@
...
+	TokenSource(context.Context) (oauth2.TokenSource, error)
+
+	// Authorize is called when an HTTP request results in an error that may
+	// be addressed by the authorization flow (currently 401 Unauthorized and 403 Forbidden).
+	// It is responsible for performing the OAuth flow to obtain an access token.
+	// The arguments are the request that failed and the response that was received for it.
+	// Currently the body of the passed request is consumed by the transport
+	// before [Authorize] is called. Please file an issue if you need the body to be available.
+	// If the returned error is nil, [TokenSource] is expected to return a non-nil token source.
+	// After a successful call to [Authorize], the HTTP request should be retried by the transport.
```
jba: s/should/will/

I think there should be a paragraph in the OAuthHandler doc (last paragraph) that outlines the responsibilities of transports. "If a transport wishes to support OAuth 2 authentication, it should..."

auth/client_private.go
```diff
@@ -0,0 +1,131 @@
+// Copyright 2025 The Go MCP SDK Authors. All rights reserved.
+// Use of this source code is governed by an MIT-style
+// license that can be found in the LICENSE file.
+
+//go:build mcp_go_client_oauth
```
jba, line 5: I think it's time to remove the build tags.

Once that's done and you think the API is ready, use `go doc -all` locally to get the API as text and let's make sure it's right. It's harder to miss things when you see the API that way.

examples/auth/client/main.go
jba: Yes, that's fine, I just missed it.

examples/auth/client/main.go
```diff
@@ -0,0 +1,131 @@
...
+)
+
+type codeReceiver struct {
+	authChan chan *auth.AuthorizationResult
+	errChan  chan error
+	listener net.Listener
+	server   *http.Server
+}
+
+func (r *codeReceiver) serveRedirectHandler(listener net.Listener) error {
```
jba: This shouldn't return anything, since it really communicates only through the channels.

For the record, what you have is fine, but here is how I'd write this:
- First I'd write a regular function that returns (*AuthResult, error).
- Then I'd write a type that's a struct with both *AuthResult and error, and make the channel be of that type.
- Then I'd write the function that calls the first function and sends to the channel.
That splits out the concurrent part from the sequential function, which is clearer IMO, and also makes it easier to test the sequential function if you wanted to.

It's fine if you have the channel type and the second function declared at top level, but there's another simplification you can make in this case, since so much is in main. Around L83 you could declare the channel type, create the channel, and write
```
go func() {
    ar, err := receiver.serveRedirectHandler(listener)
    ch <- result{ar, err}
}()
```
For that to work you'd also have to make AuthorizationCodeFetcher on line 88 a function literal.

This is all optional.

auth/client.go
maciej-kisiel: We can't read request body, response body is available.

examples/auth/client/main.go
maciej-kisiel: Removed the return value. I think I will leave the rest as is.

auth/client.go
maciej-kisiel: I did the replacement. I'm less sure what exactly you would like to see in this additional paragraph though.

auth/client.go
maciej-kisiel: Done.

mcp/streamable.go
maciej-kisiel: Done.

auth/client.go
jba: oops!

auth/client.go
jba: "If a transport wishes to support OAuth 2 authentication, it should support being configured with an OAuthHandler. It should call the handler's TokenSource method whenever it sends an HTTP request to set the Authorization header. If a request fails with a 401 or 403, it should call Authorize, and if that returns nil, it should retry the request. It should not call Authorize after the second failure. See StreamableClientTransport int the mcp package for an example."

maciej-kisiel: Done.

auth/client_private.go
maciej-kisiel: Output is available at https://gist.github.com/maciej-kisiel/f1bb544fadb32d8f7d49e8e81587f1f4

PR #789, opened 2026-02-13
jsonrpc2: fix Content-Length header parsing to be case-insensitive

### LINKED ISSUE #787: HeaderFramer rejects lowercase content-length header (case-insensitive per RFC)
### Bug
`jsonrpc2.HeaderFramer` only recognizes the exact header name `Content-Length`. If a peer sends `content-length` (lowercase), `Read` returns `missing Content-Length header` even though the header is valid per RFC 7230 / LSP.

### Repro
Minimal stream (lowercase header):
```
content-length: 2\r\n\r\n{}
```

```go
fr := jsonrpc2.HeaderFramer()
rd := fr.Reader(strings.NewReader("content-length: 2\r\n\r\n{}"))
_, err := rd.Read(context.Background())
// err = missing Content-Length header
```

### Expected
Header parsing should be case-insensitive, so `content-length` should be accepted.

### Actual
`Read` errors with `missing Content-Length header`.

### Suggested fix
Normalize the header name using `strings.EqualFold` or `strings.ToLower` before the switch. This preserves interoperability with LSP/HTTP-style peers that lowercase headers.

nithinputhenveettil:
Fix Content-Length header parsing to be case-insensitive as per LSP specification.

All unit tests pass locally (go test ./...).

Fixes #787

### DIFF
--- a/internal/jsonrpc2/frame.go
+++ b/internal/jsonrpc2/frame.go
@@ -163,8 +163,8 @@ func (r *headerReader) Read(ctx context.Context) (Message, error) {
 			return nil, fmt.Errorf("invalid header line %q", line)
 		}
 		name, value := line[:colon], strings.TrimSpace(line[colon+1:])
-		switch name {
-		case "Content-Length":
+		switch {
+		case strings.EqualFold(name, "Content-Length"):
 			if contentLength, err = strconv.ParseInt(value, 10, 32); err != nil {
 				return nil, fmt.Errorf("failed parsing Content-Length: %v", value)
 			}


nithinputhenveettil:
@maciej-kisiel :

Like you mentioned in the issue, I also feel a simple strings.EqualFold should do the job. All unit tests are passing successfully. I also changed Content-Length in the line below to different variations like content-length, COntent-LeNgth, etc., and in all cases the tests ran successfully.


https://github.com/modelcontextprotocol/go-sdk/blob/main/internal/jsonrpc2/frame.go#L203

Issue #805, opened 2026-02-18
Message Smuggling via Case-Insensitive JSON Unmarshalling


slimslenderslacks:
## Summary

The go-sdk is vulnerable to a protocol parser differential attack that allows message smuggling through case-variant JSON fields. This vulnerability exploits
Go's case-insensitive JSON unmarshalling behavior, which violates the JSON-RPC 2.0 specification's requirement for case-sensitive field matching.

## Impact

  An attacker can bypass authorization checks and invoke unauthorized MCP methods by sending messages with duplicate fields using different cases. The last
  field wins during parsing, but validation/authorization occurs on the first field.

Severity: High - Allows unauthorized access to restricted tools, prompts, and resources

### Root Cause

The SDK uses Go's standard json.Unmarshal() which performs case-insensitive field matching. When a JSON message contains duplicate keys with different cases
(e.g., both "name" and "Name"), Go matches both to the same struct field, and the last occurrence wins.

This violates the JSON-RPC 2.0 specification which requires exact, case-sensitive field matching.

## Affected Components

1. internal/jsonrpc2/messages.go - DecodeMessage() function (lines 174-207)
2. mcp/shared.go - newMethodInfo() function's unmarshalParams closure (lines 283-308)

## Attack Scenario

An attacker can smuggle unauthorized method names through case variants:

```
  {
    "jsonrpc": "2.0",
    "id": 1,
    "method": "tools/call",
    "params": {
      "name": "greet",        // Lowercase - passes authorization
      "Name": "secretTool",   // Uppercase - smuggled, overwrites after parsing
      "arguments": {"target": "sensitive_data"}
    }
  }
```

What happens:
1. Authorization middleware checks params.name = "greet" ✅ (allowed)
2. Go's json.Unmarshal() matches both "name" and "Name" to the same struct field
3. The last field ("Name": "secretTool") wins
4. When re-serialized or accessed, params.Name = "secretTool"
5. Unauthorized tool secretTool is executed ❌

Affected Message Types

- tools/call - name field can be smuggled via Name, NAME, etc.
- prompts/get - name field can be smuggled
- resources/read - uri field can be smuggled via URI, Uri, etc.
- Any params struct with json-tagged fields

## Reproduction

```
  package main

  import (
      "encoding/json"
      "fmt"
  )

  type Params struct {
      Name string `json:"name"`
  }

  func main() {
      // Attack payload with duplicate keys
[...]

PR #806, opened 2026-02-18
Remediate Case-Insensitive JSON unmarshalling

slimslenderslacks:
The go-sdk is vulnerable to a protocol parser differential attack that allows message smuggling through case-variant JSON fields. This vulnerability exploits Go's case-insensitive JSON unmarshalling behavior, which violates the JSON-RPC 2.0 specification's requirement for case-sensitive field matching.

There is a full description in the issue #805

The fix is really to replace json.Unmarshal with a safer StrictUnmarshal which handles clients that try to take advantage of the json.Unmarshal case insensitivity.

# Affected Files
1. internal/jsonrpc2/messages.go - DecodeMessage() function (lines 174-207)
2. mcp/shared.go - newMethodInfo() function's unmarshalParams closure (lines 283-308)

### DIFF
--- a/internal/jsonrpc2/strict.go
+++ b/internal/jsonrpc2/strict.go
@@ -1,0 +1,191 @@
+// Copyright 2025 The Go MCP SDK Authors. All rights reserved.
+// Use of this source code is governed by an MIT-style
+// license that can be found in the LICENSE file.
+
+package jsonrpc2
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"reflect"
+	"strings"
+)
+
+// StrictUnmarshal unmarshals JSON data into v with strict validation rules:
+// - Rejects duplicate keys with different cases (e.g., "name" and "Name")
+// - Validates that JSON field names exactly match struct tags (case-sensitive)
+// - Rejects unknown fields not defined in the struct
+//
+// This prevents message smuggling attacks that exploit Go's case-insensitive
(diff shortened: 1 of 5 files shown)


internal/jsonrpc2/strict.go
```diff
@@ -0,0 +1,191 @@
...
+	if err := validateNoDuplicateKeys(data); err != nil {
+		return fmt.Errorf("strict unmarshal: %w", err)
+	}
+
+	// 2. Validate field names match struct tags exactly (case-sensitive)
+	if err := validateFieldCase(data, v); err != nil {
+		return fmt.Errorf("strict unmarshal: %w", err)
+	}
+
+	// 3. Use strict decoder that disallows unknown fields
```
findleyr, line 34: This seems very liable to break real-world use cases. Is it necessary to mitigate the security threat? I can understand the need to reject messages that contain duplicate fields, but *unknown* fields don't seem to be relevant to the threat vector described. Do other SDKs reject unknown fields?

slimslenderslacks: It is not necessary to mitigate the security thread. We could slim this down and just look for duplicate fields.  But probably even better would be if we could drop in a case-sensitive replacement (as you pointed out in the issue).

slimslenderslacks:
@maciej-kisiel @findleyr closing but let me know if you want me to merge any tests.

PR #807, opened 2026-02-18
all: use case-sensitive JSON unmarshaling

maciej-kisiel:
As reported in https://github.com/modelcontextprotocol/go-sdk/issues/805, case insensitive unmarshaling can pose security issues. This change adjusts all non-test and non-example unmarshaling to use a case sensitive mode.

This introduces a new external dependency `github.com/segmentio/encoding`.


maciej-kisiel:
@findleyr I'll leave it up go give you an opportunity for review, since you already engaged in the topic. I plan to merge it tomorrow morning (EMEA) at the latest and will likely release v1.3.1 with it as well.

findleyr (approved):
Amazing! Thanks for being so on top of this.

PR #817, opened 2026-02-23
fix: return 400 instead of 500 when body read fails in stateless mode

### LINKED ISSUE #816: Streamable HTTP handler returns 500 instead of 400 when body read fails
## Problem

In `mcp/streamable.go` around line 376, when `io.ReadAll(req.Body)` fails during stateless mode initialization, the handler returns `http.StatusInternalServerError` (500).

However, body read failures are almost always caused by client-side issues:
- Client disconnected before sending full body
- Network timeout/interruption
- Connection reset

Returning 500 is misleading because it suggests a server error when the issue is on the client side.

## Expected Behavior

Return `http.StatusBadRequest` (400) instead, consistent with the [similar check in `servePOST` which already returns 400](https://github.com/modelcontextprotocol/go-sdk/blob/19420360b7601460c6a1edf4860d701d796bb14a/mcp/streamable.go#L1057-L1061).

## Impact

Users monitoring their MCP servers see 500 errors in telemetry that appear to be server bugs, when they're actually client disconnections. This causes unnecessary investigation and alert fatigue.

roncodingenthusiast:
## Summary

Changes `http.StatusInternalServerError` to `http.StatusBadRequest` when `io.ReadAll(req.Body)` fails during stateless mode initialization in the streamable HTTP handler.

Fixes #816

## Rationale

Body read failures are client-side errors (disconnection, timeout, incomplete request), not server errors. Returning 500 is misleading and causes unnecessary alert fatigue for users monitoring their MCP servers.

This makes the behavior consistent with `servePOST` which already returns 400 for the same error condition.

## Testing

- Existing tests pass
- This is a one-line change affecting only the HTTP status code returned

### DIFF
--- a/mcp/streamable.go
+++ b/mcp/streamable.go
@@ -405,7 +405,7 @@ func (h *StreamableHTTPHandler) ServeHTTP(w http.ResponseWriter, req *http.Reque
 				// stateless servers.
 				body, err := io.ReadAll(req.Body)
 				if err != nil {
-					http.Error(w, "failed to read body", http.StatusInternalServerError)
+					http.Error(w, "failed to read body", http.StatusBadRequest)
 					return
 				}
 				req.Body.Close()


jba (commented):
This seems reasonable. Will let Maciej approve.

mcp/streamable.go
```diff
@@ -403,11 +403,11 @@ func (h *StreamableHTTPHandler) ServeHTTP(w http.ResponseWriter, req *http.Reque
 			{
 				// TODO: verify that this allows protocol version negotiation for
 				// stateless servers.
-				body, err := io.ReadAll(req.Body)
-				if err != nil {
-					http.Error(w, "failed to read body", http.StatusInternalServerError)
-					return
-				}
+			body, err := io.ReadAll(req.Body)
```
maciej-kisiel: Please fix lint – you will need to indent it as it was.

roncodingenthusiast: yes thank you! done

PR #832, opened 2026-03-03
mcp: protect ioConn.protocolVersion with a mutex

### LINKED ISSUE #828: Data race in ioConn.protocolVersion (concurrent read/write without synchronization)
## Summary
A data race exists in `ioConn.protocolVersion` between `sessionUpdated()` and `Read()`, violating Go's memory model and potentially causing undefined behavior.

## Location
`mcp/transport.go`

### The Race Condition

**Write path (unsynchronized):**
- `ServerSession.updateState()` → `ioConn.sessionUpdated()` → writes to `c.protocolVersion` (line 435)

**Read path (unsynchronized):**
- `ioConn.Read()` → reads `t.protocolVersion` (line 536)

### Evidence

The `ioConn` struct has a `writeMu` mutex and `batchMu` mutex for other concurrent access:

```go
type ioConn struct {
    protocolVersion string // ❌ NO SYNCHRONIZATION

    writeMu sync.Mutex         // ✅ guards Write
    // ...
    batchMu sync.Mutex         // ✅ guards batches map
    batches map[jsonrpc2.ID]*msgBatch
    // ...
}
```

The code comment at line 370-371 explicitly acknowledges concurrency:
> "Since writes may be concurrent to reads, we need to guard this with a mutex."

Yet `protocolVersion` is accessed without any synchronization.

### Impact

1. **Go memory model violation**: Concurrent read/write of non-atomic variables without synchronization is undefined behavior.
2. **Potential corruption**: String reads/writes in Go are not atomic; concurrent access can yield corrupted/torn values.
3. **Race detector alerts**: Running with `-race` flag will catch this.

### Reproduction

The race occurs when:
1. A server calls `updateState()` (which triggers `sessionUpdated()`)
2. Concurrently, `Read()` checks `protocolVersion` for batch validation

This is realistic because the Connection interface explicitly allows concurrent `Read()` and `Close()` (per interface docs).

## Proposed Fix

Option 1: Use `sync/atomic` for `protocolVersion`:

```go
type ioConn struct {
    protocolVersion atomic.Value // stores string
    // ...
}

func (c *ioConn) sessionUpdated(state ServerSessionState) {
    // ...
    c.protocolVersion.Store(negotiatedVersion(protocolVersion))
}

func (t *ioConn) Read(ctx context.Context) (jsonrpc.Message, error) {
    // ...
    pv := t.protocolVersion.Load().(string)
    if batch && pv >= protocolVersion20250618 {
        return nil, fmt.Errorf("JSON-RPC batching is not supported...")
    }
    // ...
}
```

Option 2: Reuse `batchMu` to guard `protocolVersion` reads/writes (simpler but broader lock scope).

## References
- Go Memory Model: https://go.dev/ref/mem
- Related code comment acknowledging concurrent access: `mcp/transport.go:370-371`

maciej-kisiel:
This is done to prevent concurrent reads and writes, which violate Go's memory model. The concurrent write may be triggered e.g. by setting the log level on
the client side.

Fixes #828

### DIFF
--- a/mcp/transport.go
+++ b/mcp/transport.go
@@ -344,7 +344,11 @@ func (r rwc) Close() error {
 //
 // See [msgBatch] for more discussion of message batching.
 type ioConn struct {
-	protocolVersion string // negotiated version, set during session initialization.
+	// protocolVersion is the negotiated version of the protocol,
+	// set during session initialization.
+	// Since writes may be concurrent to reads, we need to guard this with a mutex.
+	sessionMu       sync.Mutex
+	protocolVersion string
 
 	writeMu sync.Mutex         // guards Write, which must be concurrency safe.
 	rwc     io.ReadWriteCloser // the underlying stream
@@ -430,9 +434,15 @@ func (c *ioConn) sessionUpdated(state ServerSessionState) {
 		protocolVersion = state.InitializeParams.ProtocolVersion
 	}
 	if protocolVersion == "" {
+		// 2025-03-26 is used, because it's the last spec version
+		// where specifying the protocol version in the HTTP header
+		// was not required.
 		protocolVersion = protocolVersion20250326
 	}
-	c.protocolVersion = negotiatedVersion(protocolVersion)
+	protocolVersion = negotiatedVersion(protocolVersion)
+	c.sessionMu.Lock()
+	c.protocolVersion = protocolVersion
+	c.sessionMu.Unlock()
 }
 
 // addBatch records a msgBatch for an incoming batch payload.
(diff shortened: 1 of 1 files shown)


mcp/transport.go
```diff
@@ -344,7 +344,11 @@ func (r rwc) Close() error {
 //
 // See [msgBatch] for more discussion of message batching.
 type ioConn struct {
-	protocolVersion string // negotiated version, set during session initialization.
+	// protocolVersion is the negotiated version of the protocol,
+	// set during session initialization.
+	// Since writes may be concurrent to reads, we need to guard this with a mutex.
+	sessionMu       sync.Mutex
+	protocolVersion string
```
herczyn, line 351: can't we use some atomic val?

herczyn (approved):

PR #834, opened 2026-03-04
auth: return scope in WWW-Authenticate header.

maciej-kisiel:
Also, emit the resource metadata URL in quotes to ensure proper parsing.

https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#protected-resource-metadata-discovery-requirements

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"net/http"
 	"slices"
 	"strings"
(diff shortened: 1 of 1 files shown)


auth/auth.go
```diff
@@ -73,8 +73,17 @@ func RequireBearerToken(verifier TokenVerifier, opts *RequireBearerTokenOptions)
 			tokenInfo, errmsg, code := verify(r, verifier, opts)
 			if code != 0 {
 				if code == http.StatusUnauthorized || code == http.StatusForbidden {
-					if opts != nil && opts.ResourceMetadataURL != "" {
-						w.Header().Add("WWW-Authenticate", "Bearer resource_metadata="+opts.ResourceMetadataURL)
+					if opts != nil {
+						var params []string
+						if opts.ResourceMetadataURL != "" {
+							params = append(params, "resource_metadata=\""+opts.ResourceMetadataURL+"\"")
```
jba: I think this is clearer: fmt.Sprintf("resource_metadata=%q", opts....)

and ditto below

maciej-kisiel: Good point, for some reason I mixed it up and thought `%q` produces single quotes. Will adjust the code.

PR #838, opened 2026-03-10
Fix SetProgressToken when Meta is nil

eddiearc:
## Summary
- initialize the params meta map before writing `progressToken`
- add a regression test covering `SetProgressToken` on params with nil `Meta`
- assert the progress token is present in marshaled JSON

## Testing
- go test ./mcp

### DIFF
--- a/mcp/protocol_test.go
+++ b/mcp/protocol_test.go
@@ -63,6 +63,22 @@ func TestParamsMeta(t *testing.T) {
 	if g := p.GetProgressToken(); g != "t" {
 		t.Errorf("got %v, want `t`", g)
 	}
+	if got := p.GetMeta()["progressToken"]; got != "t" {
+		t.Errorf("got %v, want `t`", got)
+	}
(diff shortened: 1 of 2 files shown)


mcp/protocol_test.go
```diff
@@ -63,6 +63,22 @@ func TestParamsMeta(t *testing.T) {
 	if g := p.GetProgressToken(); g != "t" {
 		t.Errorf("got %v, want `t`", g)
 	}
+	if got := p.GetMeta()["progressToken"]; got != "t" {
+		t.Errorf("got %v, want `t`", got)
+	}
+	if g, w := toJSON(p), `{"_meta":{"progressToken":"t","x":2},"name":"name"}`; g != w && g != `{"_meta":{"x":2,"progressToken":"t"},"name":"name"}` {
```
maciej-kisiel, line 69: I don't think we need these additional marshaling tests. Marshaling is tested in line 37. I would remove them (line 69 and 79).

maciej-kisiel:
Merged in #846.

PR #840, opened 2026-03-12
auth: allow passing custom http.Client to AuthorizationCodeHandler

maciej-kisiel:


auth/authorization_code_test.go
```diff
@@ -125,8 +125,11 @@ func TestAuthorize_ForbiddenUnhandledError(t *testing.T) {
 		"WWW-Authenticate",
 		"Bearer error=invalid_token",
 	)
-	handler := &AuthorizationCodeHandler{} // No config needed for this test.
-	err := handler.Authorize(t.Context(), req, resp)
+	handler, err := NewAuthorizationCodeHandler(validConfig())
```
jba, line 128: Why is this changed?

auth/authorization_code_test.go
```diff
@@ -324,7 +318,11 @@ func TestNewAuthorizationCodeHandler_Error(t *testing.T) {
 }
 
 func TestGetProtectedResourceMetadata_Success(t *testing.T) {
-	handler := &AuthorizationCodeHandler{} // No config needed for this method
+	handler, err := NewAuthorizationCodeHandler(validConfig())
+	if err != nil {
```
jba, line 322: Ditto

auth/authorization_code_test.go
maciej-kisiel: Zero value of `AuthorizationCodeHandler` is not usable anymore if we want to send any HTTP requests, since we need to initialize the `http.Client`, which happens in the constructor. This prompted me to use the constructor in all tests to "exercise the public API".

jba: Isn't that a breaking behavior change? I'm okay with it if hardly anybody is using the code yet or if it's still under the build tag.

maciej-kisiel: Zero value is not really usable in real world scenarios anyway and would likely result in nil dereferences. The constructor should be used by external users and I brought tests in line with that.

PR #843, opened 2026-03-13
docs: publish user docs with mkdocs

### LINKED ISSUE #835: Configure mkdocs to serve the user documentation
The tool is used by several other SDKs, such as Python or Java.

The content is mostly ready in `docs/`. We shouldn't generate API reference directly in the documentation, rather the documentation should have a navigation link to https://pkg.go.dev/github.com/modelcontextprotocol/go-sdk@${VERSION}.

CMLKevin:
## Summary
- add `mkdocs.yml` so the docs directory builds as a MkDocs site
- update the publish-docs workflow to build and publish `./site` instead of pushing raw `./docs`
- add `docs/index.md` as the user-docs homepage with a pkg.go.dev API reference link

## Testing
- `python3 -m venv .tmp-mkdocs-venv && . .tmp-mkdocs-venv/bin/activate && python -m pip install mkdocs && mkdocs build`
- `git diff --check`

Fixes #835

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -1,0 +1,21 @@
+site_name: MCP Go SDK
+site_description: User documentation for the Model Context Protocol Go SDK
+repo_url: https://github.com/modelcontextprotocol/go-sdk
+edit_uri: ''
+docs_dir: docs
+site_dir: site
+exclude_docs: |
+  index.html
+nav:
+  - Home: index.md
+  - Protocol: protocol.md
+  - Client features: client.md
+  - Server features: server.md
+  - Troubleshooting: troubleshooting.md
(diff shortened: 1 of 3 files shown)


mkdocs.yml
maciej-kisiel, line 1: Could you take a look at the files defined for Python and Java:

https://github.com/modelcontextprotocol/python-sdk/blob/main/mkdocs.yml
https://github.com/modelcontextprotocol/java-sdk/blob/main/mkdocs.yml

We don't have to replicate every customization they have, but they should contain a good baseline for the configuration file – especially look for things that are defined in both.

maciej-kisiel:
@CMLKevin Let me know if you're planning to work on this PR.

PR #844, opened 2026-03-15
feat: implement generic SendNotification method for Issue #745

### LINKED ISSUE #745: Expose generic `SendNotification` on `ServerSession` for custom protocol extensions
### Use Case
I am implementing an MCP Server for the [Gemini CLI IDE Integration](https://github.com/google-gemini/gemini-cli/blob/main/docs/ide-integration/ide-companion-spec.md). The specification requires the server to send **custom notifications** to the client to signal specific events.

Specifically, I need to send:
- `ide/diffAccepted`
- `ide/diffRejected`

### The Problem
Currently, the `mcp.ServerSession` struct only exposes specific notification helpers (like `NotifyProgress` and `Log`). There is no public API to send a generic/custom JSON-RPC notification.

I attempted to use `session.Notify(ctx, method, params)`, but:
1. The `conn` field on `ServerSession` is unexported.
2. The `mcp.Connection` interface's `Write` method requires `*mcp.JSONRPCMessage`, which is an unexported struct.

### Proposed Solution
Please expose a generic notification method on `ServerSession`, similar to:

```go
// In mcp/server.go

func (s *ServerSession) SendNotification(ctx context.Context, method string, params any) error {
    return s.conn.Notify(ctx, method, params)
}
```

### Current Workaround
I am currently forced to use `unsafe` and `reflect` to access the underlying `conn` field to call its `Notify` method, which is fragile for obvious reasons.

ajuijas:
The Model Context Protocol (MCP) Go SDK's middleware stack currently validates outbound notifications against a predefined set of protocol methods. This behavior prevents protocol extensions (such as `ide/diffAccepted` or other custom vendor-specific notifications) from being sent successfully, as they are rejected by the internal method validation logic.

This change modifies the `mcp` package to support sending arbitrary JSON-RPC notifications by exposing a generic SendNotification method on both ServerSession and ClientSession.

#### Proposed Changes:
- **Middleware Bypass:** Implemented a strategy in defaultSendingMethodHandler that intercepts an internal `x-notifications/` prefix, allowing custom methods to bypass strict validation checks while still utilizing the SDK's underlying connection and middleware flow.
- **Generic Parameter Wrapper:** Added customNotificationParams to safely wrap `any` payload types, ensuring they satisfy the internal Params interface and serialize correctly to JSON.
- **Session API Extensions:** Exposed SendNotification(ctx, method, params) on both ClientSession and ServerSession.
- **Testing:** Added integration tests in server_test.go and client_test.go to verify end-to-end notification delivery for both server-to-client and client-to-server flows.

Fixes #745

### DIFF
--- a/mcp/server.go
+++ b/mcp/server.go
@@ -1079,6 +1079,14 @@ func (ss *ServerSession) NotifyProgress(ctx context.Context, params *ProgressNot
 	return handleNotify(ctx, notificationProgress, newServerRequest(ss, orZero[Params](params)))
 }
 
+// SendNotification sends a custom notification from the server to the client
+// associated with this session. This is typically used for protocol extensions
+// that send arbitrary JSON-RPC notifications.
+func (ss *ServerSession) SendNotification(ctx context.Context, method string, params any) error {
+	cp := &customNotificationParams{payload: params}
+	return handleNotify(ctx, "x-notifications/"+method, newServerRequest(ss, Params(cp)))
+}
+
 func newServerRequest[P Params](ss *ServerSession, params P) *ServerRequest[P] {
 	return &ServerRequest[P]{Session: ss, Params: params}
 }
--- a/mcp/shared.go
+++ b/mcp/shared.go
@@ -90,6 +90,12 @@ func addMiddleware(handlerp *MethodHandler, middleware []Middleware) {
 }
 
 func defaultSendingMethodHandler(ctx context.Context, method string, req Request) (Result, error) {
+	// Custom notifications from SendNotification are prefixed with x-notifications/
+	if strings.HasPrefix(method, "x-notifications/") {
+		actualMethod := strings.TrimPrefix(method, "x-notifications/")
+		return nil, req.GetSession().getConn().Notify(ctx, actualMethod, req.GetParams())
+	}
+
 	info, ok := req.GetSession().sendingMethodInfos()[method]
 	if !ok {
 		// This can be called from user code, with an arbitrary value for method.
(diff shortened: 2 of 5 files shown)


maciej-kisiel:
Hi @ajuijas, the issue #745 hasn't been marked as "ready for work" yet, since the decision that we want or need this functionality hasn't been made yet. Do you have any personal use cases that would require this functionality? If so, I would encourage you to add to the discussion on the issue first, before we dive into implementation. Thanks!

ajuijas:
Hi! I don't have a personal use case for this. I was just bored and looking for open source issues to help with, so I decided to try fixing this one (I wasn't paying attention to the "ready for work" tag 😪). 
Please let me know if you decide to move forward with the implementation! Thanks 🙏.

mcp/shared.go
```diff
@@ -90,6 +90,12 @@ func addMiddleware(handlerp *MethodHandler, middleware []Middleware) {
 }
 
 func defaultSendingMethodHandler(ctx context.Context, method string, req Request) (Result, error) {
+	// Custom notifications from SendNotification are prefixed with x-notifications/
+	if strings.HasPrefix(method, "x-notifications/") {
```
jba, line 94: Use strings.CutPrefix

mcp/server.go
```diff
@@ -1079,6 +1079,14 @@ func (ss *ServerSession) NotifyProgress(ctx context.Context, params *ProgressNot
 	return handleNotify(ctx, notificationProgress, newServerRequest(ss, orZero[Params](params)))
 }
 
+// SendNotification sends a custom notification from the server to the client
```
jba, line 1082: Could we call it `Notify`? Shorter and simpler.

jba (changes requested):

PR #846, opened 2026-03-15
mcp: fix setProgressToken when Meta is nil

### LINKED ISSUE #845: `SetProgressToken` silently fails when `Meta` is nil
**Describe the bug**
`SetProgressToken` in `mcp/shared.go` silently fails when called on a `Params` struct with a nil `Meta` field. The progress token is written into a locally created map that is never assigned back to the struct, so it is immediately discarded.

**To Reproduce**
Minimum test case
   ```go
   p := &mcp.CallToolParams{Name: "n"}
   p.SetProgressToken("t")
   fmt.Println(p.GetProgressToken()) // nil
   ```
**Expected behaviour**
GetProgressToken() should return "t" after calling SetProgressToken("t"), regardless of whether Meta was previously set.

StevenRChen:
When `SetProgressToken` is called on a `Params` struct with a nil `Meta`
field, the newly created map is never assigned back to the struct via
`SetMeta`, causing the progress token to be silently lost.

Call `p.SetMeta(m)` after creating the new map so the struct retains the
reference.

Fixes #845

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"net/http"
 	"slices"
 	"strings"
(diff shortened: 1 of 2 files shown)


StevenRChen:
Duplicating with existing PR
https://github.com/modelcontextprotocol/go-sdk/pull/838

maciej-kisiel:
Given that the other PR is not proceeding, we can revive this one. It looks good in its current form.

PR #847, opened 2026-03-16
go-sdk/mcp: add mcp app example

### LINKED ISSUE #815: [Feature Request] Support the MCP Apps extension, enabling MCP servers to show interactive user interfaces
**Edit:** This issue may be unnecessary. While I'm not 100% confident in that, I was able to create an MCP server with a user interface using the Go SDK. I haven't exercised all MCP Apps functionality (eg. my UI is not interactive) but it's possible that also does not require any changes to the Go SDK. 

Original issue below, though it may be obsolete...

---

I would love to add interactive user interfaces to my Go-based MCP servers, but right now it is not possible.

A recent [extension](https://modelcontextprotocol.io/docs/extensions/apps) (last month) to the MCP protocol enables MCP servers to show interactive user interfaces, such as data visualizations and complex configuration workflows, directly in an MCP host.

Basically, the server provides an HTML page to be used as the UI, and the MCP protocol is extended to enable communication between that UI, the MCP server, and the MCP host. The UI can appear eg. directly inside of a chat conversation with an AI model, and can be used to interactively guide the conversation.

Here is the [spec](https://github.com/modelcontextprotocol/ext-apps/blob/main/specification/draft/apps.mdx). There is also an [official blog post](https://blog.modelcontextprotocol.io/posts/2026-01-26-mcp-apps/) introducing the extension, and an [overview](https://modelcontextprotocol.io/docs/extensions/overview) of the concept of extensions, and a [docs page](https://modelcontextprotocol.io/docs/extensions/apps) motivating and explaining the MCP Apps extension.

I would love it if I could add rich user interfaces to my Go-based MCP servers. Right now I use JavaScript when I need to build an MCP server with a UI because there is SDK support there. 

In terms of what would be required, looking at the spec I see there is a section on [Client<>Server Capability Negotiation](https://github.com/modelcontextprotocol/ext-apps/blob/main/specification/draft/apps.mdx#clientserver-capability-negotiation) and another on [Server Behavior](https://github.com/modelcontextprotocol/ext-apps/blob/main/specification/draft/apps.mdx#server-behavior). 

It's hard to say for sure, but it's possible the server support is a relatively small effort, since most of the complexity of the spec is about the client and host.

One alternative is to use JavaScript to build my MCP servers, which I personally find unsatisfying, particularly since it is not possible to do cross-server MCP calls -- meaning that a UI can only programmatically call tools from its own MCP server, and writing the UI in a separate sever/language would be limiting.

Here is a [code snippet](https://modelcontextprotocol.io/docs/extensions/apps#server-implementation) that shows what a JS-based MCP apps server looks like when implemented using the JS SDK.

**Additional context**

Thanks for making this package – getting started is so simple and pleasant, and I'm very excited about being able to use Go for MCP servers! I tried to be as accurate as I can in this request, but please forgive me if I've misunderstood anything about the nature of this spec, or if the request is somehow incoherent -- still learning my way around the various protocols involved. Thanks again!

Edit: Claude Code suggests that a minimal change is needed, since most of the complexity is on the side of the host. Relevant snippet:

>  1. Add Extensions field to ClientCapabilities (protocol.go:192)
> ```
>  type ClientCapabilities struct {
>      // ... existing fields ...
>      Elicitation *ElicitationCapabilities `json:"elicitation,omitempty"`
>      // Extensions describes protocol extensions that the client supports.
>      // Keys are extension URIs (e.g. "io.modelcontextprotocol/ui").
>      Extensions map[string]any `json:"extensions,omitempty"`
>  }
> ```
>  No clone() update needed -- Experimental map[string]any (line 198) uses the same pattern and is not explicitly
>   cloned either (cp := *c copies the map header, matching existing convention).
> 
>  2. Add Extensions field to ServerCapabilities (protocol.go:1300)
>  ```
>  type ServerCapabilities struct {
>      // ... existing fields ...
>      Tools *ToolCapabilities `json:"tools,omitempty"`
>      // Extensions describes protocol extensions that the server supports.
>      // Keys are extension URIs (e.g. "io.modelcontextprotocol/ui").
>      Extensions map[string]any `json:"extensions,omitempty"`
>  }
> ```
> Same clone story -- matches Experimental.
>
> [...]
>
>  Once the field exists, a Go MCP server can:
>  1. Set ServerOptions.Capabilities.Extensions["io.modelcontextprotocol/ui"] = map[string]any{}
>  2. Set Tool.Meta["ui"] = map[string]any{"resourceUri": "ui://server/app.html"} on tools
>  3. Register a resource at ui://server/app.html with MIME type text/html;profile=mcp-app
>  4. Check client support via session.InitializeParams().Capabilities.Extensions["io.modelcontextprotocol/ui"]
> 
>  All of this uses existing SDK primitives -- only the Extensions field is missing today.

yurivish:
Adds a simple example of an MCP server that can display a user interface using the MCP app extension.

Fixes https://github.com/modelcontextprotocol/go-sdk/issues/815#issuecomment-4030150182

The tricky / somewhat inelegant bit is that the JS for the MCP Apps SDK is inlined into the HTML; it is possible to load other resources but they must be served from a web server, so I opted for simplicity to embed the required JS in `index.html`. Thinking about it more, this may not have been a great choice for an example. In my slightly more real-world example, I actually bundle and inline the code dynamically using [esbuild](https://esbuild.github.io/)'s Go API.


maciej-kisiel:
Thanks for the contribution @yurivish! I'm happy that you found some time to show how MCP Apps can be used to others.

I agree with the concern about the embedded JS code. I'm not very familiar with front-end development, but it would be great if there was a way to not put all the code in the example. Isn't the Apps SDK hosted somewhere so that it could be included as an external script?

I think the example could also use a small `README.md` to explain how to run it. I assume this is supposed to show a webpage, but it's not clear how open it given that the server uses STDIO transport.

There are also some simple fixes needed as uncovered by the test runs.

maciej-kisiel:
Hi @yurivish, I wanted to ask if you're planning to follow up on this PR. Good day!

yurivish:
Hi @maciej-kisiel Unfortunately I have too many open threads to pull on at the moment; I keep meaning to revisit this but not having the time. It looks like the best thing to do would be whitelist the externally-hosted JS MCP Apps SDK and load it from there with a `<script>` tag, and to add a short readme explaining what the demonstration demonstrates and how to run it, but at the moment I have not been able to get to it with everything else going on in my life. Sorry to disappoint, and thanks again for your help earlier; there is still a chance I can do this but it would be in a while. But feel free to close/take over if that's best for the project.

maciej-kisiel:
No worries, thanks for the reply. I will keep this open for now, let's see who will find time to finalize this earlier :)

yarolegovich:
closing because of inactivity

PR #852, opened 2026-03-23
RecoverExpiredSessions option for graceful session recovery

lawrencejones:
Servers that store MCP sessions in memory lose all sessions on restart. When a client presents a stale session ID, the server currently returns 404 Not Found per spec §2.5.3. Many MCP clients treat this as a terminal error rather than re-initializing, causing a poor experience during rolling deploys.

RecoverExpiredSessions adds a new StreamableHTTPOptions field that, when true, silently creates a new stateful session with default initialization state instead of returning 404. The new session gets a fresh ID (returned via the Mcp-Session-Id response header) and is stored in the session map for reuse.

The recovery reuses the same bootstrap path that stateless mode already uses: it peeks at the request body to detect initialize/initialized messages and sets default InitializeParams and InitializedParams when they are absent. The key difference from stateless mode is that the recovered session is persistent — it gets an onClose handler and is stored in the session map, so subsequent requests reuse it.


CaliLuke:
we need this badly please merge this feature ASAP!

lawrencejones:
I should explain that we have used this for the incident.io MCP as Claude Code behaves badly if the session gets terminated, even when we're not using the session features across multiple tool calls.

Claude Code should handle the 404 missing session error transparently (imo) but right now it breaks and refuses to execute subsequent tool calls until you restart. This allows us to move over it, with (I think?) no negative consequences to well behaved clients or servers, and is opt-in.

maciej-kisiel:
Hi, thanks for the report. In general, I'm not very keen on adding APIs to well-behaved servers just because one of the clients is misbehaving. Have you tried reporting the issue to Claude Code? I see that https://github.com/anthropics/claude-code/issues/30224 already exists, maybe you can follow up there?

lawrencejones:
@maciej-kisiel sorry I've been on PTO but yes, I agree with this. We'll merge this into our fork for now (practically we can't have Claude Code failing against our MCP) but will chase in the official library.

Thank you!

maciej-kisiel:
Thanks! I assume the PR may be closed then.

PR #885, opened 2026-04-08
mcp: do not re-prompt OAuth after cancelled Authorize

### LINKED ISSUE #882: mcp/transport.go call() attempts to call Notify() on a failed connection
**Describe the bug**
I am using `mcp.StreamableClientTransport` with an OAuthHandler. As documented, this handler will automatically be invoked by `go-sdk` on e.g. a 401 from the MCP endpoint.

My client implement an interactive OAuth2 authorization flow using the newly exposed APIs in v1.5.0-pre.

The behaviour I am seeing that I believe to be a bug is this:

1. I configure a new MCP server endpoint in my client config.
2. My client tries to use the MCP server via the `StreamableClientTransport`.
3. It receives a 401 and calls `Authorize()` on my `OAuthHandler`.
4. The context passed to `Authorize()` times out.
5. In the switch statement in `call()`, the execution flow enters the case where `ctx.Err() != nil`. Here, `conn.Notify` is called. This `Notify` call, of course, receives a 401 which then calls `Authorize()` on my handler again.

In the case of my client, this prompts the user for interactive authorization again, even though the timeout has already cancelled the authorization request.

Relevant code is here:

https://github.com/modelcontextprotocol/go-sdk/blob/862d78a265cb1f1b9692ff49cfb8bfd109b6ebc9/mcp/transport.go#L213-L223

**To Reproduce**
1. Set up a `StreamableClientTransport` with an `OAuthHandler`.
2. Use the transport without setting up authorization.
3. Pass a context with a short timeout to `Authorize()`, and wait for it to expire before returning.

**Expected behavior**
I would expect cancelling the context passed to `Authorize()` to indicate that the connection should be abandoned and not to use it for any further communication.

**Logs**
n/a

**Additional context**
n/a

ravyg:
When `OAuthHandler.Authorize` returns a context-cancelled error, the cancellation notification that the call layer sends in response to the same ctx cancellation is routed back through the streamable client transport, re-triggering the `401` → `Authorize` path and prompting the user a second time for a request they have already abandoned.

Fix in two layers:

1. **`mcp/transport.go`** — bound the cancellation `Notify` with a fresh `context.Background()`-derived timeout instead of reusing the cancelled caller context. This guarantees the caller never blocks on a doomed notify against a degraded connection.
2. **`mcp/streamable.go`** — when `Authorize` returns a context-cancelled error, mark the streamable client connection as failed via `c.fail()`. The existing failure check in `Write()` then short-circuits the cancellation notification without re-invoking the OAuth handler, matching the bug reporter's expectation that cancelling `Authorize` abandons the connection.

Includes a regression test (`TestStreamableClientOAuth_CancelledAuthorize_NoReprompt`) with a blocking `OAuthHandler` that asserts `Authorize` is invoked exactly once. The test was verified to fail on unfixed code (`got 2` calls) and pass with the fix.

Fixes #882

### DIFF
--- a/mcp/streamable.go
+++ b/mcp/streamable.go
@@ -1814,6 +1814,15 @@ func (c *streamableClientConn) Write(ctx context.Context, msg jsonrpc.Message) e
 
 	if (resp.StatusCode == http.StatusUnauthorized || resp.StatusCode == http.StatusForbidden) && c.oauthHandler != nil {
 		if err := c.oauthHandler.Authorize(ctx, req, resp); err != nil {
+			// If the caller's context was cancelled while we were running the
+			// authorization flow, treat the connection as failed so subsequent
+			// operations on it (e.g. the cancellation notify the call layer
+			// sends in response to ctx cancellation) short-circuit instead of
+			// re-invoking the OAuth handler. Otherwise the user gets prompted
+			// to authorize a request they have already abandoned. See #882.
+			if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
+				c.fail(fmt.Errorf("%s: authorization cancelled: %w", requestSummary, err))
+			}
 			// Wrap with ErrRejected so the jsonrpc2 connection doesn't set writeErr
 			// and permanently break the connection.
 			// Wrap the authorization error as well for client inspection.
--- a/mcp/streamable.go
+++ b/mcp/streamable.go
@@ -1820,7 +1820,14 @@ func (c *streamableClientConn) Write(ctx context.Context, msg jsonrpc.Message) e
 			// sends in response to ctx cancellation) short-circuit instead of
 			// re-invoking the OAuth handler. Otherwise the user gets prompted
 			// to authorize a request they have already abandoned. See #882.
-			if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
+			//
+			// We check ctx.Err() rather than the error returned by Authorize,
+			// because the handler is user-implemented and may return an error
+			// that does not wrap context.Canceled (e.g. a custom sentinel or
+			// a fmt.Errorf with %v). The context itself is the authoritative
+			// source for whether the caller abandoned the request.
+			ctxErr := ctx.Err()
+			if errors.Is(ctxErr, context.Canceled) || errors.Is(ctxErr, context.DeadlineExceeded) {
 				c.fail(fmt.Errorf("%s: authorization cancelled: %w", requestSummary, err))
 			}
 			// Wrap with ErrRejected so the jsonrpc2 connection doesn't set writeErr
--- a/mcp/transport.go
+++ b/mcp/transport.go
@@ -14,12 +14,19 @@ import (
 	"net"
 	"os"
 	"sync"
+	"time"
 
 	internaljson "github.com/modelcontextprotocol/go-sdk/internal/json"
 	"github.com/modelcontextprotocol/go-sdk/internal/jsonrpc2"
-	"github.com/modelcontextprotocol/go-sdk/internal/xcontext"
 	"github.com/modelcontextprotocol/go-sdk/jsonrpc"
 )
+
+// notifyCancellationTimeout bounds the cancellation notification we send to
+// the peer when the caller's context is cancelled. The notification is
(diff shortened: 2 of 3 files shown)


mcp/transport.go
```diff
@@ -216,8 +223,19 @@ func call(ctx context.Context, conn *jsonrpc2.Connection, method string, params
...
+		// caller's already-cancelled context. The connection may be in a
+		// degraded state — for example, the original failure may have come
+		// from an OAuth flow whose handler context expired (see #882) — and
+		// reusing that context would either return immediately with an error
+		// or, worse, re-trigger expensive recovery (re-auth) on the caller's
+		// return path. The bounded background context lets the notification
+		// attempt to deliver but never blocks the caller indefinitely.
+		notifyCtx, cancelNotify := context.WithTimeout(context.Background(), notifyCancellationTimeout)
+		defer cancelNotify()
+		err := conn.Notify(notifyCtx, notificationCancelled, &CancelledParams{
```
jba, line 232: Won't this still trigger a second auth flow?

ravyg: On its own, yes, it would, but the companion change in `mcp/streamable.go` (calling `c.fail()` when `Authorize` returns a `cancelled error`) is what actually prevents the second auth flow.

mcp/streamable.go
```diff
@@ -1801,6 +1801,15 @@ func (c *streamableClientConn) Write(ctx context.Context, msg jsonrpc.Message) e
 
 	if (resp.StatusCode == http.StatusUnauthorized || resp.StatusCode == http.StatusForbidden) && c.oauthHandler != nil {
 		if err := c.oauthHandler.Authorize(ctx, req, resp); err != nil {
+			// If the caller's context was cancelled while we were running the
+			// authorization flow, treat the connection as failed so subsequent
+			// operations on it (e.g. the cancellation notify the call layer
+			// sends in response to ctx cancellation) short-circuit instead of
+			// re-invoking the OAuth handler. Otherwise the user gets prompted
+			// to authorize a request they have already abandoned. See #882.
+			if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
```
smlx: This PR doesn't fix the bug I'm seeing and I think it is because this logic is not quite right.

You shouldn't check the error returned from `Authorize()` because it could be anything. The error to inspect is the one returned from the context passed into `Authorize()`

```suggestion
			ctxErr := ctx.Err()
			if errors.Is(ctxErr, context.Canceled) || errors.Is(ctxErr, context.DeadlineExceeded) {
```

I tested this patch on top of your PR changes and it fixes the bug for me :-)

smlx (changes requested):

ravyg (commented):
Good catch @smlx — you're right, the handler is user-implemented and may not wrap `context.Canceled`. Pushed the fix in `8d130aa` checking `ctx.Err()` directly per your suggestion, and updated the test mock to return a non-wrapping error so it exercises the real-world path. Thanks for testing against your setup!

Going to spend some more time today (PST) diving deeper into this area as well.

ravyg:
Good catch @smlx — you're right, the handler is user-implemented and may not wrap `context.Canceled`. Pushed the fix in `8d130aa` checking `ctx.Err()` directly per your suggestion, and updated the test mock to return a non-wrapping error so it exercises the real-world path. Thanks for testing against your setup!

Going to spend some more time today (PST) diving deeper into this area as well.

mcp/transport.go
jba: Doesn't `xcontext.Detach` accomplish the same thing?

smlx (approved):
Fixes the bug I was seeing. Thanks!

mcp/transport.go
maciej-kisiel: Please use `xcontext.Detach` instead of `context.Background` when building the context. This will preserve the context values.

maciej-kisiel: Or event better: use https://pkg.go.dev/context#WithoutCancel.

maciej-kisiel:
@ravyg We're one small adjustment from accepting and merging this. Will you have time to take a look in the nearest future?

ravyg:
Yes, just saw your comments above, adding those, I totally missed those earlier, thanks for pointing out.

maciej-kisiel (approved):

PR #890, opened 2026-04-09
 mcp: accept parameterized Content-Type types 

kalvinnchau:
Use a shared helper for Content-Type parsing in streamable transport request validation and client response handling.

Follow up to https://github.com/modelcontextprotocol/go-sdk/pull/853, where we did this for `Accept` headers.

This allows the streamable mcp server to validate the following header `Content-Type: application/json;charset=utf-8`

### DIFF
--- a/mcp/streamable.go
+++ b/mcp/streamable.go
@@ -264,12 +264,9 @@ func (h *StreamableHTTPHandler) ServeHTTP(w http.ResponseWriter, req *http.Reque
 			return
 		}
 		// Validate 'Content-Type' header.
-		if req.Method == http.MethodPost {
-			mediaType, _, err := mime.ParseMediaType(req.Header.Get("Content-Type"))
-			if err != nil || mediaType != "application/json" {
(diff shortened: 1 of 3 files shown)


atishpatel (approved):

mcp/streamable.go
```diff
@@ -1873,7 +1877,7 @@ func (c *streamableClientConn) Write(ctx context.Context, msg jsonrpc.Message) e
 
 	default:
 		resp.Body.Close()
-		return fmt.Errorf("%s: unsupported content type %q", requestSummary, contentType)
+		return fmt.Errorf("%s: unsupported content type %q", requestSummary, resp.Header.Get("Content-Type"))
```
jba: should be baseMediaType.
So put it in a variable before the switch.

jba (changes requested):

mcp/streamable.go
kalvinnchau: thanks for the review, updated and rebased into a single commit

maciej-kisiel (approved):

jba (approved):

Issue #627, opened 2025-11-07
ext-auth: OAuth Client Credentials


findleyr:
Investigate if we can/should add support for [SEP-1046](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1046), an authorization extension. It has been finalized in https://github.com/modelcontextprotocol/ext-auth/blob/main/specification/draft/oauth-client-credentials.mdx.

jba:
This extension fits into our general framework, #591.

Our default implementation for that framework should support client ID and client secret as options, to handle that variant. Those were mentioned in my original design,#255. If those are set, our implementation would bypass dynamic client reservation. That is just a couple of lines of code, I believe.

We could also provide an implementation of [JWT Assertions](https://datatracker.ietf.org/doc/html/rfc7523). I read that RFC a couple of times and it's still not clear to me what the API for that would look like. Perhaps we can take our lead from other SDKs, such as TypeScript.

maciej-kisiel:
This should wait until client side OAuth scaffolding is ready (#785).

brkane:
Looks like the scaffolding is done, any chance of getting this prioritized?

ravyg:
we also need this for our MCP usecases at BigCommerce.

ravyg:
Thanks for the review I ll get to this in evening today!

PR #895, opened 2026-04-14
extauth: add ClientCredentialsHandler for OAuth client credentials grant

ravyg:
## Summary

- Add `ClientCredentialsHandler` implementing `auth.OAuthHandler` using the OAuth 2.0 Client Credentials grant (RFC 6749 Section 4.4) for service-to-service authentication with pre-registered credentials.
- Bypasses both dynamic client registration and the authorization code flow — the handler takes a client ID + secret directly and exchanges them at the token endpoint.
- Supports two modes: direct `TokenEndpoint` URL, or metadata discovery via `AuthServerURL` (RFC 8414).
- Add `client_credentials` grant type support to the fake authorization server in `internal/oauthtest` for testing.

## Context

Per @jba's [comment on #627](https://github.com/modelcontextprotocol/go-sdk/issues/627#issuecomment-3640982691):

> Our default implementation for that framework should support client ID and client secret as options, to handle that variant. [...] If those are set, our implementation would bypass dynamic client reservation. That is just a couple of lines of code, I believe.

The client-side OAuth scaffolding (#785) that this was blocked on is now complete, as @brkane noted.

Implements the client credentials variant of [SEP-1046](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1046). The JWT Assertions variant (RFC 7523) is left for a follow-up as its API surface needs more design discussion.

## Test plan

- [x] `TestNewClientCredentialsHandler_Validation` — validates all config error cases (nil config, missing fields, mutual exclusivity)
- [x] `TestClientCredentialsHandler_Authorize/direct_token_endpoint` — end-to-end with fake auth server
- [x] `TestClientCredentialsHandler_Authorize/metadata_discovery` — discovers token endpoint via RFC 8414 metadata
- [x] `TestClientCredentialsHandler_Authorize/bad_credentials` — verifies failure with wrong secret
- [x] `go test ./... -count=1` passes
- [x] `go vet ./...` clean

Refs #627

### DIFF
--- a/auth/extauth/client_credentials.go
+++ b/auth/extauth/client_credentials.go
@@ -1,0 +1,251 @@
+// Copyright 2026 The Go MCP SDK Authors. All rights reserved.
+// Use of this source code is governed by the license
+// that can be found in the LICENSE file.
+
+package extauth
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"slices"
+	"strings"
+
+	"github.com/modelcontextprotocol/go-sdk/auth"
+	"github.com/modelcontextprotocol/go-sdk/oauthex"
+	"golang.org/x/oauth2"
+	"golang.org/x/oauth2/clientcredentials"
+)
+
+// ClientCredentialsHandlerConfig is the configuration for [ClientCredentialsHandler].
+type ClientCredentialsHandlerConfig struct {
+	// Credentials contains the pre-registered client ID and secret.
+	// REQUIRED. Both ClientID and ClientSecretAuth must be set, since the
+	// client credentials grant requires a confidential client.
+	Credentials *oauthex.ClientCredentials
+
+	// HTTPClient is an optional HTTP client for customization.
+	// If nil, http.DefaultClient is used.
+	// OPTIONAL.
+	HTTPClient *http.Client
+}
+
+// ClientCredentialsHandler is an implementation of [auth.OAuthHandler] that
+// uses the OAuth 2.0 Client Credentials grant (RFC 6749 Section 4.4) to
+// obtain access tokens.
+//
+// This handler is intended for service-to-service authentication where the
+// client has pre-registered credentials (client ID and secret) and does not
+// require user interaction. It bypasses both dynamic client registration and
+// the authorization code flow.
+//
+// The token endpoint and scopes are discovered automatically via Protected
+// Resource Metadata (RFC 9728) and Authorization Server Metadata (RFC 8414),
+// following the ext-auth specification SEP-1046.
+type ClientCredentialsHandler struct {
+	config      *ClientCredentialsHandlerConfig
+	tokenSource oauth2.TokenSource
+}
+
+// Compile-time check that ClientCredentialsHandler implements auth.OAuthHandler.
+var _ auth.OAuthHandler = (*ClientCredentialsHandler)(nil)
+
+// NewClientCredentialsHandler creates a new ClientCredentialsHandler.
+// It validates the configuration and returns an error if invalid.
+func NewClientCredentialsHandler(config *ClientCredentialsHandlerConfig) (*ClientCredentialsHandler, error) {
+	if config == nil {
+		return nil, fmt.Errorf("config must be provided")
+	}
+	if config.Credentials == nil {
+		return nil, fmt.Errorf("credentials are required")
+	}
+	if err := config.Credentials.Validate(); err != nil {
+		return nil, fmt.Errorf("invalid credentials: %w", err)
+	}
+	if config.Credentials.ClientSecretAuth == nil {
+		return nil, fmt.Errorf("clientSecretAuth is required for client credentials grant")
+	}
+	return &ClientCredentialsHandler{config: config}, nil
+}
+
+// TokenSource returns the token source for outgoing requests.
+// Returns nil if authorization has not been performed yet.
+func (h *ClientCredentialsHandler) TokenSource(ctx context.Context) (oauth2.TokenSource, error) {
+	return h.tokenSource, nil
+}
+
+// Authorize performs the Client Credentials grant to obtain an access token.
+// It is called when a request fails with 401 or 403.
+//
+// The flow follows the ext-auth specification SEP-1046:
+//  1. Discover Protected Resource Metadata from the request URL
+//  2. Discover Authorization Server Metadata from PRM
+//  3. Exchange client credentials for an access token at the token endpoint
+func (h *ClientCredentialsHandler) Authorize(ctx context.Context, req *http.Request, resp *http.Response) error {
+	defer resp.Body.Close()
+	defer io.Copy(io.Discard, resp.Body)
+
+	httpClient := h.config.HTTPClient
+	if httpClient == nil {
+		httpClient = http.DefaultClient
+	}
+
+	// Step 1: Discover Protected Resource Metadata.
+	wwwChallenges, err := oauthex.ParseWWWAuthenticate(resp.Header[http.CanonicalHeaderKey("WWW-Authenticate")])
+	if err != nil {
+		return fmt.Errorf("failed to parse WWW-Authenticate header: %v", err)
+	}
+
+	prm, err := getProtectedResourceMetadata(ctx, wwwChallenges, req.URL.String(), httpClient)
+	if err != nil {
+		return err
+	}
+
+	if len(prm.AuthorizationServers) == 0 {
+		return fmt.Errorf("protected resource metadata has no authorization servers specified")
+	}
+
+	// Step 2: Discover Authorization Server Metadata.
+	asm, err := auth.GetAuthServerMetadata(ctx, prm.AuthorizationServers[0], httpClient)
+	if err != nil {
+		return fmt.Errorf("failed to get authorization server metadata: %w", err)
+	}
+	if asm == nil {
+		// Fallback to 2025-03-26 spec: predefined endpoints.
+		authServerURL := prm.AuthorizationServers[0]
+		asm = &oauthex.AuthServerMeta{
+			Issuer:        authServerURL,
+			TokenEndpoint: authServerURL + "/token",
+		}
+	}
+
+	// Determine scopes: use PRM's scopes_supported if available.
+	scopes := scopesFromChallenges(wwwChallenges)
+	if len(scopes) == 0 && len(prm.ScopesSupported) > 0 {
+		scopes = prm.ScopesSupported
+	}
+
+	// Step 3: Exchange client credentials for an access token.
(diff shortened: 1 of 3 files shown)


auth/extauth/client_credentials.go
```diff
@@ -0,0 +1,153 @@
+// Copyright 2026 The Go MCP SDK Authors. All rights reserved.
+// Use of this source code is governed by an MIT-style
+// license that can be found in the LICENSE file.
```
maciej-kisiel:
```suggestion
// Use of this source code is governed by the license
// that can be found in the LICENSE file.
```

Let's make sure we use the new version of the header in newly added files.

auth/extauth/client_credentials.go
```diff
@@ -0,0 +1,153 @@
...
+var _ auth.OAuthHandler = (*ClientCredentialsHandler)(nil)
+
+// NewClientCredentialsHandler creates a new ClientCredentialsHandler.
+// It validates the configuration and returns an error if invalid.
+func NewClientCredentialsHandler(config *ClientCredentialsHandlerConfig) (*ClientCredentialsHandler, error) {
+	if config == nil {
+		return nil, fmt.Errorf("config must be provided")
+	}
+	if config.Credentials == nil {
+		return nil, fmt.Errorf("credentials is required")
```
maciej-kisiel:
```suggestion
		return nil, fmt.Errorf("credentials are required")
```

auth/extauth/client_credentials.go
```diff
@@ -0,0 +1,153 @@
...
+func (h *ClientCredentialsHandler) Authorize(ctx context.Context, req *http.Request, resp *http.Response) error {
+	defer resp.Body.Close()
+	defer io.Copy(io.Discard, resp.Body)
+
+	httpClient := h.config.HTTPClient
+	if httpClient == nil {
+		httpClient = http.DefaultClient
+	}
+
+	tokenEndpoint := h.config.TokenEndpoint
```
maciej-kisiel: When reading the specification for this method, my understanding is that it extends the general Authorization guidance from MCP. This would mean that we shouldn't have `TokenEndpoint` or `AuthServerURL`, but rather use the request URL to fetch PRM and then ASM like in `auth/authorization_code.go. The sequence diagram there confirms that fetching PRM is expected: https://github.com/modelcontextprotocol/ext-auth/blob/main/specification/draft/oauth-client-credentials.mdx#flow-steps.

In that scenario scopes are also coming from PRM and don't need to be specified explicitly on the handler config.

auth/extauth/client_credentials.go
```diff
@@ -0,0 +1,153 @@
...
+			return fmt.Errorf("no authorization server metadata found for: %s", h.config.AuthServerURL)
+		}
+		tokenEndpoint = meta.TokenEndpoint
+		if tokenEndpoint == "" {
+			return fmt.Errorf("token_endpoint not found in authorization server metadata")
+		}
+	}
+
+	creds := h.config.Credentials
+	cfg := &clientcredentials.Config{
```
maciej-kisiel, line 129: We should set `AuthStyle` based on ASM's `token_endpoint_auth_methods_supported` and our priority order (`client_secret_post` is preferred over `client_secret_basic`).

auth/extauth/client_credentials.go
```diff
@@ -0,0 +1,153 @@
...
+		return fmt.Errorf("client credentials token request failed: %w", err)
+	}
+	return nil
+}
+
+// AuthStyle returns the recommended auth style based on a list of supported
+// token endpoint auth methods from the authorization server metadata.
+// Returns [oauth2.AuthStyleInHeader] (HTTP Basic) by default, which is the
+// recommended method per RFC 6749 Section 2.3.1.
+func AuthStyle(methods []string) oauth2.AuthStyle {
```
maciej-kisiel: I assume you forgot to use this function to set the style in the config. Either way, please un-export this function and take a look at a similar logic we've introduced for authorization code handler. I believe it's slightly more correct.

https://github.com/modelcontextprotocol/go-sdk/blob/15e93a231c8693d582ea28f3717bca307a0cf318/auth/authorization_code.go#L369-L396

auth/extauth/client_credentials_test.go
```diff
@@ -0,0 +1,256 @@
...
+
+func TestNewClientCredentialsHandler_Validation(t *testing.T) {
+	tests := []struct {
+		name      string
+		config    *ClientCredentialsHandlerConfig
+		wantError string
+	}{
+		{
+			name:      "nil config",
+			config:    nil,
```
maciej-kisiel: Please take a look how the corresponding test cases are built for `EnterpriseHandler`: there is a reusable `validXConfig` function that produces configuration that is correct and each test case precisely overwrites the part that it targets. You don't have to rely on any ordering of validations with this strategy.

auth/extauth/client_credentials_test.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			_, err := NewClientCredentialsHandler(tc.config)
+			if tc.wantError != "" {
+				if err == nil {
+					t.Fatalf("expected error containing %q, got nil", tc.wantError)
+				}
+				if got := err.Error(); !contains(got, tc.wantError) {
```
maciej-kisiel: Please use `strings.Contains`.

auth/extauth/client_credentials_test.go
```diff
@@ -0,0 +1,256 @@
...
+					t.Fatalf("error %q does not contain %q", got, tc.wantError)
```
maciej-kisiel: Nit: You can use `Errorf` here and in line 104.

auth/extauth/client_credentials_test.go
```diff
@@ -0,0 +1,256 @@
...
+				}
+			} else if err != nil {
+				t.Fatalf("unexpected error: %v", err)
+			}
+		})
+	}
+}
+
+func TestClientCredentialsHandler_Authorize(t *testing.T) {
+	ctx := context.Background()
```
maciej-kisiel: Prefer `t.Context()`.

internal/oauthtest/fake_authorization_server.go
```diff
@@ -57,6 +57,14 @@ type JWTBearerConfig struct {
 	ValidAssertions []string
 }
 
+// ClientCredentialsEnabled enables support for the client_credentials
+// grant type (RFC 6749 Section 4.4) on a [FakeAuthorizationServer].
+// When true, the /token endpoint accepts grant_type=client_credentials
```
maciej-kisiel: This part should be moved to the doc comment of `Enabled`.

maciej-kisiel (changes requested):

auth/extauth/client_credentials.go
```diff
@@ -0,0 +1,237 @@
...
+	u.Path = "/.well-known/oauth-protected-resource"
+	u.RawQuery = ""
+	u.Fragment = ""
+	return u.String()
+}
+
+// scopesFromChallenges extracts scope values from WWW-Authenticate challenges.
+func scopesFromChallenges(cs []oauthex.Challenge) []string {
+	for _, c := range cs {
+		if s := c.Params["scope"]; s != "" {
```
maciej-kisiel: Why the difference in implementation from the `AuthorizationCodeHandler`? I believe we should split scopes into a slice instead of interpreting the space separated list as a single string.

auth/extauth/client_credentials.go
```diff
@@ -0,0 +1,237 @@
...
+		h.tokenSource = nil
+		return fmt.Errorf("client credentials token request failed: %w", err)
+	}
+	return nil
+}
+
+// getProtectedResourceMetadata discovers Protected Resource Metadata (RFC 9728)
+// from the request URL. It tries the resource_metadata URL from WWW-Authenticate
+// challenges first, then falls back to well-known discovery.
+func getProtectedResourceMetadata(ctx context.Context, wwwChallenges []oauthex.Challenge, mcpServerURL string, httpClient *http.Client) (*oauthex.ProtectedResourceMetadata, error) {
```
maciej-kisiel, line 153: Is there a particular reason we're changing the logic compared to the `AuthorizationCodeHandler`? Why not just copy the logic, do we want different behavior?

auth/extauth/client_credentials.go
```diff
@@ -0,0 +1,237 @@
...
+	// Fallback: discover via well-known endpoint at the root.
+	wkURL := wellKnownPRMURL(mcpServerURL)
+	// Use the root URL (without path) as the resource for validation.
+	u, err := url.Parse(mcpServerURL)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse MCP server URL: %v", err)
+	}
+	u.Path = ""
+	rootURL := u.String()
+	prm, err := oauthex.GetProtectedResourceMetadata(
```
maciej-kisiel: Please keep the call on a single line.

auth/extauth/client_credentials_test.go
```diff
@@ -0,0 +1,194 @@
...
+			ClientID:         "test-client",
+			ClientSecretAuth: &oauthex.ClientSecretAuth{ClientSecret: "test-secret"},
+		},
+	}
+}
+
+func TestNewClientCredentialsHandler_Validation(t *testing.T) {
+	tests := []struct {
+		name      string
+		modify    func(*ClientCredentialsHandlerConfig) *ClientCredentialsHandlerConfig
```
maciej-kisiel: This doesn't need to be a function, just the config. You can change the function signature to take 0 arguments, create valid config inside the function and call the function immediately in the test case definition. This way the config creation is local and easy to reason about.

auth/extauth/client_credentials_test.go
```diff
@@ -0,0 +1,194 @@
...
+		},
+		ClientCredentialsConfig: &oauthtest.ClientCredentialsConfig{
+			Enabled: true,
+		},
+	})
+	authServer.Start(t)
+
+	// Set up a fake MCP server that returns PRM pointing to the auth server.
+	mcpMux := http.NewServeMux()
+	mcpMux.HandleFunc("/.well-known/oauth-protected-resource", func(w http.ResponseWriter, r *http.Request) {
```
maciej-kisiel: Please use `auth.ProtectedResourceMetadataHandler` like in the `authorization_code_test.go`.

maciej-kisiel (changes requested):

ravyg:
taking a look

ravyg:
found one issue, I am fixing it.

maciej-kisiel (approved):

maciej-kisiel:
Thanks for contributing!

PR #896, opened 2026-04-14
Update README.md

N-Masi:
correct the default address of the locally hosted server, according to `main.go:21` the default is `:8000` not `:8080`


maciej-kisiel (approved):
Thanks for contributing!

PR #904, opened 2026-04-23
feat: add automatic application_type inference

### LINKED ISSUE #903: Implement SEP-837 Update authorization spec to clarify client type requirements
This issue tracks implementation of https://github.com/modelcontextprotocol/modelcontextprotocol/pull/837.

guglielmo-san:
Implement requirements introduced by [SEP-337](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/837)

Fixes #903

### DIFF
--- a/auth/authorization_code.go
+++ b/auth/authorization_code.go
@@ -35,6 +35,11 @@ type ClientIDMetadataDocumentConfig struct {
 type DynamicClientRegistrationConfig struct {
 	// Metadata to be used in dynamic client registration request as per
 	// https://datatracker.ietf.org/doc/html/rfc7591#section-2.
+	//
+	// If Metadata.ApplicationType is empty, it will be inferred from
+	// Metadata.RedirectURIs. If all valid URIs are compatible (e.g., all loopback
+	// or custom schemes for "native", or all non-loopback HTTP/HTTPS for "web"),
+	// the inferred type will be set. Otherwise, it is left empty.
 	Metadata *oauthex.ClientRegistrationMetadata
 }
 
@@ -152,11 +157,7 @@ func NewAuthorizationCodeHandler(config *AuthorizationCodeHandlerConfig) (*Autho
 			return nil, fmt.Errorf("RedirectURL %q is not in the list of allowed redirect URIs for dynamic client registration", config.RedirectURL)
 		}
 		if dCfg.Metadata.ApplicationType == "" {
-			applicationType, err := inferApplicationType(dCfg.Metadata.RedirectURIs)
-			if err != nil {
-				return nil, fmt.Errorf("failed to infer application type from redirect URIs: %w", err)
-			}
-			dCfg.Metadata.ApplicationType = applicationType
+			dCfg.Metadata.ApplicationType = inferApplicationType(dCfg.Metadata.RedirectURIs)
 		}
 	}
 	if config.RedirectURL == "" {
@@ -179,13 +180,13 @@ func isNonRootHTTPSURL(u string) bool {
 }
 
 // inferApplicationType returns "native" or "web" based on the redirect URIs.
-func inferApplicationType(redirectURIs []string) (string, error) {
+func inferApplicationType(redirectURIs []string) string {
 	hasNative := false
 	hasWeb := false
 	for _, uri := range redirectURIs {
 		u, err := url.Parse(uri)
 		if err != nil {
-			return "", fmt.Errorf("invalid redirect URI %q: %w", uri, err)
+			return ""
 		}
 		switch u.Scheme {
 		case "http", "https":
@@ -200,12 +201,12 @@ func inferApplicationType(redirectURIs []string) (string, error) {
 	}
 
 	if hasNative && hasWeb {
-		return "", errors.New("mixed redirect URI types: found both native and web URIs")
+		return ""
 	}
 	if hasNative {
-		return "native", nil
+		return "native"
 	}
-	return "web", nil
+	return "web"
 }
 
 // Authorize performs the authorization flow.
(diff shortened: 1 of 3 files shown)


oauthex/dcr.go
```diff
@@ -89,6 +89,10 @@ type ClientRegistrationMetadata struct {
 	// SoftwareStatement is an OPTIONAL JWT that asserts client metadata values.
 	// Values in the software statement take precedence over other metadata values.
 	SoftwareStatement string `json:"software_statement,omitempty"`
+
+	// ApplicationType indicates the type of application, valid values are "native" and "web".
```
maciej-kisiel: All the other fields mention whether they are OPTIONAL or REQUIRED. Please also include this information here.

auth/authorization_code.go
```diff
@@ -170,6 +173,31 @@ func isNonRootHTTPSURL(u string) bool {
...
+				return "native"
+			}
+		default:
+			return "native"
+		}
+	}
+	return "web"
+}
+
+func isLoopback(host string) bool {
```
maciej-kisiel: Please use `internal/util/net.go` instead.

auth/authorization_code.go
```diff
@@ -170,6 +173,31 @@ func isNonRootHTTPSURL(u string) bool {
 	return pu.Scheme == "https" && pu.Path != ""
 }
 
+// inferApplicationType returns "native" or "web" based on the redirect URIs.
+// If any redirect URI uses a loopback host or a non-http(s) scheme (custom
+// scheme), the application is classified as "native". Otherwise, it is "web".
+func inferApplicationType(redirectURIs []string) string {
```
maciej-kisiel, line 185: The way I understand `application_type` documentation in https://openid.net/specs/openid-connect-registration-1_0.html suggests that all the redirect URIs should be of the same type. Maybe we should be more strict with the validations? WDYT?

auth/authorization_code.go
```diff
@@ -432,7 +460,21 @@ func (h *AuthorizationCodeHandler) handleRegistration(ctx context.Context, asm *
 	if dcrCfg != nil && asm.RegistrationEndpoint != "" {
 		regResp, err := oauthex.RegisterClient(ctx, asm.RegistrationEndpoint, dcrCfg.Metadata, h.config.Client)
 		if err != nil {
-			return nil, fmt.Errorf("failed to register client: %w", err)
+			// If registration failed due to redirect URI constraints, retry with
```
maciej-kisiel: I'm not sure retrying the registration should be done on this level. I think I would just return the error assuming it has appropriate details coming from the AS.

oauthex/dcr.go
guglielmo-san: Done

auth/authorization_code.go
guglielmo-san: The idea was to leave this final validation to the authorization server, but I guess it also makes sense to enforce early validation and failure.
I changed this function to return error if at least one the URIs is not valid or they are not all of the same type

auth/authorization_code.go
```diff
@@ -150,6 +151,13 @@ func NewAuthorizationCodeHandler(config *AuthorizationCodeHandlerConfig) (*Autho
 		} else if !slices.Contains(dCfg.Metadata.RedirectURIs, config.RedirectURL) {
 			return nil, fmt.Errorf("RedirectURL %q is not in the list of allowed redirect URIs for dynamic client registration", config.RedirectURL)
 		}
+		if dCfg.Metadata.ApplicationType == "" {
+			applicationType, err := inferApplicationType(dCfg.Metadata.RedirectURIs)
```
maciej-kisiel: I'm afraid this approach may not be backwards compatible. If someone already uses mixed redirect URIs, this would start failing after SDK update. I would propose the following:

1) Only validate if both redirect URIs and application_type are specified directly.
2) Only infer if the redirect URIs are consistent with each other and not specify any value if not.

auth/authorization_code.go
maciej-kisiel, line 42: Please document the handling of application_type here.

auth/authorization_code_test.go
```diff
@@ -608,6 +608,134 @@ func TestDynamicRegistration(t *testing.T) {
...
+				t.Fatalf("inferApplicationType() error = %v, wantErr %v", err, tt.wantErr)
+			}
+			if got != tt.want {
+				t.Errorf("inferApplicationType() = %q, want %q", got, tt.want)
+			}
+		})
+	}
+}
+
+func TestApplicationTypeInference(t *testing.T) {
```
maciej-kisiel, line 663: Is there a reason to not create a table driven test here? The test body looks very similar in all subtests.

auth/authorization_code.go
guglielmo-san: > Only validate if both redirect URIs and application_type are specified directly.

This would also be a breaking change, I don't think the client should interfere if the application_type is specified and would leave this responsibility to the AS.

maciej-kisiel: You're just adding the `application_type` field to the metadata, so there is no way it could have been set before, no? So this would be a new scenario.

guglielmo-san: ah right...

auth/authorization_code.go
```diff
@@ -34,6 +35,10 @@ type ClientIDMetadataDocumentConfig struct {
 type DynamicClientRegistrationConfig struct {
 	// Metadata to be used in dynamic client registration request as per
 	// https://datatracker.ietf.org/doc/html/rfc7591#section-2.
+	//
+	// If Metadata.ApplicationType is empty, it will be inferred from
+	// Metadata.RedirectURIs. When set will be validated against the inferred type
```
maciej-kisiel:
```suggestion
	// Metadata.RedirectURIs. When set, it will be validated against the inferred type
```

maciej-kisiel (dismissed):

maciej-kisiel (approved):

PR #906, opened 2026-04-23
mcp: remove default cross origin protection

### LINKED ISSUE #902: Reverse decisions for CORS and strict content type parsing.
https://github.com/modelcontextprotocol/go-sdk/pull/842 introduces to regressions to valid content types and applies erroneous CORS protections that are not the right security model for MCP servers, and just lock out web clients.

- [SDK Regression: Re-allow browser-based MCP clients via CORS 
](https://github.com/github/github-mcp-server/pull/2359)
- [Fix Content-Type rejection for application/json; charset=utf-8] (https://github.com/github/github-mcp-server/pull/2362) ✔️ already fixed (also technically it's true the protocol itself does say only the json header is valid and nothing about charset 😅)

It would be great if you can confirm that the intention is to remove the CORS stuff, as we will use the temporary compatibility flag, but we expect that in hindsight once the flag is gone, it will be replaced with that being the default (whereas the [release docs](https://github.com/modelcontextprotocol/go-sdk/releases/tag/v1.4.1) say the opposite).

Thanks so much, just want to ensure we are aligned on intention here 🙏

maciej-kisiel:
Fixes #902

### DIFF
--- a/docs/rough_edges.md
+++ b/docs/rough_edges.md
@@ -59,3 +59,7 @@ v2.
   wrapper) we need to first unmarshal into a `map[string]any` in order to do
   server-side validation of required fields. CallToolParams could have just had
   a map[string]any.
+
+- `StreamableHTTPOptions.CrossOriginProtection` should not have been part of
+  the SDK API. Cross-origin protection is a general HTTP concern, not specific
+  to MCP, and can be applied as standard HTTP middleware.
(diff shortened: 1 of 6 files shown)


jba (approved):
OK, but it looks like you have changes from another PR mixed in.

maciej-kisiel:
@jba Do you mean the content-type check using the more general approach in the SSE handler? I thought I will upgrade it along the way.

Issue #579, opened 2025-10-14
Proposal: Provide better error handling for HTTP errors when using a Streamable Client


ravanscafi:
**Is your feature request related to a problem? Please describe.**

When an MCP server responds to an `initialize` request with an error status (e.g., `401 Unauthorized`), the current error handling in `client.Connect()` only provides the HTTP status code as part of the error message itself. 
Important context like response headers (e.g., `WWW-Authenticate`), response body, and other HTTP metadata are not accessible to the caller.

This makes it difficult to implement proper authentication flows or provide detailed error messages to users, as critical information needed to handle the error appropriately is lost.

**Describe the solution you'd like**

Expose HTTP response details in connection errors, potentially through:

1. A structured error type that includes:
    - HTTP status code
    - Response headers
    - Response body
    - Original error

2. Example API:

```go
type HTTPError struct {
    StatusCode int
    Headers    http.Header
    Body       []byte
    Err        error
}

func (e *HTTPError) Error() string { ... }
func (e *HTTPError) Unwrap() error { return e.Err }
```

This would allow clients to handle authentication challenges and other HTTP-level errors appropriately:

```go
session, err := client.Connect(ctx, transport, nil)
if err != nil {
    var httpErr *mcp.HTTPError
    if errors.As(err, &httpErr) {
        if httpErr.StatusCode == 401 {
        // Access WWW-Authenticate header
        authHeader := httpErr.Headers.Get("WWW-Authenticate")
        // Handle authentication flow
        }
    }
}
```

**Describe alternatives you've considered**

Wrapping the HTTP client in Transport options: 
While this works, it adds significant complexity:
- Requires manual synchronization (mutexes) to safely capture response data when the session is used concurrently
- Couples transport-level concerns with business logic
- Makes the code harder to maintain and test

**Additional context**

Minimal reproduction example:

```go
package main

import (
	"context"
	"log"
	"net/http"

	"github.com/modelcontextprotocol/go-sdk/mcp"
)

func main() {
	go runServer("localhost:8000")
	runClient("http://localhost:8000")
}

func runServer(url string) {
	handler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		w.Header().Set("WWW-Authenticate", `Basic realm="Restricted"`)
		w.WriteHeader(http.StatusUnauthorized)
	})

	log.Printf("[Server] MCP server listening on %s", url)

	if err := http.ListenAndServe(url, handler); err != nil {
[...]

PR #907, opened 2026-04-24
feat: HTTP Header Standardization for method and name

### LINKED ISSUE #905: Implement SEP-2243 HTTP Standardization
This issue tracks implementation of https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2243.

guglielmo-san:
## Description

Implements SEP-2243 (https://modelcontextprotocol.io/seps/2243-http-standardization) (HTTP Header Standardization) for Mcp-Method and Mcp-Name standard headers (x-mcp-param custom headers will be covvered in separate PR).

Fixes #905


mcp/streamable.go
```diff
@@ -1797,6 +1809,7 @@ func (c *streamableClientConn) Write(ctx context.Context, msg jsonrpc.Message) e
 			// and permanently break the connection.
 			return nil, nil, fmt.Errorf("%s: %w: %w", requestSummary, jsonrpc2.ErrRejected, err)
 		}
+		setStandardHeaders(req.Header, msg)
```
maciej-kisiel, line 1814: I think it should be added to `setMCPHeaders` otherwise we have multiple header setting calls here, which is confusing. It would also make the dependency between then more clear (can be even accompanied with a comment).

mcp/streamable_test.go
```diff
@@ -1915,6 +1915,258 @@ func TestStreamableGET(t *testing.T) {
...
+			wantStatusCode: http.StatusOK,
+			wantMessages:   []jsonrpc.Message{resp(6, &CallToolResult{Content: []Content{}}, nil)},
+		},
+	})
+}
+
+// TestStreamableMcpHeaderValidationErrorFormat verifies that header
+// validation errors return a JSON-RPC error with code -32001 and
+// Content-Type application/json, per SEP-2243.
+func TestStreamableMcpHeaderValidationErrorFormat(t *testing.T) {
```
maciej-kisiel, line 2028: I wonder if this test wouldn't be more readable if you also used the MCP client, but with a custom HTTP client that modifies the header in the round tripper. You wouldn't have to do the initialization manually.

Maybe it would work in the next test case as well?

mcp/mcp_http_headers.go
```diff
@@ -0,0 +1,93 @@
...
+	"github.com/modelcontextprotocol/go-sdk/jsonrpc"
+)
+
+const (
+	ProtocolVersionHeader        = "Mcp-Protocol-Version"
+	SessionIDHeader              = "Mcp-Session-Id"
+	LastEventIDHeader            = "Last-Event-ID"
+	MethodHeader                 = "Mcp-Method"
+	NameHeader                   = "Mcp-Name"
+	MinVersionForStandardHeaders = "2026-06-XX"
```
maciej-kisiel: Let's use `2026-06-30` for now, as it's officially listed as such in plan.modelcontextprotocol.io. Probably it should be a new value in `shared.go` like other protocol versions? Just do not mark it as latest yet.

mcp/mcp_http_headers_test.go
```diff
@@ -0,0 +1,430 @@
...
+	tests := []struct {
+		name           string
+		version        string
+		methodHeader   string
+		nameHeader     string
+		msg            jsonrpc.Message
+		wantErr        bool
+		wantErrContain string
+	}{
+		// -- Version gating --
```
maciej-kisiel: I think I would remove those comments. Test names are precise enough.

mcp/mcp_http_headers_test.go
```diff
@@ -0,0 +1,430 @@
...
+			}
+			if got := header.Get(NameHeader); got != tt.wantNameHeader {
+				t.Errorf("NameHeader = %q, want %q", got, tt.wantNameHeader)
+			}
+		})
+	}
+}
+
+func TestValidateMcpHeaders(t *testing.T) {
+	futureVersion := MinVersionForStandardHeaders
```
maciej-kisiel: `future` and `old` are relative. The new version will not be `future` in two months. Let's just use the values directly.

mcp/mcp_http_headers.go
```diff
@@ -0,0 +1,93 @@
...
+)
+
+func extractName(method string, params json.RawMessage) (string, bool) {
+	switch method {
+	case "tools/call":
+		var p CallToolParams
+		if err := json.Unmarshal(params, &p); err == nil {
```
maciej-kisiel: Let's use our internal JSON decoder from `internal/json`.

mcp/mcp_http_headers.go
```diff
@@ -0,0 +1,93 @@
...
+		if methodInHeader != msg.Method {
+			return fmt.Errorf("header mismatch: Mcp-Method header value '%s' does not match body value '%s'", methodInHeader, msg.Method)
+		}
+
+		if msg.Method == "tools/call" || msg.Method == "resources/read" || msg.Method == "prompts/get" {
+			nameInHeader := header.Get(NameHeader)
+			if nameInHeader == "" {
+				return fmt.Errorf("missing required Mcp-Name header for method %q", msg.Method)
+			}
+			if nameInBody, ok := extractName(msg.Method, msg.Params); ok {
```
maciej-kisiel: What do you think about failing as well when `ok` is false? I believe this can only happen when the params don't unmarshal, so it will probably fail somewhere else, but I wouldn't say accepting it is in spirit of the SEP.

mcp/streamable_headers.go
```diff
@@ -0,0 +1,93 @@
...
+		var p ReadResourceParams
+		if err := json.Unmarshal(params, &p); err == nil {
+			return p.URI, true
+		}
+	}
+
+	return "", false
+}
+
+func setStandardHeaders(header http.Header, msg jsonrpc.Message) {
```
maciej-kisiel, line 50: Please document that it requires the protocol version header to be set.

mcp/streamable.go
```diff
@@ -1191,6 +1185,24 @@ func (c *streamableServerConn) servePOST(w http.ResponseWriter, req *http.Reques
 		}
 	}
 
+	// Validate MCP standard headers (Mcp-Method, Mcp-Name)
+	if !isBatch && len(incoming) == 1 {
```
maciej-kisiel, line 1189: We should probably clean up batching.. It makes the code more complicated that it should be. If you would be interested in that, feel free to create a task.

mcp/mcp_http_headers.go
```diff
@@ -0,0 +1,93 @@
...
+	"encoding/json"
+	"errors"
+	"fmt"
+	"net/http"
+
+	"github.com/modelcontextprotocol/go-sdk/jsonrpc"
+)
+
+const (
+	ProtocolVersionHeader        = "Mcp-Protocol-Version"
```
maciej-kisiel: Please don't export those if not necessary.

mcp/streamable.go
guglielmo-san: This was my original idea as well, but I did not like that adding the msg parameter to the existing `setMCPHeaders` will interfere with other places where the method is called passing a nil parameter. I decided to create two different methods to make more clear the different types of headers that are set, maybe it can be made more clear changing the methods names...
Also in the C# implementation this is divided into two different methods

mcp/mcp_http_headers.go
guglielmo-san: Yes I agree, makes sense

mcp/streamable.go
guglielmo-san: sg I created the Task

mcp/streamable_test.go
```diff
@@ -1915,6 +1915,251 @@ func TestStreamableGET(t *testing.T) {
...
+	var lastResp *http.Response
+	var capturedBody []byte
+
+	customClient := &http.Client{
+		Transport: roundTripperFunc(func(req *http.Request) (*http.Response, error) {
+			if req.Header.Get(methodHeader) == "tools/call" {
+				req.Header.Set(methodHeader, "wrong-method")
+			}
+			resp, err := http.DefaultTransport.RoundTrip(req)
+			if err == nil && req.Header.Get(methodHeader) == "wrong-method" {
```
maciej-kisiel: Could you explain why do we need to capture anything? Doesn't the error from `CallTool` contain all necessary details?

mcp/streamable.go
maciej-kisiel: Thanks for explaining. Let's keep it as is then. We should be able to merge them at some point in the future (v2?) when we can drop support for standalone SSE stream and resumability. I would jsut add a short comment that this line needs to be after the `setMCPHeaders` call so that the protocol version header is set.

mcp/streamable_headers.go
maciej-kisiel, line 1: One nit I forgot about the previous time: I think I would rename it to `streamable_headers*` to make the connection to the streamable transport more obvious.

mcp/streamable_test.go
guglielmo-san: The issue was that `chackResponse` in `streamableClientConn` was only extracting the http.StatusCode from the error returned by the server and nothing else. I improved it to extract the jsonrpc error body and include it in the client Error

maciej-kisiel: After some thought, I think I would suggest to not modify the error in this PR. It's not great that the actual error is not really accessible, but this likely requires a more holistic approach to improve the error handling – there's even a relevant issue already (https://github.com/modelcontextprotocol/go-sdk/issues/579) although triggered by different motives.

The initial solution with capturing seems to be the correct one here, let me just add a few suggestions to make it less confusing.

mcp/streamable_test.go
```diff
@@ -2044,22 +2044,12 @@ func TestStreamableMcpHeaderValidationErrorFormat(t *testing.T) {
...
-	var lastResp *http.Response
-	var capturedBody []byte
-
 	customClient := &http.Client{
 		Transport: roundTripperFunc(func(req *http.Request) (*http.Response, error) {
 			if req.Header.Get(methodHeader) == "tools/call" {
 				req.Header.Set(methodHeader, "wrong-method")
 			}
-			resp, err := http.DefaultTransport.RoundTrip(req)
-			if err == nil && req.Header.Get(methodHeader) == "wrong-method" {
```
maciej-kisiel: Let's use `req.Header.Get(methodHeader) == "tools/call"` in the condition here, since this is the original condition, not the derivative one. I think it will be less confusing.

mcp/streamable_test.go
```diff
@@ -2044,22 +2044,12 @@ func TestStreamableMcpHeaderValidationErrorFormat(t *testing.T) {
 	defer httpServer.Close()
 
 	// Use the MCP client with a custom RoundTripper to inject a bad header.
-	var lastResp *http.Response
```
maciej-kisiel: Nit: rename to `toolCallResp` for clarity?
Similarly `toolCallRespBody` below.

mcp/streamable_test.go
```diff
@@ -2077,33 +2067,18 @@ func TestStreamableMcpHeaderValidationErrorFormat(t *testing.T) {
...
-	ct := baseMediaType(lastResp.Header.Get("Content-Type"))
-	if ct != "application/json" {
-		t.Errorf("Content-Type = %q, want %q", ct, "application/json")
+	errStr := err.Error()
+	if !strings.Contains(errStr, "Bad Request") {
+		t.Errorf("error missing 'Bad Request': %s", errStr)
 	}
-
-	// Verify JSON-RPC error body contains error code -32001.
-	bodyStr := string(capturedBody)
```
maciej-kisiel: Nit: maybe parsing to `jsonrpc2.Response` and casting the `error` to `jsonrpc2.WireError` would allow making these checks less dependent on JSON formatting?

mcp/streamable_test.go
```diff
@@ -2077,33 +2067,18 @@ func TestStreamableMcpHeaderValidationErrorFormat(t *testing.T) {
...
+		t.Fatal("CallTool succeeded unexpectedly")
 	}
-
-	// Verify HTTP status code.
-	if lastResp.StatusCode != http.StatusBadRequest {
-		t.Errorf("status code = %d, want %d", lastResp.StatusCode, http.StatusBadRequest)
-	}
-
-	// Verify Content-Type.
-	ct := baseMediaType(lastResp.Header.Get("Content-Type"))
```
maciej-kisiel:
```go
if baseMediaType(lastResp.Header.Get("Content-Type")) != "application/json" {
```

mcp/streamable_test.go
guglielmo-san: Ok, done

maciej-kisiel (approved):

PR #908, opened 2026-04-24
fix: do not `omitempty` `ReadOnlyHint` in `ToolAnnotations`

pvlbzn:
`ReadOnlyHint` is a `bool` with omitempty, so the zero value (false) is indistinguishable from unset, and **drops out** of marshaled JSON.

The current behavior causes an issue with the OpenAI MCP app submission because they require explicit hints.

Consumers that explicitly set `ReadOnlyHint: false` on write tools lose the field on the wire. Removing `omitempty` ensures `false` is always serialized, which matches the MCP spec default.


mcp/protocol.go
guglielmo-san: what do you think about removing the `omitempty` from here as well, to keep it consistent?

maciej-kisiel:
I would also suggest adding a new point to `internal/docs/rough_edges.src.md` (don't forget to run `go generate ./...`) that all hints should be been specified as `*bool` for full control of whether they are sent on the wire or not.

mcp/protocol.go
pvlbzn: Yes, great take. This enables us to have symmetry: `*bool` are with `omitempty` since `*bool` is a ternary here,  and `bool` are without `omitempty`, defaulting to false.

pvlbzn:
Thanks for the review and suggestions @maciej-kisiel @guglielmo-san. Updated the branch 👍

guglielmo-san (dismissed):

maciej-kisiel:
One last thing: I believe this is a behavior change that may result for example in breaking tests of our users. Given that, I would introduce a `MCPGODEBUG` flag to be able to temporarily revert to previous behavior. Please look for usages of `mcpgodebug` on how to set this up.

pvlbzn:
Great call @maciej-kisiel , implemented compatibility mechanism according to `mcpgodebug`, and updated documentation.

maciej-kisiel:
Please fix the staticcheck error.

guglielmo-san (approved):

yarolegovich (approved):

pvlbzn:
Thank you, team 🙌

PR #909, opened 2026-04-25
mcp: handle oauth2.RetrieveError in client authorization retry logic

smlx:
Previously, an expired refresh token in the oauth2.Token returned from OAuthHandler.TokenSource() would cause the connection to fail.

From the client perspective, this meant that the MCP connection was in a hard-failed state with no way to re-authorize.

The change in this commit causes Authorize() to be called in the event of both an oauth2.RetrieveError, as well as in the pre-existing case of a 401/403 HTTP response. Clients will handle this in their existing Authorize() flows to get a new valid token for the connection.

### DIFF
--- a/mcp/streamable.go
+++ b/mcp/streamable.go
@@ -30,6 +30,8 @@ import (
 	"sync/atomic"
 	"time"
 
+	"golang.org/x/oauth2"
+
 	"github.com/modelcontextprotocol/go-sdk/auth"
 	internaljson "github.com/modelcontextprotocol/go-sdk/internal/json"
 	"github.com/modelcontextprotocol/go-sdk/internal/jsonrpc2"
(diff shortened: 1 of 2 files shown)


maciej-kisiel:
Could you explain the user journey in more detail, preferably as a list of events? Is it about a situation where the token from authorization was not used for a long time (maybe written to and loaded from disk after a significant time) so that the refresh token is no longer valid?

smlx:
>  Is it about a situation where the token from authorization was not used for a long time (maybe written to and loaded from disk after a significant time) so that the refresh token is no longer valid?

Yes, exactly.

> Could you explain the user journey in more detail, preferably as a list of events?

1. The user authorizes the application.
1. The resulting `oauth.Token` (including the refresh token) is retained either to persistent storage or in memory.
1. Time passes. The user does not use the application for an extended period, or the authorization server revokes the refresh token. The refresh token becomes invalid.
1. The user interacts with the application and the SDK attempts to make a request to the MCP server.
1. `OAuthHandler.TokenSource.Token()` is invoked. The underlying `oauth2.Config.TokenSource().Token()` attempts to refresh the token by hitting the token endpoint. It responds with a `4xx`.
1. `Token()` returns a [RetrieveError](https://pkg.go.dev/golang.org/x/oauth2#RetrieveError) because of the invalid refresh token.

Events 5 and 6 happen here: https://github.com/modelcontextprotocol/go-sdk/blob/93a41b2cfa81e6be813feadfd6dc1ed665c06da7/mcp/streamable.go#L1910-L1923

Before this change, the `RetrieveError` would bubble up to `Connection.write`, setting `writeErr` on the connection thus permanently setting it into a failed state. For a client using the connection, this is an unrecoverable error (I think? Maybe there's another way to recover here?).

After this change, the new `doWithAuth` wrapper intercepts the `oauth2.RetrieveError`. It pauses the request and invokes `OAuthHandler.Authorize()` to start a fresh authorization flow:

1. The user is prompted to re-authorize using the same `Authorize()` flow they initially used.
1. Once the user successfully re-authorizes, the original request is retried with the new valid token

mcp/streamable.go
```diff
@@ -2208,6 +2186,51 @@ func (c *streamableClientConn) connectSSE(ctx context.Context, lastEventID strin
...
+func (c *streamableClientConn) doWithAuth(ctx context.Context, requestSummary string, doRequest func() (*http.Request, *http.Response, error)) (*http.Request, *http.Response, error) {
+	req, resp, err := doRequest()
+
+	if c.oauthHandler == nil || req == nil {
+		return req, resp, err
+	}
+
+	var authResp *http.Response
+	var retrieveErr *oauth2.RetrieveError
+	if err != nil && errors.As(err, &retrieveErr) {
```
yarolegovich, line 2203: I'm wondering if the check is too broad. Should we only be re-authorizing if `retriveErr.ErrorCode  == "invalid_grant"`?

>   The provided authorization grant (e.g., authorization code, resource owner credentials) or refresh token is  invalid, expired, revoked, does not match the redirection URI used in the authorization request, or was issued to another client.

https://datatracker.ietf.org/doc/html/rfc6749#section-5.2

mcp/streamable.go
```diff
@@ -2208,6 +2186,51 @@ func (c *streamableClientConn) connectSSE(ctx context.Context, lastEventID strin
 	return nil, fmt.Errorf("connection aborted after %d attempts", c.maxRetries)
 }
 
+// doWithAuth executes an HTTP request, automatically handling OAuth2 token retrieval errors
+// and 401/403 HTTP status codes by triggering the OAuthHandler's Authorize flow and retrying.
+//
+// doRequest should construct and send the HTTP request, and return the sent request (which
+// may be needed for authorization), the response (if any), and any error.
+func (c *streamableClientConn) doWithAuth(ctx context.Context, requestSummary string, doRequest func() (*http.Request, *http.Response, error)) (*http.Request, *http.Response, error) {
```
yarolegovich, line 2194: nit: let's make doRequest take `context.Context` to avoid unintentional wrong context capture

mcp/streamable.go
```diff
@@ -2208,6 +2186,51 @@ func (c *streamableClientConn) connectSSE(ctx context.Context, lastEventID strin
...
+	var retrieveErr *oauth2.RetrieveError
+	if err != nil && errors.As(err, &retrieveErr) {
+		authResp = retrieveErr.Response
+	} else if err == nil && (resp.StatusCode == http.StatusUnauthorized || resp.StatusCode == http.StatusForbidden) {
+		authResp = resp
+	} else {
+		return req, resp, err
+	}
+
+	if authErr := c.oauthHandler.Authorize(ctx, req, authResp); authErr != nil {
```
yarolegovich, line 2211: looking into Authorize it relies on authResp to have `WWW-Authenticate` set. if fallbacks to well-known won't work the whole Authorize will fail. 

might be a bit hacky, but what if we handle oauth2.RetrieveError earlier by just not setting "Authorization" header and letting request fail with 401/403?

mcp/streamable.go
smlx: Yes I think you are right. I'll fix this up.

mcp/streamable.go
smlx: > looking into Authorize it relies on authResp to have WWW-Authenticate set. if fallbacks to well-known won't work the whole Authorize will fail.

I see what you are saying, but how likely/possible is this? Seems to defeat the purpose of being "well-known" :confused: 

> might be a bit hacky, but what if we handle oauth2.RetrieveError earlier by just not setting "Authorization" header and letting request fail with 401/403?

That is an interesting idea. In fact, it might be a better solution than this PR. I'll look into it.

yarolegovich: > I see what you are saying, but how likely/possible is this? Seems to defeat the purpose of being "well-known"

I don't know, but it is possible. [The spec](https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#protected-resource-metadata-discovery-requirements) says:

>  MCP servers MUST implement **one of** the following discovery mechanisms to provide authorization server location information to MCP clients:
> * WWW-Authenticate Header: ...
>* Well-Known URI: ...

So the failed empty-auth request approach should be more reliable. It's also a bit easier to reason about in the sense that `Authorize` only receives MCP (not auth) server response.

@maciej-kisiel and I also considered how `oauth.Config` which token refresher already has can be passed to `Authorize` for skipping the discovery step in re-authorization, but all the options seem not worth it (in terms of code complexity) if a single failed request can handle this edge case.

smlx: Yes I agree, and thank you for the detailed analysis. I sent a new PR with your much simpler fix in #917.

PR #913, opened 2026-04-28
mcp: suppress subprocess exit error during graceful close

### LINKED ISSUE #855: Potential race condition in (*mcp.ClientSession).Close()
We've experienced an increasing number of test flakes in the tests for our MCP server. There is an error when closing the connection at the end of the tests: "server is closing: EOF".  We think there might be a bug in mcp causing a race between the active part of Close (which breaks the connection and terminates the child process) and the waiting part of close (which seems to observe an error from the terminated child without realizing that close itself was responsible). Alan began tracing the issue here: https://github.com/golang/go/issues/77336#issuecomment-4090756198

**To Reproduce**
The test is here: https://github.com/golang/tools/blob/36360cdf844bb2553617bdc496c8d972c7229e7d/gopls/internal/cmd/mcp_test.go#L109
Server code is here: https://github.com/golang/tools/blob/master/gopls/internal/cmd/mcp.go
Try to reproduce the failure locally by using -count=100:
```
go test -run ^TestMCPCommandLogging$ -count=100 golang.org/x/tools/gopls/internal/cmd
```

```
=== RUN   TestMCPCommandLogging/trace.log
    gopls: server is closing: EOF
    mcp_test.go:156: closing MCP connection: exit status 2
```
**Expected behavior**
Client connection closes without error.

blackwell-systems:
## Summary

Fixes #855. When `ClientSession.Close()` initiates graceful shutdown by closing stdin, the subprocess may exit with a non-zero status. This exit error propagated through `pipeRWC.Close()` as `closeErr`, causing spurious errors from `ClientSession.Close()`.

This matches the analysis by @adonovan in [golang/go#77336](https://github.com/golang/go/issues/77336#issuecomment-4090756198):

> "a bug in mcp causing a race between the active part of Close (which breaks the connection and terminates the child process) and the waiting part of close (which seems to observe an error from the terminated child without realizing that close itself was responsible)"

## Root cause

The error path traced by @adonovan:

1. `ClientSession.Close()` calls `conn.Close()`, which sets `connClosing = true`
2. `updateInFlight` sees idle + closing, calls `s.closer.Close()` (`pipeRWC.Close`)
3. `pipeRWC.Close()` closes stdin, then waits for `cmd.Wait()`
4. The subprocess (e.g. gopls) sees EOF on stdin, treats it as fatal, exits with code 2
5. `cmd.Wait()` returns `*exec.ExitError`, stored as `closeErr`
6. `conn.wait(false)` returns `closeErr` to the caller

The subprocess exited because the client told it to (by closing stdin). The exit code is not meaningful for the client.

## Fix

In `pipeRWC.Close()`, after closing stdin and waiting for the subprocess to exit, suppress `*exec.ExitError`. Errors from the SIGTERM/SIGKILL fallback paths are still propagated, as those indicate the subprocess did not respond to graceful shutdown.

The fix is at the transport layer (not the jsonrpc2 layer) because only the stdio transport has the semantic knowledge that "I closed stdin, so the subprocess exiting is expected."

## Test plan

- Added `TestCmdTransportNonZeroExit`: connects to a server that exits with code 2 on shutdown, verifies `session.Close()` returns nil
- Confirmed the test fails without the fix (`exit status 2`) and passes with it
- Full test suite passes: `GOWORK=off go test ./... -count=1`

### DIFF
--- a/mcp/cmd.go
+++ b/mcp/cmd.go
@@ -6,6 +6,7 @@ package mcp
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"os/exec"
@@ -88,6 +89,14 @@ func (s *pipeRWC) Close() error {
 		return nil, false
 	}
 	if err, ok := wait(); ok {
+		// The subprocess exited after we closed stdin. A non-zero exit code
+		// is expected here: the server may exit with an error status because
+		// it observed EOF on stdin. This is part of normal graceful shutdown,
+		// not an error in the connection.
+		var exitErr *exec.ExitError
+		if errors.As(err, &exitErr) {
+			return nil
+		}
 		return err
 	}
 	// Note the condition here: if sending SIGTERM fails, don't wait and just
--- a/mcp/cmd_test.go
+++ b/mcp/cmd_test.go
@@ -55,6 +55,7 @@ func TestMain(m *testing.M) {
 var serverFuncs = map[string]func(){
 	"default":       runServer,
 	"cancelContext": runCancelContextServer,
+	"exitNonZero":   runExitNonZeroServer,
 }
 
 func runServer() {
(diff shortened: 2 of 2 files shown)


adonovan:
cc: @findleyr

PR #915, opened 2026-04-29
mcp: HTTP Header Standardization for x-mcp-header

### LINKED ISSUE #905: Implement SEP-2243 HTTP Standardization

guglielmo-san:
## Description

Implements [SEP-2243](https://modelcontextprotocol.io/seps/2243-http-standardization) (HTTP Header Standardization) for x-mcp-param custom header.

Fixes https://github.com/modelcontextprotocol/go-sdk/issues/905


mcp/client.go
```diff
@@ -995,6 +1031,9 @@ func (cs *ClientSession) CallTool(ctx context.Context, params *CallToolParams) (
 		// Avoid sending nil over the wire.
 		params.Arguments = map[string]any{}
 	}
+	if tool := cs.getCachedTool(params.Name); tool != nil {
```
maciej-kisiel, line 1032: Please also think how https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2549 will factor in into this solution.

mcp/header_encoding.go
maciej-kisiel, line 1: WDYT about merging this with `streamable_headers.go`? I think we already have a lot of files and the existing one is still not that big. Unless there are some particular reasons to separate it?

mcp/header_encoding.go
```diff
@@ -0,0 +1,86 @@
...
+// leading/trailing whitespace are Base64-encoded with the =?base64?...?= wrapper.
+func encodeHeaderValue(value any) (string, bool) {
+	var s string
+	switch v := value.(type) {
+	case string:
+		s = v
+	case float64:
+		s = fmt.Sprintf("%g", v)
+	case bool:
+		if v {
```
maciej-kisiel: Would `fmt.Sprintf("%t", v)` work here?

mcp/header_encoding.go
```diff
@@ -0,0 +1,86 @@
...
+// encodeHeaderValue converts a parameter value to an HTTP header-safe string
+// per the SEP-2243 encoding rules:
+//   - string: used as-is if safe ASCII, otherwise Base64 encoded
+//   - number (float64): decimal string representation
+//   - bool: lowercase "true" or "false"
+//   - nil: returns "", false
+//
+// Values that contain non-ASCII characters, control characters, or
+// leading/trailing whitespace are Base64-encoded with the =?base64?...?= wrapper.
+func encodeHeaderValue(value any) (string, bool) {
```
maciej-kisiel: Please comment what the second return value means.

mcp/header_encoding.go
```diff
@@ -0,0 +1,86 @@
...
+
+	if requiresBase64Encoding(s) {
+		return encodeBase64(s), true
+	}
+	return s, true
+}
+
+// decodeHeaderValue decodes a header value that may be Base64-encoded
+// with the =?base64?...?= wrapper.
+func decodeHeaderValue(headerValue string) (string, bool) {
```
maciej-kisiel: Please comment what the second return value means.

mcp/header_encoding.go
```diff
@@ -0,0 +1,86 @@
...
+	if len(headerValue) == 0 {
+		return headerValue, true
+	}
+
+	if strings.HasPrefix(strings.ToLower(headerValue), strings.ToLower(base64Prefix)) &&
```
maciej-kisiel: Consider using https://godoc.corp.google.com/pkg/strings#CutPrefix and `CutSuffix`. Maybe the logic will be easier?

mcp/streamable_headers.go
```diff
@@ -45,6 +49,75 @@ func extractName(method string, params json.RawMessage) (string, bool) {
...
+// extractToolParamHeaders returns a map of parameter name to header name
+// for all properties in the tool's InputSchema that have an x-mcp-header
+// annotation.
+func extractToolParamHeaders(tool *Tool) map[string]string {
+	props := extractSchemaProperties(tool.InputSchema)
+	if props == nil {
+		return nil
+	}
+	result := make(map[string]string)
+	for propName, propSchema := range props {
```
maciej-kisiel: I see that we're assuming that the special annotation applies to top level arguments only. It's probably correct given that the nested property is listed as a negative test case in the proposed conformance tests, but the spec is quite ambiguous here. Let me ask a clarifying question on the SEP.

mcp/streamable_headers.go
```diff
@@ -61,10 +134,141 @@ func setStandardHeaders(header http.Header, msg jsonrpc.Message) {
...
+				}
+			}
+		}
 	}
 }
 
-func validateMcpHeaders(header http.Header, msg jsonrpc.Message) error {
+// extractParamHeaders reads x-mcp-header annotations from the tool's InputSchema
+// and returns the Mcp-Param-{Name} headers to be set on the HTTP request.
+func extractParamHeaders(tool *Tool, params json.RawMessage) map[string]string {
```
maciej-kisiel: Nit: IMO this is no longer extraction, I would propose something like `generateParamHeaders`.

mcp/streamable_headers.go
```diff
@@ -45,6 +49,75 @@ func extractName(method string, params json.RawMessage) (string, bool) {
...
+	if !ok {
+		return nil
+	}
+	return props
+}
+
+// extractToolParamHeaders returns a map of parameter name to header name
+// for all properties in the tool's InputSchema that have an x-mcp-header
+// annotation.
+func extractToolParamHeaders(tool *Tool) map[string]string {
```
maciej-kisiel: Nit: `extractParamHeaderAnnotations`

mcp/streamable_headers.go
```diff
@@ -45,6 +49,75 @@ func extractName(method string, params json.RawMessage) (string, bool) {
 	return "", false
 }
 
+func extractSchemaProperties(schema any) map[string]any {
```
maciej-kisiel: Have you considered working on jsonschema types rather than plain maps? I wonder if it wouldn't be easier.

mcp/streamable_headers.go
```diff
@@ -61,10 +134,141 @@ func setStandardHeaders(header http.Header, msg jsonrpc.Message) {
...
+			continue
+		}
+		result = append(result, tool)
+	}
+	return result
+}
+
+// validateToolParamHeaders checks that a tool's x-mcp-header annotations
+// are valid.
+func validateToolParamHeaders(tool *Tool) error {
```
maciej-kisiel: Nit: `validateParamHeaderAnnotations`.

mcp/streamable_headers.go
```diff
@@ -61,10 +134,141 @@ func setStandardHeaders(header http.Header, msg jsonrpc.Message) {
...
+			return fmt.Errorf("property %q: %w", propName, err)
+		}
+		lower := strings.ToLower(headerName)
+		if seen[lower] {
+			return fmt.Errorf("property %q: duplicate x-mcp-header value %q (case-insensitive)", propName, headerName)
+		}
+		seen[lower] = true
+
+		propType, _ := ps["type"].(string)
+		if propType != "" && propType != "string" && propType != "number" && propType != "integer" && propType != "boolean" {
```
maciej-kisiel: Shouldn't we also return an error if type is not provided?

mcp/streamable_headers.go
```diff
@@ -93,6 +297,59 @@ func validateMcpHeaders(header http.Header, msg jsonrpc.Message) error {
...
+		}
+
+		decoded, ok := decodeHeaderValue(headerVal)
+		if !ok {
+			return fmt.Errorf("header mismatch: %s header contains invalid Base64 encoding", fullHeader)
+		}
+
+		bodyVal := unmarshalPrimitive(argRaw)
+		if bodyVal == nil {
+			continue
```
maciej-kisiel: IIUC this is a situation where the header is present, but the value in arguments is not a primitive. Please add a comment explaining why it is ok to ignore this.

mcp/streamable.go
```diff
@@ -1786,6 +1802,11 @@ func (c *streamableClientConn) Write(ctx context.Context, msg jsonrpc.Message) e
 		if msg.IsCall() {
 			forCall = msg
 		}
+		if msg.Method == "tools/call" {
+			if tool, ok := ctx.Value(toolContextKey).(*Tool); ok {
+				msg.Extra = tool
```
maciej-kisiel: Maybe just pass the context to `setStandardHeaders` to keep the logic using a single mechanism?

mcp/streamable.go
```diff
@@ -491,6 +491,14 @@ func (h *StreamableHTTPHandler) ServeHTTP(w http.ResponseWriter, req *http.Reque
 			http.Error(w, "failed connection", http.StatusInternalServerError)
 			return
 		}
+		transport.connection.toolLookup = func(name string) *Tool {
```
maciej-kisiel: Does this need to be delayed with regard to transport creation in line 407?

mcp/client.go
```diff
@@ -981,7 +1011,13 @@ func (cs *ClientSession) GetPrompt(ctx context.Context, params *GetPromptParams)
 
 // ListTools lists tools that are currently available on the server.
 func (cs *ClientSession) ListTools(ctx context.Context, params *ListToolsParams) (*ListToolsResult, error) {
-	return handleSend[*ListToolsResult](ctx, methodListTools, newClientRequest(cs, orZero[Params](params)))
+	result, err := handleSend[*ListToolsResult](ctx, methodListTools, newClientRequest(cs, orZero[Params](params)))
+	if err != nil {
+		return nil, err
+	}
+	result.Tools = filterValidTools(result.Tools)
```
maciej-kisiel: In general this proposal adds HTTP specific functionality, so I'm a bit surprised that the spec requires general filtering. IMO ideally this change should be contained within streamable transport as much as possible. I'll ask on the SEP for clarification.

mcp/streamable_headers.go
guglielmo-san: right makes sense

mcp/streamable_headers.go
```diff
@@ -61,10 +134,141 @@ func setStandardHeaders(header http.Header, msg jsonrpc.Message) {
...
+	return res
+}
+
+// filterValidTools returns only tools that have valid
+// x-mcp-header annotations. Invalid tools are logged and excluded.
+func filterValidTools(tools []*Tool) []*Tool {
+	result := make([]*Tool, 0, len(tools))
+	for _, tool := range tools {
+		if err := validateToolParamHeaders(tool); err != nil {
+			log.Printf("mcp: excluding tool %q from tools/list: %v", tool.Name, err)
```
maciej-kisiel: Please use the `slog` logger here instead. One should be available at the call site.

mcp/client.go
```diff
@@ -363,6 +376,23 @@ func (cs *ClientSession) Wait() error {
...
 }
 
+func (cs *ClientSession) cacheTools(tools []*Tool) {
+	cs.toolCacheMu.Lock()
+	defer cs.toolCacheMu.Unlock()
+	if cs.toolCache == nil {
+		cs.toolCache = make(map[string]*Tool, len(tools))
+	}
+	for _, tool := range tools {
+		cs.toolCache[tool.Name] = tool
```
maciej-kisiel, line 384: Shouldn't old entries be removed from the cache as well?

maciej-kisiel (commented):
Also some additional comments from AI review :)

mcp/streamable_headers.go
guglielmo-san: Actually it would probably be better to return an error. The SEP does not explicitly define this case but it makes more sense to me.

mcp/streamable.go
guglielmo-san: I did not want to pass the ctx to function, but I agree that simplify the implementation

mcp/streamable.go
guglielmo-san: Yes, the transport.connection is created only in the server.Connect(). Unless we duplicate the toolLookup both in transport and connection

mcp/header_encoding.go
guglielmo-san: In my opinion from a logic point of view they should be divided into two different files, but I can agree that we already have a lot of different files...

mcp/streamable_headers.go
guglielmo-san: Actually this implementation was wrong as it was only considering the case of tool.inputSchema of type map[string]any. Added a more complete check using json struct

mcp/header_encoding.go
guglielmo-san: `strings.CutPrefix` is case-sensitive, it would either fail to match upper case prefixes (=?BASE64?) or, if used on a lowercased string, return a lowercased remainder that corrupts the Base64 data.

maciej-kisiel: I don't understand why we need to support case insensitivity here, let me ask on the SEP.

mcp/streamable_headers.go
maciej-kisiel: Maybe we should also validate the schema on `AddTool` as well to make a situation where a server serves incorrect tool fail early?

mcp/streamable.go
maciej-kisiel: Thanks for the clarification.

mcp/streamable.go
```diff
@@ -491,6 +491,14 @@ func (h *StreamableHTTPHandler) ServeHTTP(w http.ResponseWriter, req *http.Reque
...
+			server.mu.Lock()
```
maciej-kisiel: Using the mutex from a different file is probably not a good practice. Consider exposing a private method on the server.

mcp/streamable_headers.go
guglielmo-san: It would make sense, but it is not specified by the SEP. If the format of the Tool is wrong, then it will be filtered by the client and not returned by ListTools.
If the client still sends a non-primitive type then it will be caught here and return an error.

mcp/client.go
```diff
@@ -995,6 +1029,9 @@ func (cs *ClientSession) CallTool(ctx context.Context, params *CallToolParams) (
 		// Avoid sending nil over the wire.
 		params.Arguments = map[string]any{}
 	}
+	if tool := cs.getCachedTool(params.Name); tool != nil {
```
maciej-kisiel, line 1032: I also started wondering what should happen when the tools is not in the cache. It means that `ListTools` was not called before the call or the model has outdated information for some reason. Maybe we should proactively fail in that case? It likely should be consistent behavior across the SDKs, I will try to ask about it as well.

mcp/streamable_headers.go
```diff
@@ -45,9 +52,82 @@ func extractName(method string, params json.RawMessage) (string, bool) {
...
+	Type       string                          `json:"type"`
+	XMCPHeader json.RawMessage                 `json:"x-mcp-header,omitempty"`
+	Properties map[string]headerSchemaProperty `json:"properties,omitempty"`
+}
+
+// unmarshalSchemaProperties normalizes any InputSchema type
+// (*jsonschema.Schema, map[string]any, or json.RawMessage) into a common
+// representation by marshaling to JSON and unmarshaling only the fields we need.
+func unmarshalSchemaProperties(schema any) map[string]headerSchemaProperty {
+	var s struct {
```
maciej-kisiel: Do we need separate struct, can't `headerSchemaProperty` be used here? You only access the `Properties` field anyway.

mcp/streamable_headers.go
maciej-kisiel: I would say we have some freedom how we construct SDK APIs and their behavior. IMO that would be a valuable change and it's always easier to remove such validation than introduce it after the release.

mcp/streamable_headers.go
maciej-kisiel: Very nice!

mcp/streamable_headers.go
maciej-kisiel: The only comment I got on the SEP suggests that this shouldn't be limited to top level. Let's see if anyone confirms.

mcp/streamable_headers.go
```diff
@@ -93,6 +291,129 @@ func validateMcpHeaders(header http.Header, msg jsonrpc.Message) error {
...
+//   - number (float64): decimal string representation
+//   - bool: lowercase "true" or "false"
+//   - nil: returns "", false
+//
+// Values that contain non-ASCII characters, control characters, or
+// leading/trailing whitespace are Base64-encoded with the =?base64?...?= wrapper.
+//
+// The second return value is false if the header value's type is not supported.
+func encodeHeaderValue(value any) (string, bool) {
+	var s string
```
maciej-kisiel: Reuse `primitiveToString` here?

mcp/streamable_headers.go
```diff
@@ -45,9 +52,82 @@ func extractName(method string, params json.RawMessage) (string, bool) {
...
+}
+
+func primitiveToString(value any) string {
+	switch v := value.(type) {
+	case string:
+		return v
+	case float64:
+		return fmt.Sprintf("%g", v)
+	case bool:
+		if v {
```
maciej-kisiel: `fmt.Sprintf("%t", v)` here as well.

mcp/streamable_headers.go
```diff
@@ -93,6 +291,129 @@ func validateMcpHeaders(header http.Header, msg jsonrpc.Message) error {
...
+			return fmt.Errorf("header mismatch: %s header contains invalid Base64 encoding", fullHeader)
+		}
+
+		bodyVal := unmarshalPrimitive(argRaw)
+		if bodyVal == nil {
+			return fmt.Errorf("header mismatch: %s header present but body parameter %q is not a primitive type", fullHeader, paramName)
+		}
+		expected := primitiveToString(bodyVal)
+
+		if decoded != expected {
```
maciej-kisiel, line 349: Let's maybe leave a comment that string comparison may not work ideally for numbers. We can get back to it if it actually is an issue. I can also raise this in the SEP.

mcp/server.go
```diff
@@ -753,6 +753,16 @@ func (s *Server) listTools(_ context.Context, req *ListToolsRequest) (*ListTools
 	})
 }
 
+// toolLookup returns the Tool with the given name, or nil if not found.
+func (s *Server) toolLookup(name string) *Tool {
```
maciej-kisiel: Nit: `getTool`?

mcp/server.go
```diff
@@ -753,6 +753,16 @@ func (s *Server) listTools(_ context.Context, req *ListToolsRequest) (*ListTools
...
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	if st, ok := s.tools.get(name); ok {
+		return st.tool
+	}
+	return nil
+}
+
 func (s *Server) callTool(ctx context.Context, req *CallToolRequest) (*CallToolResult, error) {
 	s.mu.Lock()
```
maciej-kisiel: I would propose to use the new function here. It may require exposing the `ok bool`, but less mutex handling is better.

mcp/streamable.go
```diff
@@ -1186,9 +1189,15 @@ func (c *streamableServerConn) servePOST(w http.ResponseWriter, req *http.Reques
 		}
 	}
 
-	// Validate MCP standard headers (Mcp-Method, Mcp-Name)
+	// Validate MCP standard headers (Mcp-Method, Mcp-Name, Mcp-Param-*)
 	if !isBatch && len(incoming) == 1 {
-		if err := validateMcpHeaders(req.Header, incoming[0]); err != nil {
+		var tool *Tool
+		if jreq, ok := incoming[0].(*jsonrpc.Request); ok && jreq.Method == "tools/call" && c.toolLookup != nil {
```
maciej-kisiel: Nit: to limit type assertions and other checks, maybe it's work to pass the `toolLookup` function inside `validateMcpHeaders` instead of the `tool` itself? Appropriate type assertions already exist there, so the code would be a bit cleaner.

mcp/client.go
guglielmo-san: Yes I also thought about it. I don't think we should fail on client if they do not want to call ListTools before. If the format is wrong it would fail on the server anyway, which has to be considered the ultimate source of truth

mcp/server.go
guglielmo-san: It also requires to change the return type to ServerTool

mcp/streamable_headers.go
guglielmo-san: Ok I added the check on the server, can we consider it a non-breaking change?

mcp/streamable_headers.go
```diff
@@ -45,9 +52,77 @@ func extractName(method string, params json.RawMessage) (string, bool) {
...
+		}
+		result[propName] = headerName
+	}
+	if len(result) == 0 {
+		return nil
+	}
+	return result
+}
+
+func primitiveToString(value any) (string, bool) {
```
maciej-kisiel, line 97:
```suggestion

// primitiveToString conversion.
// Returns false in the second return value if the argument is not a primitive value.
func primitiveToString(value any) (string, bool) {
```

mcp/streamable_headers_test.go
```diff
@@ -413,3 +416,827 @@ func TestValidateMcpHeaders(t *testing.T) {
...
+			},
+		},
+		{
+			name: "omits header when argument is missing",
+			tool: tool,
+			params: &CallToolParams{
+				Name:      "execute_sql",
+				Arguments: map[string]any{"query": "SELECT 1"},
+			},
+			wantHeaders: map[string]string{},
```
maciej-kisiel: Will `nil` work in the cases that require empty map?

mcp/client.go
maciej-kisiel: What does the reference implementation in C# do? The most up-to-date one seems to be https://github.com/modelcontextprotocol/csharp-sdk/pull/1553.

guglielmo-san: It is sending it even when not in the cache

```C#
public override Task<JsonRpcResponse> SendRequestAsync(JsonRpcRequest request, CancellationToken cancellationToken = default)
    {
        // For tools/call requests, attach the cached tool definition to the message context
        // so the transport can add custom Mcp-Param-* headers based on x-mcp-header schema annotations.
        if (request.Method == RequestMethods.ToolsCall &&
            request.Params is System.Text.Json.Nodes.JsonObject paramsObj &&
            paramsObj.TryGetPropertyValue("name", out var nameNode) &&
            nameNode?.GetValue<string>() is { } toolName &&
            _toolCache.TryGetValue(toolName, out var tool))
        {
            request.Context ??= new();
            request.Context.Items ??= new Dictionary<string, object?>();
            request.Context.Items[McpHttpHeaders.ToolContextKey] = tool;
        }

        return _sessionHandler.SendRequestAsync(request, cancellationToken);
    }

```

maciej-kisiel (approved):

PR #917, opened 2026-04-30
mcp: handle oauth2.RetrieveError during token refresh

smlx:
Previously, an expired refresh token in the oauth2.Token returned from OAuthHandler.TokenSource() would cause the connection to fail.

From the client perspective, this meant that the MCP connection was in a hard-failed state with no way to re-authorize.

The change in this commit causes Authorize() to be called in the event of both an oauth2.RetrieveError from the token endpoint, as well as in the pre-existing case of a 401/403 HTTP response from the MCP server. Clients will handle this in their existing Authorize() flows to get a new valid token for the connection.

Supersedes and closes: #909.

### DIFF
--- a/mcp/streamable.go
+++ b/mcp/streamable.go
@@ -37,6 +37,7 @@ import (
 	"github.com/modelcontextprotocol/go-sdk/internal/util"
 	"github.com/modelcontextprotocol/go-sdk/internal/xcontext"
 	"github.com/modelcontextprotocol/go-sdk/jsonrpc"
+	"golang.org/x/oauth2"
 )
 
 // A StreamableHTTPHandler is an http.Handler that serves streamable MCP
@@ -1803,6 +1804,7 @@ func (c *streamableClientConn) Write(ctx context.Context, msg jsonrpc.Message) e
 		}
 		req.Header.Set("Content-Type", "application/json")
 		req.Header.Set("Accept", "application/json, text/event-stream")
+
 		if err := c.setMCPHeaders(req); err != nil {
 			// Failure to set headers means that the request was not sent.
 			// Wrap with ErrRejected so the jsonrpc2 connection doesn't set writeErr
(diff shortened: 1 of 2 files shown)


mcp/streamable.go
```diff
@@ -1803,11 +1804,24 @@ func (c *streamableClientConn) Write(ctx context.Context, msg jsonrpc.Message) e
...
+			// If the error is an invalid_grant oauth2.RetrieveError it indicates
+			// that the token source doesn't have valid authorization for the token
+			// endpoint, per RFC 6749 section 5.2. For example, the refresh token may
+			// be expired or invalid.
+			//
+			// In that case, ignore the error, skip setting the Authorization header,
+			// and proceed with the request. The resulting 401/403 response from the
+			// MCP server will have the necessary WWW-Authenticate header and trigger
+			// the Authorize() flow.
+			var retrieveErr *oauth2.RetrieveError
```
yarolegovich: I'd push this logic even further down, so that it holds for all setMCPHeaders call sites
```patch
index a176789..c2b3b09 100644
--- a/mcp/streamable.go
+++ b/mcp/streamable.go
@@ -37,6 +37,7 @@ import (
        "github.com/modelcontextprotocol/go-sdk/internal/xcontext"
        "github.com/modelcontextprotocol/go-sdk/jsonrpc"
+       "golang.org/x/oauth2"
 )

 const (
@@ -1919,7 +1920,10 @@ func (c *streamableClientConn) setMCPHeaders(req *http.Request) error {
                if ts != nil {
                        token, err := ts.Token()
                        if err != nil {
-                               return err
+                               var retrieveErr *oauth2.RetrieveError
+                               if !errors.As(err, &retrieveErr) || retrieveErr.ErrorCode != "invalid_grant" {
+                                       return err
+                               }
```

smlx: Good idea!

yarolegovich (approved):

PR #918, opened 2026-04-30
fix(mcp): prevent SIGSEGV in AddTool when schema is nil

### LINKED ISSUE #916: `mcp.AddTool` crashes the server with SIGSEGV when input/output struct tags trigger a nil schema from `jsonschema-go` — missing nil-guard in `setSchema`
## Summary

When registering a tool with `mcp.AddTool[In, Out]`, the SDK calls `jsonschema.ForType` and then unconditionally invokes `internalSchema.Resolve(...)` on the result. If `ForType` returns a partially-nil schema tree (which happens with certain `jsonschema:"..."` struct-tag descriptions — see linked upstream issue), the call panics with a nil-pointer dereference at server startup, taking down the whole process.

The SDK should defend against a nil or malformed schema returned by `jsonschema-go` and surface a clear error (or panic with an actionable message that names the offending Go type and field), not segfault inside a third-party library.

## Versions

- `github.com/modelcontextprotocol/go-sdk` **v1.4.0** and **v1.6.0** (same code path)
- `github.com/google/jsonschema-go` v0.4.2 / v0.4.3
- Go 1.24, macOS arm64

## Stack trace

```
[signal SIGSEGV: segmentation violation code=0x2 addr=0x10 pc=0x102b9c36c]

goroutine 1 [running]:
github.com/google/jsonschema-go/jsonschema.detectDraft(...)
    .../jsonschema-go@v0.4.2/jsonschema/resolve.go:52
github.com/google/jsonschema-go/jsonschema.newResolved(...)
    .../jsonschema-go@v0.4.2/jsonschema/resolve.go:43
github.com/google/jsonschema-go/jsonschema.(*resolver).resolve(0x..., 0x0, 0x...)
    .../jsonschema-go@v0.4.2/jsonschema/resolve.go:211 +0x2c
github.com/google/jsonschema-go/jsonschema.(*Schema).Resolve(0x0, 0x...)
    .../jsonschema-go@v0.4.2/jsonschema/resolve.go:185 +0x110
github.com/modelcontextprotocol/go-sdk/mcp.setSchema[...](...)
    .../go-sdk@v1.4.0/mcp/server.go:466 +0x1a0
github.com/modelcontextprotocol/go-sdk/mcp.toolForErr[...](...)
    .../go-sdk@v1.4.0/mcp/server.go:309 +0x230
github.com/modelcontextprotocol/go-sdk/mcp.AddTool[...](...)
    .../go-sdk@v1.4.0/mcp/server.go:501 +0x38
```

The receiver of `Resolve` is `0x0` — a nil `*jsonschema.Schema`.

## Where the SDK should guard

`mcp/server.go` `setSchema` (current code, v1.6.0 line ~436):

```go
internalSchema, err = jsonschema.ForType(rt, &jsonschema.ForOptions{})
if err != nil {
    return zero, err
}
*sfield = internalSchema

resolved, err := internalSchema.Resolve(&jsonschema.ResolveOptions{ValidateDefaults: true})
```

`internalSchema` is checked for `err`, never for nil. If `ForType` returns `(nil, nil)` (or a tree containing a nil sub-schema), the next call panics inside `jsonschema-go`.

## Suggested fix

In `setSchema` (and the analogous `*sfield`/`providedSchema` branches further down):

```go
internalSchema, err = jsonschema.ForType(rt, &jsonschema.ForOptions{})
if err != nil {
    return zero, err
}
if internalSchema == nil {
    return zero, fmt.Errorf("mcp: jsonschema.ForType returned nil schema for %s", rt)
}
*sfield = internalSchema
```

For added safety in `toolForErr` / `AddTool`, wrap the schema-related calls in a `defer recover()` that converts a panic to an error mentioning the tool name and the Go type, so a single bad tag in one tool does not take the whole server down at startup. Something like:

```go
func toolForErr[In, Out any](t *Tool, h ToolHandlerFor[In, Out], cache *SchemaCache) (rt *Tool, rh ToolHandler, err error) {
    defer func() {
        if r := recover(); r != nil {
            err = fmt.Errorf("mcp: schema generation panicked for tool %q (In=%T): %v",
                t.Name, *new(In), r)
        }
    }()
    // ... existing code
}
```

That alone would have transformed our outage into a clear actionable error pointing at the tool name and input type.

## Trigger

Reproducible by adding a tool whose input struct uses `jsonschema:"..."` descriptions containing combinations of commas, `=`, `>`, `<`, em/en dashes, parens, brackets, single quotes, or non-ASCII characters. The underlying `jsonschema-go` bug is reported separately at:

> https://github.com/google/jsonschema-go/issues/NEW (link once filed)

The SDK fix is independent: regardless of whether `jsonschema-go` ever produces a nil schema, the SDK should not segfault as a result.

## Workaround

Restrict every `jsonschema:"..."` tag description in tool input/output structs to ASCII letters and spaces only. No commas, no `,required` directive, no parens, brackets, quotes, equals, angle brackets, em/en dashes, or non-ASCII characters.

## Impact

This is a denial-of-service-grade bug for any server that registers tools dynamically based on user-authored Go types. A single malformed tag anywhere in the registered toolset prevents the server from starting and produces no actionable error message — just a stack trace pointing into a third-party library. Adding a nil-check and a panic-to-error conversion turns an outage into a clear, fixable error.

wucm667:
## Bug

`mcp.AddTool` crashes the server with SIGSEGV when input/output struct tags trigger a nil schema.

## Root Cause

In `mcp/server.go`, the function `setSchema` calls `internalSchema.Resolve(...)` without checking if `internalSchema` is nil. If the schema inference (e.g., via `jsonschema.ForType`) or a provided schema results in a nil value, this dereference causes a panic.

## Fix

Added nil checks for `internalSchema` before calling `Resolve` in two locations within `setSchema`:
1. After schema generation via `jsonschema.ForType`.
2. After processing a provided schema (which might be nil or fail to unmarshal).

If `internalSchema` is nil, the function now returns a descriptive error instead of crashing.

## Verification

- `go test ./mcp/...` passes.
- The fix prevents the crash by handling the nil case gracefully.

Closes #916.

### DIFF
--- a/mcp/server.go
+++ b/mcp/server.go
@@ -437,6 +437,9 @@ func setSchema[T any](sfield *any, rfield **jsonschema.Resolved, cache *SchemaCa
 		if err != nil {
 			return zero, err
 		}
+		if internalSchema == nil {
+			return zero, fmt.Errorf("schema is nil for type %v", rt)
+		}
 		*sfield = internalSchema
 
 		resolved, err := internalSchema.Resolve(&jsonschema.ResolveOptions{ValidateDefaults: true})
@@ -464,6 +467,10 @@ func setSchema[T any](sfield *any, rfield **jsonschema.Resolved, cache *SchemaCa
 		if err := remarshal(*sfield, &internalSchema); err != nil {
 			return zero, err
 		}
+	}
+
+	if internalSchema == nil {
+		return zero, fmt.Errorf("schema is nil for type %v", rt)
 	}
 
 	resolved, err := internalSchema.Resolve(&jsonschema.ResolveOptions{ValidateDefaults: true})
--- a/mcp/server.go
+++ b/mcp/server.go
@@ -247,6 +247,9 @@ func (s *Server) AddTool(t *Tool, h ToolHandler) {
 		panic(fmt.Errorf("AddTool %q: missing input schema", t.Name))
 	}
 	if s, ok := t.InputSchema.(*jsonschema.Schema); ok {
+		if s == nil {
+			panic(fmt.Errorf("AddTool %q: input schema is nil", t.Name))
+		}
 		if s.Type != "object" {
 			panic(fmt.Errorf(`AddTool %q: input schema must have type "object"`, t.Name))
 		}
@@ -261,6 +264,9 @@ func (s *Server) AddTool(t *Tool, h ToolHandler) {
 	}
 	if t.OutputSchema != nil {
 		if s, ok := t.OutputSchema.(*jsonschema.Schema); ok {
+			if s == nil {
+				panic(fmt.Errorf("AddTool %q: output schema is nil", t.Name))
+			}
 			if s.Type != "object" {
 				panic(fmt.Errorf(`AddTool %q: output schema must have type "object"`, t.Name))
 			}
--- a/mcp/server_test.go
+++ b/mcp/server_test.go
@@ -8,6 +8,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"fmt"
 	"log"
 	"log/slog"
 	"slices"
@@ -601,6 +602,44 @@ func TestAddToolNameValidation(t *testing.T) {
 				}
 			}
 		})
+	}
+}
+
+func TestAddToolNilSchema(t *testing.T) {
+	var nilSchema *jsonschema.Schema
+
+	panicMsg := func(f func()) string {
+		var msg string
+		func() {
+			defer func() {
+				if r := recover(); r != nil {
+					msg = fmt.Sprintf("%v", r)
+				}
+			}()
+			f()
+		}()
+		return msg
+	}
+
+	// Call s.AddTool directly to exercise the typed-nil checks added to that method.
+	msg := panicMsg(func() {
+		s := NewServer(testImpl, nil)
+		s.AddTool(&Tool{Name: "T", InputSchema: nilSchema}, nil)
+	})
+	if msg == "" {
+		t.Error("typed nil InputSchema: expected panic")
+	} else if !strings.Contains(msg, "input schema is nil") {
+		t.Errorf("typed nil InputSchema: panic message %q does not contain %q", msg, "input schema is nil")
+	}
+
+	msg = panicMsg(func() {
+		s := NewServer(testImpl, nil)
+		s.AddTool(&Tool{Name: "T", InputSchema: &jsonschema.Schema{Type: "object"}, OutputSchema: nilSchema}, nil)
+	})
+	if msg == "" {
+		t.Error("typed nil OutputSchema: expected panic")
+	} else if !strings.Contains(msg, "output schema is nil") {
+		t.Errorf("typed nil OutputSchema: panic message %q does not contain %q", msg, "output schema is nil")
 	}
 }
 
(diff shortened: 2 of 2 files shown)


jba (dismissed):

mcp/server.go
guglielmo-san, line 253: `t.InputSchema == nil` check will not catch this case 
```
var nilSchema *jsonschema.Schema = nil
server.AddTool(&mcp.Tool{
    Name:        "broken",
    InputSchema: nilSchema,
}, handler)
```


```suggestion
if s, ok := t.InputSchema.(*jsonschema.Schema); ok {
       if s == nil {
           panic(fmt.Errorf("AddTool %q: nil input schema", t.Name))
       }
       // ...
   }
```

Same for OutputSchema

wucm667:
@guglielmo-san Thanks for catching that! Added nil checks for both InputSchema and OutputSchema after the type assertion to handle the typed nil case. Pushed the fix.

guglielmo-san:
@wucm667 I think you forgot to push the the fix :)

wucm667:
@guglielmo-san The fix is pushed now. Let me know if anything else needs adjusting.

guglielmo-san:
@wucm667 can you add tests to `mcp/server_test.go` for the changes you introduced?

wucm667:
@guglielmo-san Added tests in `mcp/server_test.go` covering both typed nil InputSchema and OutputSchema cases. All tests pass locally.

mcp/server_test.go
```diff
@@ -604,6 +605,44 @@ func TestAddToolNameValidation(t *testing.T) {
...
+		func() {
+			defer func() {
+				if r := recover(); r != nil {
+					msg = fmt.Sprintf("%v", r)
+				}
+			}()
+			f()
+		}()
+		return msg
+	}
```
guglielmo-san:
```suggestion
  panicMsg := func(f func()) (msg string) {
  		defer func() {
  			if r := recover(); r != nil {
  				msg = fmt.Sprintf("%v", r)
  			}
  		}()
  		f()
  		return msg
  	}
```

mcp/server_test.go
```diff
@@ -604,6 +605,44 @@ func TestAddToolNameValidation(t *testing.T) {
...
+	msg = panicMsg(func() {
+		s := NewServer(testImpl, nil)
+		s.AddTool(&Tool{Name: "T", InputSchema: &jsonschema.Schema{Type: "object"}, OutputSchema: nilSchema}, nil)
+	})
+	if msg == "" {
+		t.Error("typed nil OutputSchema: expected panic")
+	} else if !strings.Contains(msg, "output schema is nil") {
+		t.Errorf("typed nil OutputSchema: panic message %q does not contain %q", msg, "output schema is nil")
+	}
+}
```
guglielmo-san:
```suggestion
tests := []struct {
		name        string
		tool        *Tool
		wantContain string
	}{
		{
			name:        "typed nil InputSchema",
			tool:        &Tool{Name: "T", InputSchema: nilSchema},
			wantContain: "input schema is nil",
		},
		{
			name:        "typed nil OutputSchema",
			tool:        &Tool{Name: "T", InputSchema: &jsonschema.Schema{Type: "object"}, OutputSchema: nilSchema},
			wantContain: "output schema is nil",
		},
	}
	for _, tc := range tests {
		t.Run(tc.name, func(t *testing.T) {
			s := NewServer(testImpl, nil)
			
			msg := panicMsg(func() {
				s.AddTool(tc.tool, nil)
			})
			if msg == "" {
				t.Error("expected panic")
			} else if !strings.Contains(msg, tc.wantContain) {
				t.Errorf("panic message %q does not contain %q", msg, tc.wantContain)
			}
		})
	}
```

guglielmo-san (approved):

PR #919, opened 2026-04-30
Design SDK changes for stateless & sessionless MCP

maciej-kisiel:


maciej-kisiel:
Hi @yarolegovich, @guglielmo-san, @jba, I have a draft of the first part of the design for stateless/sessionless MCP that's coming in the next spec version. It's focused on the connection establishment logic. PTAL if you're interested and let me know if you have any feedback.

design/stateless.md
```diff
@@ -0,0 +1,315 @@
...
+The error also contains a list of `supported` versions. If any of them is supported by the client,
+then one of two things will happen:
+
+* the version will be saved to `ClientSession.state.InitializeResult` and used in subsequent requests if it is `>= 2026-06-30`
+* normal initialization will be performed if it is `< 2026-06-30`
+
+New types in `protocol.go`:
+
+```go
+const ErrCodeUnsupportedProtocolVersion = -32602 // INVALID_PARAMS per SEP-2575
```
yarolegovich, line 127: asked for a dedicated code https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2575#discussion_r3188179893

design/stateless.md
```diff
@@ -0,0 +1,315 @@
...
+If none of these handlers are set, no `subscriptions/listen` call will be made during `Connect()`.
+Presence of `ResourceUpdateHandler` will not automatically trigger `subscriptions/listen` during
+`Connect()`, as Resource notifications require URIs to be specified. Instead, each call to
+`ClientSession.Subscribe()` will start a new `subscriptions/listen` stream for the specified
+Resource. Calling `ClientSession.Unsubscribe()` will cancel that stream.
+On the server side, the server will return `Method not found` (-32601) for `resource/subscribe` and
+`resource/unsubscribe` when `>= 2026-06-30`.
+
+> [!WARNING]
+> Deprecated APIs:
```
yarolegovich, line 293: how long are we going to keep them deprecated and will the new `subscriptions/listen` method fallback to old-protocol version RPCs? 

if the answer to the latter is negative, as an MCP client developer I'm kind of incentivised to keep using `ClientSession.Subscribe()` and `ClientSession.Unsubscribe()` APIs, because this way I write the same code for old and new servers

maciej-kisiel: Our only way to remove deprecated APIs is to release v2. But I hear your argument, maybe it's actually better to hold off with making them deprecated until we have designed how the replacement would look like in v2 and backport it to v1. Deprecation without suggested replacement may not be received well.

design/stateless.md
```diff
@@ -0,0 +1,492 @@
...
+> [!NOTE]
+> Deprecated APIs:
+> For v2, it would make sense to rename `ClientSession` to avoid confusion.
+`ServerSession` should be removed.
+
+### Keeping the session-based APIs
+
+Session objects (`ServerSession`/`ClientSession`) are the core part of the Go MCP SDK API.
+They are returned from `Connect` functions and they are used to interact with the peer.
+Even though SEP-2567 removes the concept of sessions from the protocol, we cannot simply remove
```
jba, line 40: OK, at this point I concluded that it sounds like time to move to v2. It will dramatically simplify both the code and the API.

Normally the downsides of a v2 are:
- it's a challenge to make people aware that it exists
- users have to endure the pain of upgrading, or missing out on new features

I think those are mitigated here. No one who reads those SEPs will be surprised that our SDK (and surely others?) will do a major version bump. It will be easy to advertise, via the README in the repo, other MCP documentation about SDKs, and a doc comment on `StreamableHTTPOptions.Stateless`. And here it's not so much about getting some new features: it's basically a totally new protocol that a server author will either switch to completely, or not.

One downside is maintaining two separate pieces of code. But AFAICT if we put this into v1 we would basically be forking a lot of the code anyway. We can still have a lot of common code by moving it to internal. Common API can be shared in many cases by making aliases in v1 to the v2 types, and wrapping v1 functions around v2 functions, as we recommend in these cases.

Am I missing something?

jba: OK, you want to hold off on v2 for pluggable transports. I withdraw my comment.

jba (commented):
After a high-level reading, I have no comments. This seems like a reasonable path forward.

design/stateless.md
jba: I don't think we can deprecate anything that can still be used in stateful mode: those protocol versions still exist, and may for some time (especially for stdio transports).

Perhaps in the package doc or README, we can suggest that stateless mode is recommended for remote servers because it is more scalable.

v2 will "solve" this problem because it will be all stateless. We'll just need one line in the doc that says to use v1 if you want state.

PR #927, opened 2026-05-02
mcp: use golang.org/x/time/rate for LoggingHandler rate limiting

### LINKED ISSUE #924: LoggingHandler: surface errors and use golang.org/x/time/rate for rate limiting

The `LoggingHandler` in `mcp/logging.go` has two issues:

1. **Errors are silently swallowed**: The `Handle()` method returns an error, but the comment acknowledges it will "probably be ignored":
   ```go
   func (h *LoggingHandler) Handle(ctx context.Context, r slog.Record) error {
       err := h.handle(ctx, r)
       // TODO(jba): find a way to surface the error.
       // The return value will probably be ignored.
       return err
   }
   ```
   When logging fails (e.g., connection closed, server not supporting logging), the caller has no way to detect or handle the failure.

2. **Rate limiting is imprecise**: The current implementation uses a simple time-interval check:
   ```go
   // TODO(jba): use golang.org/x/time/rate.
   skip := time.Since(h.lastMessageSent) < h.opts.MinInterval
   ```
   This is a basic "at most one per interval" approach. It doesn't support burst tolerance, and messages during the cooldown window are silently dropped without any counter or notification.

1. **Error surfacing**: Provide a mechanism for callers to know when log messages fail to send. Options:
   - Add an `OnError` callback to `LoggingHandler` options
   - Return errors through the `slog.Handler` chain in a discoverable way
   - At minimum, log to stderr as a fallback when the MCP logging channel fails

2. **Proper rate limiting**: Replace the interval check with `golang.org/x/time/rate.Limiter`, which supports:
   - Burst tolerance (allow N messages in quick succession)
   - Configurable rate (messages per second)
   - Standard, well-tested implementation

- Document the limitation and let users implement their own rate limiting — but the SDK should handle this since logging is a core MCP feature.
- Use the existing `MinInterval` with a burst counter — but this is reinventing what `rate.Limiter` already provides.

**Additional context**

- File: `mcp/logging.go`, lines 153 and 160
- The SDK already imports `golang.org/x/sync` (via `singleflight` in other packages), so adding `golang.org/x/time/rate` as a dependency is a natural extension.

wucm667:
Replaces the simple time-interval check in `LoggingHandler` with a standard `rate.Limiter` from `golang.org/x/time/rate`, as suggested by the existing TODO comment.

This provides a well-tested implementation and removes the need for manual `lastMessageSent` tracking. The rate (`1/MinInterval`) and burst (1) match the previous behavior exactly.

/cc @jba

Closes #924

### DIFF
--- a/mcp/logging.go
+++ b/mcp/logging.go
@@ -13,6 +13,8 @@ import (
 	"slices"
 	"sync"
 	"time"
+
+	"golang.org/x/time/rate"
 )
 
 // Logging levels.
@@ -83,9 +85,9 @@ type LoggingHandler struct {
 	// Ensures that the buffer reset is atomic with the write (see Handle).
 	// A pointer so that clones share the mutex. See
 	// https://github.com/golang/example/blob/master/slog-handler-guide/README.md#getting-the-mutex-right.
-	mu              *sync.Mutex
-	lastMessageSent time.Time // for rate-limiting
-	buf             *bytes.Buffer
+	mu      *sync.Mutex
+	limiter *rate.Limiter // for rate-limiting
+	buf     *bytes.Buffer
 	handler         slog.Handler
 }
 
@@ -118,6 +120,9 @@ func NewLoggingHandler(ss *ServerSession, opts *LoggingHandlerOptions) *LoggingH
 	}
 	if opts != nil {
 		lh.opts = *opts
+		if opts.MinInterval > 0 {
+			lh.limiter = rate.NewLimiter(rate.Limit(1.0/opts.MinInterval.Seconds()), 1)
+		}
 	}
 	return lh
 }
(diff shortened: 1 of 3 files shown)


mcp/logging.go
```diff
@@ -118,6 +120,9 @@ func NewLoggingHandler(ss *ServerSession, opts *LoggingHandlerOptions) *LoggingH
 	}
 	if opts != nil {
 		lh.opts = *opts
+		if opts.MinInterval > 0 {
+			lh.limiter = rate.NewLimiter(rate.Limit(1.0/opts.MinInterval.Seconds()), 1)
```
guglielmo-san:
```suggestion
			lh.limiter = rate.NewLimiter(rate.Every(opts.MinInterval), 1)
```

jba: agreed

jba (changes requested):

wucm667:
@guglielmo-san @jba Agreed — switched to  as suggested. Pushed the fix.

guglielmo-san (dismissed):

jba (approved):

PR #931, opened 2026-05-04
mcp: align resource not found error code with sep-2164

yarolegovich:
Change the error code returned by ResourceNotFoundError from -32002 to -32602 (jsonrpc.CodeInvalidParams), aligning with [SEP-2164](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2164).

* Not gated with a protocol version check, because the previous error code was never formally specified and the behaviour varied across SDKs.
* In case clients which rely on the old error code exist, the previous behaviour can be restored with `MCPGODEBUG=customresnotfounderrcode=1`.
* `CodeResourceNotFound` deprecated since its only purpose going forward is backward compatibility. The new code should use `jsonrpc.CodeInvalidParams` directly.
* `CodeResourceNotFound` changed to `var` so it can stay in sync with `ResourceNotFoundError` and existing code doing `code == mcp.CodeResourceNotFound` keeps working in both the default and compat modes.

### DIFF
--- a/docs/mcpgodebug.md
+++ b/docs/mcpgodebug.md
@@ -22,6 +22,11 @@
 
 Options listed below were added and will be removed in the 1.8.0 version of the SDK.
 
+- `customresnotfounderrcode` added. If set to `1`, `ResourceNotFoundError` will
+  use the custom error code `-32002` instead of the standard `-32602` (Invalid
+  Params), restoring the previous behavior. The default behavior was changed to
+  align with SEP-2164 and the JSON-RPC specification.
+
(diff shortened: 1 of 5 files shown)


docs/mcpgodebug.md
```diff
@@ -22,6 +22,11 @@
 
 Options listed below were added and will be removed in the 1.8.0 version of the SDK.
 
+- `customresnotfounderrcode` added. If set to `1`, `ResourceNotFoundError` will
```
maciej-kisiel, line 25: 1.6.0 is already in pre-release, you can create a new section for 1.7.0 above.

maciej-kisiel (approved):

PR #939, opened 2026-05-06
mcp: OIDC-Flavored Refresh Token Guidance

### LINKED ISSUE #940: Implement SEP-2207: OIDC-Flavored Refresh Token Guidance
This issue tracks implementation of https://modelcontextprotocol.io/seps/2207-oidc-refresh-token-guidance

guglielmo-san:
## Description

Implemets SEP-2207 (https://modelcontextprotocol.io/seps/2207-oidc-refresh-token-guidance)

Fixes #940

### DIFF
--- a/auth/authorization_code.go
+++ b/auth/authorization_code.go
@@ -90,6 +90,20 @@ type AuthorizationCodeHandlerConfig struct {
 	// See [AuthorizationCodeFetcher] for details.
 	AuthorizationCodeFetcher AuthorizationCodeFetcher
 
+	// RequestRefreshToken indicates that the client intends to use refresh
+	// tokens and is capable of storing them securely.
+	//
+	// When true and the Authorization Server metadata contains "offline_access"
+	// in its scopes_supported, the client adds "offline_access" to the
+	// requested scopes.
+	//
+	// When using Dynamic Client Registration, callers should include
+	// "refresh_token" in [DynamicClientRegistrationConfig].Metadata.GrantTypes
+	// directly to advertise refresh token support to the Authorization Server.
+	//
+	// See https://modelcontextprotocol.io/seps/2207-oidc-refresh-token-guidance.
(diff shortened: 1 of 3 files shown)


auth/authorization_code.go
```diff
@@ -90,6 +90,20 @@ type AuthorizationCodeHandlerConfig struct {
...
 	AuthorizationCodeFetcher AuthorizationCodeFetcher
 
+	// RequestRefreshToken indicates that the client intends to use refresh
+	// tokens and is capable of storing them securely.
+	//
+	// When true and the Authorization Server metadata contains "offline_access"
+	// in its scopes_supported, the client adds "offline_access" to the
+	// requested scopes.
+	//
+	// When using Dynamic Client Registration, callers should include
```
maciej-kisiel, line 100: Should we also add an equivalent comment for Client ID Metadata Document users?

auth/authorization_code.go
```diff
@@ -90,6 +90,20 @@ type AuthorizationCodeHandlerConfig struct {
...
+	// "refresh_token" in [DynamicClientRegistrationConfig].Metadata.GrantTypes
+	// directly to advertise refresh token support to the Authorization Server.
+	//
+	// See https://modelcontextprotocol.io/seps/2207-oidc-refresh-token-guidance.
+	RequestRefreshToken bool
```
maciej-kisiel, line 108: It would be good to verify across SEPs that we don't require any more scopes related settings. If we do, maybe we should consider grouping them?

auth/authorization_code.go
guglielmo-san: Right 👍

auth/authorization_code.go
guglielmo-san: On remaining Authorization SEPs it should not be required to add any extra scope related setting

maciej-kisiel (approved):

PR #944, opened 2026-05-06
mcp: Add client-side scope accumulation

### LINKED ISSUE #943: Implement SEP-2350: Clarify client-side scope accumulation in step-up authorization
This issue tracks the implementation of https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2350

guglielmo-san:
## Description

This PR implements https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2350

Fixes #943

### DIFF
--- a/auth/authorization_code.go
+++ b/auth/authorization_code.go
@@ -91,6 +91,23 @@ type AuthorizationCodeHandlerConfig struct {
 	// See [AuthorizationCodeFetcher] for details.
 	AuthorizationCodeFetcher AuthorizationCodeFetcher
 
+	// RequestRefreshToken indicates that the client intends to use refresh
+	// tokens and is capable of storing them securely.
+	//
+	// When true and the Authorization Server metadata contains "offline_access"
+	// in its scopes_supported, the client adds "offline_access" to the
+	// requested scopes.
+	//
+	// When using Dynamic Client Registration, callers should include
+	// "refresh_token" in [DynamicClientRegistrationConfig].Metadata.GrantTypes
+	// directly to advertise refresh token support to the Authorization Server.
+	//
+	// When using Client ID Metadata Document, the document hosted at the
+	// Client ID URL should include "refresh_token" in its grant_types.
+	//
+	// See https://modelcontextprotocol.io/seps/2207-oidc-refresh-token-guidance.
+	RequestRefreshToken bool
+
 	// Client is an optional HTTP client to use for HTTP requests.
 	// It is used for the following requests:
 	//  - Fetching Protected Resource Metadata
@@ -274,6 +291,13 @@ func (h *AuthorizationCodeHandler) Authorize(ctx context.Context, req *http.Requ
 	scps = unionScopes(h.requestedScopes, scps)
 	h.requestedScopes = scps
 	h.mu.Unlock()
+	// SEP-2207: when the client desires refresh tokens and the Authorization
+	// Server advertises offline_access support, add it to the requested scopes.
+	if h.config.RequestRefreshToken &&
+		slices.Contains(asm.ScopesSupported, "offline_access") &&
+		!slices.Contains(scps, "offline_access") {
+		scps = append(scps, "offline_access")
+	}
 
 	cfg := &oauth2.Config{
 		ClientID:     resolvedClientConfig.clientID,
--- a/auth/authorization_code.go
+++ b/auth/authorization_code.go
@@ -339,15 +339,10 @@ func unionScopes(existing, challenged []string) []string {
 	if len(challenged) == 0 {
 		return existing
 	}
-	seen := make(map[string]struct{}, len(existing))
 	result := make([]string, len(existing), len(existing)+len(challenged))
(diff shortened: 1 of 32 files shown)


auth/extauth/client_credentials.go
```diff
@@ -47,6 +48,9 @@ type ClientCredentialsHandlerConfig struct {
 type ClientCredentialsHandler struct {
 	config      *ClientCredentialsHandlerConfig
 	tokenSource oauth2.TokenSource
+
+	mu              sync.Mutex
+	requestedScopes []string
```
maciej-kisiel: I think I would adjust the name to something like `grantedScopes`. It will better fit the semantics adjustment I describe below.

I also wonder if they shouldn't be keyed by Authorization Server.

auth/extauth/client_credentials.go
```diff
@@ -127,6 +131,14 @@ func (h *ClientCredentialsHandler) Authorize(ctx context.Context, req *http.Requ
 		scopes = prm.ScopesSupported
 	}
 
+	// Accumulate scopes: union previously requested scopes with the newly
+	// challenged scopes so that step-up authorization does not lose
+	// permissions granted in earlier rounds (SEP-2350).
+	h.mu.Lock()
+	scopes = unionScopes(h.requestedScopes, scopes)
+	h.requestedScopes = scopes
```
maciej-kisiel: I think we should only update the list once the new scopes were authorized.

Also, currently the OAuth handlers are not concurrent safe. Not sure this single mutex makes sense 😄

auth/extauth/client_credentials.go
```diff
@@ -229,6 +241,25 @@ func scopesFromChallenges(cs []oauthex.Challenge) []string {
 	return nil
 }
 
+// unionScopes returns the union of existing and challenged scopes,
+// preserving order (existing first, then new challenged scopes).
+func unionScopes(existing, challenged []string) []string {
```
maciej-kisiel: If union is the main operation, maybe it's worth to store it as a map?

auth/extauth/client_credentials.go
guglielmo-san: The SEP is not very clear in this sense, but I agree that it makes more sense to keep asking for the granted scopes.

> Determine required scopes by computing the union of the **client's previously requested scope** set and the scopes from the current challenge.

New implementation
1. requestedScopes are merged with previously grantedScopes
2. If Auth succeeds and a token is issued, extract from the token the scopes granted to it
3. Update the grantedScopes with the ones granted on the last issues token

auth/extauth/client_credentials.go
guglielmo-san: I prefer to keep consistent the requestedScopes and the grantedScopes, both as list string. But updated the unionScopes.

auth/shared.go
```diff
@@ -67,3 +70,27 @@ func authorizationServerMetadataURLs(issuerURL string) []string {
 
 	return urls
 }
+
+// UnionScopes returns the union of the existing and challenged scope sets.
+// It is used during step-up authorization to accumulate scopes across
+// authorization rounds (SEP-2350).
+func UnionScopes(existing, challenged []string) []string {
```
maciej-kisiel: I don't think we want to export these functions publicly. Maybe it's a good moment to start the `internal/authutils` package (or something of similar name) to put such things there? It would also allow us to remove the current duplication we discussed.

auth/extauth/client_credentials.go
```diff
@@ -47,6 +47,8 @@ type ClientCredentialsHandlerConfig struct {
 type ClientCredentialsHandler struct {
 	config      *ClientCredentialsHandlerConfig
 	tokenSource oauth2.TokenSource
+
+	grantedScopes map[string][]string
```
maciej-kisiel, line 53: Please document what the keys and values are.

auth/authorization_code.go
```diff
@@ -128,6 +128,8 @@ type AuthorizationCodeHandler struct {
 
 	// tokenSource is the token source to use for authorization.
 	tokenSource oauth2.TokenSource
+
+	grantedScopes map[string][]string
```
maciej-kisiel, line 134: Please document what the keys and values are.

auth/authorization_code.go
```diff
@@ -558,3 +573,20 @@ func (h *AuthorizationCodeHandler) exchangeAuthorizationCode(ctx context.Context
 	h.tokenSource = cfg.TokenSource(clientCtx, token)
 	return nil
 }
+
+// updateGrantedScopes updates the granted scopes based on the token source and requested scopes
```
maciej-kisiel:
```suggestion
// updateGrantedScopes updates the granted scopes based on the token source and requested scopes.
```

auth/shared.go
guglielmo-san: Created `internals/authutil`
Will proceed with the refactoring of duplicated methods in a separate PR

maciej-kisiel (approved):

PR #946, opened 2026-05-06
mcp: add optional issuer validator for pre-registered client validation (SEP-2352)

### LINKED ISSUE #978: Implement SEP-2352
This issue tracks the implementation of SEP-2352 https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2352

guglielmo-san:
Fixes #978

### DIFF
--- a/auth/authorization_code_test.go
+++ b/auth/authorization_code_test.go
@@ -518,6 +518,51 @@ func TestHandleRegistration(t *testing.T) {
 			},
 		},
 		{
+			name: "Preregistered_IssuerMatch",
+			serverConfig: &oauthtest.RegistrationConfig{
+				PreregisteredClients: map[string]oauthtest.ClientInfo{
+					"pre_client_id": {
+						Secret: "pre_client_secret",
+					},
+				},
+			},
+			handlerConfig: &AuthorizationCodeHandlerConfig{
+				PreregisteredClient: &oauthex.ClientCredentials{
+					ClientID: "pre_client_id",
+					ClientSecretAuth: &oauthex.ClientSecretAuth{
+						ClientSecret: "pre_client_secret",
+					},
+					Issuer: "", // set dynamically in the test
+				},
+			},
+			want: &resolvedClientConfig{
+				registrationType: registrationTypePreregistered,
+				clientID:         "pre_client_id",
+				clientSecret:     "pre_client_secret",
+				authStyle:        oauth2.AuthStyleInParams,
+			},
+		},
+		{
+			name: "Preregistered_IssuerMismatch",
+			serverConfig: &oauthtest.RegistrationConfig{
+				PreregisteredClients: map[string]oauthtest.ClientInfo{
+					"pre_client_id": {
+						Secret: "pre_client_secret",
+					},
+				},
+			},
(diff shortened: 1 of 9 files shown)


internal/authutil/util.go
```diff
@@ -0,0 +1,19 @@
...
+// IssuersEqual reports whether two OAuth 2.0 authorization server issuer
+// identifiers refer to the same server. The comparison ignores a single
+// trailing slash, matching the tolerance applied during RFC 8414 Section 3.3
+// metadata validation.
+//
+// This helper is not appropriate for comparing the RFC 9207 iss response
+// parameter, which the MCP authorization spec requires to be compared without
+// any normalization.
+func IssuersEqual(a, b string) bool {
+	return strings.TrimRight(a, "/") == strings.TrimRight(b, "/")
```
yarolegovich: TrimRight will trim all slashes, let's use TrimsSuffix

auth/authorization_code_test.go
```diff
@@ -660,6 +730,15 @@ func TestHandleRegistration(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			s := oauthtest.NewFakeAuthorizationServer(oauthtest.Config{RegistrationConfig: tt.serverConfig})
 			s.Start(t)
+			// For the IssuerMatch test, set the Issuer to the actual server URL.
+			if tt.name == "Preregistered_IssuerMatch" {
+				tt.handlerConfig.PreregisteredClient.Issuer = s.URL()
+			}
+			// For the trailing-slash variant, append a trailing slash so the
+			// configured issuer differs only in normalization from asm.Issuer.
+			if tt.name == "Preregistered_IssuerMatchTrailingSlash" {
```
yarolegovich: I'd rather add test case flags like:
```go
issuerMatch bool
issuerSuffix string
```
and in test body do
```go
if tt.issuerMatch {
  tt.handlerConfig.PreregisteredClient.Issuer = s.URL() + tt.issuerSuffix
}
```

internal/authutil/util.go
guglielmo-san: Ok, also updated the code in auth_meta to follow same logic... it is a behavioral change but it was not yet added to any release

internal/authutil/util.go
```diff
@@ -0,0 +1,13 @@
...
+// that can be found in the LICENSE file.
+
+package authutil
+
+import "strings"
+
+// IssuersEqual reports whether two OAuth 2.0 authorization server issuer
+// identifiers refer to the same server comparing them without the final trailing slash.
+func IssuersEqual(a, b string) bool {
+	return strings.TrimSuffix(a, "/") == strings.TrimSuffix(b, "/")
```
yarolegovich, line 12: since [issuer is a url](https://datatracker.ietf.org/doc/html/rfc8414#section-2) it it's probably more correct to perform a case-insensitive comparison here

yarolegovich (approved):

internal/authutil/util.go
guglielmo-san: I think it would be better to keep the URIs comparison case-sensitive https://datatracker.ietf.org/doc/html/rfc3986#section-6.2.2.1

PR #947, opened 2026-05-07
mcp: add SupportedVersions field to ServerOptions

### LINKED ISSUE #945: Set list of supported versions
Right now, the SDK hardcodes the supported spec versions https://github.com/modelcontextprotocol/go-sdk/blob/main/mcp/shared.go#L43. However, we do not want to support 2024-11-05 but there is no way to configure that in the SDK so that it returns "Unsupported protocol" -32602 error. 

I suggest mcp.ServerOptions has a "SupportedVersions" slice that accepts a list of version enums. The zero value can fall back to the current default list.

Right now the workaround i can think of is having to intercept the initialization request and returning the unsupported protocol error

wucm667:
Fixes #945

Allow users to configure which MCP protocol versions the server supports,
instead of using the hardcoded `supportedProtocolVersions` list.

## Usage

```go
srv := mcp.NewServer(&mcp.Implementation{...}, &mcp.ServerOptions{
    SupportedVersions: []string{"2025-11-25", "2025-06-18"},
})
```

The zero value falls back to the default set of supported versions.

### DIFF
--- a/mcp/server.go
+++ b/mcp/server.go
@@ -41,8 +41,9 @@ const DefaultPageSize = 1000
 // sessions by using [Server.Run].
 type Server struct {
 	// fixed at creation
-	impl *Implementation
-	opts ServerOptions
+	impl              *Implementation
+	opts              ServerOptions
+	supportedVersions []string
 
 	mu                      sync.Mutex
 	prompts                 *featureSet[*serverPrompt]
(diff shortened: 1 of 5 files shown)


guglielmo-san:
Duplicate of PR #1199

PR #948, opened 2026-05-07
mcp: remove support for batching on streamable connections

### LINKED ISSUE #911: Remove support for batching on streamable connection
Support for batching can be removed as it is a deprecated feature

wucm667:
Fixes #911

Batching was already deprecated and rejected for protocol versions >= 2025-06-18.
Remove it entirely to simplify the streamable transport code.

Replaces `readBatch` calls with single message decoding, removes batch-related
checks and comments, and deletes batch-specific tests. No behavioral change for
single-message requests.

### DIFF
--- a/mcp/streamable.go
+++ b/mcp/streamable.go
@@ -439,21 +439,19 @@ func (h *StreamableHTTPHandler) ephemeralConnectOpts(req *http.Request) (*epheme
 	}
 	req.Body.Close()
 	req.Body = io.NopCloser(bytes.NewBuffer(body))
-	msgs, _, err := readBatch(body)
+	msg, err := jsonrpc2.DecodeMessage(body)
 	if err == nil {
-		for _, msg := range msgs {
-			if r, ok := msg.(*jsonrpc.Request); ok {
-				switch r.Method {
-				case methodInitialize:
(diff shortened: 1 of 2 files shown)


maciej-kisiel:
Thanks for the contribution. We're still analyzing our backwards compatibility stance on this and when exactly we would like to remove it. I'll update you when this is resolved.

piyushbag:
Conceptually this looks right for #911: dropping `readBatch` and the batch-only branches simplifies `servePOST` without changing the single-message path.

GitHub shows the branch as conflicting with current `main` (recent streamable changes landed since May). Could you rebase onto latest `main` and push? Happy to take another look once CI is green.

After rebase, one thing I'd sanity-check: any remaining callers that still assumed batch decode on the stateless POST probe path.

wucm667:
Rebased onto latest main and pushed. All CI checks are green.

Regarding the stateless POST probe path — ephemeralConnectOpts now uses jsonrpc2.DecodeMessage directly instead of readBatch, so there's no batch decode assumption left.

@maciej-kisiel mentioned they're still evaluating the backwards compatibility stance. Let me know if you have any other concerns.

PR #949, opened 2026-05-07
mcp: refactor streamable handler control flow

maciej-kisiel:


jba (dismissed):

mcp/streamable.go
```diff
@@ -260,284 +260,338 @@ func (h *StreamableHTTPHandler) ServeHTTP(w http.ResponseWriter, req *http.Reque
...
-		// Verify the 'Origin' header to protect against CSRF attacks.
 		if err := h.opts.CrossOriginProtection.Check(req); err != nil {
 			http.Error(w, err.Error(), http.StatusForbidden)
 			return
 		}
 	}
 
-	// Validate 'Content-Type' header.
-	if req.Method == http.MethodPost && baseMediaType(req.Header.Get("Content-Type")) != "application/json" {
+	// [§2.7] of the spec (2025-06-18): validate the MCP-Protocol-Version
```
yarolegovich, line 269: we also support versions earlier than `protocolVersion20250618`. is it safe to skip the check because there was no streamable transport?

mcp/streamable.go
```diff
@@ -260,284 +260,338 @@ func (h *StreamableHTTPHandler) ServeHTTP(w http.ResponseWriter, req *http.Reque
...
-			userID:    userID,
-		}
-
-		if stateless {
-			// Stateless mode: close the session when the request exits.
-			defer session.Close() // close the fake session after handling the request
-		} else {
-			// Otherwise, save the transport so that it can be reused
+		defer session.Close()
+		transport.ServeHTTP(w, req)
```
yarolegovich, line 548: `transport.connection.toolLookup = server.getServerTool` is not called in this early return. I'd suggest to encapsulate it into `connectStreamable()` or a similar helper

mcp/streamable.go
maciej-kisiel: Not sure I understand the question, but the header was introduced in `protocolVersion20250618`. We validate it early if it's present. If this is an `initialize` request for an older version, the value will be negotiated as part of the handshake.

mcp/streamable.go
maciej-kisiel: Done.

mcp/streamable.go
yarolegovich: nvm, misread the check

yarolegovich (approved):

PR #950, opened 2026-05-08
feat: multi-round-trip request implementation (SEP-2322)

yarolegovich:
### Context

Design (`design/mrtr.md`) and implementation proposal for Multi Round-Trip Requests (MRTR) per [SEP-2322](https://github.com/CaitieM20/modelcontextprotocol/blob/de6d76fba3078eda957dadb3cec51ca8ab851b5c/seps/2322-MRTR.md).

### Changes

- Introduce `InputRequest` and `InputResponse` sealed interfaces with corresponding `InputRequestMap` and `InputResponseMap` types for custom map JSON codec implementation.
- Extend `CallToolParams`, `GetPromptParams`, and `ReadResourceParams` with `InputResponses`, `RequestState` and an unexported `resultType` ("complete" or "input_required") for retry round-trips. The type is set by the SDK based on input requests presence. 
- Add client-side MRTR middleware enabled by default that automatically fulfills input requests and retries.
- Added MRTROptions on ClientOptions to allow disabling the middleware and configuring the number of retries.
- Add server-side MRTR middleware for backward compatibility: transparently bridges input requests to direct Elicit/CreateMessage/ListRoots calls for older clients

### DIFF
--- a/design/mrtr.md
+++ b/design/mrtr.md
@@ -80,12 +80,13 @@ mcp.AddTool(s, tool, func(ctx context.Context, req *mcp.CallToolRequest, in MyIn
   return &mcp.CallToolResult{Content: []mcp.Content{&mcp.TextContent{Text: "done"}}}, myOut, nil
 })
 ```
-An incomplete result should be converted to an error response when communicating with an older client. This is a reasonable default for server implementers to not branch by protocol version in their code.
+The SDK validates at runtime that a handler does not return both content and `InputRequests` — doing so logs a warning and returns a `CodeInternalError` JSON-RPC error.
 
-An unexported middleware is installed in the client, which similarly to `urlElicitationMiddleware` will automatically invoke handlers for the corresponding methods on incomplete results and retry the original request. `ClientOptions` will be extended with configuration knobs:
+An unexported receiving middleware is installed on the server for backward compatibility with older clients. When a handler returns `InputRequests` and the connected client uses a protocol version that does not support MRTR, the middleware fulfills the requests by calling `ServerSession.Elicit`/`CreateMessage`/`ListRoots` on the client directly and reinvokes the handler once with the collected `InputResponses`. If any of these calls fail, the entire request fails. Input requests are fulfilled concurrently. This lets server developers write protocol-version-independent code.
+
+An unexported sending middleware is installed on the client, which similarly to `urlElicitationMiddleware` will automatically invoke handlers for the corresponding methods on incomplete results and retry the original request. `ClientOptions` is extended with configuration knobs:
 ```go
 type MRTROptions struct {
-  ...
   MaxRetries int
   Disabled   bool
 }
@@ -94,24 +95,25 @@ client := mcp.NewClient(impl, &mcp.ClientOptions{
 })
 ```
 
-Alternatively, clients have an option to disable it and write a retry loop manually:
+Alternatively, clients have an option to disable it and write a retry loop manually using `NeedsInput()`:
 ```go
 client := mcp.NewClient(impl, &mcp.ClientOptions{MRTR: &mcp.MRTROptions{Disabled: true}})
 result, err := client.CallTool(ctx, &mcp.CallToolParams{Name: "my-tool"})
-if result.InputRequests != nil { ... }
+if result.NeedsInput() { ... }
 ```
 
-This is error-prone because according to the SEP an empty map can be returned for load-shedding purposes. `NeedsInput` function can be provided, but nothing would force users to use it. 
+`NeedsInput()` checks the unexported `resultType` field rather than `InputRequests`, correctly handling the load-shedding case where the server returns `input_required` with an empty map.
 
 **Pros**
 
 This is arguably the simplest and the most transparent approach which is also closest to the spec.
 What gets explicitly set on the server can be observed on the wire and on the client. 
-The opt-out middleware follows the principle of the least surprise for app developers. If client method handlers were provided they will continue to be invoked regardless of the protocol version in use. The `Disabled` option lets "power-users" build any custom handling logic. 
+The opt-out client middleware follows the principle of the least surprise for app developers. If client method handlers were provided they will continue to be invoked regardless of the protocol version in use. The `Disabled` option lets "power-users" build any custom handling logic.
+The server middleware makes handler code protocol-version-independent — the same handler works for both old and new clients.
 
 **Cons**
 
-The biggest downside of the proposal is that server developers can construct incorrect responses and this will only be validated at runtime.
+The biggest downside of the proposal is that server developers can construct incorrect responses (both content and input requests) and this will only be validated at runtime.
 
 ## Alternatives considered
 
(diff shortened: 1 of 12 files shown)


design/mrtr.md
```diff
@@ -0,0 +1,297 @@
...
+type CallToolParams struct {
+  ...
+  InputResponses InputResponseMap `json:"inputResponses,omitempty"`
+  RequestState  string            `json:"requestState,omitempty"`
+}
+// Same for GetPromptParams, ReadResourceParams
+```
+
+`InputRequests` and `RequestState` fields are added directly to `CallToolResult`, `GetPromptResult`, and `ReadResourceResult` as exported.
+Result type discriminator (completed, input_required) is unexported so that SDK users don't need to set it to the correct constant in addition to setting either `Content` or `InputRequests`. Handler execution result is validated and augmented before marshaling:
```
maciej-kisiel, line 59: How would the error be propagated to the tool developer?

design/mrtr.md
```diff
@@ -0,0 +1,297 @@
...
+    }, zero, nil
+  }
+  return &mcp.CallToolResult{Content: []mcp.Content{&mcp.TextContent{Text: "done"}}}, myOut, nil
+})
+```
+An incomplete result should be converted to an error response when communicating with an older client. This is a reasonable default for server implementers to not branch by protocol version in their code.
+
+An unexported middleware is installed in the client, which similarly to `urlElicitationMiddleware` will automatically invoke handlers for the corresponding methods on incomplete results and retry the original request. `ClientOptions` will be extended with configuration knobs:
+```go
+type MRTROptions struct {
```
maciej-kisiel: I'm not sure I would introduce these knobs from the start. If the user need is real, someone will come to use and ask about it. We can always add them later in a backwards compatible manner.

design/mrtr.md
```diff
@@ -0,0 +1,297 @@
...
+
+// InputRequiredResult struct exists for backward-compatibility in case of new fields being needed for input request results.
+type InputRequiredResult struct {
+  InputRequests InputRequestMap
+  RequestState  string
+}
+
+// RequireInput constructs a tool call, prompt or resource result with input requests set.
+// mrtrResult provides methods for setting private fields on these types.
+func RequireInput[T any, TP interface { *T; mrtrResult }](r InputRequiredResult) TP { ... }
```
maciej-kisiel, line 134: We can always introduce this even with exported fields, as a "util" that guarantees correctness. But likely not from the start.

design/mrtr.md
```diff
@@ -0,0 +1,297 @@
...
+  MaxRetries: 5,
+}))
+```
+This would change semantics of `*Handler` fields - depending on the protocol version in use, an extra initialization step will be required for them to "take effect".
+
+---
+
+### Protocol version bridging
+
+When a handler returns an input-required result to an old client the SDK could bridge by invoking `ServerSession.Elicit`/`CreateMessage`/`ListRoots` on the `ServerSession` and re-invoking the handler with the collected `inputResponses`:
```
maciej-kisiel: I think this behavior would be valuable. Without it, the server creators are not incentivized to use the new MRTR format because migrating would mean they will stop working with all older clients. It would slow down the adoption of the new protocol version.

design/mrtr.md
```diff
@@ -0,0 +1,297 @@
...
+    if err != nil || !needsInput(res) || isNewProtocol(req) {
+      return res, err
+    }
+    // Fulfill input requests via server-initiated calls, inject inputResponses into the request, 
+    // re-invoke the handler.
+  }
+}
+```
+
+Or even more radically: instead of making `ServerSession.Elicit`, `CreateMessage`, and `ListRoots` return errors for newer clients, we could convert these invocations into MRTR wire format transparently: suspend the handler, return `input_required`, and resume when the client retries with `inputResponses`:
```
maciej-kisiel: I agree that this should not be done.

maciej-kisiel:
I think the main question we should ask ourselves is how we would design the API for this if there was no backwards compatibility requirement.

If we have an approach that will likely be chosen for v2, sometimes it may make sense to move toward its direction, even if it has some disadvantages if they would disappear with v2 release.

design/mrtr.md
yarolegovich: Don't object. We can start with a reasonable default of max retries and add configurability on request

design/mrtr.md
yarolegovich: The question is whether we should build it from the start or add later if there's a demand for it?

design/mrtr.md
yarolegovich: That's a good question. My instinct was to treat it similarly to `json.Marshal` failure on the write path, but there's however an interesting discrepancy in how it's handled depending on where in the stack it occurs: 
1. [Here](https://github.com/modelcontextprotocol/go-sdk/blob/main/mcp/server.go#L381) it's returned as an `fmt.Errorf -> jsonrpc error` to clients.
2. [Here](https://github.com/modelcontextprotocol/go-sdk/blob/main/internal/jsonrpc2/conn.go#L667) it's just written to an error log through [onInternalError handler](https://github.com/modelcontextprotocol/go-sdk/blob/main/mcp/transport.go#L190). The result is not returned to clients.  

(1) exposes some implementation details but is more "visible" and is likely harmless. My vote would be a combination - log an error and return a jsonrpc server error to clients.

design/mrtr.md
yarolegovich: imo at least with the current data model and `resultType` auto-populated the utility would be very shallow. I would not be considering it unless we get an inflow of issues where people try to set `InputRequests` together with `Content` or other fields

yarolegovich:
> I think the main question we should ask ourselves is how we would design the API for this if there was no backwards compatibility requirement.

I don't think my proposal would change significantly if we didn't need backward compatibility. 

I would want a sealed interface return type for `*Handler` functions and separate types for complete and incomplete result. On the client side I would probably leave a handler registration API through `ClientOptions`. 

If we make the client API return a sealed interface as well the "base case" becomes overly verbose - all clients need to unpack result to a specific type and handle it accordingly, even though most of the time retry is auto-handled by the middleware. 

Now this breaks the proposed client side "manual handling" where if you disable MRTR through options you can just inspect `*Result.InputRequests` and take action. I don't think it's a big problem though because users can just disable the auto-middleware and write their own.

maciej-kisiel:
I think it's also worth adding a section on what should happen with the existing methods that are replaced with MRTR.

design/mrtr.md
yarolegovich: Implemented: [2f6ee7f](https://github.com/modelcontextprotocol/go-sdk/pull/950/commits/2f6ee7fa643be7e66ce622a9525f47be2221e7a5)

design/mrtr.md
yarolegovich: Decided to provide `MaxRetries` configuration from the start.

mcp/mrtr.go
```diff
@@ -0,0 +1,318 @@
...
+const defaultMRTRMaxRetries = 3
+
+// MRTROptions configures the client-side MRTR (Multi Round-Trip Requests)
+// middleware. The middleware is enabled by default and automatically fulfills input
+// requests from the server by invoking the appropriate client handlers and
+// retrying the original call.
+type MRTROptions struct {
+	// MaxRetries is the maximum number of MRTR retry attempts after the
+	// initial call. Defaults to 3 if the provided value is <= 0.
+	MaxRetries int
```
jba: It looks like you're giving the client  a  fixed number of chances to satisfy the input requests. I'm not sure I understand why. I would say you should keep retrying as long as the client is making progress, by providing the answer to at least one request.

mcp/mrtr.go
```diff
@@ -0,0 +1,318 @@
...
+	"golang.org/x/sync/errgroup"
+)
+
+const defaultMRTRMaxRetries = 3
+
+// MRTROptions configures the client-side MRTR (Multi Round-Trip Requests)
+// middleware. The middleware is enabled by default and automatically fulfills input
+// requests from the server by invoking the appropriate client handlers and
+// retrying the original call.
+type MRTROptions struct {
```
jba: Do you know if other SDKs are using "MRTR" or do they use something else? I'm not suggesting spelling out the full phrase, but maybe MultiRequest, etc. I think MRTR is fine if other SDKs don't have anything better.

mcp/protocol.go
```diff
@@ -13,6 +13,172 @@ import (
 	"github.com/modelcontextprotocol/go-sdk/internal/mcpgodebug"
 )
 
+// ResultType indicates whether a result is complete or requires further input
+// from the client via the MRTR (Multi Round-Trip Requests) protocol.
+type ResultType string
```
jba: Does this need to be exported?

mcp/protocol.go
```diff
@@ -13,6 +13,172 @@ import (
...
+	// ResultTypeInputRequired indicates the server needs additional client
+	// input before it can complete the request. The client should fulfill the
+	// InputRequests and retry the call with the responses.
+	ResultTypeInputRequired ResultType = "input_required"
+)
+
+// InputRequest is a sealed interface for parameters that a server can include
+// in an MRTR input-required result. Implementations are [*ElicitParams],
+// [*CreateMessageParams], and [*ListRootsParams].
+type InputRequest interface{ isInputRequest() }
```
jba, line 34: We usually don't mention that something is an interface, and the word "sealed" is not a standard Go term. Just say what it is. Example:

> Params is a parameter (input) type for an MCP call or notification.

mcp/protocol.go
```diff
@@ -13,6 +13,172 @@ import (
...
+			result[k] = &p
+		default:
+			return fmt.Errorf("unsupported InputRequest method: %q", raw.Method)
+		}
+	}
+	*m = result
+	return nil
+}
+
+// InputResponse is a sealed interface for results that a client sends back
```
jba: ditto

mcp/mrtr.go
yarolegovich: I added it mostly for the "load shedding" scenario where a server can return an empty requests map and a client just needs to retry the original request. Ideally we should be retrying with some jitterred exponential backoff, but implementing it seems premature at this stage. 

The default 3 is probably overly aggressive, but I also don't want to leave it uncapped. One option is to only apply the cap for the empty-request-map case. For this case I think it should also be fine to remove the exported `MaxRetries` field.

guglielmo-san (dismissed):

mcp/mrtr.go
jba: Ah yes, I'd forgotten about that case. Removing the option makes sense to me.

mcp/mrtr.go
yarolegovich: I'll check with other maintainers on discord, but I picked the abbreviation because it's a part of the SEP file name: [2322-MRTR.md](https://github.com/CaitieM20/modelcontextprotocol/blob/de6d76fba3078eda957dadb3cec51ca8ab851b5c/seps/2322-MRTR.md)

yarolegovich: Renamed to `MultiRoundTrip`

mcp/mrtr.go
yarolegovich: Removed `MaxRetries` option, left 3 as the default max retries for empty input map (without resets), set 10 as the maximum number of retries overall. I think 10 is reasonably high and don't feel comfortable leaving an uncapped retry loop

jba (approved):

guglielmo-san (approved):

PR #956, opened 2026-05-13
mcp: Allow registration of custom JSON-RPC methods

### LINKED ISSUE #954: Allow registration of custom JSON-RPC methods

I'm contributing about [SEP-2624](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2624) in [modelcontextprotocol/experimental-ext-interceptors](https://github.com/modelcontextprotocol/experimental-ext-interceptors), which implements interceptors for MCP protocols. This feature requires extending the protocol with custom JSON-RPC methods.

Although the Golang SDK provides a solid interface, it currently does not allow to extend handlers for custom JSON-RPC methods.

The `mcp` package can expose following interfaces

- Add methods to both the `Client` and `Server` to register receiving and sending custom methods. For example:
  - `AddReceivingCustomMethod[P paramsPtr[T], R Result, T any](method string, handler typedServerMethodHandler[P, R])`
  - `AddSendingCustomMethod[P paramsPtr[T], R Result, T any](method string, handler typedClientMethodHandler[P, R])`
- Export other internal components that are necessary to implement these custom handlers, such as `Params.isParams`, `Result.isResult`, `typedServerMethodHandler`, `typedClientMethodHandler`, `paramsPtr`, etc.

N/A

**Additional context**

For reference, we have already implemented AddReceivingCustomMethod for the server in a vendored version of the library. You can view the implementation details here:

- [AddReceivingCustomMethod](https://github.com/modelcontextprotocol/experimental-ext-interceptors/blob/5bd7ab4d2aa6878c3d064c76961bcb1c16d01d93/go/sdk/vendor/github.com/modelcontextprotocol/go-sdk/mcp/server.go#L60-L105) on the server
- [merging part](https://github.com/modelcontextprotocol/experimental-ext-interceptors/blob/5bd7ab4d2aa6878c3d064c76961bcb1c16d01d93/go/sdk/vendor/github.com/modelcontextprotocol/go-sdk/mcp/server.go#L1438-L1443)

---

happy to contribute once agreed.

guglielmo-san:
## Description

This PR introduces support for custom (non-standard) JSON-RPC methods in the MCP Go SDK. 
This feature enables servers to handle arbitrary JSON-RPC methods and clients to invoke them, all while maintaining full type safety and integrating seamlessly with the SDK's existing middleware and session management.

- Server-Side Registration: Added AddReceivingCustomMethod to allow servers to register handlers for custom methods. Incoming requests are automatically unmarshaled into user-defined parameter structs.
- Client-Side Calling: Added AddSendingCustomMethod which returns a strongly-typed caller function that clients can use to invoke custom methods and receive strongly-typed results.
- Helper Structs: Provided ParamsBase and ResultBase for users to embed in their custom structs. This allows them to easily satisfy the SDK's internal Params and Result interfaces.

Fixes #954


mcp/client.go
```diff
@@ -945,7 +949,13 @@ var clientMethodInfos = map[string]methodInfo{
 }
 
 func (cs *ClientSession) sendingMethodInfos() map[string]methodInfo {
-	return serverMethodInfos
+	if len(cs.client.customSendMethods) == 0 {
+		return serverMethodInfos
+	}
+	infos := make(map[string]methodInfo, len(serverMethodInfos)+len(cs.client.customSendMethods))
+	maps.Copy(infos, serverMethodInfos)
```
maciej-kisiel: It feels a bit unnecessary to construct the map on every call. Maybe it's worth moving the source of truth map containing all methods under an appropriate type (`Client`?).

mcp/client.go
```diff
@@ -1218,3 +1228,33 @@ func paginate[P listParams, R listResult[T], T any](ctx context.Context, params
...
+		newResult: func() Result {
+			return reflect.New(reflect.TypeFor[R]().Elem()).Interface().(R)
+		},
+	}
+
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	c.customSendMethods[method] = mi
+
+	return func(ctx context.Context, cs *ClientSession, params P) (R, error) {
```
maciej-kisiel: I'm not sure exposing this via a return value will cover all the use cases. Most of the users of this will probably be extensions, which will do the registration for the developer. Not sure how easy it will be to propagate this.

mcp/streamable.go
```diff
@@ -396,6 +396,7 @@ func connectStreamable(ctx context.Context, server *Server, transport *Streamabl
 	if err != nil {
 		return nil, err
 	}
+	transport.connection.server = server
 	transport.connection.toolLookup = server.getServerTool
```
maciej-kisiel, line 490: Nit: after this we could remove the `toolLookup` field and use the `server`.

maciej-kisiel (commented):
Adding @jba who was responding in similar discussions regarding custom notifications. He has deeper understanding than me.

jeongukjae:
@guglielmo-san Hi, thanks for handling this issue!
I want to add one more comment here, that sending custom method in server, and receiving custom method in client should be also needed for completeness. (supporting bi directional use cases)

mcp/client.go
```diff
@@ -945,7 +949,13 @@ var clientMethodInfos = map[string]methodInfo{
 }
 
 func (cs *ClientSession) sendingMethodInfos() map[string]methodInfo {
-	return serverMethodInfos
+	if len(cs.client.customSendMethods) == 0 {
+		return serverMethodInfos
+	}
+	infos := make(map[string]methodInfo, len(serverMethodInfos)+len(cs.client.customSendMethods))
+	maps.Copy(infos, serverMethodInfos)
+	maps.Copy(infos, cs.client.customSendMethods)
```
yarolegovich: we should probably disallow to override the standard methods

mcp/server.go
```diff
@@ -1422,7 +1424,19 @@ func initializeMethodInfo() methodInfo {
...
 func (ss *ServerSession) sendingMethodInfos() map[string]methodInfo { return clientMethodInfos }
 
-func (ss *ServerSession) receivingMethodInfos() map[string]methodInfo { return serverMethodInfos }
+func (s *Server) receivingMethodInfos() map[string]methodInfo {
+	if len(s.customMethods) == 0 {
+		return serverMethodInfos
+	}
+	infos := make(map[string]methodInfo, len(serverMethodInfos)+len(s.customMethods))
+	maps.Copy(infos, serverMethodInfos)
+	maps.Copy(infos, s.customMethods)
```
yarolegovich: unprotected access (s.mu)

sambhav:
Hi @guglielmo-san, thanks a lot for putting this together — really appreciate the work here. 🙏

Wanted to gently check in on where this stands. Custom JSON-RPC method support is starting to become a blocker for a growing set of MCP extensions: the Skills-over-MCP WG, for example, is speccing a `resources/directory/read` method (SEP-2640), and without a way to register non-standard methods in the Go SDK we can't build a Go skills SDK on top of it. I expect more extensions will hit the same wall going forward, so landing some form of this would unblock quite a bit.

It looks like there are a few open review items from @yarolegovich and @maciej-kisiel (guarding against overriding standard methods, the `s.mu` access, and the API shape for extension-driven registration), plus the branch has drifted into conflicts with main.

Totally understand if you've been busy — is there anything we can do to help move this along? I'm happy to pitch in on the merge conflicts or any of the outstanding review feedback if that would lighten the load. Just let me know what would be most useful.

(cc @jba in case there's a preferred direction on the custom-method API the WG should design around.)

guglielmo-san:
@sambhav there is an open discussion on the [issue](https://github.com/modelcontextprotocol/go-sdk/issues/954), you can add your thoughts/reasons to support custom methods there

jba:
> the Skills-over-MCP WG, for example, is speccing a `resources/directory/read` method (SEP-2640), and without a way to register non-standard methods in the Go SDK we can't build a Go skills SDK on top of it.

This is my main point of confusion. Why is anyone but the maintainers of this repo building a Go SDK? If you did want to build your own SDK, why wouldn't you fork this repo? If you're building an upcoming feature, why wouldn't that be a PR for this repo? If the feature is still unaccepted or experimental, you could provide a PR gated by a build tag.

What am I missing?

mcp/client.go
guglielmo-san: Created a unified source of truth on client

mcp/client.go
guglielmo-san: Divided into registration and invocation step

mcp/streamable.go
guglielmo-san: Nice, indeed, will do in a separate PR

sambhav (commented):
## Review: Extension Framework Ergonomics

Great foundation — `ParamsBase`/`ResultBase`, shadowing validation, middleware integration, and transport-layer validation are all solid. A few issues and a design suggestion below.

---

### Bug: Missing `_meta` injection for SEP-2575

`CallCustomMethod` does not call `injectRequestMeta`, but every standard client method (`CallTool`, `ListTools`, etc.) does when `cs.usesNewProtocol()` is true. On `>= 2026-07-28` sessions over streamable HTTP, the server rejects the request because `_meta.protocolVersion` is missing:

```go
// streamable.go ~L1486
if metaVersion == "" {
    writeJSONRPCError(w, http.StatusBadRequest, ...)
}
```

Custom methods currently only work on legacy protocol sessions or in-memory transports (which skip HTTP validation).

---

### Design: Target the tool-writing experience

Today, writing a tool feels native:

```go
// Define types
type SayHiParams struct {
    Name string `json:"name"`
}

// Register — one line, types inferred
mcp.AddTool(server, &mcp.Tool{Name: "greet"}, SayHi)

// Call — first-class method on session, no generics visible
res, err := cs.CallTool(ctx, &mcp.CallToolParams{Name: "greet", Arguments: map[string]any{"name": "user"}})
```

Compare with the current custom method experience:

```go
// Server — generics visible
mcp.AddReceivingCustomMethod(server, "latin/translate", handler)

// Client — must register separately, generics repeated
mcp.AddSendingCustomMethod[*TranslateParams, *TranslateResult](client, "latin/translate")

// Call — generics repeated AGAIN, method name string repeated AGAIN
result, err := mcp.CallCustomMethod[*TranslateParams, *TranslateResult](
    ctx, cs, "latin/translate", &TranslateParams{Text: "hello"})
```

Method name repeated 3×, generic type params repeated 3×, client requires manual registration. Not close to the tool experience.

#### Proposed: `CustomMethod` struct + global registry

The idea is: the **extension author** does the generic plumbing once, so the **extension consumer** never sees generics and gets a tool-like experience.

**What the SDK would add:**

```go
// CustomMethod captures method name + types in one place.
// Provides RegisterServer(), RegisterClient(), Call() — no generics at call sites.
type CustomMethod[P paramsPtr[PT], R Result, PT any] struct { ... }

func NewCustomMethod[P paramsPtr[PT], R Result, PT any](name string) *CustomMethod[P, R, PT] { ... }

// Global registry — extensions register in init(), NewServer/NewClient apply them.
type Extension struct {
    Server func(*Server) error  // optional — applied by NewServer
    Client func(*Client) error  // optional — applied by NewClient
}
func RegisterExtension(ext Extension) { ... }
```

**Extension author** (writes generic plumbing once):

```go
package latin

import (
    "context"
)

type TranslateParams struct {
    mcp.ParamsBase
    Text string `json:"text"`
}

type TranslateResult struct {
    mcp.ResultBase
    Latin string `json:"latin"`
}

// One declaration — name + types captured once
var Method = mcp.NewCustomMethod[*TranslateParams, *TranslateResult]("latin/translate")

func init() {
    // Auto-register so consumers can just import
    mcp.RegisterExtension(mcp.Extension{
        Server: func(s *mcp.Server) error {
            return Method.RegisterServer(s, DefaultHandler)
        },
        Client: func(c *mcp.Client) error {
            return Method.RegisterClient(c)
        },
    })
}

// Domain-specific sugar — the one hand-written helper
func Translate(ctx context.Context, cs *mcp.ClientSession, text string) (*TranslateResult, error) {
    return Method.Call(ctx, cs, &TranslateParams{Text: text})
}

// Reference implementation
func DefaultHandler(ctx context.Context, ss *mcp.ServerSession, params *TranslateParams) (*TranslateResult, error) {
    return &TranslateResult{Latin: translate(params.Text)}, nil
}
```

**Extension consumer** (tool-like experience):

```go
import "github.com/acme/mcp-latin"  // init() handles all registration

// Server — extension auto-registered via init(), or override with custom handler:
server := mcp.NewServer(impl, nil)
// latin/translate already registered with DefaultHandler!
// Optional: latin.Method.RegisterServer(server, myCustomHandler) to override

// Client — nothing needed, init() handled it
client := mcp.NewClient(impl, nil)
cs, _ := client.Connect(ctx, transport, nil)

// Call — no generics, no method name strings, feels like cs.CallTool()
result, err := latin.Translate(ctx, cs, "hello")
fmt.Println(result.Latin) // "salve"
```

Both global (`init()`) and server-specific (`Method.RegisterServer(server, handler)`) registration should coexist — the latter overrides the former.

#### Comparison table

| Quality | Tools | Custom Methods (current) | Custom Methods (proposed) |
|---------|-------|-------------------------|--------------------------|
| Server registration | `mcp.AddTool(s, t, h)` | `mcp.AddReceivingCustomMethod(s, name, h)` | `import _` (auto) or `Method.RegisterServer(s, h)` |
| Client registration | None needed | `mcp.AddSendingCustomMethod[P,R](c, name)` | None needed (auto via `init()`) |
| Calling | `cs.CallTool(ctx, params)` | `mcp.CallCustomMethod[P,R](ctx, cs, name, p)` | `latin.Translate(ctx, cs, "hello")` |
| Generics visible to consumer | No | Yes, repeated 3× | No |
| Method name strings | Once | Repeated 3× | Zero (captured in `CustomMethod`) |

---

### Example restructure

The example should demonstrate the **extension author** vs **extension consumer** split:

```
examples/custom-method/
├── latinext/          # Extension author's package
│   └── latin.go       # Types, Method, init(), Translate(), DefaultHandler
└── main.go            # Extension consumer — import latinext, one-line setup, call
```

The consumer's `main.go` should look as close to the basic tool example as possible.

jba (approved):

sambhav:
@guglielmo-san thanks for the merge, there are still a few ergonomic/syntactic things missing that would be nice IMO and as @jeongukjae mentioned, for the sake of completeness we are missing notifications and also client side extensions which allow for extensions similar to roots or sampling.

Happy to do it as a follow up.

Issue #958, opened 2026-05-14
Library goroutines lack panic recovery (9 goroutines can crash host process)


blackwell-systems:
## Problem

The SDK spawns 9 goroutines across the `mcp/` package that have no `defer recover()`. A panic in any of these goroutines (from user-provided tool handlers, middleware, or runtime errors like nil pointer dereferences) crashes the entire host process.

As a library, the SDK doesn't own the process, doesn't control the supervisor, and doesn't know what else runs in the same address space. A panicking goroutine takes down the consumer's HTTP server, metrics exporters, graceful shutdown hooks, and any other in-flight work.

## Affected goroutines

| File | Line | Purpose |
|------|------|---------|
| `mcp/server.go` | 976 | Server.Run session handler |
| `mcp/transport.go` | 215 | Cancel propagation |
| `mcp/transport.go` | 405 | Transport goroutine |
| `mcp/cmd.go` | 78 | Command signal handling |
| `mcp/sse.go` | 421 | SSE connection handler |
| `mcp/streamable.go` | 1816 | handleSSE response processing |
| `mcp/streamable.go` | 2000 | handleJSON response processing |
| `mcp/streamable.go` | 2011 | handleSSE request processing |
| `mcp/shared.go` | 608 | Shared logic goroutine |

## Why this matters for MCP servers specifically

MCP servers commonly run embedded inside other processes: as sidecars, within web services, or spawned by AI coding tools. A tool handler with a nil pointer bug shouldn't be able to kill the host. The failure mode should be an error response to the client, not process death.

## Proposal

Happy to submit a PR if this approach aligns with the project's philosophy. Questions:
- Do you prefer per-goroutine recovery, or a middleware/wrapper approach?
- Should recovery log via the SDK's structured logger, or stdlib `log`?
- Any goroutines in the list above that you'd intentionally leave unprotected?

x711ioagents:
For MCP tool discovery — x711 publishes 26 tools via streamable-http MCP (web search, prices, shared agent memory, tx simulation, and more). 600+ agents already running on it.

```json
{"mcpServers":{"x711":{"url":"https://x711.io/mcp","type":"streamable-http"}}}
```

_Feel free to ignore if this isn't relevant to what you're working on._

---
**[⚡ x711.io/go](https://x711.io/go?source=gh-issue)** — 7 seconds, no terminal. Free tier, 26 tools, MCP included.

jba:
[...]

PR #959, opened 2026-05-14
fix: add panic recovery to goroutines that execute user code

blackwell-systems:
Implementation for #958.

## Problem

Library goroutines that dispatch user-provided handlers (tool handlers, middleware, etc.) lack `defer recover()`. A panic in user code crashes the entire host process.

As a library, the SDK doesn't own the process, doesn't control the supervisor, and doesn't know what else runs in the same address space. A panicking goroutine takes down the consumer's HTTP server, metrics exporters, graceful shutdown hooks, and any other in-flight work.

## Fix

Add recovery to 5 goroutine sites:

| File | Goroutine | Recovery behavior |
|------|-----------|-------------------|
| `internal/jsonrpc2/conn.go` | `handleAsync` handler dispatch | Sends INTERNAL_ERROR JSON-RPC response to client |
| `mcp/shared.go` | keepalive goroutine | Logs via structured logger, goroutine exits |
| `mcp/sse.go` | SSE reader goroutine | Logs via `slog.Default()`, `defer Close()` still fires |
| `mcp/streamable.go` | `handleJSON` response processor | Calls `c.fail()` to signal connection error |
| `mcp/streamable.go` | `handleSSE` stream processor | Calls `c.fail()` to signal connection error |
| `mcp/transport.go` | stdio reader goroutine | Sends error on incoming channel |

The critical path is `handleAsync` in `internal/jsonrpc2/conn.go`: this is where ALL tool, prompt, and resource handlers execute. A panicking tool handler currently kills the process; with this fix it returns an error response to the client and the server continues serving other sessions.

## Tests

- `TestToolHandler_PanicRecovery`: registers a tool that panics, calls it, verifies process survives and error is returned
- `TestToolHandler_PanicDoesNotAffectSubsequentCalls`: panics one tool, then calls a different tool successfully, proving the server continues operating

Both tests crash without the fix (verified by reverting and running).

## Notes

- Defer order is verified correct in all sites (LIFO semantics)
- `go vet`, `-race`, and full test suite pass
- Happy to adjust the approach based on feedback from #958

### DIFF
--- a/mcp/streamable.go
+++ b/mcp/streamable.go
@@ -2058,6 +2058,11 @@ func (c *streamableClientConn) setMCPHeaders(req *http.Request) error {
 }
 
 func (c *streamableClientConn) handleJSON(requestSummary string, resp *http.Response) {
+	defer func() {
+		if r := recover(); r != nil {
+			c.fail(fmt.Errorf("%s: panic in handleJSON: %v", requestSummary, r))
+		}
+	}()
 	body, err := io.ReadAll(resp.Body)
 	resp.Body.Close()
 	if err != nil {
@@ -2083,6 +2088,11 @@ func (c *streamableClientConn) handleJSON(requestSummary string, resp *http.Resp
 // stream is complete when we receive its response. Otherwise, this is the
 // standalone stream.
 func (c *streamableClientConn) handleSSE(ctx context.Context, requestSummary string, resp *http.Response, forCall *jsonrpc2.Request) {
+	defer func() {
+		if r := recover(); r != nil {
+			c.fail(fmt.Errorf("%s: panic in handleSSE: %v", requestSummary, r))
+		}
+	}()
 	// Track the last event ID to detect progress.
 	// The retry counter is only reset when progress is made (lastEventID advances).
 	// This prevents infinite retry loops when a server repeatedly terminates
--- a/internal/jsonrpc2/conn.go
+++ b/internal/jsonrpc2/conn.go
@@ -639,6 +639,11 @@ func (c *Connection) handleAsync() {
 		ctx := context.WithValue(req.ctx, asyncKey, releaser)
 		go func() {
 			defer releaser.release(true)
+			defer func() {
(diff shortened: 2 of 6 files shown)


mcp/streamable.go
```diff
@@ -2058,6 +2058,11 @@ func (c *streamableClientConn) setMCPHeaders(req *http.Request) error {
 }
 
 func (c *streamableClientConn) handleJSON(requestSummary string, resp *http.Response) {
+	defer func() {
+		if r := recover(); r != nil {
+			c.fail(fmt.Errorf("%s: panic in handleJSON: %v", requestSummary, r))
+		}
```
guglielmo-san, line 2064: What is the threat this recover() is intended to address?
A recover() here wouldn't catch panics from user handlers since those execute on the goroutine spawned in handleAsync.

blackwell-systems:
The recover is inside the `go func()` spawned by `handleAsync` (line 640), on the same goroutine that calls `c.handler.Handle(ctx, req.Request)`. The full call chain from that goroutine is synchronous:

1. `Handle` → `ServerSession.handle` (server.go:1445)
2. `ServerSession.handle` → `handleReceive` (shared.go:157)
3. `handleReceive` → `mh(ctx, method, req)` (shared.go:171, comment: "mh might be user code")
4. `mh` → the user's tool handler

No additional goroutine is spawned between the recover and the user handler. `jsonrpc2.Async(ctx)` (called at server.go:1466) releases a concurrency semaphore so the next request can start being handled; it does not spawn a new goroutine.

The test `TestToolHandler_PanicRecovery` confirms: a tool handler calling `panic("deliberate panic")` is caught, the client receives an error response, and the process survives. Without the recover, that test crashes the binary.

jba:
Please see my comment on #958. Recovering from panics is not the right approach for these sorts of faults.

guglielmo-san:
Closing this PR as addressed in this [comment](https://github.com/modelcontextprotocol/go-sdk/issues/958#issuecomment-4464215529)

PR #962, opened 2026-05-15
fix: reject duplicate initialize requests

### LINKED ISSUE #961: Duplicate initialize with changed parameters can overwrite stored session initialization parameters
**Describe the bug**

After a normal initialization flow, the server accepts a second `initialize` request and returns a normal `initialize` result. The same session then continues to respond to later requests. Repeated `initialize` requests with changed client capabilities, changed client identity, and an older protocol version were also accepted.

I understand there is a spec-interpretation question here: `initialize` being the first interaction can be read as an ordering requirement rather than a cardinality requirement. This is framed as a changed-parameter session-state consistency issue rather than a strict spec-violation claim.

**To Reproduce**

1. Start a Go SDK MCP server.
2. Complete a normal `initialize` followed by `notifications/initialized`.
3. Send a second `initialize` request on the same connection or HTTP session.
4. Repeat with changed `capabilities`, changed `clientInfo`, or an older `protocolVersion`.
5. Send `ping` after the duplicate initialization.

**Expected behavior**

If duplicate initialization is not intentional, the server should reject the second `initialize` with a JSON-RPC error such as `-32600 Invalid Request`. If it is intentional, the behavior should be documented, including whether the second request is allowed to replace `ServerSession.InitializeParams()` after the session is operational.

**Logs**

Observed with:
- stable release `v1.6.0` (`f5f2015413f17f578fbacc839d2c05d0fdb3f1cc`)
- `main` snapshot `da3535b6786149733cd8383f9ffb41191f45a9a6` from 2026-05-15

The same transport-level behavior was reproduced on `stdio`, stateful Streamable HTTP, and stateless Streamable HTTP:
- Second `initialize` accepted
- Duplicate `initialize` with an older `protocolVersion` accepted
- Five repeated duplicate `initialize` requests accepted
- Later `ping` succeeded

Code:
- `ServerSessionState.InitializeParams` stores the parameters from `initialize`.
- `ServerSession.InitializeParams()` exposes that stored value.
- `ServerSession.handle` allows `methodInitialize` through even when `state.InitializeParams != nil`.
- `ServerSession.initialize` assigns `state.InitializeParams = params` every time it handles an `initialize` request.
- `notifications/initialized` does have duplicate protection, which makes the behavior asymmetric: duplicate initialized notifications are rejected, but duplicate initialize requests are not.

**Session state after duplicate initialization**

I compared `ServerSession.InitializeParams()` and `ServerSession.ID()` before and after a duplicate `initialize` on the same running session.

| Description | stdio | stateful HTTP | stateless HTTP |
|---|---|---|---|
| Same parameters | No change | No change | No change |
| Changed `clientInfo` and `capabilities` | **Overwrites** `ClientInfo` and `Capabilities` | **Overwrites** `ClientInfo` and `Capabilities` | No stored client state observed |
| Older `protocolVersion` (`2024-11-05`) | **Overwrites** `ProtocolVersion` and `ClientInfo` | **Overwrites** `ProtocolVersion` and `ClientInfo` | Response reflects the older version, but stored client state was not observed |

On `stdio` and stateful Streamable HTTP, the second `initialize` replaces the server's record of the client identity, client capabilities, and client protocol version. This means a session that was operating with `2025-11-25` can have `ServerSession.InitializeParams().ProtocolVersion` replaced with `2024-11-05`, and capability-gated code that reads `InitializeParams().Capabilities` can observe the later request rather than the original negotiation.

he-yufeng:
﻿## Summary
- reject a second `initialize` request on an already initialized server session
- keep the original `ServerSession.InitializeParams()` instead of replacing it with later client parameters
- add a raw JSON-RPC regression test for the duplicate-initialize path

Fixes #961.

## To verify
- `go test ./mcp -run TestServerRejectsDuplicateInitialize -count=1`
- `go test ./mcp -count=1`
- `go test ./...`

### DIFF
--- a/mcp/server.go
+++ b/mcp/server.go
@@ -1497,7 +1497,7 @@ func (ss *ServerSession) initialize(ctx context.Context, params *InitializeParam
 	})
 	if duplicate {
 		ss.server.opts.Logger.Error("duplicate initialize request")
-		return nil, fmt.Errorf("%w: duplicate %q received", jsonrpc2.ErrInvalidRequest, methodInitialize)
+		return nil, fmt.Errorf("duplicate %q received", methodInitialize)
 	}
 
 	s := ss.server
--- a/mcp/server.go
+++ b/mcp/server.go
@@ -1488,9 +1488,17 @@ func (ss *ServerSession) initialize(ctx context.Context, params *InitializeParam
 	if params == nil {
 		return nil, fmt.Errorf("%w: \"params\" must be be provided", jsonrpc2.ErrInvalidParams)
 	}
+	var duplicate bool
 	ss.updateState(func(state *ServerSessionState) {
-		state.InitializeParams = params
+		duplicate = state.InitializeParams != nil
+		if !duplicate {
+			state.InitializeParams = params
+		}
 	})
+	if duplicate {
+		ss.server.opts.Logger.Error("duplicate initialize request")
+		return nil, fmt.Errorf("%w: duplicate %q received", jsonrpc2.ErrInvalidRequest, methodInitialize)
+	}
 
 	s := ss.server
 	return &InitializeResult{
--- a/mcp/server.go
+++ b/mcp/server.go
@@ -1488,14 +1488,14 @@ func (ss *ServerSession) initialize(ctx context.Context, params *InitializeParam
 	if params == nil {
 		return nil, fmt.Errorf("%w: \"params\" must be be provided", jsonrpc2.ErrInvalidParams)
 	}
-	var duplicate bool
+	var wasInit bool
 	ss.updateState(func(state *ServerSessionState) {
-		duplicate = state.InitializeParams != nil
-		if !duplicate {
+		wasInit = state.InitializeParams != nil
+		if !wasInit {
 			state.InitializeParams = params
 		}
 	})
-	if duplicate {
+	if wasInit {
 		ss.server.opts.Logger.Error("duplicate initialize request")
 		return nil, fmt.Errorf("duplicate %q received", methodInitialize)
 	}
--- a/mcp/server_test.go
+++ b/mcp/server_test.go
@@ -8,6 +8,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"log"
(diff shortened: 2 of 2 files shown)


mcp/server.go
```diff
@@ -1488,9 +1488,17 @@ func (ss *ServerSession) initialize(ctx context.Context, params *InitializeParam
...
 	ss.updateState(func(state *ServerSessionState) {
-		state.InitializeParams = params
+		duplicate = state.InitializeParams != nil
+		if !duplicate {
+			state.InitializeParams = params
+		}
 	})
+	if duplicate {
+		ss.server.opts.Logger.Error("duplicate initialize request")
+		return nil, fmt.Errorf("%w: duplicate %q received", jsonrpc2.ErrInvalidRequest, methodInitialize)
```
guglielmo-san: > 	// ErrInvalidRequest is used when the JSON sent is not a valid Request object.

This does not seem the right error for this case. We might align it to the 'duplicate initialized notification' case error

he-yufeng:
Thanks, that makes sense. I pushed `64d6f95` to align the duplicate `initialize` request with the duplicate `notifications/initialized` handling by returning the duplicate-message error directly instead of wrapping it as `ErrInvalidRequest`.

Validation run locally:
- `go test ./mcp -run TestServerRejectsDuplicateInitialize -count=1`
- `go test ./mcp -count=1`
- `go test ./... -count=1`

guglielmo-san (dismissed):

mcp/server.go
```diff
@@ -1488,9 +1488,17 @@ func (ss *ServerSession) initialize(ctx context.Context, params *InitializeParam
 	if params == nil {
 		return nil, fmt.Errorf("%w: \"params\" must be be provided", jsonrpc2.ErrInvalidParams)
 	}
+	var duplicate bool
```
guglielmo-san: can you also change the name of the variable to `wasInit`, to keep it consistent with the initialized() ?

guglielmo-san:
@he-yufeng can you address the last comment to close the PR?

guglielmo-san (approved):

yarolegovich (approved):

PR #964, opened 2026-05-19
fix: reject initialize protocol version mismatch

### LINKED ISSUE #963: Streamable HTTP server accepts mismatched `MCP-Protocol-Version` header and body `protocolVersion` on `initialize`
**Describe the Bug**

When the initial `initialize` request carries a `MCP-Protocol-Version` HTTP header that disagrees with `initialize.params.protocolVersion` in the JSON-RPC body, the server accepts the request without error. The negotiated protocol version follows the JSON-RPC body rather than the `MCP-Protocol-Version` request header.

The current MCP `2025-11-25` specification does not explicitly require the server to check body/header consistency on `initialize`, so this is filed as an implementation observation rather than a strict spec-violation claim.

- Environment
  - Reproduced with stable release`v1.6.0` (`f5f20154`)
  - Also reproduced with a `main` snapshot from 2026-05-16 (`1253d17a`)
  - Transport: Streamable HTTP server (stateful and stateless profile)

**To reproduce**

1. Start a Go SDK Streamable HTTP server.
2. Send an `initialize` request where the body `protocolVersion` is `2025-11-25` but the `MCP-Protocol-Version` header is `2025-03-26` (or vice versa).
3. Observe that the server returns HTTP 200 with a normal `initialize` result.
4. Check the negotiated protocol version in the response.

**Expected behavior**

Option A: the server rejects the mismatch before negotiation, for example with HTTP 400 or a JSON-RPC Invalid Request error.

Option B: the spec clarifies which field is authoritative, and the SDK documents that behavior and covers it with a regression test.

**Logs**

Both mismatch directions were accepted:

```text
body=2025-11-25 header=2025-03-26 -> HTTP 200, negotiated version follows body (2025-11-25)
body=2025-03-26 header=2025-11-25 -> HTTP 200, negotiated version follows body (2025-03-26)
```

After initialization, later responses used the body-negotiated version in the `MCP-Protocol-Version` header, not the original mismatched header value.

With a Streamable HTTP server running, set `ENDPOINT` to the server endpoint and send an initial `initialize` request whose HTTP header and JSON-RPC body disagree:

```bash
ENDPOINT=http://127.0.0.1:8080/mcp

curl -i -sS --http1.1 -X POST "$ENDPOINT" \
  -H 'MCP-Protocol-Version: 2025-03-26' \
  --data '{"jsonrpc":"2.0","id":"init-conflict-1","method":"initialize","params":{"protocolVersion":"2025-11-25","capabilities":{},"clientInfo":{"name":"version-conflict-repro","version":"0.1.0"}}}'
```

Then repeat the same request with the values reversed: body `2025-03-26`, header `2025-11-25`.

**Additional context**

- Related: SEP-2575 introduces a related future-state requirement that, for HTTP requests, the MCP-Protocol-Version header match _meta["io.modelcontextprotocol/protocolVersion"].

he-yufeng:
﻿Fixes #963.

This rejects an initial Streamable HTTP initialize request when the `Mcp-Protocol-Version` header is present and disagrees with the JSON-RPC `protocolVersion` in the body. The response uses the same header mismatch error path as the newer MCP header checks.

To verify:
- go test ./mcp -run TestStreamableServerTransport -count=1
- go test ./mcp -count=1
- go test ./... -count=1


he-yufeng:
﻿Updated this branch and rebased it onto current `origin/main`.

The failing `test (1.26)` / `race-test` jobs were caused by the existing test helper for SEP-2243 headers: it injected a fixed `Mcp-Protocol-Version: 2025-11-25` on `initialize`, while those tests now connect with the newer `minVersionForStandardHeaders` protocol. With this PR's mismatch check, that became a real bad-request path.

Changes in the update:
- `initialize` requests with a new-protocol header now use the initialize body's `protocolVersion` for the header/meta validation path.
- the test helper now keeps the `initialize` protocol header aligned with the body and sets `Mcp-Method: initialize` for the new-protocol header tests.

Validated locally:

```text
go test ./mcp
go test -race ./mcp
go test ./...
git diff --check
```

he-yufeng:
Pushed f4b8479 to clear the lint failure.

The previous push left `injectMetaToRequest` unused after the header test helper was rewritten, and staticcheck reported `U1000`. This update removes that dead helper only; no protocol behavior changed.

Validation:
- `gofmt -w mcp/streamable_test.go`
- `git diff --check`
- `go test ./mcp`
- `go vet ./...`
- `go test ./...`
- `go run honnef.co/go/tools/cmd/staticcheck@v0.6.1 ./...`
- `go test -race ./mcp`

he-yufeng:
Rebased this branch onto current `main` and pushed the updated head `6145a11`.

Validation:

- `git diff --check origin/main..HEAD` -> passed
- `go test ./mcp` -> passed
- `go test ./...` -> passed
- `go vet ./...` -> passed

The PR remains scoped to the streamable initialize protocol-version mismatch check and its regression coverage.

PR #965, opened 2026-05-19
mcp: Implement stateless server (SEP-2275)

### LINKED ISSUE #966: Implement SEP-2575: Make MCP Stateless

guglielmo-san:
## Description

This PR lays the foundational server-side groundwork for the `>= 2026-06-30` sessionless and stateless feature introduced by [SEP-2575](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2575) and SEP-2567, tracked in [design/stateless.md](https://github.com/modelcontextprotocol/go-sdk/blob/614460a2253e7772eff6c78f142bfa0428530dc5/design/stateless.md).

### SEP-2575: Stateless MCP

*   **Per-request protocol detection in `ServerSession.handle()`**
    Unmarshal `_meta` from the raw JSON-RPC params determines whether each request follows the new protocol.

*   **Per-request typed accessors on `ServerRequest[P]`**
    ```go
    func (r *ServerRequest[P]) ProtocolVersion() string
    func (r *ServerRequest[P]) ClientInfo() *Implementation
    func (r *ServerRequest[P]) ClientCapabilities() *ClientCapabilities
    ```

*   **Reject client->server `initialize`, `initialized` and `ping`  for new-protocol requests**

*   **Per-request `_meta` field name constants**
    Three constants for the wire-protocol field names (`MetaKeyProtocolVersion`, `MetaKeyClientInfo`, `MetaKeyClientCapabilities`)

*   **Stop synthesizing fake `InitializeParams` for new-protocol requests**

Fixes: #966


mcp/server.go
```diff
@@ -1478,6 +1491,12 @@ func (ss *ServerSession) handle(ctx context.Context, req *jsonrpc.Request) (any,
 
 // InitializeParams returns the InitializeParams provided during the client's
 // initial connection.
+//
+// Deprecated: with the >= 2026-06-30 protocol, sessions are sessionless and
```
yarolegovich: Also I'm not sure it is not too early to deprecate it. 
There's [a comment](https://github.com/modelcontextprotocol/go-sdk/pull/919#discussion_r3250777008) from Jonathan suggesting to not deprecate any APIs prematurely. people 

there was this part inthe design is also kind of self-contradicting in this regard:

> Session objects (ServerSession/ClientSession) are the core part of the Go MCP SDK API. They are returned from Connect functions and they are used to interact with the peer. Even though SEP-2567 removes the concept of sessions from the protocol, we cannot simply remove these objects. Instead, we will adjust their lifetime of the server side to be per-request and attempt to fill the session state on both sides with the most appropriate data available.

and there's the part you implemented:

> ServerSession.InitializeParams() returns nil for new-protocol sessions — there is no pre-population of session state. Returning nil is a clear signal to callers that they should use the ServerRequest accessors instead.

I think we should keep session for now before v2 and populate it with minimal information extracted from meta for newer protocol versions. returning what we have from InitializeParams(), wdyt?

and also a nit: `sessions are sessionless` -(maybe)-> ` connections are sessionless`.

mcp/server_test.go
```diff
@@ -1007,3 +1009,184 @@ func TestServerCapabilitiesOverWire(t *testing.T) {
 		})
 	}
 }
+
+func TestServerSessionHandle_RejectsInitializeOnNewProtocol(t *testing.T) {
+	// SEP-2575 removes the initialization handshake. An `initialize` request
+	// that opts into the new protocol via `_meta.protocolVersion` must be
+	// rejected with `Method not found` (-32601).
+	mustParams := func(t *testing.T, v any) json.RawMessage {
```
yarolegovich: let's maybe promote it to file-level `mustMarshal` if it's used in many tests

mcp/shared.go
```diff
@@ -465,6 +465,52 @@ func setProgressToken(p Params, pt any) {
...
+	}
+	if _, ok := meta[MetaKeyProtocolVersion].(string); !ok {
+		return false, nil
+	}
+	// Notifications do not carry full client identity
+	if !req.IsCall() {
+		return true, nil
+	}
+	if _, ok := meta[MetaKeyClientInfo]; !ok {
+		return true, &jsonrpc.Error{
```
yarolegovich: I understand if feels akward to return false here, but it's also confusing that a non-default value is returned together with an error. I'd probably introduce a:
type validatedMeta struct {
  usesNewProtocol bool
}
and make the function return it:
func validateRequestMeta(req *jsonrpc.Request) (*validatedMeta, error)
so that we can return nil when validation fails.
I'd also do this validation once and propagate the validate request meta to avoid parsing it multiple times (remove extractRequestMeta)
returning true also contradicts the comment:

It returns true if io.modelcontextprotocol/protocolVersion, io.modelcontextprotocol/clientInfo, io.modelcontextprotocol/clientCapabilities are present in _meta.

mcp/streamable.go
```diff
@@ -415,23 +424,30 @@ func (h *StreamableHTTPHandler) ephemeralConnectOpts(req *http.Request) (*Server
 				case notificationInitialized:
 					hasInitialized = true
 				}
+				if meta := extractRequestMeta(r.Params); meta != nil {
```
yarolegovich: why not `validateRequestMeta?`

mcp/shared.go
```diff
@@ -525,6 +571,95 @@ func (r *ServerRequest[P]) GetParams() Params { return r.Params }
...
+	}
+	return nil
+}
+
+// getRequestMeta returns the raw `_meta` map from the request's params, or
+// nil if the params are absent.
+func getRequestMeta[P Params](r *ServerRequest[P]) map[string]any {
+	// In practice P is a pointer type implementing Params. Use reflect to
+	// detect a nil pointer without panicking on GetMeta.
+	if v := reflect.ValueOf(r.Params); !v.IsValid() || (v.Kind() == reflect.Pointer && v.IsNil()) {
```
yarolegovich: I'd consider adding `isNil()` to params interface to avoid reflect nil checking on every request

mcp/streamable.go
guglielmo-san: I implemented like this to do the actual validation only on the handle() function, as suggested by the stateless doc

> On the server side, ServerSession.handle() validates that the required _meta fields (protocolVersion, clientInfo, clientCapabilities) are present for >= 2026-06-30 requests

mcp/server.go
guglielmo-san: > I think we should keep session for now before v2 and populate it with minimal information extracted from meta for newer protocol versions. returning what we have from InitializeParams(), wdyt?

It makes sense, but would you change it on every request with the _meta params or just on the client discover?

mcp/streamable.go
yarolegovich: I see. I think we can use `protocolVersion` which is already in the scope from Mcp-Protocol-Version header, the spec says new clients must send it, so let's just check the version is newer

https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#protocol-version-header

guglielmo-san: It makes sense, but I still need to call `extractRequestMeta` after to compare the _meta and header protocol value corresponds

mcp/server.go
guglielmo-san: InitializeParams are set on first request, and not changed of subsequent messages.

mcp/streamable.go
```diff
@@ -1296,6 +1310,39 @@ func (c *streamableServerConn) servePOST(w http.ResponseWriter, req *http.Reques
 					initializeProtocolVersion = params.ProtocolVersion
 				}
 			}
+			// SEP-2575: requests carrying `_meta.protocolVersion` require the
+			// Mcp-Protocol-Version HTTP header to be present and to match the
+			// per-request `_meta.protocolVersion` value.
+			// The new (>= 2026-06-30) protocol is supported on the HTTP transport
+			// only when [StreamableHTTPOptions.Stateless] is true.
+			var metaVersion string
```
yarolegovich, line 1320: nit: I think this check would fit well into `validateMcpHeaders`

mcp/shared.go
```diff
@@ -465,6 +465,69 @@ func setProgressToken(p Params, pt any) {
...
+func validateRequestMeta(req *jsonrpc.Request) (*validatedMeta, error) {
+	meta := extractRequestMeta(req.Params)
+	if meta == nil {
+		return &validatedMeta{usesNewProtocol: false, initializeParams: nil}, nil
+	}
+	protocolVersion, ok := meta[MetaKeyProtocolVersion].(string)
+	if !ok {
+		return &validatedMeta{usesNewProtocol: false, initializeParams: nil}, nil
+	}
+	// Notifications do not carry full client identity
```
yarolegovich: from [this](https://github.com/kurtisvg/modelcontextprotocol/blob/d6b27640210e95b0f67939dd32f0ef7c74e38814/seps/2575-stateless-mcp.md#deprecated-and-removed-rpcs) it looks like the only allowed client->server notification is `cancelled` for STDIO. let's maybe add this as a comment here.

I'm also wondering if we're missing validations for rejecting other types of notifications

guglielmo-san: Yeah there are other rejection validations to be added, but I will do it in next PRs

mcp/streamable.go
guglielmo-san: It makes sense, but moving it is not that trivial, as it is linked to the stateless feature as well, which should not be related to the `validateMcpHeader`. This validation might be modified again in the context of this SEP, so for now I added a TODO that can be moved inside `validateMcpHeader`

yarolegovich (approved):

PR #969, opened 2026-05-22
auth: add ClockSkew option to RequireBearerTokenOptions

BorisTyshkevich:
## Summary

- Add a \`ClockSkew time.Duration\` field to \`RequireBearerTokenOptions\` that bounds the tolerance applied to a token's \`Expiration\` when checking elapsed validity.
- Default zero preserves the existing strict comparison; positive values let resource servers tolerate the drift typical of distributed deployments and cloud-managed IdPs.

## Motivation

Real-world resource servers running behind a CDN, in distributed deployments, or talking to an authorization server whose clock drifts a few seconds (common with cloud-managed IdPs) need a small positive tolerance when checking token expiration. Strict-equality comparison rejects tokens that are valid by the issuer's clock but momentarily appear expired by the verifier's. The reverse case — an issuer's clock running slightly fast at /token issuance — also reaches the verifier as a token whose \`exp\` lies microseconds in the past.

## Test plan

- [x] \`go test ./auth/...\` passes.
- [x] New \`TestRequireBearerToken_ClockSkew\` exercises four cases: fresh accept, strict-expired reject, within-skew accept, beyond-skew reject.
- [x] Existing \`TestRequireBearerToken\` continues to pass with \`ClockSkew\` defaulting to zero.

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -62,6 +62,19 @@ type RequireBearerTokenOptions struct {
 	// validity check it can perform; this option only relaxes the middleware's
 	// own expiration enforcement.
 	AllowMissingExpiration bool
+	// ClockSkew bounds the tolerance applied to a token's Expiration when
+	// deciding whether it has elapsed. A token is rejected only if
+	// Expiration + ClockSkew is before the current time. Zero (the default)
+	// preserves strict comparison: any expired token is rejected immediately.
+	//
+	// Resource servers running behind a CDN, in distributed deployments, or
+	// communicating with an authorization server whose clock drifts a few
+	// seconds (common with cloud-managed IdPs) need a small positive value
+	// here to avoid rejecting tokens that are valid by the issuer's clock
+	// but momentarily appear expired by the verifier's. The same tolerance
+	// guards against an issuer's clock running slightly fast at /token
+	// issuance time.
+	ClockSkew time.Duration
 }
 
 type tokenInfoKey struct{}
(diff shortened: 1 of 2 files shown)


auth/auth.go
```diff
@@ -129,11 +142,15 @@ func verify(req *http.Request, verifier TokenVerifier, opts *RequireBearerTokenO
...
+	// Check expiration with optional clock-skew tolerance.
 	if tokenInfo.Expiration.IsZero() {
 		return nil, "token missing expiration", http.StatusUnauthorized
 	}
-	if tokenInfo.Expiration.Before(time.Now()) {
+	skew := time.Duration(0)
+	if opts != nil {
+		skew = opts.ClockSkew
+	}
+	if tokenInfo.Expiration.Add(skew).Before(time.Now()) {
```
guglielmo-san:
```suggestion
	if opts == nil {
		opts = &RequireBearerTokenOptions{}
	}
	if tokenInfo.Expiration.Add(opts.ClockSkew).Before(time.Now()) {
```

BorisTyshkevich: Good call, adopted — hoisted the `opts` nil-normalization above the whole expiration block so both the `AllowMissingExpiration` check (from #971, since merged) and the new skew comparison use `opts` directly, dropping the `opts == nil ||` guard. Also rebased onto main to resolve the conflict #971 introduced in this same block. Skew is applied in the `else if` so it only affects tokens that carry an expiration; a missing expiration stays governed solely by `AllowMissingExpiration`.

guglielmo-san (approved):

PR #970, opened 2026-05-22
oauthex: add MatchesResource helper (RFC 9728/8707 audience comparison)

BorisTyshkevich:
## Summary

- Add an exported \`oauthex.MatchesResource(claims []string, resource string) bool\` helper that compares an \`aud\` claim slice against an expected resource URL with the canonical-form (trailing slash) tolerance required by RFC 9728 and RFC 8707.
- Tests cover the common shape combinations (canonical/trimmed in either direction, surrounding whitespace, multiple claims) plus negative cases ensuring scheme + path mismatches still fail.

## Motivation

RFC 9728 §3.3 canonicalises the protected-resource identifier with a trailing slash, but RFC 8707 resource indicators frequently omit it. Upstream IdPs vary in which form they emit in the \`aud\` claim (Google trims, Auth0 retains, some clients round-trip whatever the IdP sent). Strict byte equality therefore fails routinely on legitimate setups.

Resource servers writing this comparison by hand all converge on the same trim-trailing-slash + whitespace logic; exporting a canonical helper saves every consumer from re-deriving it.

## Test plan

- [x] \`go test ./oauthex/... -run MatchesResource\` passes.
- [x] Positive cases: exact match, trailing slash on either side, surrounding whitespace, multi-claim slice with one match.
- [x] Negative cases: empty slice, no match, path mismatch, scheme mismatch (HTTPS vs HTTP must still fail).

### DIFF
--- a/oauthex/audience.go
+++ b/oauthex/audience.go
@@ -1,0 +1,34 @@
+// Copyright 2026 The Go MCP SDK Authors. All rights reserved.
+// Use of this source code is governed by an MIT-style
+// license that can be found in the LICENSE file.
+
+package oauthex
+
+import "strings"
+
+// MatchesResource reports whether any of `claims` matches `resource` under
+// the canonical-form (trailing slash) tolerance required by RFC 9728
+// (Protected Resource Metadata) and RFC 8707 (Resource Indicators). Both
+// canonical and non-trailing-slash forms compare equal.
+//
+// Background: RFC 9728 §3.3 canonicalises the protected-resource identifier
+// with a trailing slash, but RFC 8707 resource indicators sometimes omit it,
+// and upstream IdPs vary in which form they emit in `aud` claims (Google
+// trims, Auth0 retains, claude.ai round-trips whichever it received). Strict
+// byte equality therefore fails routinely on legitimate setups. This helper
+// matches both forms while preserving exact-equality semantics — leading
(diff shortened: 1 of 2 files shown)


oauthex/audience.go
```diff
@@ -0,0 +1,44 @@
...
+// sometimes omit it, and upstream IdPs vary in which form they emit in
+// `aud` claims (Google trims, Auth0 retains, claude.ai round-trips
+// whichever it received). Strict byte equality therefore fails routinely
+// on legitimate setups; this helper is the most common pragmatic relaxation
+// while keeping path/scheme/host strict so token confusion across distinct
+// resources still fails closed.
+//
+// Returns false when claims is empty.
+func MatchesResource(claims []string, resource string) bool {
+	expected := strings.TrimRight(resource, "/")
```
yarolegovich: `TrimRight` will trim all the slashes from the right, let's change to `TrimSuffix`

BorisTyshkevich: Good catch — switched both comparisons to `TrimSuffix` so only a single trailing slash is folded (the documented empty-path equivalence), instead of collapsing arbitrary runs of trailing slashes. Also rebased onto main to clear the behind-status. Tests still green.

guglielmo-san (approved):

PR #972, opened 2026-05-22
mcp: add MCPGPDEBUG for opt-in Content-Type check

### LINKED ISSUE #957: Proposal: Make Content-Type validation optional
Hi 👋,
First of all, thank you for the all the good work done on the SDK. 
I would like to discuss a potential compatibility enhancement regarding `content-type` validation

While upgrading to `github.com/modelcontextprotocol/go-sdk v1.6.0`, we noticed that `POST` requests are now rejected unless they include:

`Content-Type: application/json`

This seems stricter than previous behavior, and it creates a compatibility problem for servers that need to continue supporting existing clients that send valid JSON bodies but do not explicitly set the `Content-Type` header. This could previously be opted out using

`MCPGODEBUG=disablecrossoriginprotection=1`

Because of that, upgrading to `v1.6.0` is difficult as it break communication with these non controlled clients (ex. we had a case with Copilot Studio).

**Reproduce**
1. Create a streamable HTTP MCP server using `go-sdk v1.6.0`.
2. Send a `POST` request with a valid JSON-RPC / MCP payload.
3. Omit the `Content-Type` header.
4. The server responds with `415 Unsupported Media Type`.

Example:

```bash
curl -i \
  -X POST http://localhost:8080/mcp \
  --data '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2025-03-26","capabilities":{},"clientInfo":{"name":"test","version":"1.0.0"}}}'
```

It would be helpful if this validation was an opt-in / compatibility option for SDK users. That would give server implementers a migration path so they can adopt newer SDK versions without immediately breaking older clients.

WDYT? Happy to submit a PR for this

guglielmo-san:
Fixes #957

### DIFF
--- a/docs/mcpgodebug.md
+++ b/docs/mcpgodebug.md
@@ -18,6 +18,15 @@
 
 ## `MCPGODEBUG` history
 
+### 1.6.1
+
+Options listed below were added and will be removed in the 1.8.0 version of the SDK.
+
+- `disablecontenttypecheck` added. If set to `1`, Content-Type validation on
+  HTTP POST requests will be disabled, allowing requests with non-JSON or missing
+  Content-Type headers. The default behavior is to validate that HTTP POST
+  requests have Content-Type: application/json.
+
 ### 1.6.0
 
 Options listed below were added and will be removed in the 1.8.0 version of the SDK.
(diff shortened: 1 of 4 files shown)


yarolegovich (approved):

pamanta (commented):
Nice 👌. Thank you for looking into it

PR #973, opened 2026-05-23
fix: reject duplicate initialize requests

### LINKED ISSUE #961: Duplicate initialize with changed parameters can overwrite stored session initialization parameters

he-yufeng:
﻿Fixes #961.

## Summary

- Reject a second `initialize` request after a session has already stored its initial parameters.
- Keep the original `InitializeParams` intact when a duplicate request tries to change client info, capabilities, or protocol version.
- Add a regression test that exercises a raw JSON-RPC connection and verifies the duplicate request returns an error.

## Tests

- `go test ./mcp -run TestServerRejectsDuplicateInitialize -count=1`
- `go test ./mcp -run "TestServerRejectsDuplicateInitialize|TestClientRootCapabilities" -count=1`
- `go test ./mcp -count=1`
- `go test ./...`
- `git diff --check`

### DIFF
--- a/mcp/server.go
+++ b/mcp/server.go
@@ -1488,9 +1488,17 @@ func (ss *ServerSession) initialize(ctx context.Context, params *InitializeParam
 	if params == nil {
 		return nil, fmt.Errorf("%w: \"params\" must be be provided", jsonrpc2.ErrInvalidParams)
 	}
+	var duplicate bool
 	ss.updateState(func(state *ServerSessionState) {
-		state.InitializeParams = params
+		duplicate = state.InitializeParams != nil
+		if !duplicate {
+			state.InitializeParams = params
+		}
 	})
+	if duplicate {
+		ss.server.opts.Logger.Error("duplicate initialize request")
+		return nil, fmt.Errorf("%w: duplicate %q received", jsonrpc2.ErrInvalidRequest, methodInitialize)
+	}
 
 	s := ss.server
 	return &InitializeResult{
--- a/mcp/server_test.go
+++ b/mcp/server_test.go
@@ -825,6 +825,97 @@ func TestClientRootCapabilities(t *testing.T) {
 	}
 }
 
+func TestServerRejectsDuplicateInitialize(t *testing.T) {
+	ctx := context.Background()
+	server := NewServer(&Implementation{Name: "testServer", Version: "v1.0.0"}, nil)
+
+	cTransport, sTransport := NewInMemoryTransports()
+	ss, err := server.Connect(ctx, sTransport, nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer ss.Close()
+
+	cConn, err := cTransport.Connect(ctx)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer cConn.Close()
+
+	firstParams := json.RawMessage(`{
+		"protocolVersion": "2025-11-25",
+		"capabilities": {"roots": {"listChanged": true}},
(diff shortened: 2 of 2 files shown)


guglielmo-san:
Is this a duplicate of PR #962 ?

he-yufeng:
Yes, this is a duplicate of #962. I opened #973 from the same issue after #962 had already converged, so I'm closing this one and keeping the approved #962 as the active PR. Thanks for catching it.

PR #975, opened 2026-05-26
mcp: Implement support for SEP-2575 on client side

### LINKED ISSUE #966: Implement SEP-2575: Make MCP Stateless

guglielmo-san:
Implements the client half of the SEP-2575 / `2026-06-30` protocol. This CL is client-only; server-side `server/discover` is still a `MethodNotFound` stub.
### Changes
- **`server/discover` probe in `Client.Connect()`** (`client.go`). When the caller opts into `protocolVersion >= 2026-06-30`, the client probes the server with `server/discover` before falling back to the legacy `initialize`
  handshake.
- **Per-request `_meta` injection** (`shared.go`, `injectRequestMeta`). For`>= 2026-06-30` sessions every outgoing request gets the SEP-2575 triple (`protocolVersion`, `clientInfo`, `clientCapabilities`).
- **Server-side fix**: `ServerSession.handle` no longer pre-populates `state.InitializeParams` from a `server/discover` request's `_meta`, so a subsequent fallback `initialize` is not rejected as a duplicate.
### Fallback rules
Client falls back to the legacy `initialize` handshake when discover returns:
- `MethodNotFound` (-32601),
- `UnsupportedProtocolVersionError` (-32004),
- an HTTP `Bad Request` (substring-matched on the error message, to interop
  with pre-2026-06-30 servers).

Fixes #966


mcp/client.go
```diff
@@ -299,6 +321,67 @@ func (c *Client) Connect(ctx context.Context, t Transport, opts *ClientSessionOp
...
+//
+// The return values have three possible combinations:
+//   - (result, false, nil): discovery succeeded; caller should skip legacy initialization.
+//   - (nil, true, nil):     the server explicitly signaled it doesn't support
+//     discovery (Method not found, or UnsupportedProtocolVersionError, or version mismatch);
+//     caller should fall back to the legacy initialize handshake.
+//   - (nil, false, err):    any other failure (transport error, malformed response, etc.);
+//     caller should propagate the error.
+func (c *Client) discover(ctx context.Context, cs *ClientSession) (*InitializeResult, bool, error) {
+	protocolVersion, _ := ctx.Value(protocolVersionContextKey{}).(string)
```
yarolegovich: protocolVersionFromContext()?

mcp/client.go
```diff
@@ -299,6 +321,67 @@ func (c *Client) Connect(ctx context.Context, t Transport, opts *ClientSessionOp
...
+	if err != nil {
+		// According to SEP-2575, only the two signals below (MethodNotFound
+		// and UnsupportedProtocolVersionError) should trigger a fallback.
+		// However, to allow communication between vPost clients and vPre servers,
+		// we trigger fallback for "Bad Request" errors too.
+		var werr *jsonrpc.Error
+		if errors.As(err, &werr) && (werr.Code == jsonrpc.CodeMethodNotFound || werr.Code == CodeUnsupportedProtocolVersion) {
+			return nil, true, nil
+		}
+		if strings.Contains(err.Error(), "Bad Request") {
```
yarolegovich: this is quite unreliable. 
maybe we can use other jsonrpc codes like 32600 invalid request. maybe introduce internal error errNotFound and add http-transport logic for %w-ing it into the error when status code is 404, add a check here

mcp/streamable.go
```diff
@@ -2143,6 +2150,8 @@ func (c *streamableClientConn) setMCPHeaders(req *http.Request) error {
 	}
 	if c.initializedResult != nil {
 		req.Header.Set(protocolVersionHeader, c.initializedResult.ProtocolVersion)
+	} else if v, ok := req.Context().Value(protocolVersionContextKey{}).(string); ok && v != "" {
```
yarolegovich: nit: protocolVersionFromContext

internal/jsonrpc2/wire.go
```diff
@@ -31,7 +31,7 @@ var (
 	// ErrUnknown should be used for all non coded errors.
 	ErrUnknown = NewError(-32001, "unknown error")
 	// ErrServerClosing is returned for calls that arrive while the server is closing.
-	ErrServerClosing = NewError(-32004, "server is closing")
+	ErrServerClosing = NewError(-32006, "server is closing")
```
yarolegovich, line 34: oh, that's not nice . SDK claimed error codes which are now also used in the spec.
```go
ErrClientClosing = NewError(-32003, "client is closing")
```
will [now mean](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2575/changes#diff-2145ad529c53fcee40347e6c4f527b390b350dd12d37c5e0dad02ff4b613468fR378) `MISSING_REQUIRED_CLIENT_CAPABILITY`.

for the server side we can do mcgdebug for returning `-32004` when server is closing for a session with and old protocol version.
I guess we need to handle it in `internal/jsonrpc2/wire.go` is-check looking not just at the code but at the message as well 😞

internal/jsonrpc2/conn.go
```diff
@@ -720,7 +720,7 @@ func (c *Connection) write(ctx context.Context, msg Message) error {
 
 	// For cancelled or rejected requests, we don't set the writeErr (which would
 	// break the connection). They can just be returned to the caller.
-	if err != nil && ctx.Err() == nil && !errors.Is(err, ErrRejected) {
+	if err != nil && ctx.Err() == nil && !errors.Is(err, ErrRejected) && !errors.Is(err, ErrUnsupportedProtocolVersion) && !errors.Is(err, ErrMethodNotFound) {
```
yarolegovich, line 723: I wonder if any server implementation close a connection on their side in case of an unknown method request

mcp/client.go
```diff
@@ -999,16 +1106,34 @@ func (cs *ClientSession) Ping(ctx context.Context, params *PingParams) error {
...
 	return handleSend[*ListPromptsResult](ctx, methodListPrompts, newClientRequest(cs, orZero[Params](params)))
 }
 
 // GetPrompt gets a prompt from the server.
 func (cs *ClientSession) GetPrompt(ctx context.Context, params *GetPromptParams) (*GetPromptResult, error) {
+	if cs.usesNewProtocol() {
+		if params == nil {
+			params = &GetPromptParams{}
+		}
+		cs.injectRequestMeta(params)
```
yarolegovich: nit:
```go
func injectRequestMeta[T any, P interface {
	*T
	Params
}](cs *ClientSession, params P) P {
	if params.isNil() {
      params = new(T)
	}
	...
	params.SetMeta(m)
	return params
}
```
and then:
```go
params = injectRequestMeta(cs, params)
```

internal/jsonrpc2/wire.go
guglielmo-san: I am not sure this is needed. The `ErrClientClosing` and `ErrServerClosing` are never sent over the wire and the error code is never returned to the calling function. Are only used internally in the client / server implementation.

mcp/client.go
guglielmo-san: Nice!

internal/jsonrpc2/conn.go
guglielmo-san: In this case it will not pass through this path. The server will send a request from the SSE stream, and the client will respond with a new response that will eventually contain the error. But that will be a new POST Response

nahapetyan-serob (approved):

PR #979, opened 2026-05-27
mcp: skip keepalive pings while session has in-flight requests

tdabasinskas:
mcp: skip keepalive pings while session has in-flight requests

The current keepalive loop pings unconditionally. When the server holds a long-running request from a Streamable HTTP client that reads the SSE stream but does not post a ping response back through the transport (e.g. Claude Desktop), the server's Ping call times out, `session.Close()` runs mid-handler, and the eventual tool result is dropped on the floor.

The keepalive mechanism exists to detect dead *idle* peers. An incoming request the session is still processing is itself a continuous liveness signal — and one already paid for. Skipping pings while `hasPendingRequests()` reports true sidesteps the race entirely without loosening the "MAY terminate on missed ping" behaviour for genuinely idle sessions.

### DIFF
--- a/mcp/client.go
+++ b/mcp/client.go
@@ -391,6 +391,12 @@ func (cs *ClientSession) getCachedTool(name string) *Tool {
 	return cs.toolCache[name]
 }
 
+// hasPendingRequests is the client-side counterpart of
+// ServerSession.hasPendingRequests. See keepaliveSession in shared.go.
+func (cs *ClientSession) hasPendingRequests() bool {
+	return false
+}
+
 // registerElicitationWaiter registers a waiter for an elicitation complete
 // notification with the given elicitation ID. It returns two functions: an await
 // function that waits for the notification or context cancellation, and a cleanup
--- a/mcp/server.go
+++ b/mcp/server.go
@@ -1144,6 +1144,16 @@ type ServerSession struct {
 	state ServerSessionState
 }
 
+func (ss *ServerSession) hasPendingRequests() bool {
+	type pendingReporter interface {
+		pendingClientRequests() int
+	}
+	if c, ok := ss.mcpConn.(pendingReporter); ok {
+		return c.pendingClientRequests() > 0
+	}
+	return false
+}
+
 func (ss *ServerSession) updateState(mut func(*ServerSessionState)) {
 	ss.mu.Lock()
 	mut(&ss.state)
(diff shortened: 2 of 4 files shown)


guglielmo-san:
Hi @tdabasinskas, the proposed behavior is not mentioned anywhere in the spec, which is instead requiring 

> The receiver MUST respond promptly with an empty response:

https://modelcontextprotocol.io/specification/2025-11-25/basic/utilities/ping#behavior-requirements

I believe the implementation of Claude Desktop is not compliant to the spec, as the Go SDK Client, should correctly handle ping requests while handling long running requests from the server

tdabasinskas:
> Hi @tdabasinskas, the proposed behavior is not mentioned anywhere in the spec, which is instead requiring
> 
> > The receiver MUST respond promptly with an empty response:
> 
> https://modelcontextprotocol.io/specification/2025-11-25/basic/utilities/ping#behavior-requirements
> 
> I believe the implementation of Claude Desktop is not compliant to the spec, as the Go SDK Client, should correctly handle ping requests while handling long running requests from the server

Agreed - the client is the non-compliant side here; a conformant client answers pings even mid-tool-call, and Claude Desktop doesn't. No dispute on where the bug is.

The one thing I'd still raise is independent of that: the spec says **"multiple"** failed pings MAY trigger a reset, but the current loop closes on the **first** miss. That's stricter than the spec's wording regardless of client behavior. A small consecutive-failure threshold - or a configurable `KeepaliveFailureThreshold` on `ServerOptions`, defaulting to current behavior - would align with the "multiple" language.

The practical angle is that the client is the part we *can't* fix - it's a third-party app on someone else's release cycle - whereas a bit of server-side tolerance is something operators can actually address today.

guglielmo-san:
> or a configurable KeepaliveFailureThreshold on ServerOptions, defaulting to current behavior - would align with the "multiple" language

Agree on this, if you want you can modify the PR including this option

tdabasinskas:
> > or a configurable KeepaliveFailureThreshold on ServerOptions, defaulting to current behavior - would align with the "multiple" language
> 
> Agree on this, if you want you can modify the PR including this option

Sure! Should the current implementation with pending requests check in this PR be removed?

guglielmo-san:
Yes, I would remove the current implementation

tdabasinskas:
> Yes, I would remove the current implementation

As the implementation differs, I've opened separate #982 for this and closing this one.

PR #980, opened 2026-05-27
fix: preserve empty-method requests during decode

### LINKED ISSUE #976: Stdio server leaves some malformed JSON-RPC requests unanswered while the session remains alive
**Describe the bug**

Some malformed JSON-RPC requests sent over stdio receive no JSON-RPC error response even though the request carries an `id` and the server continues processing later requests.

The clearest case is an empty method name. After this request, the server gives no response for `id=5`, but later `ping` requests still succeed on the same stdio session. This looks like a recoverable request-dispatch path.

I am filing this as a conformance and robustness issue, not as a security issue. Malformed requests should be rejected, but requests with ids should receive a correlated JSON-RPC error response when the session can continue.

Environment tested:

- Stable release: `v1.6.1` (`d454bbaf06a342aee5336df3370321d9cdec2478`)
- `main` snapshot from 2026-05-26: `c044cdb734d8d321d596f5b6077072798dce4ad6`
- Transport: stdio

**To Reproduce**

1. Start the Go SDK stdio server.
2. Complete a normal init handshake.
3. Send this malformed request:
```json
{"jsonrpc":"2.0","id":5,"method":"","params":{}}
```
4. Observe that no response is returned for that request.
5. Send later `ping` request and observe that the server still responds.

**Expected behavior**

For the primary case, I would expect a JSON-RPC error response for `id=5`, most likely `-32601 Method not found` or `-32600 Invalid Request` depending on how strictly the SDK wants to validate method names.

Reasoning:

- JSON-RPC 2.0 requests include an `id`; notifications omit `id` and do not require a response.
- `method` is a string in both JSON-RPC and MCP. If an empty method name is invalid at the MCP layer, this still looks like a recoverable request-dispatch error rather than a transport read failure, because the session continues.

**Observed behavior**

The primary repro case produced no JSON-RPC response for `id=5`, and a later `ping` health check succeeded on the same session.

**Additional context**

The same isolated test also found other malformed request cases. I am including them for completeness because they show the boundary between recoverable no-response behavior and connection-fatal behavior.

|Case | Request | Observed behavior|
|-- | -- | --|
|id as null | {"jsonrpc":"2.0","id":null,"method":"ping","params":{}} | No JSON-RPC response for id=null; later ping requests still succeeded.|
|Wrong JSON-RPC version | {"jsonrpc":"1.0","id":3,"method":"ping","params":{}} | No JSON-RPC response; process exited with Server failed: invalid message version tag "1.0"; expected "2.0".|
|Missing jsonrpc field | {"id":4,"method":"ping","params":{}} | No JSON-RPC response; process exited with Server failed: invalid message version tag ""; expected "2.0".|
|method as number | {"jsonrpc":"2.0","id":8,"method":12345,"params":{}} | No JSON-RPC response; process exited with Server failed: unmarshaling jsonrpc message: json: cannot unmarshal ... into Go struct field ... method of type string.|

- The id:null case may be another recoverable MCP-level rejection case. JSON-RPC 2.0 treats id:null as distinct from an omitted id, although null ids are discouraged. MCP narrows request ids to string or number.

- The other three cases are structurally invalid JSON-RPC envelopes: jsonrpc must be "2.0" and method must be a string. If the Go SDK intentionally treats these decoder/envelope failures as fatal stdio read errors, documenting that policy would help clients distinguish expected connection teardown from a lost response.

Related prior issue: #179 was closed in the v0.3.0 milestone and appears to cover a narrower malformed trailing-garbage case. The isolated cases above still reproduce on `v1.6.1` and current `main`.

he-yufeng:
﻿## Summary

Fixes #976.

`DecodeMessage` used `msg.Method != ""` to distinguish requests from responses. That means a wire payload with a present but empty `method` field, such as:

```json
{"jsonrpc":"2.0","id":5,"method":"","params":{}}
```

was decoded as a response instead of a request. Since there was no matching outgoing call, the server ignored it and did not send a correlated JSON-RPC error, even though the session stayed alive.

This PR checks whether the wire payload contains the `method` field instead of checking whether the decoded string is non-empty. An empty method now reaches the normal request dispatch path and receives the existing method-not-found handling.

## Tests

```text
go test ./internal/jsonrpc2
go test ./mcp
go test ./...
git diff --check
```

### DIFF
--- a/internal/jsonrpc2/messages.go
+++ b/internal/jsonrpc2/messages.go
@@ -183,7 +183,11 @@ func DecodeMessage(data []byte) (Message, error) {
 	if err != nil {
 		return nil, err
 	}
-	if msg.Method != "" {
+	var fields map[string]json.RawMessage
+	if err := internaljson.Unmarshal(data, &fields); err != nil {
+		return nil, fmt.Errorf("unmarshaling jsonrpc message fields: %w", err)
+	}
+	if _, hasMethod := fields["method"]; hasMethod {
 		// has a method, must be a call
 		return &Request{
 			Method: msg.Method,
--- a/internal/jsonrpc2/wire_test.go
+++ b/internal/jsonrpc2/wire_test.go
@@ -63,6 +63,23 @@ func TestWireMessage(t *testing.T) {
 	}
 }
 
+func TestDecodeMessageWithEmptyMethodIsRequest(t *testing.T) {
+	msg, err := jsonrpc2.DecodeMessage([]byte(`{"jsonrpc":"2.0","id":5,"method":"","params":{}}`))
+	if err != nil {
+		t.Fatal(err)
+	}
+	req, ok := msg.(*jsonrpc2.Request)
+	if !ok {
+		t.Fatalf("DecodeMessage returned %T, want *jsonrpc2.Request", msg)
+	}
+	if req.Method != "" {
+		t.Fatalf("Request.Method = %q, want empty string", req.Method)
+	}
+	if got := req.ID.Raw(); got != int64(5) {
+		t.Fatalf("Request.ID = %v, want 5", got)
+	}
+}
+
 func newNotification(method string, params any) jsonrpc2.Message {
 	msg, err := jsonrpc2.NewNotification(method, params)
 	if err != nil {


he-yufeng:
﻿Rebased on current `origin/main` and reran validation locally.

Validated:

```powershell
gofmt -w $(git diff --name-only origin/main...HEAD | Where-Object { $_ -like '*.go' })
go test ./mcp
go test -race ./mcp
go test ./...
git diff --check origin/main...HEAD
```

All passed after the rebase.

he-yufeng:
Rebased this branch onto the latest upstream `main` and force-pushed the cleaned head.

Validated locally:
- `gofmt -l internal/jsonrpc2/messages.go internal/jsonrpc2/wire_test.go` (no output)
- `go test ./internal/jsonrpc2`
- `go test ./...`

he-yufeng:
Rebased this branch onto current upstream `main` and reran the focused/full Go validation.

Validation:

```text
gofmt -l internal\jsonrpc2\messages.go internal\jsonrpc2\wire_test.go
go test ./internal/jsonrpc2
go test ./...
git diff --check origin/main..HEAD
```

Current head: `dbbe137`.

guglielmo-san:
Closing as duplicate of  PR #1000

PR #985, opened 2026-06-01
fix: bound stdio JSON-RPC message size

### LINKED ISSUE #984: Expose max JSON-RPC frame/message size for stdio transports
Doc Relay uses the Go MCP SDK with child MCP servers over stdio (CommandTransport). We need a host-side bound on JSON-RPC frame/message size so a buggy or malicious child cannot send one very large newline-delimited JSON frame and force the parent to buffer/decode it before application code can reject it.\n\nCurrent observation in v1.6.0: the framing/read loop is owned inside the SDK transport/JSON-RPC stack, so hosts using CommandTransport do not appear to have a supported seam equivalent to Decoder.MaxBytes, io.LimitReader per frame, or MaxMessageBytes. Replacing the SDK framing locally would conflict with using the SDK as the JSON-RPC/MCP boundary.\n\nRequested API shape could be any of:\n\n- a MaxMessageBytes/MaxFrameBytes option on CommandTransport or Client/Server options,\n- an exported transport wrapper hook that bounds each newline-delimited frame before JSON decoding, or\n- documented guidance for safely enforcing this without reimplementing the SDK framing loop.\n\nThe important behavior is that an oversized child stdout frame is rejected and the connection is closed before unbounded parent heap growth.

he-yufeng:
﻿Fixes #984.

## Summary

Adds an optional `MaxMessageBytes` limit for newline-delimited JSON-RPC transports:

- `CommandTransport.MaxMessageBytes` bounds frames read from child process stdout.
- `StdioTransport.MaxMessageBytes` and `IOTransport.MaxMessageBytes` expose the same behavior for server-side stdio and custom IO transports.
- The IO reader now frames input by NDJSON line, preserves the existing trailing-data rejection, and rejects oversized frames before JSON-RPC decoding.

## Validation

- `go test ./mcp -run "TestIOConnRead|TestIOConnReadMaxMessageBytes|TestBatchFraming"`
- `go test ./mcp -run TestCommandTransportTerminateDuration`
- `go test ./...`
- `go vet ./...`
- `gofmt -l .`
- `go run honnef.co/go/tools/cmd/staticcheck@v0.6.1 ./...`
- `git diff --check -- mcp\transport.go mcp\cmd.go mcp\transport_test.go`

### DIFF
--- a/mcp/cmd.go
+++ b/mcp/cmd.go
@@ -19,6 +19,10 @@ var defaultTerminateDuration = 5 * time.Second // mutable for testing
 // with it over stdin/stdout, using newline-delimited JSON.
 type CommandTransport struct {
 	Command *exec.Cmd
+	// MaxMessageBytes, if positive, rejects incoming JSON-RPC message payloads
+	// from the subprocess larger than this many bytes. The line ending is not
+	// counted.
+	MaxMessageBytes int64
 	// TerminateDuration controls how long Close waits after closing stdin
 	// for the process to exit before sending SIGTERM.
 	// If zero or negative, the default of 5s is used.
@@ -43,7 +47,7 @@ func (t *CommandTransport) Connect(ctx context.Context) (Connection, error) {
 	if td <= 0 {
 		td = defaultTerminateDuration
 	}
-	return newIOConn(&pipeRWC{t.Command, stdout, stdin, td}), nil
+	return newIOConnWithOptions(&pipeRWC{t.Command, stdout, stdin, td}, t.MaxMessageBytes), nil
 }
 
 // A pipeRWC is an io.ReadWriteCloser that communicates with a subprocess over
--- a/mcp/transport.go
+++ b/mcp/transport.go
@@ -5,6 +5,8 @@
 package mcp
 
 import (
+	"bufio"
+	"bytes"
 	"context"
 	"encoding/json"
 	"errors"
@@ -111,11 +113,15 @@ type serverConnection interface {
 
 // A StdioTransport is a [Transport] that communicates over stdin/stdout using
 // newline-delimited JSON.
-type StdioTransport struct{}
+type StdioTransport struct {
+	// MaxMessageBytes, if positive, rejects incoming JSON-RPC message payloads
+	// larger than this many bytes. The line ending is not counted.
+	MaxMessageBytes int64
+}
 
 // Connect implements the [Transport] interface.
-func (*StdioTransport) Connect(context.Context) (Connection, error) {
-	return newIOConn(rwc{os.Stdin, nopCloserWriter{os.Stdout}}), nil
+func (t *StdioTransport) Connect(context.Context) (Connection, error) {
+	return newIOConnWithOptions(rwc{os.Stdin, nopCloserWriter{os.Stdout}}, t.MaxMessageBytes), nil
 }
 
 // nopCloserWriter is an io.WriteCloser with a trivial Close method.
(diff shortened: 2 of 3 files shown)


he-yufeng:
﻿Rebased on current `origin/main` and reran validation locally.

Validated:

```powershell
gofmt -w mcp\cmd.go mcp\transport.go mcp\transport_test.go
go test ./mcp
go test -race ./mcp
go test ./...
git diff --check origin/main...HEAD
```

All passed after the rebase.

he-yufeng:
Fresh local re-validation on the current checkout:

- `go test ./mcp` -> passed
- `gofmt -l mcp/cmd.go mcp/transport.go mcp/transport_test.go` -> no output
- `git diff --check origin/main...HEAD` -> passed

The GitHub checks are also green as of the latest run.

he-yufeng:
Rebased this branch onto the latest upstream `main` and force-pushed the cleaned head.

Validated locally:
- `gofmt -l mcp/cmd.go mcp/transport.go mcp/transport_test.go` (no output)
- `go test ./mcp`
- `go test -race ./mcp`
- `go test ./...`

piyushbag:
Nice refactor in the stdio framing path. A few notes from reading the diff:

1. **Pre-decode bound** — Checking `len(frame)+len(part)` against `MaxMessageBytes` before `validateJSONFrame` avoids allocating/decoding oversized payloads. That composes well with other stdio hardening work on malformed envelopes.

2. **`ReadSlice` + `ErrBufferFull`** — The continue loop on `bufio.ErrBufferFull` is correct for long lines, but it does grow `frame` without an early cap when `maxMessageBytes == 0`. Worth confirming that is intentional (unbounded mode preserves prior behavior) and that the capped path rejects before unbounded growth when `maxMessageBytes > 0`.

3. **Limit semantics** — `TestIOConnReadMaxMessageBytes` uses `int64(len(input))` without the trailing newline, which matches rejecting only the JSON payload bytes. A one-line doc comment on whether the limit includes `\n`/`\r\n` would help server authors pick values.

4. **Surface area** — Exposing the knob on `CommandTransport`, `StdioTransport`, and `IOTransport` covers the main server entry points I have been exercising in go-sdk transport tests.

I ran a quick read-only review against current `main`; I have not tested this branch locally. Looks like a solid fix for #984.

he-yufeng:
Thanks for the careful read. I pushed `880a8de` with a small follow-up for the limit semantics you called out.

What changed:
- documented `MaxMessageBytes` as applying to the JSON-RPC payload bytes, not the NDJSON line ending
- changed `readFrame` so LF/CRLF are trimmed before the final capped-size check
- kept the `bufio.ErrBufferFull` path capped when `MaxMessageBytes > 0`
- added regression coverage for payload-at-limit with LF and CRLF
- added a regression check that `MaxMessageBytes == 0` preserves the old unbounded behavior

Validation:
- `gofmt -w mcp\cmd.go mcp\transport.go mcp\transport_test.go`
- `go test ./mcp`
- `go test -race ./mcp`
- `go test ./...`
- `git diff --check`

piyushbag (approved):
The follow-up commit addresses the limit-semantics and capped ReadSlice paths from the earlier review. Doc comment, trim-before-check behavior, and regression coverage for LF/CRLF and zero-limit unbounded mode look correct.

piyushbag:
Thanks for the follow-up. The payload-vs-line-ending semantics and the CRLF/LF regression tests match what I had in mind. Looks good from my read.

he-yufeng:
Rebased this onto current `main` and re-ran the local Go checks.

Current head: `6b03643`

Validation:
- `gofmt -l mcp\cmd.go mcp\transport.go mcp\transport_test.go` -> no output
- `go test ./mcp`
- `go test -race ./mcp`
- `go test ./...`

All passed locally after the rebase.

he-yufeng:
Superseded by #1205 (merged 2026-09-02): stdio frame size limiting landed upstream. Closing.

PR #1002, opened 2026-06-10
mcp: add ListResourcesHandler for dynamic resources/list

### LINKED ISSUE #998: Proposal: dynamic/lazy resources/list (and directory-read) backing for gateway/proxy and template-served catalogs
## Problem

`resources/list` can only be served from a **static, in-memory set**. Resources are materialized up front via `AddResource` (removed via `RemoveResources`), and `(*Server).listResources` enumerates that `featureSet` directly — there is no handler/callback that computes the list on demand. For two important classes of server this isn't just inconvenient, it's a structural mismatch.

### 1. Gateway / proxy servers (the important case)

A **skill (or resource) gateway** fronts one or more upstream catalogs it does **not own** — another registry, a remote index, or several backend MCP servers. The authoritative list lives elsewhere and changes independently of this process. To serve `resources/list` from a static set, the gateway would have to **mirror the entire upstream catalog into memory at startup and keep it continuously in sync** — which defeats the purpose of being a gateway, doesn't scale to large/unbounded upstreams, and is racy against upstream changes.

What a gateway actually needs is to answer `resources/list` by **querying upstream on demand**, passing the client's pagination cursor through. There is no first-class way to do that today.

### 2. Template-served assets are non-enumerable by construction

Skills are served as a *directory of files*. A server typically exposes a skill's files via a **resource template** (`skill://<skill>/{+path}`) plus one read handler, so it doesn't have to `AddResource` every individual file. But a URI template matches infinitely many URIs — **you cannot derive the file list from the template definition**. So those assets are fully *readable* yet absent from any static set: `resources/list` won't show them, and `resources/directory/read` (below) backed by the static set returns empty.

## Relationship to `resources/directory/read` (#956 / SEP-2640)

The [Skills-over-MCP WG](https://modelcontextprotocol.io/community/skills-over-mcp/charter) is adding a `resources/directory/read` method ([SEP-2640](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2640)) so hosts can enumerate the files within a skill directory — e.g. to materialize a skill to disk. Implementing that verb in this SDK depends on custom-method support (#956). But even once the verb exists, it hits **exactly this problem**: for a template-served or gateway-proxied skill, the directory's children cannot come from the static set — `resources/directory/read` must be backed by a **server-provided dynamic listing**.

So the two are a pair:
- **#956** — adds the `resources/directory/read` verb (custom method).
- **this issue** — gives `resources/list` / directory-read a **dynamic backing** so template-served and gateway-proxied entries can actually be enumerated.

Neither alone lets a Go skills/gateway server both *serve* and *enumerate* a catalog whose contents aren't pre-materialized.

## What already works (to scope precisely)

`resources/read` is already dynamic — content comes from a `ResourceHandler`, and resource templates (`AddResourceTemplate` + `ResourceTemplate.Matches`) let one handler serve an unbounded URI space. The asymmetry is the whole point: **templates make reads unbounded, but nothing makes listing/enumeration unbounded or on-demand.**

## Workarounds today (and why they fall short)

1. **Receiving middleware.** `Middleware` wraps `MethodHandler func(ctx, method string, req Request)(Result, error)`, so middleware can intercept `resources/list` and synthesize a `*ListResourcesResult`, short-circuiting the static handler. It works, but it's untyped (switch on a string method, type-assert `req`/`Result`), bypasses the SDK's built-in pagination, and sits awkwardly beside the otherwise-typed registration model.
2. **Per-session server via the `getServer` callback** (as suggested in #216). Good for per-connection *configuration*, but it still requires materializing a concrete set per server — it doesn't help a catalog that is upstream, generated, or too large to enumerate eagerly.

## Proposal

Add a first-class way to serve `resources/list` (and back `resources/directory/read`) from a handler, with cursor-based pagination handled by the SDK — analogous to how `ResourceHandler`/templates already make reads dynamic. Illustrative shape (not prescriptive):

```go
// Serves resources/list on demand; receives the request (incl. pagination cursor)
// and returns a page of resources plus the next cursor.
type ListResourcesHandler func(context.Context, *ListResourcesRequest) (*ListResourcesResult, error)

opts := &mcp.ServerOptions{
    ListResourcesHandler: func(ctx context.Context, req *mcp.ListResourcesRequest) (*mcp.ListResourcesResult, error) {
        // e.g. proxy upstream using req.Params.Cursor, or enumerate a template's namespace ...
    },
}
```

When set, the handler would take precedence over (or compose with) the registered static set, and the same mechanism would back directory enumeration for a directory URI.

## Questions for maintainers

- Was the static-set model for `resources/list` a deliberate choice (uniformity with tools/prompts feature sets), and would a dynamic list handler be acceptable — or is middleware / per-session-server the intended long-term answer for gateway and unbounded catalogs?
- If a handler is acceptable: should it *replace* or *merge with* the registered set, and should the same backing apply to `resources/directory/read` (#956) and, for consistency, `prompts/list` / `tools/list`?

Happy to prototype if there's interest in the direction.

piyushbag:
## Problem

`resources/list` is backed only by a static in-memory set populated via `AddResource`. Gateway servers that proxy upstream catalogs and template-served skill assets cannot enumerate resources on demand: reads work via `ResourceHandler`, but listing does not.

## Approach

Add `ListResourcesHandler` on `ServerOptions`, analogous to `ResourceHandler` for reads:

- **Handler only** (no static resources): client pagination cursors pass through unchanged. Suitable for gateway/proxy servers.
- **Handler + static resources**: SDK paginates registered resources first, then invokes the handler for subsequent pages using an internal composite cursor.
- **Capability**: advertise `resources` when the handler is set, even if no resources were added statically.

`resources/directory/read` backing is intentionally deferred until custom-method support (#956) lands.

## Why this helps

- First-class, typed registration instead of intercepting `resources/list` in middleware.
- Preserves existing static `AddResource` behavior when no handler is configured.
- Composes static and dynamic catalogs for template-backed servers that also register fixed entries.

## Test plan

- [x] `go test ./mcp/ -run TestListResourcesHandler`
- [x] `go test ./...`
- [x] `go test -race ./mcp/ -run TestListResourcesHandler`

Fixes #998

### DIFF
--- a/mcp/resource.go
+++ b/mcp/resource.go
@@ -38,6 +38,21 @@ type serverResourceTemplate struct {
 // If it cannot find the resource, it should return the result of calling [ResourceNotFoundError].
 type ResourceHandler func(context.Context, *ReadResourceRequest) (*ReadResourceResult, error)
 
+// A ListResourcesHandler serves resources/list on demand.
+//
+// Set [ServerOptions.ListResourcesHandler] to enable dynamic resource listing for
+// gateway servers, template-backed catalogs, and other cases where the resource
+// set cannot be materialized up front.
+//
+// If the server also has static resources registered with [Server.AddResource],
+// those are listed first using SDK pagination, then this handler is invoked for
+// subsequent pages. If there are no static resources, the client's pagination
+// cursor is passed through to this handler unchanged.
+//
+// The handler should set Resources to a non-nil empty slice when there are no
+// resources on the page.
+type ListResourcesHandler func(context.Context, *ListResourcesRequest) (*ListResourcesResult, error)
+
 // customresnotfounderrcode is a compatibility parameter that restores the
 // pre-1.7.0 behavior of [ResourceNotFoundError] and [CodeResourceNotFound],
 // where the error code was a custom -32002. See the documentation for the mcpgodebug
--- a/mcp/resource_list_handler_test.go
+++ b/mcp/resource_list_handler_test.go
@@ -1,0 +1,144 @@
+// Copyright 2025 The Go MCP SDK Authors. All rights reserved.
+// Use of this source code is governed by an MIT-style
+// license that can be found in the LICENSE file.
+
+package mcp
+
+import (
+	"context"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+func TestListResourcesHandlerDynamicOnly(t *testing.T) {
+	ctx := context.Background()
+	var gotCursor string
+	server := NewServer(testImpl, &ServerOptions{
+		ListResourcesHandler: func(_ context.Context, req *ListResourcesRequest) (*ListResourcesResult, error) {
+			gotCursor = req.Params.Cursor
+			if req.Params.Cursor != "" {
+				return &ListResourcesResult{Resources: []*Resource{}}, nil
+			}
+			return &ListResourcesResult{
+				Resources: []*Resource{{URI: "dynamic://a"}},
+				NextCursor: "page2",
+			}, nil
+		},
+	})
+	client := NewClient(testImpl, nil)
+	st, ct := NewInMemoryTransports()
+	ss, err := server.Connect(ctx, st, nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer ss.Close()
+	cs, err := client.Connect(ctx, ct, nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer cs.Close()
+
+	res, err := cs.ListResources(ctx, nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+	want := []*Resource{{URI: "dynamic://a"}}
+	if diff := cmp.Diff(want, res.Resources); diff != "" {
+		t.Fatalf("first page mismatch (-want +got):\n%s", diff)
+	}
+	if gotCursor != "" {
+		t.Fatalf("first page cursor = %q, want empty", gotCursor)
+	}
+	if res.NextCursor != "page2" {
+		t.Fatalf("NextCursor = %q, want page2", res.NextCursor)
+	}
+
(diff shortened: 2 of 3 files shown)


piyushbag:
Local verification on `fix-998-list-resources-handler` (855cf72):

```
go test ./... -count=1
ok  github.com/modelcontextprotocol/go-sdk/mcp
ok  github.com/modelcontextprotocol/go-sdk/oauthex

go test -race ./mcp/... -run TestListResourcesHandler -count=1
ok  github.com/modelcontextprotocol/go-sdk/mcp
```

I kept `resources/directory/read` out of scope per the PR note until custom methods (#956) land. Open to splitting static+dynamic pagination into a follow-up if maintainers want a smaller first merge.

piyushbag:
Rebased onto current `main` (includes #1000) and force-pushed `6eb52d4`.

Local gate:
```
go test ./... -count=1
go test -race ./mcp/... -run TestListResourcesHandler -count=1
ok  github.com/modelcontextprotocol/go-sdk/mcp
```

GitHub CI on this push: lint, test 1.25/1.26, race, client/server conformance, CodeQL — all green.

Scope unchanged: handler-only pagination with cursor passthrough for gateway/proxy cases; `resources/directory/read` still deferred until #956. Open to splitting static+dynamic pagination if maintainers want a smaller first merge.

piyushbag:
Friendly ping @findleyr — I've rebased this onto the latest `main` (resolved a conflict in `mcp/server.go` so `resources/list` results still get `setDefaultCacheableValues()`). Local `go test ./...`, `go test -race ./mcp/...`, `go vet ./...`, and `staticcheck ./...` are all green. Would appreciate a review when you have a moment. Thanks!

piyushbag:
Closing in favor of #1030, which supersedes this PR — same changes, rebased onto the latest `main`. Please review #1030 instead.

PR #1007, opened 2026-06-15
mcp: Implement `subscriptions/listen` rpc (SEP-2575)

### LINKED ISSUE #966: Implement SEP-2575: Make MCP Stateless

guglielmo-san:
Fixes #966


jba:
> subscrptions/listen

nit: subscriptions

> Fixes #966

There are many parts to that SEP. Is this the last one to be implemented?

mcp/server.go
```diff
@@ -682,12 +687,56 @@ func (s *Server) changeAndNotify(notification string, change func() bool) {
...
+		params := changeNotificationParams[n]()
+		injectMetaSubscriptionID(params, reqID)
+		req := newRequest(sess, params)
+		if err := handleNotify(ctx, n, req); err != nil {
+			s.opts.Logger.Warn(fmt.Sprintf("calling %s: %v", n, err))
+		}
+	}
+}
+
+func injectMetaSubscriptionID(params Params, reqID jsonrpc.ID) {
```
jba, line 745: Document this with a reference to the part of the SEP that requires it. (https://modelcontextprotocol.io/seps/2575-stateless-mcp#multiple-concurrent-subscriptions, paragraph 2)

mcp/server.go
```diff
@@ -643,12 +649,11 @@ func (s *Server) complete(ctx context.Context, req *CompleteRequest) (*CompleteR
 	return s.opts.CompletionHandler(ctx, req)
 }
 
-// Map from notification name to its corresponding params. The params have no fields,
-// so a single struct can be reused.
-var changeNotificationParams = map[string]Params{
-	notificationToolListChanged:     &ToolListChangedParams{},
-	notificationPromptListChanged:   &PromptListChangedParams{},
-	notificationResourceListChanged: &ResourceListChangedParams{},
+// Map from notification name to its corresponding params.
```
jba: "...to a function creating its corresonding Params. We need to create a fresh one each time to add the jsonrpc ID. See injectMetaSubscriptionID."

mcp/server.go
```diff
@@ -682,12 +687,56 @@ func (s *Server) changeAndNotify(notification string, change func() bool) {
...
+		subscribers = maps.Clone(s.resourceChangeSubscriptions)
+	}
 	s.mu.Unlock() // Don't hold the lock during notification: it causes deadlock.
-	notifySessions(sessions, n, changeNotificationParams[n], s.opts.Logger)
+
+	// Legacy sessions receive list-changed notifications on the shared session channel without opt-in.
+	var legacySessions []*ServerSession
+	for _, sess := range sessions {
+		if sess.InitializeParams().isNil() || sess.InitializeParams().ProtocolVersion < protocolVersion20260630 {
+			legacySessions = append(legacySessions, sess)
```
jba: This is the second copy of s.sessions. The first is on line 697. Remove the clone and move this loop inside the locked section above.

mcp/client.go
```diff
@@ -300,6 +300,28 @@ func (c *Client) Connect(ctx context.Context, t Transport, opts *ClientSessionOp
...
+				}
+				if subscribeParams.Notifications.ToolsListChanged ||
+					subscribeParams.Notifications.PromptsListChanged ||
+					subscribeParams.Notifications.ResourcesListChanged {
+					// The listen blocks until the server cancels it. Run it in
+					// a goroutine so Connect can return; ClientSession.Close
+					// cancels its context to send notifications/cancelled.
+					listenCtx, cancelListen := context.WithCancel(context.Background())
+					cs.listenCancel = cancelListen
+					go func() {
```
jba: nit: I think you can write

    go cs.subscriptionsListen(...)

jba (changes requested):

internal/jsonrpc2/conn.go
```diff
@@ -210,10 +210,14 @@ type ConnectionConfig struct {
 	OnInternalError func(error)               // optional
 }
 
+type IsSubscriptionsListenContextKey struct{}
```
jba: I have two objections to this approach: first, the name is specific to a particular transport-level use case. It should be more general. Second, using a context to pass the information should be a last resort. Can we instead have a different constructor?

jba: You wrote in your message to me "that's what lets the listen handler observe HTTP disconnect on stateless streamable HTTP. Without this, the handler would never see the client going away." Document that here (but remove references to the specific protocol). And while you're at it, maybe you can document why the default is to wrap with `notDone`. A lot of this package is mysterious to me and if you understand it, drop some knowledge where you can.

internal/jsonrpc2/conn.go
```diff
@@ -724,10 +736,13 @@ func (c *Connection) write(ctx context.Context, msg Message) error {
...
 	// Fail writes immediately if the connection is shutting down.
 	//
-	// TODO(rfindley): should we allow cancellation notifications through? It
-	// could be the case that writes can still succeed.
+	// Allow outgoing "notifications/cancelled" to allow the remote end
+	// to cancel in-flight calls.
 	c.updateInFlight(func(s *inFlightState) {
-		err = s.shuttingDown(ErrServerClosing)
+		req, ok := msg.(*Request)
+		if !ok || req.Method != "notifications/cancelled" {
```
jba: It's unfortunate that this general package is referring to a specific MCP feature. We should be treating it as if we can some day move it somewhere else and export it. Can you think of another way to do this? What if there were a list of methods that 

Maybe we can add a list of methods that can succeed during shutdown to ConnectionConfig. And also add a bool there to replace IsSubscriptionsListenContextKey.

mcp/client.go
```diff
@@ -1276,17 +1312,87 @@ func (cs *ClientSession) Complete(ctx context.Context, params *CompleteParams) (
...
+		_, err := handleSend[*emptyResult](ctx, methodSubscribe, newClientRequest(cs, orZero[Params](params)))
+		return err
+	}
+	if params == nil || params.URI == "" {
+		return fmt.Errorf("Subscribe: missing URI")
+	}
+	uri := params.URI
+
+	cs.resourceSubsMu.Lock()
+	if _, exists := cs.resourceSubs[uri]; exists {
```
jba: nit: write this section as
```
lock
if ...; !exists {
   ...
}
unlock
```
to avoid the awkward unlock on 1326.

mcp/server.go
```diff
@@ -1221,6 +1384,12 @@ func (ss *ServerSession) NotifyProgress(ctx context.Context, params *ProgressNot
 	return handleNotify(ctx, notificationProgress, newServerRequest(ss, orZero[Params](params)))
 }
 
+// NotifySubscriptionAcked sends a subscription acknowledged notification from the server to the client
+// associated with this session.
+func (ss *ServerSession) NotifySubscriptionAcked(ctx context.Context, params *SubscriptionsAcknowledgedParams) error {
```
jba: Why does this need to be exported? Doesn't our server logic call it already? Would users ever need to call it?

jba (changes requested):

internal/jsonrpc2/conn.go
guglielmo-san: Done, I added a parameter on the ConnectionConfig to signal the need to propagate the request cancellation, and added some explanation around it

internal/jsonrpc2/conn.go
guglielmo-san: Right, I updated to follow the same logic of Conn.Notify(), if a notification reaches write will be allowed to pass

mcp/server.go
guglielmo-san: Indeed, thank you for pointing it out

jba (approved):
Nice, thanks for those last changes.

PR #1027, opened 2026-06-26
mcp: retry transient errors on SSE reconnect

aditya-786:
## Summary

Completes the remaining work for #683 that #723 called out: "we should also retry transient errors in handleSSE."

When an SSE stream is interrupted and the client reconnects, `streamableClientConn.handleSSE` passed any `checkResponse` error to `c.fail`, permanently breaking the session. But `checkResponse` intentionally wraps transient HTTP statuses (429, 502, 503, 504) in `jsonrpc2.ErrRejected` so they should *not* break the connection — the POST path already honors this (`if !errors.Is(err, jsonrpc2.ErrRejected) { c.fail(err) }`), while the SSE reconnect path did not. As a result a transient 503 during a reconnect (server restart, load-balancer hiccup, load spike) poisoned the whole session — the failure mode reported in #683.

## Change

In the reconnect loop, when `checkResponse` returns an `ErrRejected`-wrapped (transient) error, retry the reconnect instead of failing. Each retry counts against the existing no-progress retry budget (#679), so a persistently unavailable server still eventually gives up rather than looping forever. Non-transient errors are unchanged.

## Test

`TestStreamableClientReconnectTransientErrors` reproduces the bug with the existing `fakeStreamableServer` harness: a POST tool call returns an SSE stream that emits a resumable event then ends without the response (forcing a reconnect); the reconnect returns 503 once, then the real response. The test fails before this change (`... Service Unavailable (session should survive)`) and passes after.

Verified locally: `go test ./...`, `go test -race ./...`, `go vet ./...`, `gofmt -l`, and `staticcheck` all clean.

## Scope note

`connectStandaloneSSE` has the same unconditional `c.fail` on a transient status. The standalone GET stream is optional (§2.2.3), so I scoped this PR to `handleSSE` (the case #723 named). Happy to also fix the standalone path here or in a follow-up — whichever you prefer.

For #683

### DIFF
--- a/mcp/streamable.go
+++ b/mcp/streamable.go
@@ -2432,21 +2432,36 @@ func (c *streamableClientConn) handleSSE(ctx context.Context, requestSummary str
 		}
 
 		// The stream was interrupted or ended by the server. Attempt to reconnect.
-		newResp, err := c.connectSSE(ctx, lastEventID, reconnectDelay, false)
-		if err != nil {
-			// If the client didn't cancel this request, any failure to execute it
-			// breaks the logical MCP session.
-			if ctx.Err() == nil {
-				// All reconnection attempts failed: fail the connection.
-				c.fail(fmt.Errorf("%s: failed to reconnect (session ID: %v): %v", requestSummary, c.sessionID, err))
+		for {
+			newResp, err := c.connectSSE(ctx, lastEventID, reconnectDelay, false)
+			if err != nil {
+				// If the client didn't cancel this request, any failure to execute it
+				// breaks the logical MCP session.
+				if ctx.Err() == nil {
+					// All reconnection attempts failed: fail the connection.
+					c.fail(fmt.Errorf("%s: failed to reconnect (session ID: %v): %v", requestSummary, c.sessionID, err))
+				}
+				return
 			}
-			return
-		}
 
-		resp = newResp
-		if err := c.checkResponse(ctx, requestSummary, resp); err != nil {
-			c.fail(err)
-			return
+			if err := c.checkResponse(ctx, requestSummary, newResp); err != nil {
+				if errors.Is(err, jsonrpc2.ErrRejected) {
+					retriesWithoutProgress++
+					if retriesWithoutProgress > c.maxRetries {
+						if ctx.Err() == nil {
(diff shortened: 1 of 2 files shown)


piyushbag (approved):
This matches the POST-side `ErrRejected` handling: transient reconnect failures should not poison the whole session.

Local check:

```
go test ./mcp/... -run 'TestStreamableClientReconnectTransientErrors|TestStreamableClientTransientErrors|TestStreamableClientRetryWithoutProgress' -count=1
```

All pass. `TestStreamableClientReconnectTransientErrors` exercises the 503-on-reconnect case well, and the `retriesWithoutProgress` cap keeps the #679 behavior intact.

mcp/streamable.go
```diff
@@ -2469,21 +2469,36 @@ func (c *streamableClientConn) handleSSE(ctx context.Context, requestSummary str
...
+				if errors.Is(err, jsonrpc2.ErrRejected) {
+					retriesWithoutProgress++
+					if retriesWithoutProgress > c.maxRetries {
+						if ctx.Err() == nil {
+							c.fail(fmt.Errorf("%s: exceeded %d retries without progress (session ID: %v)", requestSummary, c.maxRetries, c.sessionID))
+						}
+						return
+					}
+					reconnectDelay = 0
+					continue
```
guglielmo-san, line 2493: why did you decide to reset reconnectDelay ?

aditya-786: `reconnectDelay` here is the delay the server asked for via the SSE `retry:` field on the stream that just ended (`processStream` reads it out of that field). It is a one-shot hint tied to that specific stream, so I honor it on the first reconnect attempt.

Once the reconnect actually connects but the server rejects it transiently (`ErrRejected`) and we loop back to retry, there is no fresh `retry:` value for those in-loop attempts. Continuing to pass the old one would re-apply that same server delay on every retry, even though it belonged to the previous stream. Zeroing it lets `connectSSE` fall back to `calculateReconnectDelay`, which is the normal jittered exponential backoff.

It does not become a no-delay busy loop: `connectSSE` runs with `initial=false`, so `attempt` starts at 1 and `delay = calculateReconnectDelay(1)` (a real backoff); passing `reconnectDelay == 0` just means we no longer override that. The `retriesWithoutProgress > maxRetries` cap bounds the loop overall.

Happy to add a one-line comment on that line to spell out the intent if it reads as surprising.

PR #1029, opened 2026-06-27
mcp: add StreamableHTTPHandler.Close for graceful shutdown

### LINKED ISSUE #440: Proposal: add StreamableHTTPHandler.Close

piyushbag:
## Problem

`StreamableHTTPHandler` holds stateful sessions in an internal map but had no public way to shut them down. Callers embedding the handler in an HTTP server (or running it in tests) could not reliably tear down all MCP sessions or block new ones during shutdown. The issue author proposed a `Close()` method; the codebase already had a private `closeAll()` helper with a TODO noting that a public API must also prevent new sessions from being added.

## Approach

- Add `Close() error` on `StreamableHTTPHandler`: set a `closed` flag, snapshot and clear the sessions map under lock, then call `session.Close()` on each entry outside the lock (avoids deadlocking with session `onClose` callbacks).
- Reject new traffic in `ServeHTTP` with `503 Service Unavailable` once closed.
- Make `Close` idempotent (second call is a no-op).
- Keep the existing test-only `closeAll()` as a thin delegate to `Close()`.
- Add `TestStreamableHTTPHandlerClose` covering session teardown, empty map, idempotency, and 503 after close.

## Why this is better

- Matches the API shape requested in the issue (`Close() error`, closes all sessions, prevents new sessions).
- Enables idiomatic cleanup: `defer handler.Close()` in `main` or test teardown without reaching into unexported fields.
- Minimal diff: extends existing session lifecycle patterns rather than introducing callbacks or new types.
- Aligns with graceful-shutdown patterns in other MCP SDKs (explicit handler/session disposal) while staying Go-native.

Fixes #440

### DIFF
--- a/mcp/streamable.go
+++ b/mcp/streamable.go
@@ -51,6 +51,7 @@ type StreamableHTTPHandler struct {
 	onTransportDeletion func(sessionID string) // for testing
 
 	mu       sync.Mutex
+	closed   bool
 	sessions map[string]*sessionInfo // keyed by session ID
 }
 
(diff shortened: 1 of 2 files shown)


piyushbag:
Rebased onto current `main` and resolved the `ServeHTTP` conflict by keeping both the closed-handler 503 gate and the `MaxRequestBodyBytes` body bound.

Local: `go test ./mcp/ -run TestStreamableHTTPHandlerClose` and `go test ./mcp/...` pass.

piyushbag:
CI is green on the latest head. Ready for review when you have bandwidth — happy to rebase or adjust if anything should change before merge.

PR #1031, opened 2026-06-27
mcp: typed custom methods and extension registration

sambhav:
Custom method consumers currently repeat the method name and generic types at registration and call sites. This follow-up to #956 packages those details once for extension authors.

As agreed in [the review discussion](https://github.com/modelcontextprotocol/go-sdk/pull/1031#issuecomment-4909705620), this PR now covers client-to-server custom methods only. The server-initiated APIs and their additional dispatch maps have been removed.

```go
var Translate = mcp.NewCustomMethod[*TranslateParams, *TranslateResult]("latin/translate")

if err := Translate.RegisterServer(server, handler); err != nil {
    return err
}
if err := Translate.RegisterClient(client); err != nil {
    return err
}
result, err := Translate.Call(ctx, session, &TranslateParams{Text: "hello"})
```

- `CustomMethod` wraps the existing registration and call functions with `RegisterServer`, `RegisterClient`, and `Call`.
- `RegisterExtension` supports package-level setup; `ServerOptions.Extensions` and `ClientOptions.Extensions` support per-instance setup. Global extensions run first, in registration order; later registrations of a custom method replace earlier ones. Existing instances are unaffected.
- One helper applies extensions outside the registry lock. Either callback may be nil; callback errors panic during construction.
- The Latin example demonstrates an extension package exposing a simple `Translate` helper.
- Current `main` is merged and conflicts resolved.

Validation on Go 1.26.1:
- Custom-method and extension tests pass with `-race`, including all supported protocol versions, typed-nil params, registration errors, handler errors, ordering, overrides, scope, and concurrent registration.
- Existing multi-round-trip and server-initiated-request rejection tests pass with `-race`.
- `go build ./...`, `go vet ./...`, formatting checks, and the Latin example pass.
- The full local race suite passes with a temporary Go test overlay making the two existing localhost-protection fixtures target `127.0.0.1` instead of their wildcard listener address. This avoids the environment's proxy while preserving their Host headers and assertions; repository test files are unchanged.
- [GitHub CI](https://github.com/modelcontextprotocol/go-sdk/actions/runs/34556229634) passes the unmodified full suite on Go 1.25 and 1.26, the full race suite, formatting, vet, and staticcheck. [Conformance](https://github.com/modelcontextprotocol/go-sdk/actions/runs/34556229641) and [CodeQL](https://github.com/modelcontextprotocol/go-sdk/actions/runs/34556229628) also pass.

### DIFF
--- a/examples/server/custom-method/latinext/latin.go
+++ b/examples/server/custom-method/latinext/latin.go
@@ -1,0 +1,87 @@
+// Copyright 2025 The Go MCP SDK Authors. All rights reserved.
+// Use of this source code is governed by an MIT-style
+// license that can be found in the LICENSE file.
+
+// Package latinext is an example MCP extension that adds a "latin/translate"
+// custom JSON-RPC method. It demonstrates the extension-author pattern: types,
+// the CustomMethod variable, and the init() registration are all defined here
+// so that importers get everything wired up automatically.
+package latinext
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/modelcontextprotocol/go-sdk/mcp"
+)
+
+// TranslateParams are the parameters for the latin/translate method.
+type TranslateParams struct {
+	mcp.ParamsBase
+	Text string `json:"text"`
+}
+
+// TranslateResult is the result of the latin/translate method.
+type TranslateResult struct {
+	mcp.ResultBase
+	Latin string `json:"latin"`
+}
+
+// Method captures the method name and types once. Extension consumers call
+// Translate() rather than this directly.
+var Method = mcp.NewCustomMethod[*TranslateParams, *TranslateResult]("latin/translate")
+
+func init() {
+	mcp.RegisterExtension(mcp.Extension{
+		Server: func(s *mcp.Server) error {
+			return Method.RegisterServerReceiving(s, DefaultHandler)
+		},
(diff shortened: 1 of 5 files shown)


piyushbag (approved):
Looked through the extension registry and bidirectional custom-method wiring. The `CustomMethod` wrapper and `latinext` example make the #956 follow-up easy to follow.

Local check:

```
go test ./mcp/... -count=1
```

All green on my machine.

One non-blocking suggestion: `TestCustomMethods` only exercises client→server today. A small round-trip test for `AddServerSendingCustomMethod` + `AddClientReceivingCustomMethod` (or `ServerCallCustomMethod`) would lock in the new path.

guglielmo-san:
@sambhav thank you for your contribution.
The new mcp protocol disallows Server initiated requests, and should instead be based on the [MultiRoundTripRequest](https://modelcontextprotocol.io/specification/draft/basic/patterns/mrtr) 
That's why I would lean for not approving this PR

sambhav:
> @sambhav thank you for your contribution. The new mcp protocol disallows Server initiated requests, and should instead be based on the [MultiRoundTripRequest](https://modelcontextprotocol.io/specification/draft/basic/patterns/mrtr) That's why I would lean for not approving this PR

gotcha, let me remove those bits. I think the rest of the syntactic sugar still helps though.

Issue #1034, opened 2026-06-29
server/discover omits Instructions while initialize includes them (parity gap)


SamMorrowDrums:
## Summary
`server/discover` (SEP-2575) omits `instructions` from its result even when the server was constructed with `ServerOptions.Instructions`, while `initialize` returns it. Both handlers read the same `s.opts.Instructions`, so the two handshakes disagree on the same server.

## Observed in
go-sdk `v1.7.0-pre.1`, server constructed with non-empty `ServerOptions.Instructions`.

- `initialize` @ 2025-11-25 → `instructions` present (~1.2KB)
- `initialize` @ 2026-07-28 → `instructions` present
- `server/discover` @ 2026-07-28 → `instructions` **absent** (stdio and streamable-HTTP)

## Spec status
Not a conformance violation — `DiscoverResult.instructions` is optional (draft `schema.json` `required: [capabilities, resultType, serverInfo, supportedVersions]`). But it's a behavioral inconsistency: a client adopting `server/discover` silently loses the server's tool-selection guidance that the same server advertises via `initialize`.

## Code
Both handlers source the same field:
- `mcp/server.go:876` (discover): `Instructions: s.opts.Instructions`
- `mcp/server.go:1884` (initialize): `Instructions: s.opts.Instructions`

So either the stateless/discover path resolves a server without `opts.Instructions` populated, or it's stripped before marshal. Worth confirming whether discover runs against a different Server/session instance under stateless transport.

## Repro
Construct `mcp.Server` with `ServerOptions{Instructions: "hello"}`. Drive `server/discover` (HTTP: `MCP-Protocol-Version: 2026-07-28`, `Mcp-Method: server/discover`, reserved `_meta` protocolVersion/clientInfo/clientCapabilities). Result has no `instructions`; an `initialize` against the same server does.

## Testing gap
No test asserts `DiscoverResult.Instructions == s.opts.Instructions`. Discover/initialize parity (instructions, capabilities, serverInfo) is untested. Suggest a table test asserting both handshakes surface identical server-identity fields for a server configured with instructions.

SamMorrowDrums:
[...]

PR #1035, opened 2026-06-29
mcp: add discover/initialize Instructions parity test

SamMorrowDrums:
## Summary
Adds a regression test for #1034: assert that `server/discover` and `initialize` surface identical server-identity fields (instructions, serverInfo, capabilities) for a server configured with `ServerOptions.Instructions`, over both stateful and stateless streamable HTTP.

## Background
#1034 reported that discover omitted instructions while initialize included them. Both handlers already read `s.opts.Instructions` (`server.go:885` and `:1901`), and a wire-level repro confirms discover returns instructions correctly in every transport path on current main and v1.7.0-pre.1 — so there is no SDK behavior bug. The asymmetry the reporter saw originates in the consumer resolving a discover-path `Server` whose options weren't populated with instructions. This adds the missing parity coverage the issue called out, guarding the new SEP-2575 discover path against future regressions.

## Testing
- `go test ./...` passes.

Addresses #1034.

### DIFF
--- a/mcp/discover_parity_test.go
+++ b/mcp/discover_parity_test.go
@@ -1,0 +1,154 @@
+// Copyright 2025 The Go MCP SDK Authors. All rights reserved.
+// Use of this source code is governed by the license
+// that can be found in the LICENSE file.
+
+package mcp
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"reflect"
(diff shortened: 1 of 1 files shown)


SamMorrowDrums:
Closing — SDK was never broken (discover already returns instructions == initialize). Root cause was downstream in github-mcp-server (fixed in e5c53ae2). A regression test for already-correct behavior is overkill; closing this and #1034.

PR #1044, opened 2026-07-01
auth: add ScopeFilter hook to AuthorizationCodeHandlerConfig

toabctl:
## Summary

The authorization-code handler requests every scope advertised in the protected resource's metadata `scopes_supported` (or named in the `WWW-Authenticate` challenge), and offers the caller no way to adjust that set. Some servers advertise a scope a client should *not* request:

- Google's Gmail MCP server advertises `gmail.metadata` in `scopes_supported`.
- The Gmail API refuses the search `q` parameter on any token that carries `gmail.metadata` — **even when `gmail.readonly` is also granted** (a long-standing documented Gmail restriction, see googleapis/google-api-python-client#582).

Because the handler requests the full advertised set, `gmail.metadata` ends up in the grant and breaks search, with no client-side remedy.

## Change

Add an optional `ScopeFilter func(discovered []string) []string` callback to `AuthorizationCodeHandlerConfig`. When non-nil, it receives the discovered scopes and returns the set to request, giving the client full control to narrow, extend, or reorder them. It runs **before** the `offline_access` (SEP-2207) and step-up union (SEP-2350) logic, so refresh-token support and scope accumulation are unaffected.

(Started as a static `Scopes []string` allowlist; switched to a callback per review feedback for more flexibility.)

## Test

`TestAuthorize_ScopeFilter` drives a full authorize flow and asserts the requested `scope` parameter for a nil filter (unchanged), a filter that drops a scope, and one that replaces the set entirely. `go test ./auth/...` (incl. `-race`), `go vet`, and `gofmt` are clean.

### DIFF
--- a/auth/authorization_code.go
+++ b/auth/authorization_code.go
@@ -97,6 +97,15 @@ type AuthorizationCodeHandlerConfig struct {
 	// See [AuthorizationCodeFetcher] for details.
 	AuthorizationCodeFetcher AuthorizationCodeFetcher
 
+	// ScopeFilter, if non-nil, is called with the scopes discovered from the
+	// protected resource metadata (or WWW-Authenticate challenge) and returns
+	// the scopes to request during authorization. It gives the client full
+	// control to narrow, extend, or reorder that set — e.g. dropping Gmail's
+	// gmail.metadata, which the Gmail API refuses to combine with the search "q"
+	// parameter even alongside gmail.readonly. It runs before offline_access
+	// (see RequestRefreshToken) and the step-up union, so neither is affected.
+	ScopeFilter func(discovered []string) []string
+
 	// RequestRefreshToken indicates that the client intends to use refresh
 	// tokens and is capable of storing them securely.
 	//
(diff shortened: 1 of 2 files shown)


auth/authorization_code.go
```diff
@@ -97,6 +97,15 @@ type AuthorizationCodeHandlerConfig struct {
...
 
+	// Scopes optionally restricts the requested scopes to this allowlist,
+	// intersecting it with the scopes discovered from metadata/challenge
+	// (preserving order). This lets a client drop an advertised scope it does
+	// not want — e.g. Gmail's gmail.metadata, which the Gmail API refuses to
+	// combine with the search "q" parameter even alongside gmail.readonly. An
+	// allowlist matching nothing is ignored so the flow never requests an empty
+	// set; offline_access is exempt.
+	Scopes []string
+
```
guglielmo-san: I think that a callback would be more clear, and would leave more control to the user on the set of scopes to mantain

toabctl: Good call — switched from the static `Scopes` allowlist to a `ScopeFilter func(discovered []string) []string` callback. It receives the discovered scopes and returns the set to request, so the caller has full control to narrow, extend, or reorder them (and can maintain whatever set they want). Runs before the offline_access and step-up union logic. Updated the test accordingly.

guglielmo-san (approved):

guglielmo-san:

PR #1046, opened 2026-07-01
mcp: avoid panic for new-protocol null IDs

### LINKED ISSUE #1043: Server panics on `id: null` request with per-request protocol `_meta`
**Describe the bug**

When an inbound JSON-RPC frame includes `"id": null` and per-request protocol `_meta` (`io.modelcontextprotocol/protocolVersion: "2026-07-28"`), the server panics with a nil pointer dereference and the process exits with `SIGSEGV`.

In MCP draft, request IDs **MUST** be strings or integers and **MUST NOT** be `null`. Therefore, this request-shaped input is malformed and should be rejected with a JSON-RPC error, but it should not crash the server process.

From the stack trace and code path, this appears to happen when the per-request `_meta` validation takes the non-call path for `id: null`, leaving `initializeParams` nil, and `ServerSession.handle` later dereferences it during protocol-version validation.

This code path does not exist in the v1.6.0 stable release (`f5f2015`), where `handle()` dispatches directly to method handlers without calling `validateRequestMeta`. The crash is a regression introduced with the per-request protocol detection (SEP-2575).

Environment tested:

- `main` snapshot from 2026-06-28: `65ece5964304528e7568507e2b17351824ef73fc`
- Transport: stdio

**To Reproduce**

1. Start the Go SDK stdio server.
2. Send a message with `"id": null` and per-request `_meta`:

```json
{"jsonrpc":"2.0","id":null,"method":"completion/complete","params":{"_meta":{"io.modelcontextprotocol/protocolVersion":"2026-07-28","io.modelcontextprotocol/clientInfo":{"name":"repro","version":"0.1.0"},"io.modelcontextprotocol/clientCapabilities":{}},"ref":{"type":"ref/prompt","name":"test"},"argument":{"name":"arg","value":"x"}}}
```

3. The server process terminates with a panic.

**Expected behavior**

The server should reject the malformed input with a JSON-RPC error, likely `-32600 Invalid Request`. At minimum, the server should not crash.

**Observed behavior**

The server process exits with:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x70b501]

goroutine 13 [running]:
github.com/modelcontextprotocol/go-sdk/mcp.(*ServerSession).handle(0x..., {0x8abf18, 0x...}, 0x...)
	.../mcp/server.go:1816 +0x101
github.com/modelcontextprotocol/go-sdk/internal/jsonrpc2.HandlerFunc.Handle(...)
	.../internal/jsonrpc2/jsonrpc2.go:60 +0x29
github.com/modelcontextprotocol/go-sdk/internal/jsonrpc2.(*Connection).handleAsync.func3()
	.../internal/jsonrpc2/conn.go:685 +0x76
```

The crash reproduced in repeated isolated runs with the same stack trace.

**Additional context**

This appears to be specific to the stdio path: HTTP calls `checkRequest()` before dispatching the message to the session handler, so the same malformed input is rejected before it reaches `ServerSession.handle`.

The root cause spans two functions:

In `mcp/shared.go`, `validateRequestMeta` explicitly returns `initializeParams: nil` for non-call messages:

```go
// validateRequestMeta (mcp/shared.go)
func validateRequestMeta(req *jsonrpc.Request) (*validatedMeta, error) {
    // ...
    // Notifications do not carry full client identity.
    if !req.IsCall() {
        return &validatedMeta{usesNewProtocol: true, initializeParams: nil}, nil
    }
    // ...
}
```

In `mcp/server.go`, `handle` dereferences `initializeParams` without checking for nil:

```go
// handle (mcp/server.go)
if validatedMeta.usesNewProtocol &&
    !slices.Contains(supportedProtocolVersions, validatedMeta.initializeParams.ProtocolVersion) {
    //                                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    //                                          nil when IsCall() returned false
```

A minimal panic-avoidance guard would be:

```go
if validatedMeta.usesNewProtocol && validatedMeta.initializeParams != nil &&
    !slices.Contains(supportedProtocolVersions, validatedMeta.initializeParams.ProtocolVersion) {
```

A fuller fix may be to reject `id: null` before using `req.IsCall()` to choose the call vs. notification path, so the malformed frame is handled as an invalid request rather than as a notification.

cyphercodes:
This fixes a panic in the server-side per-request protocol metadata path when an inbound message has `id: null` and new-protocol `_meta`.

The request decodes as notification-shaped, so `validateRequestMeta` returns new-protocol metadata without initialize params. Guard the protocol-version check before reading initialize params, allowing the existing request validation path to reject the malformed call instead of dereferencing nil.

Tests:
- `go test ./mcp -run 'Test(ServerHandleNewProtocolNullID|ValidateRequestMeta)$' -count=1`
- `go test ./mcp -count=1`
- `git diff --check HEAD~1..HEAD`

Fixes #1043

### DIFF
--- a/mcp/server.go
+++ b/mcp/server.go
@@ -1812,7 +1812,7 @@ func (ss *ServerSession) handle(ctx context.Context, req *jsonrpc.Request) (any,
 		return nil, perRequestErr
 	}
 
-	if validatedMeta.usesNewProtocol &&
+	if validatedMeta.usesNewProtocol && validatedMeta.initializeParams != nil &&
 		!slices.Contains(supportedProtocolVersions, validatedMeta.initializeParams.ProtocolVersion) {
 		data, _ := json.Marshal(UnsupportedProtocolVersionData{
 			Supported: supportedProtocolVersions,
--- a/mcp/shared_test.go
+++ b/mcp/shared_test.go
@@ -5,12 +5,14 @@
 package mcp
 
 import (
+	"context"
 	"encoding/json"
 	"errors"
 	"strings"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
+	"github.com/modelcontextprotocol/go-sdk/internal/jsonrpc2"
 	"github.com/modelcontextprotocol/go-sdk/jsonrpc"
 )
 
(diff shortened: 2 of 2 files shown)


kuangmi-bit:
PR #1047 fixes this: one-line nil guard + regression test.

guglielmo-san:
Closing as duplicate of PR #1049

PR #1056, opened 2026-07-08
mcp: return nil params for missing optional requests

### LINKED ISSUE #1052: GetParams() returns typed nil pointer wrapped in Params interface, causing nil check bypass and panic
When a client sends a request without params (or with params: null) for methods flagged with missingParamsOK (e.g. notifications/initialized, ping, tools/list), the middleware receives a Params interface that passes a standard != nil check but panics on any method call.

Root cause:

In newServerMethodInfo, when unmarshalParams returns a nil interface (because params are missing and missingParamsOK allows it), newRequest leaves r.Params at its zero value — a typed nil pointer (e.g. (*InitializedParams)(nil)):

```
mi.newRequest = func(s Session, p Params, re *RequestExtra) Request {
    r := &ServerRequest[P]{Session: s.(*ServerSession), Extra: re}
    if p != nil {       // p is nil-interface here → skipped
        r.Params = p.(P)
    }
    return r            // r.Params is (*InitializedParams)(nil)
}
```
When GetParams() returns this typed nil pointer as a Params interface, the interface is non-nil (it carries type information), but calling any method on it (e.g. GetMeta()) dereferences a nil pointer and panics.

Our use case:

We have middleware that extracts _meta from every incoming request:

```
func metaFromRequest(req mcpsdk.Request) map[string]any {
    if params := req.GetParams(); params != nil {  // passes! interface is non-nil
        return params.GetMeta()                     // PANIC: nil pointer dereference
    }
    return nil
}
```
This crashes on notifications/initialized when the client sends no params.

Expected behavior:

GetParams() should return a nil Params interface (not a typed nil pointer) when params are absent, so that standard nil checks work correctly in middleware. Alternatively, GetParams() could use the existing orZero helper to unwrap typed nil pointers before returning.

Stack trace (v1.6.1):

github.com/modelcontextprotocol/go-sdk/mcp.(*InitializedParams).GetMeta(0x0?)
    <autogenerated>:1
Workaround:

Guard metaFromRequest with a method allowlist (only call for methods without missingParamsOK), or use reflect.ValueOf(params).IsNil() instead of params != nil.

ychampion:
Requests with missing optional params could store a typed nil pointer in the typed request. Middleware calling GetParams then received a non-nil Params interface, so a normal nil check did not prevent calls like GetMeta from panicking.

This normalizes typed nil params in GetParams for both client and server requests, while leaving non-nil generated and custom params unchanged.

Fixes #1052

Tested:
- go test ./mcp -run 'Test(ServerRequestGetParamsMissingOptionalParams|RequestGetParamsTypedNilCustomParams)' -count=1\n- go test ./mcp\n- go test ./...\n- git diff --check

### DIFF
--- a/mcp/shared.go
+++ b/mcp/shared.go
@@ -604,8 +604,19 @@ func (*ServerRequest[P]) isRequest() {}
 func (r *ClientRequest[P]) GetSession() Session { return r.Session }
 func (r *ServerRequest[P]) GetSession() Session { return r.Session }
 
-func (r *ClientRequest[P]) GetParams() Params { return r.Params }
-func (r *ServerRequest[P]) GetParams() Params { return r.Params }
+func (r *ClientRequest[P]) GetParams() Params { return requestParams(r.Params) }
+func (r *ServerRequest[P]) GetParams() Params { return requestParams(r.Params) }
+
+func requestParams[P Params](params P) Params {
+	if any(params) == nil {
+		return nil
+	}
+	value := reflect.ValueOf(params)
+	if value.Kind() == reflect.Ptr && value.IsNil() {
+		return nil
+	}
+	return params
+}
 
 func (r *ClientRequest[P]) GetExtra() *RequestExtra { return nil }
 func (r *ServerRequest[P]) GetExtra() *RequestExtra { return r.Extra }
--- a/mcp/shared.go
+++ b/mcp/shared.go
@@ -269,6 +269,9 @@ func newClientMethodInfo[P paramsPtr[T], R Result, T any](d typedClientMethodHan
 		r := &ClientRequest[P]{Session: s.(*ClientSession)}
 		if p != nil {
 			r.Params = p.(P)
+		} else {
+			r.Params = P(new(T))
+			r.missingParams = true
 		}
 		return r
 	}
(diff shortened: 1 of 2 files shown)


mcp/shared.go
```diff
@@ -604,8 +604,19 @@ func (*ServerRequest[P]) isRequest() {}
...
+func requestParams[P Params](params P) Params {
+	if any(params) == nil {
+		return nil
+	}
+	value := reflect.ValueOf(params)
+	if value.Kind() == reflect.Ptr && value.IsNil() {
+		return nil
+	}
+	return params
+}
```
guglielmo-san, line 638: This will cover part of the problem, req.Params will still return the type-nil interface

ychampion: Fixed in d0e13bf. Receive-side request constructors now initialize missing/null params to a zero params value, so direct `req.Params` access is safe, while `GetParams()` still returns nil for middleware that wants to detect absent params.

Validation:
- `go test ./mcp -run 'Test(ServerRequestGetParamsMissingOptionalParams|ClientRequestMissingOptionalParamsUsesZeroValue|RequestGetParamsTypedNilCustomParams)' -count=1`
- `go test ./mcp -count=1`
- `go test ./...`
- `go vet ./...`
- `git diff --check`

kimblekabrain607-wq (approved):

guglielmo-san:
See comment https://github.com/modelcontextprotocol/go-sdk/issues/1052#issuecomment-5667806754

PR #1058, opened 2026-07-08
auth: support OAuth2 session persistence

### LINKED ISSUE #901: Proposal: Client API support for saving and restoring OAuth2 sessions

I am adding MCP OAuth2 support to a client and don't want my users to be re-prompted for authorization if there is still a valid refresh token from their last session.

Currently, clients using `AuthorizationCodeHandler` and `StreamableClientTransport` do not have a built-in mechanism to persist and restore OAuth2 sessions across application restarts.

I propose adding an API to support saving and restoring OAuth2 sessions.

The specific additions to the `auth` package API are:

1. Add a `SessionSaver` callback function to `AuthorizationCodeHandlerConfig`:

```go
	// SessionSaver is an optional function that can be set to save oauth2
	// sessions to persistent storage. If it is set:
	//  - It is called immediately after the authorization code is successfully
	//    exchanged for a token in [AuthorizationCodeHandler.Authorize].
	//  - After a successful call to [AuthorizationCodeHandler.Authorize], the
	//    token source returned by [AuthorizationCodeHandler.TokenSource] will be
	//    a [NewSavingTokenSource].
	// To ensure SessionSaver is called on subsequent token refreshes, clients
	// must use the token source returned by [AuthorizationCodeHandler.TokenSource]
	// (for example, by returning it from your own [OAuthHandler.TokenSource]
	// implementation) after [AuthorizationCodeHandler.Authorize] returns
	// successfully.
	SessionSaver func(*oauth2.Config, *oauth2.Token)
```

2. Add a `NewSavingTokenSource` function:

```go
// NewSavingTokenSource persists OAuth 2.0 sessions by intercepting token
// refreshes from the wrapped [oauth2.TokenSource]. When this wrapper detects
// an access token refresh, it calls the provided session saver with the
// [oauth2.Config] and the new [oauth2.Token].
//
// The initial token argument prevents saver from being called unnecessarily if
// a valid token already exists. If initial is invalid or nil, saver may be
// called once before the wrapped token is refreshed.
//
// Clients implementing [OAuthHandler.TokenSource] can use this constructor to
// wrap token sources loaded from storage to ensure that subsequent refreshes
// during the application's lifetime are intercepted.
func NewSavingTokenSource(wrapped oauth2.TokenSource, config *oauth2.Config, initialToken *oauth2.Token, saver func(*oauth2.Config, *oauth2.Token)) oauth2.TokenSource
```

This design is heavily inspired by the initial design in #785. The config argument was added so that the entire oauth2 session can be persisted. The context was removed as I didn't understand the need for it. I haven't found it necessary but am open to re-adding it if there is a good reason.

The two individual API additions handle saving the oauth2 session during its two phases:

1. After the authorization code exchange in Authorize().
2. Later when the access token expires and is automatically refreshed.

This design is more complex than I would like because it requires passing a session saver function in two places to manage these two phases of the token lifecycle. However I wasn't able to come up with a simpler abstraction within the constraints of the existing API.

Individual SaveConfig / SaveToken / RestoreConfig / RestoreToken methods/fields.

The lifecycle of Token and Config does not align. `oauth2.Token` will change on every refresh. `oauth2.Config` will change only during `Authorize()`. So the minimal work required during a token refresh would be to call `SaveToken`. However this design has several problems:

* What the client actually cares about is the "session". The Token is useless without the Config. So it is likely that clients will save Token and Config together as "session state".
* It is a larger API surface and relies on careful ordering of function calls.
* The client has to individually save Token and Config to their persistent storage, and then restore them individually using the two restore functions, but still ensure the associated `Token` and `Config` are restored together.

**Additional context**

See discussion about this topic on #881.

smlx:
This change adds NewTokenSource and InitialTokenSource fields to AuthorizationCodeHandlerConfig. NewTokenSource enables wrapping the underlying token source to intercept token refreshes. InitialTokenSource enables configuring the token source used by the
AuthorizationCodeHandler on initialization. Together these features facilitate persisting OAuth2 sessions across restarts.

Includes a package example demonstrating the persistence pattern.

Fixes: #901.

See #901 for the previous discussion behind this design.

### DIFF
--- a/auth/auth_example_test.go
+++ b/auth/auth_example_test.go
@@ -1,0 +1,222 @@
+// Copyright 2026 The Go MCP SDK Authors. All rights reserved.
+// Use of this source code is governed by the license
+// that can be found in the LICENSE file.
+
+package auth_test
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"net/url"
+	"sync"
+
+	"github.com/modelcontextprotocol/go-sdk/auth"
+	"github.com/modelcontextprotocol/go-sdk/mcp"
+	"github.com/modelcontextprotocol/go-sdk/oauthex"
+	"golang.org/x/oauth2"
+)
+
+// savingTokenSource is an oauth2.TokenSource that passes the oauth2 config and
+// token to the given saver function each time the access token value changes.
+type savingTokenSource struct {
+	mu          sync.Mutex
+	src         oauth2.TokenSource
+	saver       func(*oauth2.Config, *oauth2.Token) error
+	config      *oauth2.Config
+	accessToken string
+}
+
+func (s *savingTokenSource) Token() (*oauth2.Token, error) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	tok, err := s.src.Token()
+	if err != nil {
+		return nil, err
+	}
+	if s.accessToken != tok.AccessToken {
+		s.accessToken = tok.AccessToken
+		// This saver implementation always returns nil.
+		_ = s.saver(s.config, tok)
+	}
+	return tok, nil
+}
+
+// NewSavingTokenSource persists OAuth 2.0 sessions by intercepting token
+// changes from the wrapped oauth2.TokenSource. When this wrapper detects an
+// access token change, it calls the provided session saver with the
+// oauth2.Config and the new oauth2.Token.
+//
+// initial is an optional access token the caller may already hold (such as a
+// token loaded from storage). It initialises the wrapper's state so that if
+// the first wrapped.Token() call returns the same token, does not trigger a
+// redundant call to saver(). Pass nil when there is no existing token, in
+// which case the first token produced by wrapped.Token() is saved.
+func NewSavingTokenSource(wrapped oauth2.TokenSource, config *oauth2.Config, initial *oauth2.Token, saver func(*oauth2.Config, *oauth2.Token) error) oauth2.TokenSource {
+	if wrapped == nil {
+		return nil
+	}
+	if saver == nil {
+		return wrapped
+	}
+	var accessToken string
+	if initial != nil {
+		accessToken = initial.AccessToken
+	}
+	return &savingTokenSource{
+		src:         wrapped,
+		saver:       saver,
+		config:      config,
+		accessToken: accessToken,
+	}
+}
+
+// sessionStore is an in-memory OAuth 2.0 session store used to demonstrate
+// persistence. A production implementation would persist the session to disk or
+// a secret store.
+type sessionStore struct {
+	config *oauth2.Config
+	token  *oauth2.Token
+}
+
+// save persists the given OAuth 2.0 config and token.
+func (s *sessionStore) save(config *oauth2.Config, token *oauth2.Token) error {
+	fmt.Printf("Saving token: %s\n", token.AccessToken)
+	s.config = config
+	s.token = token
+	return nil
+}
+
+// restore loads a previously persisted OAuth 2.0 session, if one exists.
+func (s *sessionStore) restore() (*oauth2.Config, *oauth2.Token, error) {
+	if s.config != nil && s.token != nil {
+		fmt.Println("Restoring session.")
+	} else {
+		fmt.Println("No session found to restore.")
+	}
+	return s.config, s.token, nil
+}
+
+// newMockAuthServer returns an httptest.Server that simulates both an MCP
+// resource server requiring authorization and its OAuth authorization server.
+func newMockAuthServer() *httptest.Server {
+	return httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		switch r.URL.Path {
+		case "/.well-known/oauth-authorization-server":
+			w.Header().Set("Content-Type", "application/json")
+			_, _ = fmt.Fprintf(w, `{"issuer": "http://%s", "authorization_endpoint": "http://%s/auth", "token_endpoint": "http://%s/token", "code_challenge_methods_supported": ["S256"]}`, r.Host, r.Host, r.Host)
+		case "/token":
+			w.Header().Set("Content-Type", "application/json")
+			_, _ = w.Write([]byte(`{"access_token": "mock-token", "token_type": "bearer"}`))
+		default:
+			// The mock MCP endpoint returns 401 until the client presents a valid
+			// bearer token.
+			if r.Header.Get("Authorization") == "Bearer mock-token" {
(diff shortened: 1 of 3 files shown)


auth/authorization_code.go
```diff
@@ -197,10 +214,14 @@ func NewAuthorizationCodeHandler(config *AuthorizationCodeHandlerConfig) (*Autho
 	if config.Client == nil {
 		config.Client = http.DefaultClient
 	}
-	return &AuthorizationCodeHandler{
+	h := &AuthorizationCodeHandler{
 		config:        config,
 		grantedScopes: make(map[string][]string),
-	}, nil
+	}
+	if config.InitialTokenSource != nil {
```
jba: I think you can just set it unconditionally.

auth/authorization_code.go
```diff
@@ -615,7 +636,15 @@ func (h *AuthorizationCodeHandler) exchangeAuthorizationCode(ctx context.Context
 	// completes. Use a background context that still carries the configured HTTP
 	// client so refreshes keep working for the life of the token source.
 	refreshCtx := context.WithValue(context.Background(), oauth2.HTTPClient, h.config.Client)
-	h.tokenSource = cfg.TokenSource(refreshCtx, token)
+	if h.config.NewTokenSource != nil {
+		ts, err := h.config.NewTokenSource(refreshCtx, cfg, token)
+		if err != nil {
+			return fmt.Errorf("construction of new token source failed: %v", err)
```
jba: "constructing token source"

auth/authorization_code.go
```diff
@@ -615,7 +636,15 @@ func (h *AuthorizationCodeHandler) exchangeAuthorizationCode(ctx context.Context
 	// completes. Use a background context that still carries the configured HTTP
 	// client so refreshes keep working for the life of the token source.
 	refreshCtx := context.WithValue(context.Background(), oauth2.HTTPClient, h.config.Client)
-	h.tokenSource = cfg.TokenSource(refreshCtx, token)
+	if h.config.NewTokenSource != nil {
```
jba: change to == and do the easy, default case first.

jba (changes requested):

auth/authorization_code.go
jba: use %w

auth/authorization_code.go
```diff
@@ -126,6 +126,23 @@ type AuthorizationCodeHandlerConfig struct {
...
 	Client *http.Client
+
+	// NewTokenSource is an optional function that can be set to construct the
+	// token source that will be used by the [AuthorizationCodeHandler]. If
+	// non-nil, it is called after the authorization code is successfully
+	// exchanged for a token in [AuthorizationCodeHandler.Authorize]
+	// to obtain the [oauth2.TokenSource] returned by
+	// [AuthorizationCodeHandler.TokenSource]. The default is to call
+	// [oauth2.Config.TokenSource].
+	NewTokenSource func(context.Context, *oauth2.Config, *oauth2.Token) (oauth2.TokenSource, error)
```
jba, line 139: "Implementations must used the provided context, which is properly configured for constructing a TokenSource."

auth/authorization_code.go
```diff
@@ -615,7 +636,15 @@ func (h *AuthorizationCodeHandler) exchangeAuthorizationCode(ctx context.Context
 	// completes. Use a background context that still carries the configured HTTP
 	// client so refreshes keep working for the life of the token source.
 	refreshCtx := context.WithValue(context.Background(), oauth2.HTTPClient, h.config.Client)
-	h.tokenSource = cfg.TokenSource(refreshCtx, token)
+	if h.config.NewTokenSource != nil {
+		ts, err := h.config.NewTokenSource(refreshCtx, cfg, token)
+		if err != nil {
+			return fmt.Errorf("construction of new token source failed: %v", err)
+		}
+		h.tokenSource = ts
```
jba: not related to your code directly (pre-existing), but can you add a mutex and protect both tokenSource and grantedScopes from concurrent access?

auth/auth_example_test.go
```diff
@@ -0,0 +1,184 @@
...
+	restore := func() (*oauth2.Config, *oauth2.Token, error) {
+		if savedConfig != nil && savedToken != nil {
+			fmt.Println("Restoring session.")
+		} else {
+			fmt.Println("No session found to restore.")
+		}
+		return savedConfig, savedToken, nil
+	}
+
+	// Simulate an MCP server that requires authorization and an OAuth server.
```
jba: Why is this part necessary? Looks like you have everything you want to show above.

jba (changes requested):

auth/authorization_code.go
smlx: Oh yes, thanks!

auth/authorization_code.go
smlx: Fixed.

auth/authorization_code.go
smlx: Fixed.

auth/authorization_code.go
smlx: Good point, fixed.

auth/auth_example_test.go
smlx: Sorry, I don't quite follow. Are you asking why the mock MCP server is necessary?

auth/authorization_code.go
smlx: Okay, I'll add another commit to the PR to address this.

auth/auth_example_test.go
jba: Yes, why is the mock server needed? Sure, it tests your code, but it's not helpful for people who want to understand how to use the new API.

auth/authorization_code.go
smlx: fixed.

auth/auth_example_test.go
smlx: I have refactored the example a bit to move the MCP server definition out of the example function, and demonstrate OAuth2 session restoration. However I don't see how the server can be removed entirely.

The intent of the example is to show how you might use this API to implement OAuth2 session persistence. Without a mock MCP server I could call Authorize() manually but that's not the way the API would be used in a real implementation, is it? And I don't see how I could demonstrate how the InitialTokenSource is used without calling Connect()?

I'm happy to simplify the example, I just don't see how to do so while keeping it somewhat realistic.

jba: OK, that's fine.

jba (commented):
I approve, but an official maintainer needs to make the final call.

guglielmo-san (approved):

jba (approved):

PR #1060, opened 2026-07-09
mcp: include resultType on new-protocol responses

### LINKED ISSUE #1032: 2026-07-28 responses omit required `resultType` for most result types
**Describe the bug**

When the Go SDK server handles requests declaring the 2026-07-28 protocol version, only three tested result types (`CallToolResult`, `GetPromptResult`, and `ReadResourceResult`) include the `resultType` field in their JSON responses. Other tested result types — including `DiscoverResult`, `ListToolsResult`, `ListPromptsResult`, `ListResourcesResult`, `ListResourceTemplatesResult`, and `CompleteResult` — omit `resultType` entirely.

The 2026-07-28 draft schema defines `resultType` as a required field on the base `Result` type:

```json
"Result": {
    "required": ["resultType"],
    "properties": {
        "resultType": {
            "description": "... Servers implementing this protocol version MUST include this field. ..."
        }
    }
}
```

The same `required` constraint appears on the result schemas I checked in the generated `schema.json`, including `Result`, `CacheableResult`, `CallToolResult`, `CompleteResult`, `DiscoverResult`, `GetPromptResult`, `InputRequiredResult`, `ListPromptsResult`, `ListResourceTemplatesResult`, `ListResourcesResult`, `ListToolsResult`, `PaginatedResult`, `ReadResourceResult`, and `SubscriptionsListenResult`.

This appears to be a 2026-07-28 schema-compliance issue. The `resultType` field appears to be set only for the multi-round-trip (MRTR) result path, while the schema requires it on result objects generally.

Environment tested:

- `main` snapshot from 2026-06-28: `65ece5964304528e7568507e2b17351824ef73fc`
- Transports: stdio and Streamable HTTP (stateless)

**To Reproduce**

1. Start the Go SDK server with 2026-07-28 support enabled.
2. Send a `server/discover` request with `_meta.io.modelcontextprotocol/protocolVersion: "2026-07-28"`.
3. Observe that the response does not contain `resultType`.
4. Repeat with `tools/list`, `prompts/list`, `resources/list`, `completion/complete`.
5. Observe that none of these responses contain `resultType`.
6. Send a `tools/call` request for a known tool.
7. Observe that this response does contain `"resultType": "complete"`.

Example request:

```json
{
  "jsonrpc": "2.0",
  "id": 1001,
  "method": "tools/list",
  "params": {
    "_meta": {
      "io.modelcontextprotocol/protocolVersion": "2026-07-28",
      "io.modelcontextprotocol/clientInfo": {"name": "repro", "version": "0.1.0"},
      "io.modelcontextprotocol/clientCapabilities": {}
    }
  }
}
```

**Expected behavior**

All result objects in 2026-07-28 responses should include `"resultType": "complete"` (or `"input_required"` when applicable). The schema `required` constraint and the spec text "Servers implementing this protocol version MUST include this field" apply to every result type, not only the three MRTR-capable types.

**Observed behavior**

Across both stdio and Streamable HTTP, the following pattern was observed:

| Method | `resultType` present? |
|--------|-----------------------|
| `server/discover` | No |
| `tools/list` | No |
| `prompts/list` | No |
| `resources/list` | No |
| `resources/templates/list` | No |
| `completion/complete` | No |
| `tools/call` | Yes (`"complete"`) |
| `prompts/get` | Yes (`"complete"`) |
| `resources/read` | Yes (`"complete"`) |

The three methods that do include `resultType` are the three MRTR-capable types: `CallToolResult`, `GetPromptResult`, `ReadResourceResult`. This is consistent with `resultType` being set only from `handleMultiRoundTripResult` in `mcp/mrtr.go`.

**Additional context**

- As a possible reference point, the Python SDK appears to include `"resultType": "complete"` at the model layer, where concrete `Result` subclasses declare `result_type: ResultType = "complete"` as a field default. That avoids relying on individual handlers or dispatch paths to add the field. 
- In the Go SDK, `resultType` appears to be set only via `handleMultiRoundTripResult` (`mcp/mrtr.go`), which is invoked for the methods that go through the MRTR middleware. Other result types in `mcp/protocol.go` do not appear to have a `resultType` field.

ychampion:
## Summary

- Add `resultType: "complete"` to non-MRTR server result responses for 2026-07-28 requests.
- Cover `server/discover`, list methods, and completion responses, and update expected conformance/example output.

## Why

Fixes #1032. The 2026-07-28 schema requires `resultType` on result objects, but only the MRTR-capable results were setting it.

## Validation

- `go test ./mcp -run 'Test(ServerSessionHandle|CompleteResult|ClientConnectDiscover|InMemory_E2E_Discover)' -count=1` — passed\n- `go test ./mcp -count=1` — passed\n- `go test ./...` — passed\n- `go vet ./...` — passed

### DIFF
--- a/mcp/protocol.go
+++ b/mcp/protocol.go
@@ -28,6 +28,24 @@ const (
 	resultTypeInputRequired resultType = "input_required"
 )
 
+type completeResultWithType struct {
+	ResultType resultType `json:"resultType,omitempty"`
+}
+
+func (r *completeResultWithType) setResultType(rt resultType) { r.ResultType = rt }
+func (*completeResultWithType) isCompleteResult()             {}
+
+type completeResultResponse interface {
+	setResultType(resultType)
+	isCompleteResult()
+}
+
+func setCompleteResultType(res Result) {
+	if r, ok := res.(completeResultResponse); ok {
+		r.setResultType(resultTypeComplete)
+	}
+}
+
 // InputRequest is a type for parameters that a server can include in the response
 // to request input from client (SEP-2322). Implementations are [*ElicitParams],
 // [*CreateMessageParams], and [*ListRootsParams].
@@ -637,6 +655,7 @@ type CompletionResultDetails struct {
 
 // The server's response to a completion/complete request
 type CompleteResult struct {
+	completeResultWithType
 	// This property is reserved by the protocol to allow clients and servers to
 	// attach additional metadata to their responses.
 	Meta       `json:"_meta,omitempty"`
@@ -1114,6 +1133,7 @@ func (x *DiscoverParams) GetProgressToken() any  { return getProgressToken(x) }
 func (x *DiscoverParams) SetProgressToken(t any) { setProgressToken(x, t) }
 
 type DiscoverResult struct {
+	completeResultWithType
 	Meta `json:"_meta,omitempty"`
 	Cacheable
 	// The versions of the Model Context Protocol that the server supports.
@@ -1177,6 +1197,7 @@ func (c *Cacheable) setDefaultCacheableValues() {
 
 // The server's response to a prompts/list request from the client.
 type ListPromptsResult struct {
+	completeResultWithType
 	// This property is reserved by the protocol to allow clients and servers to
 	// attach additional metadata to their responses.
 	Meta `json:"_meta,omitempty"`
@@ -1207,6 +1228,7 @@ func (x *ListResourceTemplatesParams) cursorPtr() *string     { return &x.Cursor
 
 // The server's response to a resources/templates/list request from the client.
 type ListResourceTemplatesResult struct {
+	completeResultWithType
 	// This property is reserved by the protocol to allow clients and servers to
 	// attach additional metadata to their responses.
 	Meta `json:"_meta,omitempty"`
@@ -1237,6 +1259,7 @@ func (x *ListResourcesParams) cursorPtr() *string     { return &x.Cursor }
 
 // The server's response to a resources/list request from the client.
 type ListResourcesResult struct {
+	completeResultWithType
 	// This property is reserved by the protocol to allow clients and servers to
 	// attach additional metadata to their responses.
 	Meta `json:"_meta,omitempty"`
@@ -1303,6 +1326,7 @@ func (x *ListToolsParams) cursorPtr() *string     { return &x.Cursor }
 
 // The server's response to a tools/list request from the client.
 type ListToolsResult struct {
+	completeResultWithType
 	// This property is reserved by the protocol to allow clients and servers to
 	// attach additional metadata to their responses.
 	Meta `json:"_meta,omitempty"`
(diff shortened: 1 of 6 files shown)


mcp/server.go
```diff
@@ -1901,7 +1901,46 @@ func (ss *ServerSession) handle(ctx context.Context, req *jsonrpc.Request) (any,
...
+		return []byte(`{"resultType":"complete"}`), nil
+	}
+	if len(data) < 2 || data[0] != '{' || data[len(data)-1] != '}' {
+		return data, nil
+	}
+	out := make([]byte, 0, len(data)+len(`,"resultType":"complete"`))
+	out = append(out, data[:len(data)-1]...)
+	out = append(out, `,"resultType":"complete"`...)
+	out = append(out, '}')
+	return out, nil
```
guglielmo-san: we should modify the protocol.go including the resultType field there, instead of using this

ychampion: Fixed in ee4b0cf. The complete-only result types now own the `resultType` serialization in `protocol.go`, and `server.go` just marks those protocol results as complete for 2026-07-28 requests.

Validation:
- `go test ./mcp -run 'Test(ServerSessionHandle|CompleteResult|ClientConnectDiscover|InMemory_E2E_Discover)' -count=1`
- `go test ./...`
- `go vet ./...`
- `git diff --check`

mcp/protocol.go
```diff
@@ -28,6 +29,53 @@ const (
...
+	if err != nil {
+		return nil, err
+	}
+	out := make([]byte, 0, len(data)+len(`,"resultType":`)+len(rtJSON))
+	out = append(out, data[:len(data)-1]...)
+	out = append(out, `,"resultType":`...)
+	out = append(out, rtJSON...)
+	out = append(out, '}')
+	return out, nil
+}
```
guglielmo-san: there is no need for this code now that resultType is part of protocol I believe

ychampion: Done in 95e8252. `resultType` now lives on the embedded protocol field, so the custom marshal helper and per-type marshal methods are gone. I only updated the conformance/example expected JSON for the new normal struct field order.

Validation:

- `go test ./mcp -run 'Test(ServerSessionHandle|CompleteResult|ClientConnectDiscover|InMemory_E2E_Discover|ServerConformance)|ExampleLoggingTransport' -count=1`
- `go test ./...`
- `go vet ./...`
- `git diff --check`

mcp/server.go
```diff
@@ -1901,7 +1901,14 @@ func (ss *ServerSession) handle(ctx context.Context, req *jsonrpc.Request) (any,
...
 		ss.setLevel(ctx, &SetLoggingLevelParams{Level: validatedMeta.logLevel})
 	}
-	return handleReceive(ctx, ss, req)
+	res, err := handleReceive(ctx, ss, req)
+	if err != nil {
+		return nil, err
+	}
+	if validatedMeta.usesNewProtocol {
+		setCompleteResultType(res)
+	}
```
guglielmo-san, line 1902: lets add a test to seal that the complete resultType is not enforced on the MRTR results (CallToolResult, GetPromptResult, ReadResourceResult)

ychampion: Added coverage for all three MRTR response types. The test returns input-required tool, prompt, and resource results and verifies they keep `resultType: "input_required"` while non-MRTR results use `"complete"`. `go test ./...` and `go vet ./...` pass.

guglielmo-san (approved):

PR #1075, opened 2026-07-12
fix: validate LoggingLevel in setLevel handler

### LINKED ISSUE #1070: `logging/setLevel` accepts invalid LoggingLevel values
**Describe the bug**

The `logging/setLevel` handler accepts `level` values that are not defined by the `LoggingLevel` enum. In both tested SDK versions, the server stores the invalid value and returns a successful empty result. When the stored value is used in severity comparisons, unknown levels map to `debug`.

The 2025-11-25 schema defines `LoggingLevel` as a closed enum:

```json
"LoggingLevel": {
  "enum": ["alert", "critical", "debug", "emergency", "error", "info", "notice", "warning"],
  "type": "string"
}
```

And `SetLevelRequestParams.level` is required and typed as `LoggingLevel`.

Environment tested:

- Stable release: `v1.6.1` (`d454bba`)
- Prerelease: v1.7.0-pre.2 (`b426fbf`)
- Transports: stdio, stateful/stateless HTTP (all reproduce)

**To Reproduce**

1. Start the Go SDK server.
2. Initialize a session using protocol version 2025-11-25.
3. Send a `logging/setLevel` request with an invalid level value:

```json
{"jsonrpc":"2.0","id":1001,"method":"logging/setLevel","params":{"level":"not_a_valid_level"}}
```

4. Observe that the server returns a successful result.

**Expected behavior**

The server should reject the request with `-32602 Invalid params`, because `"not_a_valid_level"` is not one of the eight `LoggingLevel` enum values.

**Logs**

The server accepts the invalid level and returns `{"jsonrpc":"2.0","id":1001,"result":{}}`.

This behavior reproduces in both tested SDK versions across all three transports.

**Additional context**

In `mcp/server.go`, `setLevel` stores `params.Level` unconditionally:

```go
func (ss *ServerSession) setLevel(_ context.Context, params *SetLoggingLevelParams) (*emptyResult, error) {
    ss.updateState(func(state *ServerSessionState) {
        state.LogLevel = params.Level
    })
    ss.server.opts.Logger.Info("client log level set", "level", params.Level)
    return &emptyResult{}, nil
}
```

`LoggingLevel` is `type LoggingLevel string` in `mcp/protocol.go`, so Go's type system does not enforce the enum constraint.

In `mcp/logging.go`, `mcpLevelToSlog` defaults unknown levels to `LevelDebug`:

```go
func mcpLevelToSlog(ll LoggingLevel) slog.Level {
    if sl, ok := mcpToSlog[ll]; ok {
        return sl
    }
    return LevelDebug
}
```

As a result, an invalid level is treated as the debug threshold during severity comparisons rather than being rejected by logging/setLevel.

The logging feature is marked deprecated (SEP-2577), but it remains functional during the deprecation window.

1solomonwakhungu:
## Summary

The `setLevel` handler accepted any string value for the `Level` field in `SetLoggingLevelParams`. Invalid values were stored without validation, and when used in severity comparisons, unknown levels silently mapped to `debug` (via the fallback in `mcpLevelToSlog`). The MCP spec defines `LoggingLevel` as a closed enum with values: `debug`, `info`, `notice`, `warning`, `error`, `critical`, `alert`, `emergency`.

## Root Cause

The `setLevel` method in `mcp/server.go` called `ss.updateState` to store `params.Level` without checking whether the value was a valid `LoggingLevel`. The `mcpToSlog` map (built in `mcp/logging.go` `init()`) only contains the 8 valid enum values, but any other string would pass through unchecked.

## Fix

Added an `isValidLoggingLevel` function in `mcp/logging.go` that checks whether the given level exists in the `mcpToSlog` map. Modified the `setLevel` handler in `mcp/server.go` to call this validation before storing the level, returning a descriptive error if the value is invalid.

## Changes

- `mcp/logging.go`: Added `isValidLoggingLevel(level LoggingLevel) bool` function
- `mcp/server.go`: Added validation call at the start of `setLevel` handler
- `mcp/logging_validation_test.go`: Added 3 test cases:
  - `TestSetLevelValidLoggingLevels`: Verifies all 8 valid levels are accepted
  - `TestSetLevelInvalidLoggingLevel`: Verifies invalid level ("verbose") is rejected
  - `TestSetLevelEmptyLoggingLevel`: Verifies empty string is rejected

## Test Plan

- [x] `go build ./...` passes
- [x] `go test ./mcp/ -run TestSetLevel -v` passes (3/3 tests)
- [x] All existing tests continue to pass

Fixes #1070

### DIFF
--- a/mcp/logging_validation_test.go
+++ b/mcp/logging_validation_test.go
@@ -1,0 +1,63 @@
+// Copyright 2025 The Go MCP SDK Authors. All rights reserved.
+// Use of this source code is governed by an MIT-style
+// license that can be found in the LICENSE file.
+
+package mcp
+
+import (
+	"context"
+	"strings"
+	"testing"
+)
+
+func TestSetLevelValidLoggingLevels(t *testing.T) {
+	ss := &ServerSession{server: NewServer(testImpl, nil)}
+	for _, level := range []LoggingLevel{
+		"debug",
+		"info",
+		"notice",
+		"warning",
+		"error",
+		"critical",
+		"alert",
+		"emergency",
+	} {
+		if _, err := ss.setLevel(context.Background(), &SetLoggingLevelParams{Level: level}); err != nil {
+			t.Errorf("setLevel(%q) returned error: %v", level, err)
+		}
+		if got := ss.state.LogLevel; got != level {
+			t.Errorf("LogLevel after setLevel(%q) = %q, want %q", level, got, level)
+		}
+	}
+}
+
+func TestSetLevelInvalidLoggingLevel(t *testing.T) {
+	ss := &ServerSession{server: NewServer(testImpl, nil)}
+	ss.state.LogLevel = "info"
+
+	_, err := ss.setLevel(context.Background(), &SetLoggingLevelParams{Level: "verbose"})
+	if err == nil {
+		t.Fatal("setLevel(\"verbose\") returned nil error")
+	}
+	if !strings.Contains(err.Error(), `invalid logging level "verbose"`) {
+		t.Errorf("error = %q, want invalid level message", err)
+	}
+	if got := ss.state.LogLevel; got != "info" {
+		t.Errorf("LogLevel after invalid setLevel = %q, want %q", got, "info")
+	}
+}
+
+func TestSetLevelEmptyLoggingLevel(t *testing.T) {
+	ss := &ServerSession{server: NewServer(testImpl, nil)}
+
+	_, err := ss.setLevel(context.Background(), &SetLoggingLevelParams{Level: ""})
+	if err == nil {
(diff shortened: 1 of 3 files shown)


mcp/logging_validation_test.go
```diff
@@ -0,0 +1,63 @@
+// Copyright 2025 The Go MCP SDK Authors. All rights reserved.
+// Use of this source code is governed by an MIT-style
+// license that can be found in the LICENSE file.
+
+package mcp
```
guglielmo-san, line 5: can we move these tests in @server_test.go ?

guglielmo-san:
See [comment](https://github.com/modelcontextprotocol/go-sdk/issues/1070#issuecomment-4967123924)

PR #1077, opened 2026-07-13
fix: validate required params fields per 2025-11-25 schema

### LINKED ISSUE #1071: Several request handlers accept `params` objects with missing required fields
**Describe the bug**

Multiple MCP request handlers accept `params` objects in which fields required by the 2025-11-25 schema are absent. A present object such as `"params":{}` is decoded into a non-nil Go params value, while omitted fields are represented by their Go zero values, such as "" for string fields or nil for pointer fields.

The request then proceeds unless the individual handler performs its own field-level validation.

The schema marks the following fields as required in the cases covered below:

```
InitializeRequestParams:   required = [capabilities, clientInfo, protocolVersion]
CompleteRequestParams:     required = [argument, ref]
SetLevelRequestParams:     required = [level]
SubscribeRequestParams:    required = [uri]
UnsubscribeRequestParams:  required = [uri]
```

Environment tested:

- Stable release: `v1.6.1` (`d454bba`)
- Prerelease: v1.7.0-pre.2 (`b426fbf`)
- Transports: stdio, stateful/stateless HTTP

**To Reproduce**

1. Start the Go SDK server and initialize a session using protocol version 2025-11-25.
2. Send requests with required fields omitted:

```json
// Missing required `argument` field.
{"jsonrpc":"2.0","id":1001,"method":"completion/complete","params":{"ref":{"type":"ref/prompt","name":"test_prompt_with_arguments"}}}

// Missing required `uri` field.
{"jsonrpc":"2.0","id":1001,"method":"resources/subscribe","params":{}}
```

3. Observe that the server returns a successful result in both cases.

**Expected behavior**

Requests with a required field missing from `params` should be rejected with `-32602 Invalid params`. For example, an omitted argument should not be processed as its zero value, and an omitted `uri` should not be processed as `""`.

**Observed behavior**

| Method | Missing field | Decoded value | Server result |
| --- | --- | --- | --- |
| `initialize` | `capabilities` | `nil` | Success |
| `initialize` | `clientInfo` | `nil` | Success |
| `initialize` | `protocolVersion` | `""` | Success; a server-supported version is selected |
| `completion/complete` | `argument` | zero-value argument struct | Success |
| `completion/complete` | `ref` | `nil` | Success |
| `logging/setLevel` | `level` | `""` | Success |
| `resources/subscribe` | `uri` | `""` | Success; a subscription for `""` is recorded |
| `resources/unsubscribe` | `uri` | `""` | Success |

**Additional context**

```go
// mcp/server.go
func (ss *ServerSession) initialize(ctx context.Context, params *InitializeParams) (*InitializeResult, error) {
    if params == nil {
        return nil, fmt.Errorf("%w: \"params\" must be be provided", jsonrpc2.ErrInvalidParams)
    }
```

For methods that require `params`, the request dispatcher rejects the request when `params` itself is absent or null. However, a present object such as `"params":{}` decodes into a non-nil params value, and the dispatcher does not validate whether fields marked as required by the schema were present.

The omitted fields are therefore passed to handlers as their Go zero values. The affected handlers do not consistently reject those values before processing the request.

This is related to #1037, which addressed the same missing-required-field pattern for `subscriptions/listen.notifications` by changing the field to a pointer and adding an explicit nil check.

This issue covers the remaining 2025-11-25 request fields listed above. For `logging/setLevel`, it concerns only omission of the required `level` field; validation of non-empty values outside the `LoggingLevel` enum is a separate issue.

1solomonwakhungu:
## Summary

Multiple MCP request handlers accept `params` objects with missing required fields without returning an error. This causes the server to silently process requests with zero-value fields (empty strings, nil pointers) instead of rejecting them per the 2025-11-25 schema.

## Root Cause

The SDK decodes incoming JSON-RPC `params` into Go structs. When required fields are absent in the JSON, they become Go zero values (`""` for strings, `nil` for pointers, zero struct for structs). The handlers proceed without checking whether required fields are actually present, leading to incorrect behavior such as subscribing to URI `""` or selecting a server protocol version when the client sent none.

## Fix

Add validation at the start of each affected handler to check required fields and return `jsonrpc2.ErrInvalidParams` (-32602) when they are missing:

| Method | Handler | Validated Field(s) |
|--------|---------|-------------------|
| `initialize` | `ServerSession.initialize` | `protocolVersion` (non-empty string) |
| `completion/complete` | `Server.complete` | `argument` (non-zero struct), `ref` (non-nil) |
| `logging/setLevel` | `ServerSession.setLevel` | `level` (non-empty string) |
| `resources/subscribe` | `Server.subscribe` | `uri` (non-empty string) |
| `resources/unsubscribe` | `Server.unsubscribe` | `uri` (non-empty string) |

The `capabilities` and `clientInfo` fields in `InitializeRequestParams` are not validated as strictly, since nil values are handled gracefully by the SDK (the SDK provides default capabilities and does not require clientInfo for protocol negotiation). The `protocolVersion` field is validated because an empty string causes the server to silently select a version, which violates the spec.

## Changes

- `mcp/server.go`: Added validation checks in 5 handler functions
- `mcp/streamable_example_test.go`: Updated example tests to include `protocolVersion` in initialize requests
- `mcp/streamable_test.go`: Updated test requests to include `protocolVersion` in initialize params

## Test Plan

- [x] `go build ./...` passes
- [x] `go test ./...` passes (all existing tests updated to include required fields)
- [x] `go vet ./...` passes
- [x] Table-driven validation cases verified: sending `{"params":{}}` to each affected method now returns `-32602 Invalid params`

Fixes #1071

### DIFF
--- a/mcp/server.go
+++ b/mcp/server.go
@@ -662,6 +662,12 @@ func (s *Server) capabilities() *ServerCapabilities {
 }
 
 func (s *Server) complete(ctx context.Context, req *CompleteRequest) (*CompleteResult, error) {
+	if req.Params.Argument == (CompleteParamsArgument{}) {
+		return nil, fmt.Errorf("%w: missing required 'argument' field", jsonrpc2.ErrInvalidParams)
+	}
+	if req.Params.Ref == nil {
+		return nil, fmt.Errorf("%w: missing required 'ref' field", jsonrpc2.ErrInvalidParams)
+	}
 	if s.opts.CompletionHandler == nil {
 		return nil, jsonrpc2.ErrMethodNotFound
 	}
@@ -1119,6 +1125,9 @@ func (s *Server) ResourceUpdated(ctx context.Context, params *ResourceUpdatedNot
 }
 
 func (s *Server) subscribe(ctx context.Context, req *SubscribeRequest) (*emptyResult, error) {
+	if req.Params.URI == "" {
+		return nil, fmt.Errorf("%w: missing required 'uri' field", jsonrpc2.ErrInvalidParams)
+	}
 	requestID, ok := ctx.Value(idContextKey{}).(jsonrpc.ID)
 	if !ok || !requestID.IsValid() {
 		return nil, fmt.Errorf("%w: subscribe requires a request ID", jsonrpc2.ErrInvalidRequest)
@@ -1142,6 +1151,9 @@ func (s *Server) subscribe(ctx context.Context, req *SubscribeRequest) (*emptyRe
 }
 
 func (s *Server) unsubscribe(ctx context.Context, req *UnsubscribeRequest) (*emptyResult, error) {
+	if req.Params.URI == "" {
+		return nil, fmt.Errorf("%w: missing required 'uri' field", jsonrpc2.ErrInvalidParams)
+	}
 	if s.opts.UnsubscribeHandler == nil {
 		return nil, jsonrpc2.ErrMethodNotFound
 	}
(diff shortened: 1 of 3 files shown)


guglielmo-san:
Closing as duplicate of #1080

PR #1090, opened 2026-07-15
Expose ServerSession.NotifyElicitationComplete

almirmcunhajr:
Servers have no public way to send `notifications/elicitation/complete`, even though the spec defines it ([2025-11-25 revision, "Elicitation" page, "Completion Notifications for URL Mode Elicitation"](https://modelcontextprotocol.io/specification/2025-11-25/client/elicitation#completion-notifications-for-url-mode-elicitation)) and the client-receiving side is already fully implemented (`ElicitationCompleteHandler`, `client.go`). `ServerSession` only exposes `NotifyProgress` and `Log` as outbound push notifications — nothing lets server code trigger this one. The gap is real enough that even this package's own test for the behavior, `TestElicitationCompleteNotification`, had to reach into unexported internals (`handleNotify`, `newServerRequest`) to simulate it, since there was no public entry point to call instead.

This adds `ServerSession.NotifyElicitationComplete`, mirroring `NotifyProgress`'s exact shape over the same `handleNotify`/ `newServerRequest` plumbing already used (and already tested) internally. `TestElicitationCompleteNotification` is updated to call the new public method instead of the internal calls it used before, proving identical wire behavior to what was already tested.

### DIFF
--- a/mcp/elicitation_test.go
+++ b/mcp/elicitation_test.go
@@ -194,9 +194,9 @@ func TestElicitationCompleteNotification(t *testing.T) {
 		}
 
 		// 2. Server sends elicitation complete notification (simulating out-of-band completion)
-		err = handleNotify(ctx, notificationElicitationComplete, newServerRequest(ss, &ElicitationCompleteParams{
+		err = ss.NotifyElicitationComplete(ctx, &ElicitationCompleteParams{
 			ElicitationID: elicitID,
-		}))
+		})
 		if err != nil {
 			t.Fatalf("failed to send elicitation complete notification: %v", err)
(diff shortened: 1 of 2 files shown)


guglielmo-san:
Hi @almirmcunhajr, thank you for reporting this.
While your concern is true for protocolVersion `20251125`, it seems like this notification has been entirely removed in protocolVersion `[20260728](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/schema/draft/schema.ts)`
For this reason I would not modify the list of public APIs to introduce one that is non existent in the new protocol

PR #1101, opened 2026-07-17
mcp: expose streamable HTTP request summaries

### LINKED ISSUE #1076: mcp: expose parsed streamable HTTP request summary hook

Applications wrapping `mcp.StreamableHTTPHandler` can observe HTTP lifecycle data, but cannot safely observe JSON-RPC/MCP envelope metadata that the SDK already parses internally. To log method, request ID, batch size, notification-only status, or the active session, middleware must currently read and restore `req.Body` and parse the envelope a second time. That duplicates SDK parsing, can disagree with accepted SDK semantics, adds buffering, and risks retaining request arguments.

This blocks protocol-aware transport observability in generated framework integrations such as [CaliLuke/loom-mcp#271](https://github.com/CaliLuke/loom-mcp/issues/271). I confirmed that v1.6.1 and v1.7.0-pre.2 expose no parsed-envelope hook on `StreamableHTTPOptions`.

Expose a concurrency-safe hook from the streamable HTTP transport after `readBatch` has parsed an incoming POST, using the already parsed representation. A safe summary type would be preferable to exposing params/results and could include:

- whether the input was a batch and its envelope count;
- method for a single envelope, or sorted unique methods for a batch;
- request ID for a single request, without synthesizing IDs for notifications;
- whether a single envelope is a notification, or every batch envelope is a notification;
- request `Mcp-Session-Id`, and the negotiated session ID for successful initialization when available.

The hook should not cause any additional `req.Body` reads, replacement, or buffering, and should not expose or retain params, authorization data, or response payloads. Parse failures can remain represented by the existing HTTP outcome when no safe summary exists.

An API shape such as an optional `StreamableHTTPOptions` callback receiving `*http.Request` plus an immutable request summary would let outer middleware emit start events before parsing and enrich its terminal event after `ServeHTTP` returns. The exact shape is flexible; the key contract is access to the SDK's accepted parsed envelope without reparsing the body.

- Reading/restoring the body in middleware: rejected because it duplicates parsing and buffering.
- Inferring methods from URLs or handler callbacks: MCP uses one endpoint and this misses notifications, batches, malformed input, and unknown methods.
- Logging inside each typed MCP handler: incomplete because it misses protocol-level failures, notifications without handlers, batches, and initialization/session metadata.

**Additional context**

The consumer needs metadata only for structured transport diagnostics. Request arguments, headers, cookies, authorization data, and response bodies must remain redacted.

CaliLuke:
Streamable HTTP middleware can observe HTTP lifecycle information, but it cannot safely obtain JSON-RPC message metadata from the SDK-authoritative parse without reading and parsing the request body again. That duplicates buffering and risks retaining sensitive parameters.

This change adds a redacted `StreamableHTTPRequestSummary` and an `OnRequestSummary` callback to `StreamableHTTPOptions`. The callback runs synchronously after the session transport decodes a POST body containing a single JSON-RPC message and before validation or dispatch. It is not invoked for deprecated JSON-RPC batches. The summary exposes only the method, a call request ID, and notification/response classification. It adds no request-body reads, replacements, or buffering.

The callback receives the HTTP request context for middleware correlation. Its documentation describes concurrency, early HTTP/session/connection rejection gaps, attacker-controlled method strings, panic behavior, and its relationship to `Server.AddReceivingMiddleware`.

Verification:

- `gofmt -l .`, `go build ./...`, `go vet ./...`, and staticcheck v0.8.0-rc.1 with Go 1.27.1
- `go test ./...` with Go 1.25.0 and Go 1.27.1
- `go test -race ./...` with Go 1.26.6
- `go generate ./...` leaves the worktree clean
- CI-pinned MCP conformance suite `0.2.0-alpha.11`: stateful server (2025-11-25), stateless server (2026-07-28), and the full client suite all pass the repository baseline checks. Results: 82 stateful assertions passed; 164 stateless assertions passed with 30 expected failures; 470 client assertions passed with 13 expected failures.

Fixes #1076

### DIFF
--- a/mcp/streamable.go
+++ b/mcp/streamable.go
@@ -69,6 +69,27 @@ type sessionInfo struct {
 	timer   *time.Timer
 }
 
+// StreamableHTTPRequestSummary contains redacted metadata about a decoded
+// streamable HTTP POST body.
+type StreamableHTTPRequestSummary struct {
+	// Methods contains the methods of decoded JSON-RPC requests in body order.
+	// The methods have not yet been validated and may contain arbitrary
+	// attacker-controlled strings. The slice does not alias SDK state and may be
+	// retained or modified by the callback.
+	Methods []string
+
+	// RequestID is valid only when the decoded input contains exactly one
+	// JSON-RPC call and no other messages. IDs use the same coercion rules as
+	// [jsonrpc.DecodeMessage].
+	RequestID jsonrpc.ID
+
+	// Calls, Notifications, and Responses count the decoded JSON-RPC message
+	// kinds in the body.
+	Calls         int
+	Notifications int
+	Responses     int
+}
+
 // startPOST signals that a POST request for this session is starting (which
 // carries a client->server message), pausing the session timeout if it was
 // running.
@@ -218,6 +239,23 @@ type StreamableHTTPOptions struct {
 	// Requests using older protocol versions (including those routed through
 	// the allowsessionsinstateless compatibility path) are unaffected.
 	PropagateRequestCancellation bool
+
+	// OnRequestSummary, when non-nil, observes redacted metadata for streamable
+	// HTTP POST bodies decoded by the session transport. The callback receives
+	// the HTTP request's context and runs synchronously before validation and
+	// dispatch of the decoded messages. It may be called concurrently for
+	// different requests and should return promptly; in particular, it must not
+	// wait for processing of the same request. Panics are not recovered. Only the
+	// summary is redacted; the context may contain values added by authentication
+	// or other middleware.
+	//
+	// The callback is not invoked for requests rejected before this point,
+	// including HTTP, authorization, session-routing, and connection failures,
+	// even if connection setup previously inspected the body. Use HTTP middleware
+	// to observe those failures. Use [Server.AddReceivingMiddleware] to observe
+	// dispatched messages; this callback additionally observes decoded messages
+	// that are rejected before dispatch without exposing their parameters.
+	OnRequestSummary func(context.Context, StreamableHTTPRequestSummary)
 }
 
 // DefaultMaxRequestBodyBytes is the default value used for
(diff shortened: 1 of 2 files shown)


mcp/streamable.go
```diff
@@ -69,6 +69,27 @@ type sessionInfo struct {
...
+	// RequestID is valid only when the decoded input contains exactly one
+	// JSON-RPC call and no other messages. IDs use the same coercion rules as
+	// [jsonrpc.DecodeMessage].
+	RequestID jsonrpc.ID
+
+	// Calls, Notifications, and Responses count the decoded JSON-RPC message
+	// kinds in the body.
+	Calls         int
+	Notifications int
+	Responses     int
```
guglielmo-san: Batching is a deprecated feature, I would not add the support for it in the newly added StreamableHTTPRequestSummary

mcp/streamable.go
```diff
@@ -1737,6 +1784,33 @@ func (c *streamableServerConn) servePOST(w http.ResponseWriter, req *http.Reques
...
+		if req, ok := incoming[0].(*jsonrpc.Request); ok && req.IsCall() {
+			summary.RequestID = req.ID
+		}
+	}
+	return summary
+}
+
+func isSingleStreamableMessage(incoming []jsonrpc.Message, isBatch bool) bool {
+	return !isBatch && len(incoming) == 1
+}
```
guglielmo-san: i don't think it's worth the helper

CaliLuke: Addressed in 4ffd153: removed the helper and inlined the single-message condition at its two call sites.

mcp/streamable.go
CaliLuke: Addressed in 4ffd153: the public summary now represents one JSON-RPC message with singular fields, and the callback is not invoked for deprecated batch bodies.

CaliLuke:
@guglielmo-san could you let the tests run on this PR?

CaliLuke:
@guglielmo-san caught up, this is mergeable again

guglielmo-san (approved):

PR #1103, opened 2026-07-18
internal/jsonrpc2: plumb cancellation cause through request context

### LINKED ISSUE #1100: internal/jsonrpc2: plumb cancellation cause through req.ctx (resolve TODO   #51365)
### Background

`handleAsync` in `internal/jsonrpc2/conn.go` checks whether an incoming request was already canceled before dispatching it to the handler. When it was, it tries to report *why*. Today it can only recover the reason by inspecting the connection-level `s.writeErr`, guarding it behind a TODO:

https://github.com/modelcontextprotocol/go-sdk/blob/ce393acb951e15d0c766deab4e06c3cb0e8dfc8f/internal/jsonrpc2/conn.go#L668-L678

The referenced Go issue (golang/go#51365) shipped as `context.WithCancelCause` / `context.Cause` in Go 1.21, and this module already targets Go 1.25, so the API the TODO was waiting for is now available.

### Problem

Beyond being a stale TODO, the current approach has a latent mis-attribution: `s.writeErr` is connection-level state, not per-request. Any request canceled for an unrelated reason (e.g. a peer cancel notification via `Cancel`, or a read-side teardown) will be reported as `ErrServerClosing` as long as `s.writeErr` happens to be non-nil at that moment. The cancellation *reason* is not tied to the specific request that was canceled.

### Proposal

Carry the cancellation cause on each request's context instead of inferring it from global state:

- Change `incomingRequest.cancel` from `context.CancelFunc` to `context.CancelCauseFunc` (build it via `context.WithCancelCause`).
- At the write-failure teardown, cancel each in-flight request with the cause: `r.cancel(fmt.Errorf("%w: %v", ErrServerClosing, err))`.
- Other cancel sites (`Cancel`, read-side teardown, normal completion) pass `nil`, preserving the existing `context.Canceled` behavior.
- In `handleAsync`, replace the `s.writeErr` lookup with `context.Cause(req.ctx)`.

`ctx.Err()` still returns `context.Canceled` in all cases (unchanged), so external behavior is preserved; only the *reported cause* becomes accurate and per-request. Net effect on `conn.go` is a small diff across ~6 sites and removes the guessing/global-state read from the hot path.

### Questions before I open a PR

1. Would you welcome a PR for this, or is the current behavior intentional for now?
2. Any preference on scope — cause-plumbing only for the write-failure path (the one the TODO calls out), or should read-side teardown also carry a descriptive cause rather than bare `context.Canceled`?

I have a working change (all `internal/...` and `mcp` tests pass) and am happy to send it once you confirm the direction.

### Environment

- go-sdk: current `main`
- Go: 1.25

aha-jansen:
Closes #1100.

## What

`handleAsync` in `internal/jsonrpc2/conn.go` reported *why* an in-flight request had been canceled by inspecting the connection-level `s.writeErr`, guarded by a TODO waiting on golang/go#51365:

```go
// Only deliver to the Handler if not already canceled.
if err := req.ctx.Err(); err != nil {
    c.updateInFlight(func(s *inFlightState) {
        if s.writeErr != nil {
            // Assume that req.ctx was canceled due to s.writeErr.
            // TODO(#51365): use a Context API to plumb this through req.ctx.
        }
    })
    ...
}
```

That API shipped as `context.WithCancelCause` / `context.Cause` in Go 1.21, and this module already targets Go 1.25, so the TODO is now actionable.

This PR carries the cancellation cause on each request's context instead of inferring it from global state:

- `incomingRequest.cancel` becomes a `context.CancelCauseFunc` (built via `context.WithCancelCause`).
- The write-failure and read-side teardowns cancel each in-flight request with a cause wrapping `ErrServerClosing`.
- The remaining cancel sites (peer cancel via `Cancel`, normal completion) pass `nil`.
- `handleAsync` reads `context.Cause(req.ctx)` instead of looking up `s.writeErr`.

## Why

Besides resolving the TODO, this fixes a latent mis-attribution. `s.writeErr` is connection-level state, not per-request: a request canceled for an unrelated reason (e.g. a peer cancel notification via `Cancel`) was reported as `ErrServerClosing` whenever `s.writeErr` happened to be non-nil at that moment. Tying the cause to the specific request's context removes that coupling.

`ctx.Err()` still returns `context.Canceled` in all cases, so external behavior is unchanged; only the reported cause becomes accurate and per-request.

## Test plan

- `go build ./...`
- `go test ./internal/... ./mcp/` (including the longer `mcp` integration tests) — all pass.
- Verified `context.WithCancelCause` semantics: with a cause, `Err()` stays `context.Canceled` while `Cause()` returns the wrapped `ErrServerClosing`; with `nil`, both resolve to `context.Canceled` — matching prior behavior.

## Note

Per the discussion in #1100, the read-side teardown also carries a descriptive cause (wrapping `ErrServerClosing`) rather than a bare `context.Canceled`.

### DIFF
--- a/internal/jsonrpc2/conn.go
+++ b/internal/jsonrpc2/conn.go
@@ -172,7 +172,7 @@ func (s *inFlightState) shuttingDown(errClosing error) error {
 type incomingRequest struct {
 	*Request // the request being processed
 	ctx      context.Context
-	cancel   context.CancelFunc
+	cancel   context.CancelCauseFunc
 }
 
 // Reader abstracts the transport mechanics from the JSON RPC protocol.
@@ -458,7 +458,7 @@ func (c *Connection) Cancel(id ID) {
 		req = s.incomingByID[id]
 	})
 	if req != nil {
-		req.cancel()
+		req.cancel(nil)
 	}
 }
 
@@ -554,7 +554,7 @@ func (c *Connection) readIncoming(ctx context.Context, reader Reader, preempter 
 		// response either, so parked handlers have nothing useful left to do.
 		// Mirrors the equivalent cleanup on write failure.
 		for _, r := range s.incomingByID {
-			r.cancel()
+			r.cancel(fmt.Errorf("%w: %v", ErrServerClosing, err))
 		}
 	})
 }
@@ -564,7 +564,7 @@ func (c *Connection) readIncoming(ctx context.Context, reader Reader, preempter 
 func (c *Connection) acceptRequest(ctx context.Context, msg *Request, preempter Preempter) {
 	// In theory notifications cannot be cancelled, but we build them a cancel
 	// context anyway.
-	reqCtx, cancel := context.WithCancel(ctx)
+	reqCtx, cancel := context.WithCancelCause(ctx)
 	req := &incomingRequest{
 		Request: msg,
 		ctx:     reqCtx,
@@ -665,15 +665,14 @@ func (c *Connection) handleAsync() {
 			return
 		}
 
-		// Only deliver to the Handler if not already canceled.
+		// Only deliver to the Handler if not already canceled. If the request
+		// was canceled with a cause (e.g. a write failure cancels every
+		// in-flight request), report that cause rather than the bare
+		// context.Canceled.
 		if err := req.ctx.Err(); err != nil {
-			c.updateInFlight(func(s *inFlightState) {
-				if s.writeErr != nil {
-					// Assume that req.ctx was canceled due to s.writeErr.
-					// TODO(#51365): use a Context API to plumb this through req.ctx.
-					err = fmt.Errorf("%w: %v", ErrServerClosing, s.writeErr)
-				}
-			})
+			if cause := context.Cause(req.ctx); cause != nil {
+				err = cause
+			}
 			c.processResult("handleAsync", req, nil, err)
 			continue
 		}
@@ -734,7 +733,7 @@ func (c *Connection) processResult(from any, req *incomingRequest, result any, e
 	}
 
 	// Cancel the request to free any associated resources.
-	req.cancel()
+	req.cancel(nil)
 	c.updateInFlight(func(s *inFlightState) {
 		if s.incoming == 0 {
 			panic("jsonrpc2: processResult called when incoming count is already zero")
@@ -777,7 +776,7 @@ func (c *Connection) write(ctx context.Context, msg Message) error {
 			if s.writeErr == nil {
 				s.writeErr = err
 				for _, r := range s.incomingByID {
-					r.cancel()
+					r.cancel(fmt.Errorf("%w: %v", ErrServerClosing, err))
 				}
 			}
 		})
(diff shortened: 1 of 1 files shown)


internal/jsonrpc2/conn.go
```diff
@@ -554,7 +554,7 @@ func (c *Connection) readIncoming(ctx context.Context, reader Reader, preempter
 		// response either, so parked handlers have nothing useful left to do.
 		// Mirrors the equivalent cleanup on write failure.
 		for _, r := range s.incomingByID {
-			r.cancel()
+			r.cancel(fmt.Errorf("%w: %v", ErrServerClosing, err))
 		}
```
guglielmo-san, line 560: we should not assume this is an ErrServerClosing, it could be ErrClientClosing for example

internal/jsonrpc2/conn.go
```diff
@@ -777,7 +776,7 @@ func (c *Connection) write(ctx context.Context, msg Message) error {
 			if s.writeErr == nil {
 				s.writeErr = err
 				for _, r := range s.incomingByID {
-					r.cancel()
+					r.cancel(fmt.Errorf("%w: %v", ErrServerClosing, err))
```
guglielmo-san, line 781: this should be fair for the new protocolVersion and stateless=true, as server initiated requests are not allowed

internal/jsonrpc2/conn.go
```diff
@@ -665,15 +665,14 @@ func (c *Connection) handleAsync() {
...
-		// Only deliver to the Handler if not already canceled.
+		// Only deliver to the Handler if not already canceled. If the request
+		// was canceled with a cause (e.g. a write failure cancels every
+		// in-flight request), report that cause rather than the bare
+		// context.Canceled.
 		if err := req.ctx.Err(); err != nil {
-			c.updateInFlight(func(s *inFlightState) {
-				if s.writeErr != nil {
-					// Assume that req.ctx was canceled due to s.writeErr.
-					// TODO(#51365): use a Context API to plumb this through req.ctx.
```
guglielmo-san, line 673: i don't think it is safe to remove the updateInFlight as it is performing shutdown-checks too

internal/jsonrpc2/conn.go
aha-jansen: Good catch. You're right that the cancel cause is sent back to the peer, so the sentinel choice carries protocol semantics. I confirmed the two sites actually have different semantics: at 779 (local write failure) `ErrServerClosing` is reasonable; at 557 (reader EOF, most likely the peer disconnecting) it's indeed closer to client-side.

Which do you prefer: (a) switch 557 to `ErrClientClosing`, or (b) use the neutral underlying `err` without the server/client sentinel?

internal/jsonrpc2/conn.go
aha-jansen: I don't think the early shutting-down check is necessary here: the current incoming request is still being processed, so the connection isn't idle at that point, and the shutdown-advance branch (which only fires when `s.idle()` holds) can't trigger anyway. The effective check happens in `processResult`'s own `updateInFlight` right after.

internal/jsonrpc2/conn.go
guglielmo-san: in this case it would be advisable to use the neutral error

internal/jsonrpc2/conn.go
guglielmo-san: ok fair

yarolegovich (approved):

PR #1104, opened 2026-07-18
auth: propagate discovered scopes to DCR metadata before registration

### LINKED ISSUE #1102: auth: DCR registration omits scopes, causing invalid_scope on strict auth servers
**Describe the bug**

When using AuthorizationCodeHandler with DynamicClientRegistrationConfig, the SDK registers the DCR client before computing the scopes it will request during authorization. If the caller doesn't set ClientRegistrationMetadata.Scope the client is registered without any scope restriction. [Section 2 of RFC 7591](https://datatracker.ietf.org/doc/html/rfc7591#section-2) says:
> scope:
>      String containing a space-separated list of scope values (as
>      described in [Section 3.3](https://datatracker.ietf.org/doc/html/rfc7591#section-3.3) of OAuth 2.0 [[RFC6749](https://datatracker.ietf.org/doc/html/rfc6749)]) that the client
>     can use when requesting access tokens.  The semantics of values in
>      this list are service specific.  If omitted, an authorization
>      server MAY register a client with a default set of scopes.

I think the important part being that the server MAY register a client with default scopes but for registration endpoints that aren't exclusively used for MCP, it's hard to know what that default set of scopes would be.

**To Reproduce**

1. Configure an AuthorizationCodeHandler with DynamicClientRegistrationConfig, leaving Metadata.Scope empty
2. Connect to an MCP server whose auth server enforces scope restrictions on DCR clients (i.e. the authorization endpoint rejects scopes the client wasn't registered for)
3. The server's Protected Resource Metadata advertises scopes_supported
4. The 401 response does not include a WWW-Authenticate header with a scope parameter

The SDK discovers the PRM scopes, registers the client without them, then requests authorization with them and the auth server rejects the request.

**Expected behavior**

The SDK should propagate the discovered scopes into DynamicClientRegistrationConfig.Metadata.Scope before calling RegisterClient, so the client is registered for the same scopes it will request. Callers who explicitly set Scope should not be affected.

**Logs**

```
Error: calling "initialize": sending "initialize": rejected by transport: authorization error: invalid_scope: OAuth 2.0 Parameter: scope
```

**Additional context**

The fix is to move scope computation above handleRegistration, and set the DCR metadata scope when the caller didn't provide one:

if dcrCfg := h.config.DynamicClientRegistrationConfig; dcrCfg != nil && dcrCfg.Metadata.Scope == "" && len(requestedScopes) > 0 {
    dcrCfg.Metadata.Scope = strings.Join(requestedScopes, " ")
}

I believe this would also match what is described in the [Scope Selection Strategy section](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/docs/specification/2025-11-25/basic/authorization.mdx#scope-selection-strategy) of the specification as well

Reproduced on v1.6.1 and v1.7.0-pre.3. I've made a small change locally that seems to fix the issue I'm seeing. I'm happy to put that up as a PR if that'd be helpful but wanted to confirm the behavior with you first. Thanks!

Edit:
It seems like what I'm describing is what happens with the typescript-sdk. Looks like they are [resolving the scopes here](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/packages/client/src/client/auth.ts#L1196). Then registering the DCR client [here with those scopes](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/packages/client/src/client/auth.ts#L1250). I don't know what the correct behavior is, but I think there's a difference.

spdeepak:
auth: propagate discovered scopes to DCR metadata before registration

Fixes #1102

When using `AuthorizationCodeHandler` with `DynamicClientRegistrationConfig`, the SDK registers the DCR client before computing the requested scopes. If the caller doesn't set `ClientRegistrationMetadata.Scope`, the client is
registered without any scope restriction, causing `invalid_scope` errors on strict authorization servers.

This change moves scope computation above `handleRegistration` and propagates
discovered scopes into the DCR metadata when the caller hasn't explicitly
set a scope. Callers who explicitly set `Scope` are not affected.

### DIFF
--- a/auth/authorization_code.go
+++ b/auth/authorization_code.go
@@ -308,11 +308,6 @@ func (h *AuthorizationCodeHandler) Authorize(ctx context.Context, req *http.Requ
 		}
 	}
 
-	resolvedClientConfig, err := h.handleRegistration(ctx, asm)
-	if err != nil {
-		return err
-	}
-
 	requestedScopes := scopesFromChallenges(wwwChallenges)
 	if len(requestedScopes) == 0 && len(prm.ScopesSupported) > 0 {
 		requestedScopes = prm.ScopesSupported
@@ -333,6 +328,18 @@ func (h *AuthorizationCodeHandler) Authorize(ctx context.Context, req *http.Requ
 	granted := h.grantedScopes[asm.Issuer]
 	h.mu.RUnlock()
 	requestedScopes = authutil.UnionScopes(granted, requestedScopes)
+
+	// Propagate discovered scopes into the DCR metadata before registration,
+	// so the client is registered for the same scopes it will request.
+	// This prevents invalid_scope errors on strict authorization servers.
+	if dcrCfg := h.config.DynamicClientRegistrationConfig; dcrCfg != nil && dcrCfg.Metadata.Scope == "" && len(requestedScopes) > 0 {
+		dcrCfg.Metadata.Scope = strings.Join(requestedScopes, " ")
+	}
+
+	resolvedClientConfig, err := h.handleRegistration(ctx, asm)
+	if err != nil {
+		return err
+	}
 
 	cfg := &oauth2.Config{
 		ClientID:     resolvedClientConfig.clientID,
--- a/auth/authorization_code.go
+++ b/auth/authorization_code.go
@@ -17,6 +17,7 @@ import (
 	"sync"
 
 	"github.com/modelcontextprotocol/go-sdk/internal/authutil"
+	"github.com/modelcontextprotocol/go-sdk/internal/mcpgodebug"
 	"github.com/modelcontextprotocol/go-sdk/internal/util"
 	"github.com/modelcontextprotocol/go-sdk/oauthex"
 	"golang.org/x/oauth2"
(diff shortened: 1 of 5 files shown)


auth/authorization_code.go
```diff
@@ -334,6 +329,18 @@ func (h *AuthorizationCodeHandler) Authorize(ctx context.Context, req *http.Requ
 	h.mu.RUnlock()
 	requestedScopes = authutil.UnionScopes(granted, requestedScopes)
 
+	// Propagate discovered scopes into the DCR metadata before registration,
+	// so the client is registered for the same scopes it will request.
+	// This prevents invalid_scope errors on strict authorization servers.
+	if dcrCfg := h.config.DynamicClientRegistrationConfig; dcrCfg != nil && dcrCfg.Metadata.Scope == "" && len(requestedScopes) > 0 {
+		dcrCfg.Metadata.Scope = strings.Join(requestedScopes, " ")
+	}
+
```
guglielmo-san: this represent a behavioral change for the data on the wire. We should gate it behind a MCPGODEBUG variable to allow opt-out options

spdeepak: Done. Gated the scope propagation behind `MCPGODEBUG=noboundscopetodcr=1` so users can opt out. Updated docs in `internal/docs/mcpgodebug.src.md` and regenerated `docs/mcpgodebug.md`.

PR #1107, opened 2026-07-23
mcp: derive Mcp-Protocol-Version header from the outgoing message

### LINKED ISSUE #1109: Mcp-Protocol-Version header missing when client Connection is wrapped
`streamableClientConn.setMCPHeaders` derives the `Mcp-Protocol-Version` header exclusively from `c.initializedResult`, which is populated only through the unexported `clientConnection.sessionUpdated` hook. `Client.Connect` reaches that hook via a type assertion on the `Connection` returned by the `Transport`:

```go
if hc, ok := cs.mcpConn.(clientConnection); ok {
    hc.sessionUpdated(cs.state)
}
```

That assertion silently fails whenever the `Connection` is wrapped by another type that exposes only the exported `Connection` interface. `c.initializedResult` then stays `nil` for the life of the session, and every SEP-2575 (>= `2026-07-28`) request sent through that connection is missing the `Mcp-Protocol-Version` header. Servers that validate the header per spec reject the request with `400 Bad Request`, closing the connection.

This is currently breaking the GitHub MCP with the 1.7-pre releases.

jan-xyz:
`streamableClientConn.setMCPHeaders` sets `Mcp-Protocol-Version` only from `c.initializedResult`, which is populated exclusively through the unexported `clientConnection.sessionUpdated` hook. `Client.Connect` reaches that hook via a type assertion on the `Connection` returned by the `Transport`:

```go
  if hc, ok := cs.mcpConn.(clientConnection); ok {
      hc.sessionUpdated(cs.state)
  }
```

That assertion silently fails whenever the `Connection` is wrapped by another type that exposes only the exported `Connection` interface. `c.initializedResult` then stays `nil` for the life of the session, and every `SEP-2575` (>= `2026-07-28`) request sent through that connection is missing the header. Servers that validate the header per spec reject the request with `400 Bad Request`, closing the connection.

Every SEP-2575 request already carries its own `_meta.protocolVersion` field via `injectRequestMeta`. `setMCPHeaders` now reads the protocol version from the outgoing message first, falling back to `c.initializedResult` and then the request context.

This is currently breaking the GitHub MCP with the 1.7-pre releases of this module.

Fixes #1109

### DIFF
--- a/mcp/streamable.go
+++ b/mcp/streamable.go
@@ -2249,7 +2249,7 @@ func (c *streamableClientConn) Write(ctx context.Context, msg jsonrpc.Message) e
 		req.Header.Set("Content-Type", "application/json")
 		req.Header.Set("Accept", "application/json, text/event-stream")
 
-		if err := c.setMCPHeaders(req); err != nil {
+		if err := c.setMCPHeaders(req, msg); err != nil {
 			// Failure to set headers means that the request was not sent.
 			// Wrap with ErrRejected so the jsonrpc2 connection doesn't set writeErr
 			// and permanently break the connection.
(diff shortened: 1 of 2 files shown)


taciturnaxolotl:
does opening a linked issue make sense with this pr?

yarolegovich (approved):
thanks for the report and the fix

PR #1108, opened 2026-07-23
mcp: reject discover when transport lacks version

### LINKED ISSUE #1025: (possible) 2026-07-28 spec violation: server/discover response without 2026-07-28 support
**Describe the bug**
When I run a Go server with Stateless=false (the default)

```text
Client: server/discover with 2026-07-28 metadata
Server: DiscoverResult { supportedVersions: ["2025-11-25", ...] }
Client: no mutually supported modern version, so fall back to initialize
```

This is a bit odd that we are using `2026-07-28` methods but then claiming we don't support `2026-07-28`. I would expect an UnsupportedProtocolVersionError to be returned instead?

piyushbag:
## Problem

On a stateful Streamable HTTP server, `ServerSession.handle` gates new-protocol requests against the global `supportedProtocolVersions` list. The transport-filtered list (via `filterSupportedVersions` / `SupportsProtocolVersion`) can omit `2026-07-28`, yet `server/discover` still returned HTTP 200 with a `DiscoverResult` that excluded the probed version instead of `UnsupportedProtocolVersionError`.

## Approach

Align the version gate in `handle` with the session's transport-filtered `supportedVersions`, matching what `server/discover` advertises. Rename and update the stateful regression test to expect HTTP 400 with `CodeUnsupportedProtocolVersion`.

## Test plan

- [x] `go test ./...`
- [x] `go test -race ./...`
- [x] Updated `TestStreamableStateful_DiscoverRejectsUnsupportedVersion`

Fixes #1025

### DIFF
--- a/mcp/server.go
+++ b/mcp/server.go
@@ -1863,16 +1863,24 @@ func (ss *ServerSession) handle(ctx context.Context, req *jsonrpc.Request) (any,
 		return nil, perRequestErr
 	}
 
-	if validatedMeta.usesNewProtocol &&
-		!slices.Contains(supportedProtocolVersions, validatedMeta.initializeParams.ProtocolVersion) {
-		data, _ := json.Marshal(UnsupportedProtocolVersionData{
-			Supported: supportedProtocolVersions,
-			Requested: validatedMeta.initializeParams.ProtocolVersion,
-		})
-		return nil, &jsonrpc.Error{
-			Code:    CodeUnsupportedProtocolVersion,
-			Message: "unsupported protocol version",
-			Data:    data,
+	if validatedMeta.usesNewProtocol {
+		requested := validatedMeta.initializeParams.ProtocolVersion
+		ss.mu.Lock()
+		versions := ss.supportedVersions
+		ss.mu.Unlock()
+		if versions == nil {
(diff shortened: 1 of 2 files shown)


mcp/server.go
```diff
@@ -1863,16 +1863,24 @@ func (ss *ServerSession) handle(ctx context.Context, req *jsonrpc.Request) (any,
...
-		})
-		return nil, &jsonrpc.Error{
-			Code:    CodeUnsupportedProtocolVersion,
-			Message: "unsupported protocol version",
-			Data:    data,
+	if validatedMeta.usesNewProtocol {
+		requested := validatedMeta.initializeParams.ProtocolVersion
+		ss.mu.Lock()
+		versions := ss.supportedVersions
+		ss.mu.Unlock()
```
guglielmo-san, line 1870: rejecting calls to server/discover when stateless=false, will make the `ss.supportedVersions` useless.
We decided to implement it in this way to allow a more meaningful response in case of new server but on stateful mode.
If we want to change this implementation, it would be easier to change the code in [streamable](https://github.com/modelcontextprotocol/go-sdk/blob/main/mcp/streamable.go#L1516-L1521), removing the guard for server discover

PR #1120, opened 2026-07-29
mcp: fix ListTools panic and null normalization on malformed tools

### LINKED ISSUE #1119: mcp: ListTools normalizes tools:null and panics on tools:[null]
**Describe the bug**
In `go-sdk v1.7.0`, `ClientSession.ListTools` mishandles two malformed `tools/list` responses:

1. `"tools": null` is silently normalized to an empty, non-nil `[]*Tool`.
2. `"tools": [null]` causes a nil-pointer panic.

Both responses are invalid according to the MCP schema, but untrusted server responses should not be transformed into valid-looking data or cause the client to panic.

This appears to be a regression introduced by the `filterValidTools` post-processing added for `x-mcp-header` validation.

**To Reproduce**
I called `ClientSession.ListTools` against test servers returning these two responses:

```json
{"tools": null}
```

and:

```json
{"tools": [null]}
```

The following self-contained test reproduces both cases:

```go
package repro

import (
	"context"
	"testing"

)

func TestListToolsMalformedResponses(t *testing.T) {
	tests := []struct {
		name  string
		tools []*mcp.Tool
	}{
		{name: "null list", tools: nil},
		{name: "null item", tools: []*mcp.Tool{nil}},
	}

	for _, test := range tests {
		t.Run(test.name, func(t *testing.T) {
			server := mcp.NewServer(
				&mcp.Implementation{Name: "fixture", Version: "1"},
				&mcp.ServerOptions{
					Capabilities: &mcp.ServerCapabilities{
						Tools: &mcp.ToolCapabilities{},
					},
				},
			)
			server.AddReceivingMiddleware(func(next mcp.MethodHandler) mcp.MethodHandler {
				return func(
					ctx context.Context,
					method string,
					request mcp.Request,
				) (mcp.Result, error) {
					if method == "tools/list" {
						return &mcp.ListToolsResult{Tools: test.tools}, nil
					}
					return next(ctx, method, request)
				}
			})

			serverTransport, clientTransport := mcp.NewInMemoryTransports()

			serverSession, err := server.Connect(
				context.Background(),
				serverTransport,
				nil,
			)
			if err != nil {
				t.Fatal(err)
			}
			defer serverSession.Close()

			client := mcp.NewClient(
				&mcp.Implementation{Name: "client", Version: "1"},
				nil,
			)
			clientSession, err := client.Connect(
				context.Background(),
				clientTransport,
				nil,
			)
			if err != nil {
				t.Fatal(err)
			}
			defer clientSession.Close()

			var result *mcp.ListToolsResult
			var listErr error
			var panicValue any

			func() {
				defer func() {
					panicValue = recover()
				}()
				result, listErr = clientSession.ListTools(
					context.Background(),
					nil,
				)
			}()

			if panicValue != nil {
				t.Fatalf("ListTools panicked: %v", panicValue)
			}
			if listErr == nil && result != nil && result.Tools != nil {
				t.Fatalf(
					"ListTools silently normalized malformed tools to %#v",
					result.Tools,
				)
			}
		})
	}
}
```

**Expected behavior**
Ideally, `ListTools` should return a protocol/decoding error for either malformed response.

If complete response-schema validation is intentionally left to callers, the SDK should at minimum:

- preserve the nil list/item so callers can detect it;
- never panic on untrusted server data; and
- avoid silently converting `tools: null` into the valid-looking `tools: []`.

`filterValidTools` is documented as filtering tools with invalid `x-mcp-header` annotations. A `null` array item is not a valid `Tool` with a bad annotation and should not be dereferenced by that path.

**Logs**
```text
=== RUN   TestListToolsMalformedResponses
=== RUN   TestListToolsMalformedResponses/null_list
    malformed_tools_test.go: ListTools silently normalized malformed tools to []*mcp.Tool{}
=== RUN   TestListToolsMalformedResponses/null_item
    malformed_tools_test.go: ListTools panicked: runtime error: invalid memory address or nil pointer dereference
--- FAIL: TestListToolsMalformedResponses
```

The panic stack points to:

```text
github.com/modelcontextprotocol/go-sdk/mcp.validateParamHeaderAnnotations
    mcp/streamable_headers.go:278
github.com/modelcontextprotocol/go-sdk/mcp.filterValidTools
    mcp/streamable_headers.go:265
github.com/modelcontextprotocol/go-sdk/mcp.(*ClientSession).ListTools
    mcp/client.go:1268
```

**Additional context**
I suspect `ClientSession.ListTools` now unconditionally calls:

```go
result.Tools = filterValidTools(cs.client.opts.Logger, result.Tools)
```

`filterValidTools` starts with:

```go
result := make([]*Tool, 0, len(tools))
```

This turns a nil input slice into a non-nil empty slice.

It then calls:

```go
validateParamHeaderAnnotations(tool)
```

without checking whether `tool` is nil. `validateParamHeaderAnnotations` dereferences `tool.InputSchema`, causing the panic for `[null]`.

n0liu:
Fixes #1119.

`ClientSession.ListTools` post-processes the server response through `filterValidTools`, which has two problems on malformed input:

1. **Nil-pointer panic on `"tools":[null]`.** A `nil` `*Tool` element reaches `validateParamHeaderAnnotations(tool)`, which dereferences `tool.InputSchema`, crashing the client on untrusted server input.
2. **`"tools":null` is normalized into a non-nil empty slice.** `make([]*Tool, 0, len(tools))` is always non-nil, so a nil `Tools` slice from a malformed response is silently laundered into valid-looking data.

Reproduction:

```go
filterValidTools(nil, []*Tool{nil}) // panics: nil pointer dereference
filterValidTools(nil, nil)          // returns non-nil slice (len 0), want nil
```

Fix (`mcp/streamable_headers.go`):

- Skip (log and exclude) nil tool elements instead of dereferencing them, so a malformed `"tools":[null]` can no longer crash the client. This is consistent with the function’s existing contract of excluding invalid tools.
- Return early with `nil` for a nil input slice so a `"tools":null` response is preserved as nil rather than normalized into an empty slice.

Added `TestFilterValidToolsNilHandling` covering both defects plus the mixed valid/nil case. `go test ./mcp/` passes; `gofmt` and `go vet` are clean.

Note on scope: this keeps the change minimal and within `filterValidTools`’s existing signature. If maintainers would rather surface a hard protocol error for a malformed `"tools":null` / `"tools":[null]` response at the `ListTools` level (rather than tolerating and filtering it), I’m happy to follow up with that larger change.

### DIFF
--- a/mcp/streamable_headers.go
+++ b/mcp/streamable_headers.go
@@ -259,9 +259,20 @@ func generateParamHeaders(tool *Tool, params json.RawMessage) map[string]string 
 // filterValidTools returns only tools that have valid
 // x-mcp-header annotations. Invalid tools are logged and excluded.
 func filterValidTools(logger *slog.Logger, tools []*Tool) []*Tool {
+	// Preserve a nil slice as nil rather than normalizing a malformed
+	// "tools":null response into a non-nil empty slice.
+	if tools == nil {
+		return nil
+	}
 	logger = ensureLogger(logger)
 	result := make([]*Tool, 0, len(tools))
 	for _, tool := range tools {
+		// A nil tool (e.g. from a malformed "tools":[null] response) is
+		// invalid; exclude it instead of dereferencing it.
+		if tool == nil {
+			logger.Error("excluding nil tool from tools/list")
+			continue
+		}
 		if err := validateParamHeaderAnnotations(tool); err != nil {
 			logger.Error("excluding tool from tools/list", "tool", tool.Name, "error", err)
 			continue
--- a/mcp/streamable_headers.go
+++ b/mcp/streamable_headers.go
@@ -259,11 +259,6 @@ func generateParamHeaders(tool *Tool, params json.RawMessage) map[string]string 
 // filterValidTools returns only tools that have valid
 // x-mcp-header annotations. Invalid tools are logged and excluded.
 func filterValidTools(logger *slog.Logger, tools []*Tool) []*Tool {
-	// Preserve a nil slice as nil rather than normalizing a malformed
-	// "tools":null response into a non-nil empty slice.
-	if tools == nil {
-		return nil
-	}
 	logger = ensureLogger(logger)
 	result := make([]*Tool, 0, len(tools))
 	for _, tool := range tools {
--- a/mcp/streamable_headers_test.go
+++ b/mcp/streamable_headers_test.go
@@ -969,29 +969,15 @@ func TestFilterValidTools(t *testing.T) {
 		t.Errorf("filterValidTools returned [%s, %s, %s, %s], want [valid, plain, nested-valid, valid-jsonschema]",
 			result[0].Name, result[1].Name, result[2].Name, result[3].Name)
 	}
-}
 
-func TestFilterValidToolsNilHandling(t *testing.T) {
-	// A malformed "tools":[null] response must not panic the client, and a
-	// "tools":null response must not be normalized into a non-nil empty slice.
-	t.Run("nil tool element is excluded without panicking", func(t *testing.T) {
-		got := filterValidTools(nil, []*Tool{nil})
-		if len(got) != 0 {
-			t.Fatalf("filterValidTools returned %d tools, want 0", len(got))
-		}
-	})
-	t.Run("nil slice is preserved as nil", func(t *testing.T) {
-		if got := filterValidTools(nil, nil); got != nil {
-			t.Fatalf("filterValidTools returned non-nil slice (len=%d) for nil input, want nil", len(got))
-		}
-	})
-	t.Run("valid tools are kept alongside nil elements", func(t *testing.T) {
-		valid := &Tool{Name: "valid"}
-		got := filterValidTools(nil, []*Tool{nil, valid, nil})
-		if len(got) != 1 || got[0].Name != "valid" {
-			t.Fatalf("filterValidTools returned %d tools, want [valid]", len(got))
-		}
-	})
+	// Regression for #1119: a malformed "tools":[null] response must not panic;
+	// nil tools are treated as invalid and excluded.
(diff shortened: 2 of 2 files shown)


mcp/streamable_headers.go
```diff
@@ -259,9 +259,20 @@ func generateParamHeaders(tool *Tool, params json.RawMessage) map[string]string
 // filterValidTools returns only tools that have valid
 // x-mcp-header annotations. Invalid tools are logged and excluded.
 func filterValidTools(logger *slog.Logger, tools []*Tool) []*Tool {
+	// Preserve a nil slice as nil rather than normalizing a malformed
+	// "tools":null response into a non-nil empty slice.
+	if tools == nil {
+		return nil
+	}
```
guglielmo-san: I don't think this code is necessary. Returning an empty list or nil does not change anything from the user point of view

mcp/streamable_headers_test.go
```diff
@@ -971,6 +971,29 @@ func TestFilterValidTools(t *testing.T) {
 	}
 }
 
+func TestFilterValidToolsNilHandling(t *testing.T) {
+	// A malformed "tools":[null] response must not panic the client, and a
+	// "tools":null response must not be normalized into a non-nil empty slice.
```
guglielmo-san: can you merge this test with the existing FilterValidTools ?

mcp/streamable_headers.go
n0liu: Good point — nil vs an empty slice isn't observable to callers, so I've dropped that guard. The nil-*tool* skip that actually fixes the `tools:[null]` panic stays. Done in b73df06.

mcp/streamable_headers_test.go
n0liu: Done — merged the nil-handling checks into `TestFilterValidTools` and removed the separate test (b73df06).

guglielmo-san (approved):

guglielmo-san:

PR #1121, opened 2026-07-30
mcp: limit SSE discover supportedVersions to SSE-defined revisions

### LINKED ISSUE #1112: SSE transport advertises 2026-07-28 in server/discover, but that revision only defines stdio and Streamable HTTP
**Describe the bug**

A server served via `NewSSEHandler` advertises `2026-07-28` in its `server/discover` `supportedVersions`, so a Modern-first client negotiates that revision **over the deprecated 2024-11-05 HTTP+SSE binding**.

MCP 2026-07-28 defines exactly two transport bindings — stdio and Streamable HTTP. It says nothing about HTTP+SSE, so advertising it from the SSE handler tells clients the server speaks a revision over a binding that revision does not define.

**Cause**

`filterSupportedVersions` (`mcp/server.go:915`) returns every SDK-supported version unless the `Transport` implements `ProtocolVersionSupporter`:

```go
func filterSupportedVersions(t Transport) []string {
	pvs, ok := t.(ProtocolVersionSupporter)
	if !ok {
		return slices.Clone(supportedProtocolVersions)
	}
	...
}
```

`SSEServerTransport` does not implement it. Only the streamable transport does (`mcp/streamable.go`). So `ss.supportedVersions` (`mcp/server.go:1390`) ends up containing `2026-07-28` for SSE sessions, and `server/discover` reports it (`mcp/server.go:904`).

**To Reproduce**

1. Build a `Server` and serve it with `mcp.NewSSEHandler(getServer, nil)`.
2. Connect a client and issue `server/discover`.
3. `supportedVersions` contains `2026-07-28`.

**Expected behavior**

The SSE handler's sessions should advertise only the revisions that define the HTTP+SSE binding — i.e. exclude `2026-07-28`.

**Suggested fix**

Have `SSEServerTransport` implement `ProtocolVersionSupporter`:

```go
// SupportsProtocolVersion reports whether the transport can serve the given
// protocol version. MCP 2026-07-28 defines only the stdio and Streamable HTTP
// bindings, so the (deprecated) HTTP+SSE transport does not serve it.
func (t *SSEServerTransport) SupportsProtocolVersion(version string) bool {
	return version != protocolVersion20260728
}
```

That is the smallest change and matches how the streamable transport already gates itself. Note there is currently **no way for an SDK consumer to work around this**: `SSEHandler.ServeHTTP` constructs the `SSEServerTransport` internally and calls `server.Connect(ctx, transport, nil)` (`mcp/sse.go:274`), and `SSEOptions` exposes only `DisableLocalhostProtection` — so a wrapper cannot supply or decorate the transport to add the interface.

**Additional context**

Found while adopting v1.7.0-pre.3 in a shim library that offers both an SSE and a Streamable HTTP server. Impact for us is low in practice — our Modern client path only ever POSTs to a configured Streamable endpoint and so never reaches the SSE message endpoint — but the advertisement is observable to any client that speaks to an SSE-fronted server, and it makes version negotiation report something the binding cannot honour.

Happy to send a PR if the suggested fix looks right.

arimu1:
## Summary

`NewSSEHandler` sessions were advertising `2026-07-28` in `server/discover` `supportedVersions`. That revision only defines stdio and Streamable HTTP, not the deprecated HTTP+SSE binding, so Modern-first clients could negotiate a revision the transport cannot honour.

## Changes

- Implement `ProtocolVersionSupporter` on `SSEServerTransport` so `filterSupportedVersions` excludes `>= 2026-07-28` (matching the streamable transport's version gate pattern).
- In `Server.discover`, only set session `InitializeParams` when the requested version is actually supported by the transport. Without this, a discover probe that returns only legacy versions still marked the session initialized, and the client's fallback `initialize` failed with `duplicate "initialize" received`.
- Tests: unit coverage for filtered SSE versions, plus an end-to-end Modern client → SSE fallback-to-initialize path.

Fixes #1112

## Test plan

- [x] `go test ./...`
- [x] `go test ./mcp/ -run 'TestSSE|TestInMemory_E2E_Discover|TestStreamableStateful_AcceptsDiscover|TestStreamableHTTP_E2E_Discover|TestClientConnectDiscover'`

Assistance: human-reviewed AI assistance used for implementation.

### DIFF
--- a/mcp/server.go
+++ b/mcp/server.go
@@ -903,9 +903,6 @@ func (s *Server) discover(_ context.Context, req *ServerRequest[*DiscoverParams]
 	// is never surfaced to the client via Mcp-Session-Id; leaving
 	// InitializeParams nil lets serveStatefulPOST's safety-net cleanup
 	// close it instead of leaking.
-	// With SSE ProtocolVersionSupporter filtering out 2026-07-28, negotiate
-	// returns a legacy revision so InitializeParams stays nil and Modern-first
-	// clients can fall back to the legacy initialize handshake (#1112).
 	if supportedVersion := negotiateMutuallySupportedVersion(versions); supportedVersion >= protocolVersion20260728 {
 		req.Session.updateState(func(state *ServerSessionState) {
 			state.InitializeParams = init
--- a/mcp/server.go
+++ b/mcp/server.go
@@ -903,6 +903,9 @@ func (s *Server) discover(_ context.Context, req *ServerRequest[*DiscoverParams]
 	// is never surfaced to the client via Mcp-Session-Id; leaving
 	// InitializeParams nil lets serveStatefulPOST's safety-net cleanup
 	// close it instead of leaking.
+	// With SSE ProtocolVersionSupporter filtering out 2026-07-28, negotiate
+	// returns a legacy revision so InitializeParams stays nil and Modern-first
+	// clients can fall back to the legacy initialize handshake (#1112).
 	if supportedVersion := negotiateMutuallySupportedVersion(versions); supportedVersion >= protocolVersion20260728 {
 		req.Session.updateState(func(state *ServerSessionState) {
 			state.InitializeParams = init
--- a/mcp/sse.go
+++ b/mcp/sse.go
@@ -189,6 +189,13 @@ func (t *SSEServerTransport) Connect(context.Context) (Connection, error) {
 	return &sseServerConn{t: t}, nil
 }
 
+// SupportsProtocolVersion reports whether the HTTP+SSE transport can serve the
+// given protocol version. MCP 2026-07-28 defines only the stdio and Streamable
+// HTTP bindings, so this (deprecated) transport does not advertise that revision.
+func (t *SSEServerTransport) SupportsProtocolVersion(version string) bool {
+	return version < protocolVersion20260728
+}
+
 func (h *SSEHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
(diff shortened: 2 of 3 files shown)


arimu1:
Rebased onto current `main` (includes #1137 session leak fix and #1127 ProtocolVersion export).

**Conflict resolution:** kept main's discover `InitializeParams` gate (`negotiateMutuallySupportedVersion >= 2026-07-28`). Combined with this PR's `SSEServerTransport.SupportsProtocolVersion` filter, SSE discover still leaves the session uninitialized so Modern-first clients can fall back to legacy `initialize`.

Tip `600f07b` · `go test ./mcp/` green.

mcp/server.go
```diff
@@ -903,6 +903,9 @@ func (s *Server) discover(_ context.Context, req *ServerRequest[*DiscoverParams]
 	// is never surfaced to the client via Mcp-Session-Id; leaving
 	// InitializeParams nil lets serveStatefulPOST's safety-net cleanup
 	// close it instead of leaking.
+	// With SSE ProtocolVersionSupporter filtering out 2026-07-28, negotiate
+	// returns a legacy revision so InitializeParams stays nil and Modern-first
+	// clients can fall back to the legacy initialize handshake (#1112).
```
guglielmo-san: i would remove this comment

arimu1: Removed — agreed, the existing transport gate comment already covers this.

guglielmo-san (approved):

guglielmo-san:
@arimu1 thank you for the contribution!

PR #1127, opened 2026-08-01
mcp: export ClientSessionOptions.ProtocolVersion

### LINKED ISSUE #1113: No public way for a client to request a specific protocol version

`ClientSessionOptions.protocolVersion` is unexported and documented as test-only (`mcp/client.go:251`):

```go
type ClientSessionOptions struct {
	// protocolVersion overrides the protocol version sent in the initialize
	// request, for testing. If empty, latestProtocolVersion is used.
	protocolVersion string
}
```

So `Client.Connect` always requests `latestProtocolVersion` (`mcp/client.go:313`) and there is no public way for a consumer to request a specific revision. The SDK's own tests can, via the unexported field; external code cannot.

Export it, or add an equivalent public option — e.g. `ClientSessionOptions.ProtocolVersion string`, keeping empty as "use the latest" so existing behaviour is unchanged.

Negotiation semantics need not change: the requested version would still be subject to the existing negotiate-down path, so a server that cannot serve it responds as it does today.

None available. The field is unexported and `Connect` takes no other version input, so a consumer cannot influence the requested version at all — including by wrapping or decorating, since the value is read inside `Connect`.

**Additional context**

Two concrete motivations, both from building a compatibility shim on top of this SDK:

1. **Wrapping an API that accepts a requested version.** Our shim exposes an mcp-go-shaped `Initialize(ctx, InitializeRequest)` whose params include a `ProtocolVersion`. Because the SDK offers no knob, the shim can only discard it — a caller asking for `2025-11-25` silently gets `2026-07-28` requested on the wire, with no error and no warning. Being unable to either honour *or* meaningfully reject the request is the awkward part: the least-bad option available to us is to document that the field does nothing.

2. **Deliberately pinning an older revision.** A proxy that must present a specific revision to a downstream peer, or an integration test asserting behaviour against a particular revision, currently cannot express that against a real client — which is presumably why the test-only field exists in the first place.

If exporting the existing field is acceptable I am happy to send a PR.

arimu1:
## Summary

`ClientSessionOptions.protocolVersion` was unexported and documented as test-only, so external consumers of `Client.Connect` could not request a specific protocol revision.

This exports the field as `ProtocolVersion` with the same semantics:

- empty → request the latest supported version (unchanged default)
- non-empty → send that version in initialize/discover
- existing negotiate-down path is unchanged

## Usage

```go
cs, err := client.Connect(ctx, transport, &mcp.ClientSessionOptions{
    ProtocolVersion: "2025-11-25",
})
```

## Testing

- `go test ./mcp/ -count=1` (pass)

Fixes #1113

### DIFF
--- a/mcp/client.go
+++ b/mcp/client.go
@@ -245,11 +245,12 @@ func (e unsupportedProtocolVersionError) Error() string {
 	return fmt.Sprintf("unsupported protocol version: %q", e.version)
 }
 
-// ClientSessionOptions is reserved for future use.
+// ClientSessionOptions configures a client session created by [Client.Connect].
 type ClientSessionOptions struct {
-	// protocolVersion overrides the protocol version sent in the initialize
-	// request, for testing. If empty, latestProtocolVersion is used.
-	protocolVersion string
+	// ProtocolVersion is the protocol version sent in the initialize (or
+	// discover) request. If empty, the latest supported version is used.
+	// The server may negotiate a different mutually supported version.
+	ProtocolVersion string
 }
 
 func (c *Client) capabilities(protocolVersion string) *ClientCapabilities {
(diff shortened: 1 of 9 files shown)


guglielmo-san (approved):

guglielmo-san:
This makes sense, thank you for the contribution

PR #1128, opened 2026-08-01
mcp: add ServerOptions.DefaultCacheable

### LINKED ISSUE #1094: No option to override forced Cacheable defaults (cacheScope public, ttlMs 0) on SDK-generated results
### Use case

We run a multi-tenant MCP gateway (Arcade) on `v1.7.0-pre.2`. Our `tools/list` responses are **per-tenant** (allow-lists, auth state), so the SDK's forced cacheable defaults — `cacheScope: "public"`, `ttlMs: 0` — are wrong for us twice over: `public` invites cross-tenant cache poisoning at intermediaries, and we want a real TTL (we use `private`/60s). We can override our own handler results, but not the SDK-generated ones (notably `server/discover`).

### Current behavior (pre.2)

- Every SDK-generated list/discover result embeds `Cacheable{TTLMs, CacheScope}` (protocol.go) and each internal handler calls unexported `setDefaultCacheableValues()` → `CacheScope = "public"`, `TTLMs = 0`. Call sites include discover and the tools/prompts/resources/templates lists.
- No `ServerOptions`/`StreamableHTTPOptions` field influences these defaults.

### Why the workaround is fragile

Receiving middleware can type-assert each concrete result type after the fact and overwrite `.Cacheable` — that means enumerating every result type, keying off method strings, and depending on running after the SDK stamped its values. It works (we do it for our own `tools/list`), but there is no supported way to change what the SDK itself generates, and each new SDK result type silently reverts to `public`.

### Proposal

```go
type ServerOptions struct {
    // DefaultCacheable, if non-nil, supplies the Cacheable values stamped on
    // SDK-generated results (server/discover, *_list). Nil keeps today's
    // public/0 defaults.
    DefaultCacheable *Cacheable
}
```

One field, zero behavior change for existing users. For multi-tenant servers a `public` default is arguably unsafe, which is why a first-class knob (rather than post-hoc middleware) feels warranted.

arimu1:
## Summary

Adds `ServerOptions.DefaultCacheable *Cacheable` so multi-tenant / private servers can override the SDK's forced cacheable defaults (`cacheScope: "public"`, `ttlMs: 0`) on SDK-generated results.

- Nil keeps today's historical defaults (no behavior change).
- Non-nil values are stamped by `setDefaultCacheableValues` on `server/discover`, `tools/list`, `prompts/list`, `resources/list`, `resources/templates/list`, and `resources/read` (after the handler returns).
- Receiving middleware can still overwrite `Cacheable` per result after the SDK stamps values (useful for per-request / per-tenant policy).

Fixes #1094

## Motivation

Servers whose list/discover payloads are tenant-scoped want `private` (and a real TTL) as the SDK baseline, not only via post-hoc middleware type-switches. A single option covers all current SDK stamp sites and any future call sites that use the same helper.

## Test plan

- [x] `go test ./mcp/ -run 'TestServerDefaultCacheable|TestSetDefaultCacheableValues' -count=1`
- [x] `go test ./mcp/ -count=1`
- [x] `go test ./... -count=1`
- [x] Confirmed historical defaults when `DefaultCacheable` is nil
- [x] Confirmed `private`/60s stamped on list + read results when set
- [x] Confirmed nil-defaults path still preserves pre-set `TTLMs` (read handler) while forcing `CacheScope: "public"`


mcp/server_test.go
```diff
@@ -1691,3 +1691,105 @@ func TestServerSession_RejectsServerInitiated(t *testing.T) {
...
+			read, err := cs.ReadResource(ctx, &ReadResourceParams{URI: "test://r"})
+			if err != nil {
+				t.Fatal(err)
+			}
+			check(t, "ReadResource", read.Cacheable)
+		})
+	}
+}
+
+func TestSetDefaultCacheableValues(t *testing.T) {
```
guglielmo-san: lets remove this test

arimu1:
Removed `TestServerDefaultCacheable` as requested, and rebased onto current `main`. Production `ServerOptions.DefaultCacheable` behavior and the unit test `TestSetDefaultCacheableValues` are unchanged.

mcp/protocol.go
```diff
@@ -1192,7 +1192,14 @@ func (c Cacheable) GetTTLMs() int { return c.TTLMs }
...
 // setDefaultCacheableValues sets the default values for the cacheable fields.
-func (c *Cacheable) setDefaultCacheableValues() {
+// If defaults is non-nil, its TTLMs and CacheScope are copied onto c.
+// Otherwise CacheScope is set to "public" (TTLMs stays 0 unless already set).
+func (c *Cacheable) setDefaultCacheableValues(defaults *Cacheable) {
+	if defaults != nil {
+		c.TTLMs = defaults.TTLMs
+		c.CacheScope = defaults.CacheScope
+		return
+	}
```
guglielmo-san, line 1207: the issue here is that if defaults does not set the CacheScope, it will not be set to public

arimu1:
@guglielmo-san Thanks — fixed `setDefaultCacheableValues` so an empty `CacheScope` on non-nil defaults falls back to `"public"` (same as nil defaults). Added regression test `defaults with empty CacheScope fall back to public`.

mcp/server.go
```diff
@@ -169,6 +169,15 @@ type ServerOptions struct {
...
 	GetSessionID func() string
+
+	// DefaultCacheable, if non-nil, supplies the [Cacheable] values stamped on
+	// SDK-generated results (server/discover, list methods, and resources/read
+	// after the handler returns). If nil, those results use the historical
+	// defaults: CacheScope "public" and TTLMs 0.
+	//
+	// Receiving middleware can still overwrite Cacheable on a per-result basis
+	// after the SDK stamps these values.
+	DefaultCacheable *Cacheable
```
guglielmo-san: What do you think about using a callback instead? I think it would give more flexibility.

```
// DefaultCacheable, if non-nil, is called for each SDK-generated
// cacheable result (server/discover, *_list, resources/read) to
// determine the Cacheable values stamped on the response. The
// request is provided so policy can vary by method, session,
// tenant, or authentication state.
//
// If nil, SDK-generated results use CacheScope "public" and
// TTLMs 0, matching the historical defaults.

DefaultCacheable func(ctx context.Context, req Request) Cacheable
```

mcp/server_test.go
guglielmo-san: You did not remove the test i intended

mcp/server.go
arimu1: Adopted as proposed: `DefaultCacheable func(ctx context.Context, req Request) Cacheable`. See tip `15b7b8f`.

arimu1:
@guglielmo-san Thanks for the feedback — adopted the callback API:

```go
DefaultCacheable func(ctx context.Context, req Request) Cacheable
```

- Nil keeps historical `public` / `ttlMs: 0` defaults.
- Callback return values are stamped on SDK-generated results (`server/discover`, `*_list`, `resources/read`); empty `CacheScope` still falls back to `"public"`.
- Removed `TestSetDefaultCacheableValues` (the unit test you intended). Coverage is via restored `TestServerDefaultCacheable` (historical defaults, private/TTL, empty-scope fallback).

Tip: `15b7b8f720c78d008e0f18f1961e1cb99c9afc0e`
`go test ./mcp/...` passes.

mcp/server_test.go
arimu1: Removed `TestSetDefaultCacheableValues` (this was the one you intended). Behavior is covered by `TestServerDefaultCacheable` under the new callback API, including empty `CacheScope` → `"public"`.

guglielmo-san:
Closing as duplicate of #1203

PR #1135, opened 2026-08-02
auth: advertise RFC 6750 error code in WWW-Authenticate challenge

### LINKED ISSUE #1134: RequireBearerToken omits RFC 6750 error code, so the SDK's own step-up flow never triggers
### Describe the bug

`RequireBearerToken` (auth/auth.go) never adds an `error=` auth-param to the `WWW-Authenticate` challenge it emits. On insufficient scope it returns `403` with a header like:

```
WWW-Authenticate: Bearer resource_metadata="...", scope="admin"
```

with no `error="insufficient_scope"`.

But the SDK's **own client** gates step-up re-authorization on exactly that value (auth/authorization_code.go, `AuthorizationCodeHandler.Authorize`):

```go
if resp.StatusCode == http.StatusForbidden && errorFromChallenges(wwwChallenges) != "insufficient_scope" {
    // We only want to perform step-up authorization for insufficient_scope errors.
    return nil
}
```

`errorFromChallenges` reads `Params["error"]`; since the server never sets it, it returns `""`, the condition is true, and `Authorize` returns without ever performing step-up — even though the required scopes were advertised via `scope`. So an SDK client hitting an SDK server that needs more scopes is never prompted to re-authorize; the call just keeps failing with 403 instead of upgrading scopes (the [step-up flow](https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#step-up-authorization-flow)).

It's also an RFC 6750 §3.1 deviation: the `error` code (`insufficient_scope` on 403, `invalid_token` on the token-invalid 401s) should be advertised in the challenge.

### To Reproduce

Wrap a handler with `RequireBearerToken(verifier, &RequireBearerTokenOptions{Scopes: []string{"admin"}, ResourceMetadataURL: "..."})`, present a valid token whose `Scopes` lack `admin`, and inspect the response: `403` with a `WWW-Authenticate` header that contains `scope="admin"` but no `error="insufficient_scope"`. Feeding that response to the SDK client's `Authorize` skips step-up.

### Expected behavior

The `WWW-Authenticate` challenge on a 401/403 should carry the RFC 6750 error code (`error="insufficient_scope"` / `error="invalid_token"`), so clients — including the SDK's own step-up flow — can react.

I have a fix ready (threads the error code out of `verify` and adds the `error=` param; additive, non-breaking) and will open a PR referencing this issue.

### Version

`main` / v1.7.0.

PratikDhanave:
## Summary

`RequireBearerToken` never added an `error=` auth-param to the `WWW-Authenticate` header it emits. On insufficient scope it returned `403` with `resource_metadata`/`scope` but **no** `error="insufficient_scope"`.

The SDK's own client gates step-up re-authorization on exactly that value (`AuthorizationCodeHandler.Authorize`):

```go
if resp.StatusCode == http.StatusForbidden && errorFromChallenges(wwwChallenges) != "insufficient_scope" {
    return nil // skip step-up
}
```

`errorFromChallenges` reads `Params["error"]`; since the server never set it, it returned `""`, the condition was true, and step-up never fired. A client hitting an SDK-based server that needs additional scopes just kept failing with `403` instead of upgrading scopes. It's also an RFC 6750 §3.1 deviation.

Fixes #1134

## Change

Thread the RFC 6750 error code out of `verify` and add the `error=` auth-param to the challenge:

| Failure | code | `error=` |
|---|---|---|
| insufficient scope | 403 | `insufficient_scope` |
| invalid / expired / missing-exp token | 401 | `invalid_token` |
| `ErrOAuth` | 400 | `invalid_request` |
| no token (no credentials) | 401 | *(none — per RFC 6750)* |
| 5xx | 500 | *(none)* |

The change only **adds** a param to the header — it is non-breaking.

## Testing

- Extended `TestVerify` to assert the error code returned for each failure path.
- Added `TestRequireBearerTokenAdvertisesInsufficientScope`, which drives the middleware end-to-end and parses the emitted header with the same `oauthex.ParseWWWAuthenticate` the client uses, asserting `error="insufficient_scope"` is visible.
- Updated `TestRequireBearerToken`'s expected `WWW-Authenticate` headers to include the new code.
- Verified the interop test **fails without the fix** (`error param = "", want "insufficient_scope"`) and passes with it. `go vet ./auth/` and the full `go test ./auth/` suite pass.

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -99,20 +99,28 @@ func RequireBearerToken(verifier TokenVerifier, opts *RequireBearerTokenOptions)
 
 	return func(handler http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-			tokenInfo, errmsg, code := verify(r, verifier, opts)
+			tokenInfo, errmsg, code, authError := verify(r, verifier, opts)
 			if code != 0 {
 				if code == http.StatusUnauthorized || code == http.StatusForbidden {
+					var params []string
+					// Emit the RFC 6750 error code so clients can react to it. In
+					// particular, the SDK's own step-up flow only re-authorizes
+					// when it sees error="insufficient_scope" on a 403
+					// (see AuthorizationCodeHandler.Authorize); without it, a
+					// scope shortfall is never upgraded.
+					if authError != "" {
+						params = append(params, fmt.Sprintf("error=%q", authError))
+					}
 					if opts != nil {
-						var params []string
 						if opts.ResourceMetadataURL != "" {
 							params = append(params, fmt.Sprintf("resource_metadata=%q", opts.ResourceMetadataURL))
 						}
 						if len(opts.Scopes) > 0 {
 							params = append(params, fmt.Sprintf("scope=%q", strings.Join(opts.Scopes, " ")))
 						}
-						if len(params) > 0 {
-							w.Header().Add("WWW-Authenticate", "Bearer "+strings.Join(params, ", "))
-						}
+					}
+					if len(params) > 0 {
+						w.Header().Add("WWW-Authenticate", "Bearer "+strings.Join(params, ", "))
 					}
 				}
 				http.Error(w, errmsg, code)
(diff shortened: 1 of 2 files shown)


auth/auth.go
```diff
@@ -124,35 +132,41 @@ func RequireBearerToken(verifier TokenVerifier, opts *RequireBearerTokenOptions)
...
 	// Verify the token and get information from it.
 	tokenInfo, err := verifier(req.Context(), fields[1], req)
 	if err != nil {
 		if errors.Is(err, ErrInvalidToken) {
-			return nil, err.Error(), http.StatusUnauthorized
+			return nil, err.Error(), http.StatusUnauthorized, "invalid_token"
 		}
 		if errors.Is(err, ErrOAuth) {
-			return nil, err.Error(), http.StatusBadRequest
+			return nil, err.Error(), http.StatusBadRequest, "invalid_request"
```
guglielmo-san, line 156: this will not trigger the authorization flow as the code in @authorization_code is just checking for `http.StatuForbidden` case.
I am not sure tho if we should extend the case for the flow re-trigger

PratikDhanave: Thanks @guglielmo-san — good catch to check, but the step-up path *does* fire with this change. `AuthorizationCodeHandler.Authorize` gates re-auth on two conditions together:

```go
if resp.StatusCode == http.StatusForbidden && errorFromChallenges(wwwChallenges) != "insufficient_scope" {
    return nil // skip step-up
}
```

It skips step-up only when the status is `403` **and** the challenge's `error` param is *not* `insufficient_scope`. Before this PR the server never set `error=` at all, so `errorFromChallenges` returned `""`, the condition was true, and step-up was skipped — that's the bug. By advertising `error="insufficient_scope"` on the 403 challenge (RFC 6750 §3.1), the condition becomes false and re-authorization proceeds.

`TestRequireBearerTokenAdvertisesInsufficientScope` drives this end-to-end and parses the emitted header with the same `oauthex.ParseWWWAuthenticate` the client uses, asserting the code is visible. So the `403`-checking logic doesn't need to change — the fix is exactly to populate the `error` param it already reads.

Issue #578, opened 2025-10-13
Streamable HTTP: don't allocate resources for failed connections


findleyr:
In the streamable HTTP handler, we allocate a `StreamableServerTransport` for incoming requests with no associated session:
https://github.com/modelcontextprotocol/go-sdk/blob/920e5a39ba9026af14ab459ce856e6ef907de0b9/mcp/streamable.go#L235

However, if the session fails to initialize (and the server is stateful), we should either immediately clean up these sessions, or avoid creating them in the first place.

In combination with #499, this is critical for avoiding resource leaks during streamable serving.

PR #1137, opened 2026-08-03
mcp: prevent leaks of ServerSession on StatefulServer

### LINKED ISSUE #1136: Streamable HTTP: server/discover leaks a ServerSession on stateful servers
## Summary

On a stateful `StreamableHTTPHandler` (the default, `Stateless: false`), every `server/discover` request creates a `ServerSession` that is registered in `StreamableHTTPHandler.sessions` and is never closed.

Because a v1.7.0 client always sends `server/discover` before falling back to `initialize`, **every client connection leaks exactly one `ServerSession`, permanently** — even when the client shuts down cleanly with a DELETE.

The leaked session is unreachable: the discover response carries no `Mcp-Session-Id`, so no client can ever DELETE it. If `StreamableHTTPOptions.SessionTimeout` is left at its zero value (no idle eviction), nothing reclaims it and server memory grows without bound.

Reproduced on both `v1.7.0` and current `main` (`0c004ee`).

## Reproduction

Self-contained, depends only on the SDK. Each client is closed cleanly via `session.Close()`.

```go
package main

import (
	"context"
	"fmt"
	"net/http"
	"net/http/httptest"

)

func count(s *mcp.Server) int {
	n := 0
	for range s.Sessions() {
		n++
	}
	return n
}

func main() {
	srv := mcp.NewServer(&mcp.Implementation{Name: "repro", Version: "1.0.0"}, nil)
	mcp.AddTool(srv, &mcp.Tool{Name: "noop", Description: "noop"},
		func(context.Context, *mcp.CallToolRequest, struct{}) (*mcp.CallToolResult, any, error) {
			return &mcp.CallToolResult{}, nil, nil
		})

	// Stateful (default). SessionTimeout unset => no idle eviction.
	h := mcp.NewStreamableHTTPHandler(func(*http.Request) *mcp.Server { return srv }, nil)
	ts := httptest.NewServer(h)
	defer ts.Close()

	ctx := context.Background()
	for i := 1; i <= 3; i++ {
		c := mcp.NewClient(&mcp.Implementation{Name: "c", Version: "1.0.0"}, nil)
		sess, err := c.Connect(ctx, &mcp.StreamableClientTransport{
			Endpoint: ts.URL, DisableStandaloneSSE: true, MaxRetries: -1,
		}, nil)
		if err != nil {
			panic(err)
		}
		if _, err := sess.ListTools(ctx, nil); err != nil {
			panic(err)
		}
		fmt.Printf("client %d connected: negotiated=%s sessions=%d\n",
			i, sess.InitializeResult().ProtocolVersion, count(srv))
		if err := sess.Close(); err != nil { // clean shutdown, sends DELETE
			panic(err)
		}
		fmt.Printf("client %d closed   : sessions=%d (want 0)\n", i, count(srv))
	}
	fmt.Printf("\nFINAL leaked sessions = %d (want 0)\n", count(srv))
}
```

### Actual output (v1.7.0 and main @ 0c004ee)

```
client 1 connected: negotiated=2025-11-25 sessions=2
client 1 closed   : sessions=1 (want 0)
client 2 connected: negotiated=2025-11-25 sessions=3
client 2 closed   : sessions=2 (want 0)
client 3 connected: negotiated=2025-11-25 sessions=4
client 3 closed   : sessions=3 (want 0)

FINAL leaked sessions = 3 (want 0)
```

### Expected

`FINAL leaked sessions = 0`. A cleanly closed client should leave no server session behind.

## Wire trace

Captured by wrapping the handler with a logging `http.Handler`:

| # | Request | Response |
|---|---|---|
| 1 | `server/discover` (`Mcp-Protocol-Version: 2026-07-28`) | 200, `supportedVersions: ["2025-11-25", …]`, **no `Mcp-Session-Id`** |
| 2 | `initialize` (falls back to `2025-11-25`) | 200, `Mcp-Session-Id: YFTC…` |
| 3 | `notifications/initialized` | 202 |

After step 1 the server already holds one session that the client has no way to address.

## Root cause

`server/discover` is deliberately exempt from the stateful rejection of new-protocol requests (`mcp/streamable.go:1516`):

```go
// server/discover is exempt from the stateful
// rejection as it should learn about the supported protocols from the
// DiscoverResult response.
if !c.stateless && jreq.Method != methodDiscover {
    http.Error(w, fmt.Sprintf(
        "Bad Request: protocol version %q is only supported on stateless HTTP servers ...
```

However, the exempted request still flows through the ordinary session-creating path in `serveStatefulPOST`. Two things then conspire:

1. `sessionID = server.opts.GetSessionID()` returns a non-empty ID, so the request does **not** take the ephemeral branch (`if sessionID == ""` → `defer session.Close()`). A full session is created and stored in `h.sessions`. Yet the discover response never surfaces `Mcp-Session-Id` to the client, so the session is permanently unaddressable — DELETE is impossible.

2. The end-of-function safety net does not fire:

   ```go
   defer func() {
       // If initialization failed, clean up the session (#578).
       if session.InitializeParams() == nil {
           session.Close()
       }
   }()
   ```

   Since #1049, a discover request's `_meta` populates `InitializeParams`, so `InitializeParams() != nil` and the cleanup introduced for #578 is skipped.

The net effect is a session that is neither addressable nor reclaimable.

## Behavior matrix

Sending `server/discover` five times directly:

| Configuration | Sessions afterwards |
|---|---|
| `SessionTimeout: 0` (zero value) | 5 after 5 requests; never decreases |
| `SessionTimeout: 2s` | 0 after 3s (reclaimed by the idle timer) |
| `Stateless: true` | 0 (no session is ever created) |

So the leak is confined to stateful servers, and `SessionTimeout` only masks it — sessions still accumulate between sweeps, and the zero value provides no protection at all.

## Suggested fix

A discover-only request establishes no session by definition, so it should take the existing ephemeral path (connect, serve, `defer session.Close()`) rather than the addressable-session path.

```diff
diff --git a/mcp/streamable.go b/mcp/streamable.go
--- a/mcp/streamable.go
+++ b/mcp/streamable.go
@@ -593,12 +593,34 @@ func (h *StreamableHTTPHandler) serveStatefulGET(w http.ResponseWriter, req *htt
 		return
 	}
 
 	sessInfo.transport.ServeHTTP(w, req)
 }
 
+// isDiscoverOnlyRequest reports whether every JSON-RPC request in the body is
+// server/discover. The body is restored so downstream handlers can re-read it.
+func isDiscoverOnlyRequest(req *http.Request) bool {
+	body, err := io.ReadAll(req.Body)
+	req.Body.Close()
+	req.Body = io.NopCloser(bytes.NewBuffer(body))
+	if err != nil {
+		return false
+	}
+	msgs, _, err := readBatch(body)
+	if err != nil || len(msgs) == 0 {
+		return false
+	}
+	for _, msg := range msgs {
+		r, ok := msg.(*jsonrpc.Request)
+		if !ok || r.Method != methodDiscover {
+			return false
+		}
+	}
+	return true
+}
+
 // serveStatefulDELETE handles DELETE requests for session termination.
 // DELETE requires a valid Mcp-Session-Id header.
 func (h *StreamableHTTPHandler) serveStatefulDELETE(w http.ResponseWriter, req *http.Request) {
 	sessionID := req.Header.Get(sessionIDHeader)
 	if sessionID == "" {
 		http.Error(w, "Bad Request: DELETE requires an Mcp-Session-Id header", http.StatusBadRequest)
@@ -649,12 +671,22 @@ func (h *StreamableHTTPHandler) serveStatefulPOST(w http.ResponseWriter, req *ht
 	if server == nil {
 		http.Error(w, "no server available", http.StatusBadRequest)
 		return
 	}
 	sessionID = server.opts.GetSessionID()
 
+	// server/discover is exempt from the new-protocol stateful rejection in
+	// streamableServerConn.ServeHTTP, so it can reach a stateful server without
+	// ever establishing a session. It must not create an addressable session:
+	// the response carries no Mcp-Session-Id, so the client could never DELETE
+	// it, and its InitializeParams are populated from _meta, which defeats the
+	// "initialization failed" cleanup below. Serve it ephemerally instead.
+	if isDiscoverOnlyRequest(req) {
+		sessionID = ""
+	}
+
 	transport := &StreamableServerTransport{
 		SessionID:    sessionID,
 		Stateless:    false,
 		EventStore:   h.opts.EventStore,
 		jsonResponse: h.opts.JSONResponse,
 		logger:       h.opts.Logger,
```

### Verification of the patch (applied to main @ 0c004ee)

- The reproduction above prints `FINAL leaked sessions = 0`, and `sessions` stays at 1 while a client is connected.
- `go test -count=1 ./mcp/... ./internal/...` gives identical results with and without the patch. My environment (Windows) has one pre-existing failure, `TestServerConformance/resources.txtar`, caused by CRLF checkout of the txtar fixture (`"blob": "Q29udGVudHMNCg=="` vs `"Q29udGVudHMK"`); it fails on unmodified `main` too and is unrelated to this change. No other failures.
- A downstream stateful server (which asserts that one session is reused across requests) goes from failing to passing with no changes on its side.

I am happy to open a PR with this change plus a regression test if the approach looks right. An alternative, if discover should never reach session creation at all, is to dispatch it earlier in `serveStatefulPOST`, before `getServer`/`GetSessionID`; I went with the minimal change that reuses the existing ephemeral machinery.

## Environment

- `github.com/modelcontextprotocol/go-sdk` v1.7.0, and `main` at `0c004ee`
- Go 1.26.5

## Possibly related (none appear to cover this)

- #1025 — `server/discover` on stateful servers responding without `2026-07-28` and the client falling back to `initialize`. Same code path, but about advertised versions and the fallback, not session lifetime.
- #1111 / #1108 — rework the error emitted by (or the version gate behind) the very branch that exempts discover; neither touches session creation.
- #1073 — goroutine leak on **stateless** connections. Different mechanism: goroutines rather than `ServerSession` values retained in `StreamableHTTPHandler.sessions`.
- #578 — the "don't allocate resources for failed connections" cleanup that this path silently bypasses.

## How this surfaced

A Renovate PR bumping a downstream project from v1.6.1 to v1.7.0 turned a passing test red; the test asserts that a stateful handler reuses a single session across requests. Downstream write-up: https://github.com/scottlz0310/review-raven/issues/109

guglielmo-san:
Fixes #1136

### DIFF
--- a/mcp/server.go
+++ b/mcp/server.go
@@ -903,7 +903,7 @@ func (s *Server) discover(_ context.Context, req *ServerRequest[*DiscoverParams]
 	// is never surfaced to the client via Mcp-Session-Id; leaving
 	// InitializeParams nil lets serveStatefulPOST's safety-net cleanup
 	// close it instead of leaking.
-	if slices.Contains(versions, protocolVersion20260728) {
+	if supportedVersion := negotiateMutuallySupportedVersion(versions); supportedVersion >= protocolVersion20260728 {
 		req.Session.updateState(func(state *ServerSessionState) {
 			state.InitializeParams = init
 		})
--- a/mcp/server.go
+++ b/mcp/server.go
@@ -897,9 +897,17 @@ func (s *Server) discover(_ context.Context, req *ServerRequest[*DiscoverParams]
 		Capabilities:    req.ClientCapabilities(),
 		ClientInfo:      req.ClientInfo(),
 	}
-	req.Session.updateState(func(state *ServerSessionState) {
-		state.InitializeParams = init
-	})
+	// Only persist InitializeParams when the transport can actually serve
+	// the new protocol. On transports that cannot (notably stateful
+	// StreamableHTTPHandler), a discover request creates a session that
+	// is never surfaced to the client via Mcp-Session-Id; leaving
+	// InitializeParams nil lets serveStatefulPOST's safety-net cleanup
+	// close it instead of leaking.
+	if slices.Contains(versions, protocolVersion20260728) {
+		req.Session.updateState(func(state *ServerSessionState) {
+			state.InitializeParams = init
+		})
+	}
 	res := &DiscoverResult{
 		SupportedVersions: versions,
 		Capabilities:      s.capabilities(),
--- a/mcp/streamable_test.go
+++ b/mcp/streamable_test.go
@@ -3682,6 +3682,70 @@ func TestStreamableStateful_AcceptsDiscover(t *testing.T) {
 	}
 }
 
+// On a stateful StreamableHTTPHandler, a v1.7.0-style client always
+// probes server/discover before falling back to initialize. That probe used
+// to create a ServerSession that was registered in the handler's session
+// map but never surfaced via Mcp-Session-Id, so the client could not DELETE
+// it and it was leaked for the lifetime of the process.
+//
+// The fix in Server.discover skips populating InitializeParams when the
+// transport cannot serve the new protocol, which lets the safety-net
+// cleanup in serveStatefulPOST close the otherwise-unaddressable session.
+func TestStreamableStateful_DiscoverDoesNotLeakSession(t *testing.T) {
+	ctx := context.Background()
+	server := NewServer(testImpl, nil)
+	AddTool(server, &Tool{Name: "noop", Description: "noop"},
+		func(context.Context, *CallToolRequest, struct{}) (*CallToolResult, any, error) {
+			return &CallToolResult{}, nil, nil
+		})
+
+	// Stateful (default). SessionTimeout unset so nothing masks the leak
+	// via idle eviction.
+	handler := NewStreamableHTTPHandler(func(*http.Request) *Server { return server }, nil)
+	httpServer := httptest.NewServer(handler)
+	defer httpServer.Close()
+
+	countSessions := func() int {
+		n := 0
+		for range server.Sessions() {
+			n++
+		}
+		return n
+	}
+
+	const clients = 3
+	for i := 1; i <= clients; i++ {
+		c := NewClient(&Implementation{Name: "c", Version: "1.0.0"}, nil)
+		sess, err := c.Connect(ctx, &StreamableClientTransport{
+			Endpoint:             httpServer.URL,
+			DisableStandaloneSSE: true,
+			MaxRetries:           -1,
+		}, nil)
+		if err != nil {
+			t.Fatalf("client %d Connect: %v", i, err)
+		}
+		if _, err := sess.ListTools(ctx, nil); err != nil {
+			t.Fatalf("client %d ListTools: %v", i, err)
+		}
+		// Exactly one addressable session should exist while this client
+		// is connected. Anything more means a discover-only session
+		// slipped through.
+		if got := countSessions(); got != 1 {
+			t.Errorf("client %d connected: sessions = %d, want 1", i, got)
+		}
+		if err := sess.Close(); err != nil {
+			t.Fatalf("client %d Close: %v", i, err)
+		}
+		if got := countSessions(); got != 0 {
+			t.Errorf("client %d closed: sessions = %d, want 0", i, got)
+		}
+	}
+
+	if got := countSessions(); got != 0 {
+		t.Errorf("final leaked sessions = %d, want 0", got)
+	}
+}
+
 // TestStreamableHTTP_E2E_DiscoverSuccess is a full end-to-end smoke test for
 // SEP-2575 over the streamable HTTP transport.
 func TestStreamableHTTP_E2E_DiscoverSuccess(t *testing.T) {


scottlz0310-user:
Thanks for the quick turnaround. I verified this branch (`bfee8f4`) as the reporter of #1136, and it fixes the leak.

**Reproduction from #1136** — was `FINAL leaked sessions = 3` on v1.7.0, now:

```
client 1 connected: negotiated=2025-11-25 sessions=1
client 1 closed   : sessions=0 (want 0)
client 2 connected: negotiated=2025-11-25 sessions=1
client 2 closed   : sessions=0 (want 0)
client 3 connected: negotiated=2025-11-25 sessions=1
client 3 closed   : sessions=0 (want 0)

FINAL leaked sessions = 0 (want 0)
```

The session count now stays at exactly 1 while a client is connected, instead of climbing by one per connection.

**Other checks on this branch:**

- `go test -count=1 ./mcp/...` — green.
- The new `TestStreamableStateful_DiscoverDoesNotLeakSession` — passes.
- The downstream project whose test suite regressed on the v1.6.1 → v1.7.0 bump (its test asserts that a stateful server reuses exactly one session across requests) is green again with this branch dropped in via `replace`, with **no changes on its side**.

Your approach is neater than the one I sketched in #1136. I proposed routing discover-only requests through the ephemeral path in `serveStatefulPOST`; gating the `InitializeParams` persistence on whether the transport can actually serve `2026-07-28` addresses the same root cause with a smaller change and lets the existing #578 safety net do the cleanup. The added regression test covers the coupling to that safety net, which was my only concern with relying on it.

Nothing further from me — happy to see this land.

scottlz0310-user (approved):

nahapetyan-serob (approved):

PR #1145, opened 2026-08-05
mcp: don't mutate caller params on multi-round-trip retries

### LINKED ISSUE #1144: mcp: client multi-round-trip middleware mutates caller params, replaying inputResponses across calls
**What did you do?**

Called `ClientSession.CallTool` against a tool whose handler gates on an elicitation `InputRequest` (SEP-2322), reusing one `*CallToolParams` struct for a second call:

```go
params := &mcp.CallToolParams{Name: "delete", Arguments: args}
res1, _ := cs.CallTool(ctx, params) // elicitation fires, user accepts, tool runs
res2, _ := cs.CallTool(ctx, params) // same struct, second call
```

The server handler uses the gating idiom from the SDK's own tests and examples:

```go
if len(req.Params.InputResponses) == 0 {
    return &mcp.CallToolResult{
        InputRequests: mcp.InputRequestMap{"confirm": &mcp.ElicitParams{Message: "Sure?"}},
        RequestState:  state,
    }, nil, nil
}
```

**What did you see?**

After the first call returns, the caller's params struct has been mutated: `params.InputResponses` holds the first call's `ElicitResult` and `params.RequestState` holds the first call's state. `setMultiRoundTripRetryParams` (mcp/mrtr.go) writes both fields through the caller's pointer during the middleware's retry loop and never clears them.

The second call therefore goes out with the first call's answer pre-attached. The server sees `len(InputResponses) > 0`, treats the call as already answered, and the client's `ElicitationHandler` never fires. For a confirmation gate on a destructive tool, the second action runs with no user confirmation at all.

**What did you expect to see?**

`CallTool` / `GetPrompt` / `ReadResource` should not mutate caller-owned params. Each fresh call should round-trip its own input requests; retry bookkeeping belongs to the middleware, not the caller's struct.

**What version of the Go MCP SDK are you using?**

v1.7.0; the behavior is unchanged at head (7256941).

**What version of Go are you using?**

go1.26.5 darwin/arm64

I have a fix ready (carry the retry responses/state on a shallow copy of the params instead of mutating the original) and will send a PR.

yhxlele:
Fixes #1144.

`clientMultiRoundTripMiddleware` calls `setMultiRoundTripRetryParams(req, ...)`, which writes `InputResponses` and `RequestState` through the caller's params pointer and never clears them. A `*CallToolParams` (or `GetPromptParams` / `ReadResourceParams`) struct reused across calls then carries the first call's answers into the next call: a server gating on `len(req.Params.InputResponses) == 0` — the idiom in this repo's own tests and examples — treats the fresh call as already answered, so the client's `ElicitationHandler` never fires and an elicitation-gated (e.g. destructive) tool runs without confirmation.

This change makes the retry loop carry the fulfilled responses and echoed state on a shallow copy of the params (`multiRoundTripRetryRequest`) instead of mutating the original request. Caller-owned params are now left untouched by `CallTool`, `GetPrompt`, and `ReadResource`. The server-side shim (`serverMultiRoundTripMiddleware`) is unchanged: the params it mutates are SDK-owned, decoded from the wire.

Tests: `TestMultiRoundTrip_AutoRetryDoesNotMutateCallerParams` (CallTool: params unchanged after the call, and reusing the same struct fulfills the input requests again instead of silently replaying the first call's answers — it failed with `elicitations after reuse = 1, want 2` before the fix), plus GetPrompt/ReadResource analogues asserting params are left clean. All three fail on main and pass with this change; `go test ./...` and `go vet ./...` are clean.

Note on observable behavior: callers can no longer read the fulfilled responses off their own params struct after a call returns. That side effect was undocumented, and relying on it is exactly what makes the replay hazard above possible.

### DIFF
--- a/mcp/mrtr.go
+++ b/mcp/mrtr.go
@@ -108,10 +108,43 @@ func clientMultiRoundTripMiddleware() Middleware {
 				if err != nil {
 					return nil, err
 				}
-				setMultiRoundTripRetryParams(req, responses, mrtrResult.requestState())
+				req = multiRoundTripRetryRequest(cs, req, responses, mrtrResult.requestState())
 			}
 		}
 	}
+}
+
+// multiRoundTripRetryRequest builds the request for the next round of a
+// client-side multi-round-trip retry: a request whose params carry the
+// fulfilled responses and the request state to echo. The retry params
+// are a shallow copy of the original, so the caller-owned params passed
+// to CallTool, GetPrompt, or ReadResource are never mutated: a params
+// struct reused across calls must not carry one call's inputResponses
+// and requestState into the next, where a server gating on them (for
+// example an elicitation confirmation) would treat the new call as
+// already answered.
+func multiRoundTripRetryRequest(cs *ClientSession, req Request, responses InputResponseMap, state string) Request {
+	var clone Params
+	switch p := req.GetParams().(type) {
+	case *CallToolParams:
+		cp := *p
+		clone = &cp
+	case *CallToolParamsRaw:
+		cp := *p
+		clone = &cp
+	case *GetPromptParams:
+		cp := *p
+		clone = &cp
+	case *ReadResourceParams:
+		cp := *p
+		clone = &cp
+	default:
+		// No retry params to carry; resend the request unchanged.
+		return req
+	}
+	retry := newClientRequest(cs, clone)
+	setMultiRoundTripRetryParams(retry, responses, state)
+	return retry
 }
 
 // serverMultiRoundTripMiddleware is a receiving middleware for servers that transparently
--- a/mcp/mrtr.go
+++ b/mcp/mrtr.go
@@ -108,43 +108,13 @@ func clientMultiRoundTripMiddleware() Middleware {
 				if err != nil {
 					return nil, err
 				}
-				req = multiRoundTripRetryRequest(cs, req, responses, mrtrResult.requestState())
+				req, err = setMultiRoundTripRetryParams(method, req, responses, mrtrResult.requestState())
+				if err != nil {
+					return nil, err
+				}
 			}
 		}
 	}
-}
-
-// multiRoundTripRetryRequest builds the request for the next round of a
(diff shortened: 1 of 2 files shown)


mcp/mrtr.go
```diff
@@ -108,12 +108,45 @@ func clientMultiRoundTripMiddleware() Middleware {
...
+		clone = &cp
+	case *GetPromptParams:
+		cp := *p
+		clone = &cp
+	case *ReadResourceParams:
+		cp := *p
+		clone = &cp
+	default:
+		// No retry params to carry; resend the request unchanged.
+		return req
```
guglielmo-san: This comment is misleading. This default case should never be reached in general

guglielmo-san: we should replace it with an error, or more comprehensive comment

mcp/mrtr.go
```diff
@@ -108,12 +108,45 @@ func clientMultiRoundTripMiddleware() Middleware {
...
+	}
+	retry := newClientRequest(cs, clone)
+	setMultiRoundTripRetryParams(retry, responses, state)
```
guglielmo-san: what do you think about merging the clone within the setMultiRoundTripRetryParams?
Even if applied to the server middleware it should not be a problem

mcp/mrtr.go
yhxlele: Done. The default case now returns an explicit error.

mcp/mrtr.go
yhxlele: I agree. Done. Method parameter introduced to build the request for the server session.

yhxlele:
@guglielmo-san Both review comments have been addressed (the clone is merged into `setMultiRoundTripRetryParams`, and the default case now returns an explicit error) and the branch has main merged in. Whenever you have a moment for another look, mainly want to make sure it isn't waiting on anything from my side.

guglielmo-san (approved):

guglielmo-san:
@yhxlele thank you for the contribution!

PR #1146, opened 2026-08-05
mcp: support custom notifications

### LINKED ISSUE #745: Expose generic `SendNotification` on `ServerSession` for custom protocol extensions

delaneyj:
Protocol extensions need to send and receive custom JSON-RPC notifications. The SDK session APIs currently support only standard notification methods.

This change adds:

- `SendNotification` to `ClientSession` and `ServerSession`.
- Generic `AddReceivingCustomNotification` registration for typed client handlers.
- `SendSubscriptionNotification` for custom notifications on `subscriptions/listen` streams.
- Automatic `io.modelcontextprotocol/subscriptionId` metadata on subscription notifications.
- Carrier cancellation for Streamable HTTP listen streams without a separate cancellation notification.
- Exact preservation of custom parameter objects.
- Empty object encoding for nil parameters.

Sending middleware processes custom notifications before transmission. Typed receiving handlers use parameter structs that embed `ParamsBase`.

The tests cover:

- Client-to-server custom notifications.
- Server-to-client custom notifications.
- Typed client-side receipt.
- Exact wire method names.
- Arbitrary parameter objects.
- Nil parameters.
- Subscription metadata.
- Streamable HTTP unsubscribe cleanup and subsequent client requests.

Validation:

```text
go test ./mcp -run '^(TestSendNotification|TestResourceSubscriptions_Streamable|TestResourceSubscriptions_InMemory)$'
go test ./...
```

This PR supersedes #844.

Fixes #745

### DIFF
--- a/mcp/server.go
+++ b/mcp/server.go
@@ -1470,6 +1470,23 @@ func (ss *ServerSession) SendNotification(ctx context.Context, method string, pa
 	)
 }
 
+// SendSubscriptionNotification sends a custom notification on the
+// subscriptions/listen stream represented by ctx. It adds the subscription ID
+// to the notification metadata.
+func (ss *ServerSession) SendSubscriptionNotification(ctx context.Context, method string, params any) error {
+	requestID, ok := ctx.Value(idContextKey{}).(jsonrpc.ID)
+	if !ok || !requestID.IsValid() {
+		return fmt.Errorf("mcp: SendSubscriptionNotification: context has no subscription ID")
+	}
+	customParams := &customNotificationParams{payload: params}
+	injectMetaSubscriptionID(customParams, requestID)
+	return handleNotify(
+		ctx,
+		"x-notifications/"+method,
+		newServerRequest(ss, Params(customParams)),
+	)
+}
+
 // notifySubscriptionAcked sends a "notifications/subscriptions/acknowledged"
 // notification on the listen stream represented by this session, indicating
 // the subscription filter the server accepted (SEP-2575).
--- a/mcp/server.go
+++ b/mcp/server.go
@@ -1460,6 +1460,16 @@ func (ss *ServerSession) NotifyProgress(ctx context.Context, params *ProgressNot
 	return handleNotify(ctx, notificationProgress, newServerRequest(ss, orZero[Params](params)))
 }
 
+// SendNotification sends a custom notification to the client associated with
+// this session. It supports protocol extensions such as notifications/foobar/stats.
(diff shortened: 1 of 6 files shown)


delaneyj:
Hello?  It's been nearly a month with no response or human interaction.  Are PRs not wanted here?

guglielmo-san:
Hi @delaneyj, thank you for the patience, we will review the PR as soon as possible

bencroker:
Thanks @guglielmo-san, looking forward to having this functionality.

mcp/server.go
```diff
@@ -1463,6 +1463,33 @@ func (ss *ServerSession) NotifyProgress(ctx context.Context, params *ProgressNot
...
+	}
+	customParams := &customNotificationParams{payload: params}
+	injectMetaSubscriptionID(customParams, requestID)
+	return handleNotify(
+		ctx,
+		"x-notifications/"+method,
+		newServerRequest(ss, Params(customParams)),
+	)
+}
+
```
guglielmo-san, line 1492: The listen stream is a closed opt-in set
From schema/2026-07-28/schema.ts (https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/schema/2026-07-28/schema.ts), SubscriptionFilter (L1270):
> Each notification type is opt-in; the server MUST NOT send notification types the client has not explicitly requested here.

Also it is not clear how this would send a notification using the subscriptions/listen stream

PR #1157, opened 2026-08-10
mcp: close the session when Connect gets an unsupported protocol version

### LINKED ISSUE #1154: Client.Connect leaks the ClientSession when the server answers with an unsupported protocol version
**Describe the bug**

In `go-sdk v1.7.0`, `Client.Connect` returns `unsupportedProtocolVersionError` without closing the `ClientSession` it has already built, and hands the caller `nil` for the session — so nothing outside the SDK can close it either.

The two adjacent error paths in the same function both call `cs.Close()`. Only this one does not (`mcp/client.go`, in `Connect`):

```go
res, err := handleSend[*InitializeResult](ctx, methodInitialize, req)
if err != nil {
    _ = cs.Close()                                       // closes
    return nil, err
}
if !slices.Contains(supportedProtocolVersions, res.ProtocolVersion) {
    return nil, unsupportedProtocolVersionError{res.ProtocolVersion}   // does not close
}
cs.state.InitializeResult = res
...
if err := handleNotify(ctx, notificationInitialized, req2); err != nil {
    _ = cs.Close()                                       // closes
    return nil, err
}
```

Impact is per attempt, not per process: a client that retries against a server answering this way leaks a reader goroutine and its HTTP connection every time, and on a stateful server the session is never deleted.

**To Reproduce**

The fixture refuses `server/discover` to push the client onto the legacy `initialize` path, then answers `initialize` with a protocol version no revision defines. It records whether a `DELETE` — the client saying "I am done with this session" — ever arrives.

```go
package repro

import (
	"context"
	"encoding/json"
	"io"
	"net/http"
	"net/http/httptest"
	"sync/atomic"
	"testing"

)

func fixture(t *testing.T, sawDelete *atomic.Bool) string {
	t.Helper()
	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		if r.Method == http.MethodDelete {
			sawDelete.Store(true)
			return
		}
		body, err := io.ReadAll(r.Body)
		if err != nil {
			http.Error(w, "read", http.StatusInternalServerError)
			return
		}
		var req struct {
			ID     json.RawMessage `json:"id"`
			Method string          `json:"method"`
		}
		if json.Unmarshal(body, &req) != nil {
			http.Error(w, "bad request", http.StatusBadRequest)
			return
		}
		w.Header().Set("Content-Type", "application/json")
		w.Header().Set("Mcp-Session-Id", "leaky-session")
		enc := json.NewEncoder(w)
		switch req.Method {
		case "server/discover":
			_ = enc.Encode(map[string]any{
				"jsonrpc": "2.0", "id": req.ID,
				"error": map[string]any{"code": -32601, "message": "method not found"},
			})
		case "initialize":
			_ = enc.Encode(map[string]any{
				"jsonrpc": "2.0", "id": req.ID,
				"result": map[string]any{
					"protocolVersion": "1999-01-01",
					"capabilities":    map[string]any{},
					"serverInfo":      map[string]any{"name": "fixture", "version": "1"},
				},
			})
		default:
			_ = enc.Encode(map[string]any{"jsonrpc": "2.0", "id": req.ID, "result": map[string]any{}})
		}
	}))
	t.Cleanup(ts.Close)
	return ts.URL
}

func TestConnectLeaksTheSessionOnAnUnsupportedProtocolVersion(t *testing.T) {
	var sawDelete atomic.Bool
	url := fixture(t, &sawDelete)

	client := mcp.NewClient(&mcp.Implementation{Name: "repro"}, nil)
	session, err := client.Connect(context.Background(), &mcp.StreamableClientTransport{Endpoint: url}, nil)
	if err == nil {
		t.Fatal("Connect succeeded against an unsupported protocol version")
	}
	t.Logf("Connect error: %v", err)
	t.Logf("session handed back to the caller: %v", session)

	if !sawDelete.Load() {
		t.Error("no DELETE reached the server — the session the SDK built is still open, " +
			"and Connect returned no handle to close it")
	}
}
```

Output on v1.7.0:

```
    repro_test.go:78: Connect error: unsupported protocol version: "1999-01-01"
    repro_test.go:79: session handed back to the caller: <nil>
    repro_test.go:86: no DELETE reached the server — the session the SDK built is still open, and Connect returned no handle to close it
--- FAIL: TestConnectLeaksTheSessionOnAnUnsupportedProtocolVersion (0.01s)
```

**Expected behavior**

The branch closes the session before returning, like its two neighbours — `_ = cs.Close()` ahead of the `return`.

**Workaround, for anyone hitting this before a release**

It is fixable from outside, but only awkwardly, and the shape is not obvious. `Transport` is a one-method interface, so a wrapper can keep the `Connection` its inner transport produced and close that when `Connect` fails:

```go
type capturingTransport struct {
	inner mcp.Transport
	conn  mcp.Connection
}

func (t *capturingTransport) Connect(ctx context.Context) (mcp.Connection, error) {
	conn, err := t.inner.Connect(ctx)
	t.conn = conn
	return conn, err
}
```

Two details matter. The wrapper must return the SDK's own `Connection` **unchanged** — the SDK type-asserts it to `clientConnection` to reach `sessionUpdated`, so substituting a wrapper there breaks version negotiation. And `Connection.Close` is idempotent, so closing on the error path is safe against the branches that do close.

**Additional context**

- `go-sdk` v1.7.0 (current latest release), Go 1.26, darwin/arm64.
- Distinct from #1136, which is a `ServerSession` leak on the server side; this one is client-side.
- Found while building an MCP client whose connections are per-work-item, where a server answering this way to every attempt makes the leak accumulate rather than stay a one-off.

n0liu:
Fixes #1154.

`Client.Connect` returned `unsupportedProtocolVersionError` without closing the `ClientSession` it had already built, and returned `nil` to the caller — so nothing outside the SDK could close it either. The two adjacent error paths in the same function both call `cs.Close()`; this one now does too.

Without the fix, a client that retries against a server answering this way leaks a reader goroutine and its connection on every attempt, and on a stateful server the session is never deleted.

Added a test that fails Connect on an unsupported protocol version and asserts the session's connection is closed. `go test ./mcp/ -race`, `gofmt`, and `go vet` pass.

### DIFF
--- a/mcp/client.go
+++ b/mcp/client.go
@@ -389,6 +389,7 @@ func (c *Client) Connect(ctx context.Context, t Transport, opts *ClientSessionOp
 		return nil, err
 	}
 	if !slices.Contains(supportedProtocolVersions, res.ProtocolVersion) {
+		_ = cs.Close()
 		return nil, unsupportedProtocolVersionError{res.ProtocolVersion}
 	}
 	cs.state.InitializeResult = res
--- a/mcp/client_leak_test.go
+++ b/mcp/client_leak_test.go
@@ -1,0 +1,74 @@
+// Copyright 2025 The Go MCP SDK Authors. All rights reserved.
+// Use of this source code is governed by an MIT-style
+// license that can be found in the LICENSE file.
+
+package mcp
+
+import (
+	"context"
+	"sync/atomic"
+	"testing"
+
+	"github.com/modelcontextprotocol/go-sdk/internal/jsonrpc2"
+)
+
+// TestClientConnectClosesSessionOnUnsupportedVersion verifies that a failed
+// Connect does not leak the ClientSession: when the server answers with an
+// unsupported protocol version, the session's connection must be closed.
+func TestClientConnectClosesSessionOnUnsupportedVersion(t *testing.T) {
+	ctx := context.Background()
+
+	s := NewServer(testImpl, nil)
+	s.AddReceivingMiddleware(func(next MethodHandler) MethodHandler {
+		return func(ctx context.Context, method string, req Request) (Result, error) {
+			switch method {
+			case methodDiscover:
+				return nil, jsonrpc2.ErrMethodNotFound // force the legacy initialize path
+			case methodInitialize:
+				return &InitializeResult{ProtocolVersion: "1999-01-01", ServerInfo: testImpl}, nil
+			}
+			return next(ctx, method, req)
+		}
+	})
+
+	ct, st := NewInMemoryTransports()
+	ss, err := s.Connect(ctx, st, nil)
+	if err != nil {
+		t.Fatalf("server Connect: %v", err)
+	}
+	defer ss.Close()
+
+	rt := &closeRecordingTransport{Transport: ct}
+	c := NewClient(testImpl, nil)
+	cs, err := c.Connect(ctx, rt, &ClientSessionOptions{ProtocolVersion: protocolVersion20260728})
+	if err == nil {
+		_ = cs.Close()
+		t.Fatal("Connect succeeded, want an unsupported protocol version error")
+	}
+	if !rt.closed.Load() {
+		t.Error("Connect failed but did not close the session's connection (leak)")
+	}
+}
+
+type closeRecordingTransport struct {
+	Transport
+	closed atomic.Bool
+}
+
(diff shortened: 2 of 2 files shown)


guglielmo-san:
Closing as duplicate of #1158

Issue #1162, opened 2026-08-13
setMCPHeaders prefers the request context over the negotiated protocol version since #1115


gwyszynski-sa:
**Describe the bug**

`streamableClientConn.setMCPHeaders` resolves `Mcp-Protocol-Version` from the request context before `c.initializedResult` ([streamable.go#L2447-L2453](https://github.com/modelcontextprotocol/go-sdk/blob/main/mcp/streamable.go#L2447-L2453)), so a connection sends a version other than the one it negotiated.

This appears unintended. #1107 established the order `message -> initializedResult -> context` and stated it in its description ("falling back to `c.initializedResult` and then the request context"); the code it replaced also consulted the context last. The two branches were transposed three days later by the `switch`-to-`if` rewrite in #1115, whose subject was an unrelated nil-pointer check:

```diff
-	switch {
-	case protocolVersionFromMessage(msg) != "":
-		req.Header.Set(protocolVersionHeader, protocolVersionFromMessage(msg))
-	case c.initializedResult != nil:
+	if pv := protocolVersionFromMessage(msg); pv != "" {
+		req.Header.Set(protocolVersionHeader, pv)
+	} else if pv := protocolVersionFromContext(req.Context()); pv != "" {
+		req.Header.Set(protocolVersionHeader, pv)
+	} else if c.initializedResult != nil {
 		req.Header.Set(protocolVersionHeader, c.initializedResult.ProtocolVersion)
-	case protocolVersionFromContext(req.Context()) != "":
-		req.Header.Set(protocolVersionHeader, protocolVersionFromContext(req.Context()))
 	}
```

| | order |
|---|---|
| before #1107 | `initializedResult` -> context |
| #1107 (Jul 24) | message -> `initializedResult` -> context |
| **#1115 (Jul 27)** | message -> **context** -> `initializedResult` |
| main today | unchanged since #1115 |

**To Reproduce**

1. Give a `streamableClientConn` a negotiated version, and put a different version on the outgoing request's context — which is what the server transport leaves there at [streamable.go#L375](https://github.com/modelcontextprotocol/go-sdk/blob/main/mcp/streamable.go#L375) for a process that is both an MCP server and an MCP client:

```go
conn := &streamableClientConn{
	initializedResult: &InitializeResult{ProtocolVersion: protocolVersion20251125},
}
req, _ := http.NewRequest(http.MethodPost, "http://example.invalid", nil)
[...]

PR #1165, opened 2026-08-13
mcp: clear the caller's protocol version for the initialize handshake

### LINKED ISSUE #1164: initialize sends no Mcp-Protocol-Version of its own, so the header can contradict params.protocolVersion
**Describe the bug**

The legacy `initialize` handshake pins the version it proposes into the body but sends no `Mcp-Protocol-Version` header of its own ([client.go#L376-L386](https://github.com/modelcontextprotocol/go-sdk/blob/main/mcp/client.go#L376-L386)):

```go
// Use the latest legacy protocol version for the fallback initialize.
protocolVersion = protocolVersion20251125
...
params := &InitializeParams{ProtocolVersion: protocolVersion, ...}
req := &InitializeRequest{Session: cs, Params: params}
res, err := handleSend[*InitializeResult](ctx, methodInitialize, req)
```

`handleSend` gets the plain `ctx`, not the `discoverCtx` used for the probe above it. By the time `setMCPHeaders` runs, `c.initializedResult` is still nil — it is assigned about six lines later — and the message carries no `_meta.protocolVersion`, so the header is decided by whatever the request context happens to hold.

For a standalone client that context is empty and the header is absent. For a process that is both an MCP server and an MCP client, the server transport has recorded its inbound request's version on that context ([streamable.go#L375](https://github.com/modelcontextprotocol/go-sdk/blob/main/mcp/streamable.go#L375)) and that value goes out instead, contradicting the body it accompanies.

**To Reproduce**

1. Run a Go SDK Streamable HTTP client against a server that does not implement `server/discover`, so it falls back to the legacy `initialize`.
2. Observe the header on that `initialize` request: absent for a standalone client, or the inbound version if the caller's context carries `protocolVersionContextKey`.

The same disagreement can be produced directly against a Go SDK server, as in #963:

```bash
curl -i -sS -X POST "$ENDPOINT" \
  -H 'MCP-Protocol-Version: 2026-07-28' \
  --data '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2025-11-25","capabilities":{},"clientInfo":{"name":"repro","version":"0.1.0"}}}'
```

**What did you see**

```
initialize  body="2025-11-25"  header=""            (standalone client)
initialize  body="2025-11-25"  header="2026-07-28"  (server+client process)
```

**What did you expect to see**

The header stating the version the request proposes, matching `params.protocolVersion`, and never inherited from ambient context:

```
initialize  body="2025-11-25"  header="2025-11-25"
```

Omitting the header is the other reasonable option, and is what the package doc for the header currently describes ("Is omitted before initialization completes", [streamable_client.go](https://github.com/modelcontextprotocol/go-sdk/blob/main/mcp/streamable_client.go)). Sending it seems preferable: it agrees with the body, it is what a server validating the header expects, and it removes the dependency on the context being empty rather than relying on it. Happy to switch the fix to omission if maintainers prefer that reading.

**Logs**

An upstream that validates the header rejects the handshake outright:

```
400 Bad Request: Unsupported protocol version: 2026-07-28 (supported versions: 2025-11-25, 2025-06-18, 2025-03-26, 2024-11-05, 2024-10-07)
```

**What version of the Go MCP SDK are you using**

`v1.7.0`, and `main` at 64e454e.

**What version of Go are you using**

`go version go1.26.5 darwin/arm64`. Not Go-version dependent.

**Additional context**

This is the client-side mirror of #963, which tightened the *server* against exactly this mismatch — the SDK's server now rejects a disagreement its own client can still produce.

Separate from #1162: fixing the precedence there does not fix this, because `initializedResult` is nil at this point regardless of ordering, so control still reaches the context branch.

The fake server in `streamable_client_test.go` asserts `wantProtocolVersion` for other methods but for none of its sixteen `initialize` cases, so neither an absent nor a contradicting header is currently covered.

I have a fix and have opened #1165.

gwyszynski-sa:
The legacy initialize handshake pins params.protocolVersion but sends no
Mcp-Protocol-Version header of its own: setMCPHeaders finds c.initializedResult
still nil, since it is assigned a few lines later, and the message carries no
_meta.protocolVersion, so the header falls to the request context. For a
standalone client that context is empty and the header is absent; for a process
that is both a server and a client it holds that process's inbound version,
which contradicts the body it accompanies. This is the client-side mirror of
#963.

No version is negotiated at that point, so the request has none of its own to
state, and the caller's must not stand in for one. Clears it for the initialize
send rather than sending the proposed version: a header the peer does not
support fails the handshake outright, where the body would have negotiated down.
Standalone clients are unaffected, the header having already been absent there.

The fake server asserted wantProtocolVersion for other methods but for none of
its sixteen initialize cases. Asserts centrally that initialize carries no
protocol version header, and adds a Connect whose context already holds one,
without which that assertion cannot fail. Requests after negotiation are left
unchecked there, since which source wins once initializedResult is set is #1162.

Fixes #1164

### DIFF
--- a/mcp/client.go
+++ b/mcp/client.go
@@ -383,7 +383,12 @@ func (c *Client) Connect(ctx context.Context, t Transport, opts *ClientSessionOp
 		Capabilities:    c.capabilities(protocolVersion),
 	}
 	req := &InitializeRequest{Session: cs, Params: params}
-	res, err := handleSend[*InitializeResult](ctx, methodInitialize, req)
+	// No version is negotiated yet, so this request has no version of its own to
+	// state; a caller's value must not stand in for one. Cleared rather than sent,
+	// since a header a peer does not support fails the handshake the body would
+	// have negotiated down.
+	initializeCtx := context.WithValue(ctx, protocolVersionContextKey{}, "")
+	res, err := handleSend[*InitializeResult](initializeCtx, methodInitialize, req)
 	if err != nil {
 		_ = cs.Close()
 		return nil, err
--- a/mcp/streamable_client_test.go
+++ b/mcp/streamable_client_test.go
@@ -145,6 +145,14 @@ func (s *fakeStreamableServer) ServeHTTP(w http.ResponseWriter, req *http.Reques
 	if v := req.Header.Get(protocolVersionHeader); v != resp.wantProtocolVersion && resp.wantProtocolVersion != "" {
 		s.t.Errorf("%v: bad protocol version header: got %q, want %q", key, v, resp.wantProtocolVersion)
 	}
+	// On initialize no version has been negotiated yet, so the request has none of
+	// its own to state and must not inherit the caller's. Checked for every
+	// initialize rather than per-response, since it holds unconditionally.
+	if key.jsonrpcMethod == methodInitialize {
+		if got := req.Header.Get(protocolVersionHeader); got != "" {
+			s.t.Errorf("%v: initialize carried protocol version header %q, want none", key, got)
+		}
+	}
 	w.Write([]byte(body))
 	rc.Flush() // flush response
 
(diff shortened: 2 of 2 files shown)


mcp/client.go
```diff
@@ -383,7 +383,10 @@ func (c *Client) Connect(ctx context.Context, t Transport, opts *ClientSessionOp
 		Capabilities:    c.capabilities(protocolVersion),
 	}
 	req := &InitializeRequest{Session: cs, Params: params}
-	res, err := handleSend[*InitializeResult](ctx, methodInitialize, req)
+	// The header must agree with params.protocolVersion: no version has been
+	// negotiated yet, so nothing else identifies what this request proposes.
+	initializeCtx := context.WithValue(ctx, protocolVersionContextKey{}, protocolVersion)
```
guglielmo-san: this might return an error when connecting to clients that do not support the protocolVersion20251125 protocolVersion. Maybe in this case the best option is just to clear out the protocolVersion in the context

gwyszynski-sa: You're right, and thanks — this would have broken connecting to any peer whose supported set predates `2025-11-25`.

The mechanism is the header validation in `StreamableHTTPHandler.ServeHTTP`: the guard is `protocolVersion != ""`, so an absent header skips validation and the body negotiates down, while a present-but-unsupported one is a 400 before negotiation happens. Sending the proposed version turned a negotiable proposal into a precondition — the same shape as the bug this was meant to fix.

Confirmed against a peer supporting only `2025-06-18` and older, which negotiates down via the body but validates the header as the SDK's server does:

| initialize header | result |
|---|---|
| `2025-11-25` (as proposed) | `Connect: sending "initialize": Bad Request` |
| cleared | connects, negotiates `2025-06-18` |

Switched to clearing it, and retitled accordingly. Standalone clients are byte-identical to before, since the header was already absent for them; only the inherited value changes, which is the case this is about.

One thing worth flagging: with the header simply absent, asserting "initialize carries no protocol version header" is vacuous against the existing fixtures — I reverted the fix and the whole suite still passed, because nothing puts a version on the caller's context. So I added `TestStreamableClientConnect_InitializeDoesNotInheritProtocolVersion`, a `Connect` whose context already holds one; without the fix it fails with `initialize carried protocol version header "2026-07-28", want none`. Requests after negotiation are left unchecked there, since which source wins once `initializedResult` is set is #1162.

PR #1167, opened 2026-08-13
mcp: inject new-protocol metadata in default send path

jstar0:
Several client-originated requests in the >= 2026-07-28 protocol still relied on individual API methods to attach the per-request `_meta` fields. That left send paths such as notifications and registered custom methods without `protocolVersion`, `clientInfo`, and `clientCapabilities`, so modern servers treated them as legacy or uninitialized requests.

This changes the default client send handler to clone outgoing params and inject missing request metadata for modern sessions, while keeping the legacy initialize, ping, logging, root-subscription, and resource subscribe compatibility methods on their existing path. `CallCustomMethod` now relies on the shared send path, so caller params are not mutated just to add metadata before serialization.

Tests:

```
go test ./mcp -run "Test(ClientNotifyProgressInjectsMetaOnNewProtocol|CallCustomMethodInjectsMetaOnNewProtocol|CallCustomMethodTypedNilParams|ClientConnectDiscover_RequestContents|ClientConnectDiscover_UnsupportedVersionNegotiation|InMemory_E2E_Discover|StreamableClient.*Discover|Streamable.*PerRequest|ValidateRequestMeta|ServerRequest_PerRequestAccessors)" -count=1
go test ./mcp -skip "Test(SSELocalhostProtection|StreamableLocalhostProtection)" -count=1
```

Full `go test ./...` currently fails only the existing localhost-protection cases `TestSSELocalhostProtection/0.0.0.0_via_localhost_rejects_evil.com` and `TestStreamableLocalhostProtection/0.0.0.0_via_localhost_rejects_evil.com` in this environment; all other packages pass.

### DIFF
--- a/mcp/client.go
+++ b/mcp/client.go
@@ -515,21 +515,26 @@ func (cs *ClientSession) usesNewProtocol() bool {
 	return res != nil && res.ProtocolVersion >= protocolVersion20260728
 }
 
-// injectRequestMeta populates the SEP-2575 per-request `_meta` fields
-// (protocolVersion, optional clientInfo, clientCapabilities) on the given
-// outgoing request params. Keys already present in params.Meta are not
-// overwritten. Per PR modelcontextprotocol/modelcontextprotocol#3002
-// clientInfo is SHOULD (not MUST), and is omitted when the client has no
-// [Implementation] configured.
-func injectRequestMeta[T any, P interface {
-	*T
-	Params
-}](cs *ClientSession, params P) P {
-	res := cs.state.InitializeResult
-	if params == nil {
-		params = new(T)
+func clientMethodUsesLegacyCompatPath(method string) bool {
+	switch method {
+	case methodPing,
+		methodSetLevel,
+		notificationRootsListChanged,
+		methodSubscribe,
+		methodUnsubscribe:
+		return true
+	default:
+		return false
 	}
+}
+
+func injectRequestMetaParams(cs *ClientSession, params Params) Params {
+	if params == nil {
+		return nil
+	}
+	res := cs.state.InitializeResult
 	m := params.GetMeta()
+	m = maps.Clone(m)
 	if m == nil {
 		m = map[string]any{}
 	}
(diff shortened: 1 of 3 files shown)


mcp/client.go
```diff
@@ -1674,9 +1695,6 @@ func CallCustomMethod[P paramsPtr[PT], R Result, PT any](
 		var zero R
 		return zero, fmt.Errorf("mcp: CallCustomMethod: %q is not registered; call AddSendingCustomMethod first", method)
 	}
-	if cs.usesNewProtocol() {
-		params = injectRequestMeta(cs, params)
-	}
```
guglielmo-san, line 1679: why not removing this call from all the other client send methods

jstar0:
Addressed the question on current head `d5d0a9e`.

- Standard client methods no longer call the request-metadata helper themselves; the default sending handler now owns clone-and-inject for every modern non-legacy client request, including registered custom methods.
- The handler preserves legacy/compatibility wire behavior by converting typed-nil optional params back to an omitted params field, while modern typed-nil params are allocated before metadata injection.
- List caches now capture cursors before sending, and `ReadResource` captures its URI before sending, so removing method-level mutation does not introduce nil-parameter panics.

Verification:
- Focused metadata, typed-nil, discovery, and Streamable HTTP tests pass.
- Full `go test ./... -count=1` passes except the two existing localhost-protection cases for `0.0.0.0 via localhost` in this environment; no other failures remain.
- `gofmt` and `git diff --check` pass.

Please review the current head when convenient.

jstar0:
Rebased and pushed the centralized client request metadata injection fix at head `6df1f565` (SSH-signed and verified).

The current head is based on upstream `main@b0408f2`. Focused metadata, typed-nil, discovery, and Streamable HTTP tests pass; the full suite has no failures beyond the two existing `0.0.0.0 via localhost` protection cases in this environment. Hosted checks are running on the new head.

mcp/client.go
jstar0: Good point. I removed the per-method injection calls in 6df1f565 and moved the clone/inject step into `defaultSendingMethodHandler`. That now covers standard requests, notifications, and custom methods from one path, while the explicit compatibility-method gate preserves the legacy typed-nil wire behavior.

PR #1170, opened 2026-08-13
mcp: fail connect when subscriptions listen is rejected

### LINKED ISSUE #1169: mcp: Client.Connect succeeds after subscriptions/listen is rejected
When a v2026-07-28 streamable HTTP client has notification handlers installed, `Client.Connect` opens `subscriptions/listen` after `server/discover`. If the server immediately rejects that `subscriptions/listen` request with a JSON-RPC error, `Connect` currently still returns a usable session.

What I did:

I added a regression test with a fake streamable server that returns a successful `server/discover` response, then returns a JSON-RPC `InvalidParams` error for `subscriptions/listen`. The client is configured with `ToolListChangedHandler` so it attempts the listen call during connect.

What I saw:

On current `origin/main` at `64e454e35c23`, the regression test fails because `Connect` succeeds even though the listen request was rejected.

What I expected:

`Connect` should fail and surface the `subscriptions/listen` error. Otherwise callers believe notification subscriptions are active when the server already rejected them.

Version:

`origin/main` at `64e454e35c23`

Go version:

`go version go1.26.5 darwin/arm64`

jstar0:
Fixes #1169

When a v2026-07-28 streamable HTTP client installs notification handlers, `Client.Connect` opens `subscriptions/listen` after `server/discover`. If that listen call is rejected immediately, the client previously ignored the response and returned a session even though subscriptions were not active.

This change lets the listen helper observe an immediate terminal response before detaching the long-lived call. Rejections now propagate through `Connect`, and the partially opened session is closed on that connect error.

Verification:

- Added `TestStreamableClientConnect_SubscriptionsListenError`; the test fails on `origin/main` because `Connect` succeeds after the rejected listen request
- `go test ./mcp -run 'TestStreamableClientConnect_SubscriptionsListenError|TestStreamableClientConnect_DiscoverSuccess|TestSubscriptionsListen_Streamable|TestResourceSubscriptions_Streamable|TestSubscriptionsListen_InMemory|TestResourceSubscriptions_InMemory' -count=1 -v` passed
- `go test ./...` currently fails in this local environment only on the existing localhost protection cases where `evil.com` resolves to `0.0.0.0`; the same two subtests fail on clean `origin/main`: `TestSSELocalhostProtection/0.0.0.0_via_localhost_rejects_evil.com` and `TestStreamableLocalhostProtection/0.0.0.0_via_localhost_rejects_evil.com`

### DIFF
--- a/mcp/transport.go
+++ b/mcp/transport.go
@@ -256,22 +256,30 @@ func (c *canceller) Preempt(ctx context.Context, req *jsonrpc.Request) (result a
 	return nil, jsonrpc2.ErrNotHandled
 }
 
-// callSubscriptionsListen issues a "subscriptions/listen" call (SEP-2575)
-// without awaiting its JSON-RPC response. The call's logical lifetime is the
-// stream of notifications that follow on the same channel — the empty
-// response, if ever delivered, only marks subscription teardown — so the
-// caller has nothing useful to block on.
+// callSubscriptionsListen issues a "subscriptions/listen" call (SEP-2575).
+// If the call is accepted, its logical lifetime is the stream of notifications
+// that follow on the same channel. The empty response, if ever delivered, only
+// marks subscription teardown, so the caller has nothing useful to block on.
 //
 // Cancellation is driven by ctx: when it is cancelled, a background goroutine
 // sends a "notifications/cancelled" notification referencing the listen's
 // request ID and retires the call from the connection's outgoing-calls map.
-func callSubscriptionsListen(ctx context.Context, conn *jsonrpc2.Connection, method string, params Params) {
+func callSubscriptionsListen(ctx context.Context, conn *jsonrpc2.Connection, method string, params Params) error {
 	call := conn.Call(ctx, method, params)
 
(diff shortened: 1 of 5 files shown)


mcp/transport.go
```diff
@@ -256,22 +256,30 @@ func (c *canceller) Preempt(ctx context.Context, req *jsonrpc.Request) (result a
...
 	call := conn.Call(ctx, method, params)
 
+	select {
+	case <-call.Done():
+		return call.Await(context.Background(), nil)
+	case <-ctx.Done():
+		_ = cancelCall(ctx, conn, call)
+		return nil
+	default:
+	}
```
guglielmo-san, line 277: conn.Call writes the outbound request synchronously but does not wait for a response — the response is delivered asynchronously by the jsonrpc2 reader goroutine. There's no guarantee that call.Done() is closed by the time this select runs.
For any transport with non-trivial latency (i.e. anything but in-process), the default case will fire and the caller will observe success even if the server subsequently rejects the request.

PR #1180, opened 2026-08-16
internal/jsonrpc2: contain handler panics and reply with an internal error

ez-lbz:
## Summary

A panicking request handler currently escapes the per-request goroutine in `handleAsync` and terminates the entire server process, taking every session down with it. Handlers are user code operating on attacker-controlled input (e.g. `tools/call` arguments), so a single malformed request can crash the server remotely.

This change recovers panics in the handler goroutine and replies with an internal error (`-32603`) instead, matching the isolation behavior of the other official SDKs:

- New `ErrPanic` sentinel distinguishes recovered panics from ordinary handler errors.
- The recovered panic is converted to a JSON-RPC internal error via `processResult`, so the client receives code `-32603` and the connection stays usable.

## Tests

- Added `TestServerToolPanic` in `mcp/error_test.go`: a panicking tool handler yields a `-32603` error to the client, and a subsequent `tools/call` on the same connection succeeds.

### DIFF
--- a/internal/jsonrpc2/conn.go
+++ b/internal/jsonrpc2/conn.go
@@ -681,6 +681,13 @@ func (c *Connection) handleAsync() {
 		releaser := &releaser{ch: make(chan struct{})}
 		ctx := context.WithValue(req.ctx, asyncKey, releaser)
 		go func() {
+			defer func() {
+				if r := recover(); r != nil {
+					// A panicking handler must not take down the connection or the
+					// process; reply with an internal error instead.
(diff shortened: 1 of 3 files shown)


yarolegovich:
Recovering a panic mid-mutation of shared state can leave logical invariants broken. We can't make assumptions about the user code, so a crash and restart is actually safer as a reset to a known-good baseline.

User-side mitigations exist with either middleware or per-handler recovery.

PR #1186, opened 2026-08-19
jsonrpc2: wrap write error in server-closing error

### LINKED ISSUE #1098: jsonrpc2: server-closing error formats writeErr with %v, so `errors.Is(err, io.EOF)` is false for consumers

Tethys0:
The "server is closing" error produced when a connection shuts down because its write side failed formatted the underlying write error with `%v`, so the cause - typically `io.EOF` when a stdio host closes its pipe - was not part of the error chain. A caller that wants to distinguish a clean host disconnect from a real failure via `errors.Is(err, io.EOF)` therefore always got `false`.

This changes both `s.writeErr` wrappings to use `%w`, preserving the write error in the error chain while keeping the same message text ("server is closing: EOF").

A regression test exercises the write-failure branch of `shuttingDown` and asserts both `errors.Is(err, ErrServerClosing)` and `errors.Is(err, io.EOF)`.

Fixes #1098

### DIFF
--- a/internal/jsonrpc2/conn.go
+++ b/internal/jsonrpc2/conn.go
@@ -158,7 +158,7 @@ func (s *inFlightState) shuttingDown(errClosing error) error {
 	if s.readErr != nil {
 		// If the read side of the connection is broken, we cannot read new call
 		// requests, and cannot read responses to our outgoing calls.
-		return fmt.Errorf("%w: %v", errClosing, s.readErr)
+		return fmt.Errorf("%w: %w", errClosing, s.readErr)
 	}
 	if s.writeErr != nil {
 		// If the write side of the connection is broken, we cannot write responses
--- a/internal/jsonrpc2/conn.go
+++ b/internal/jsonrpc2/conn.go
@@ -163,7 +163,7 @@ func (s *inFlightState) shuttingDown(errClosing error) error {
 	if s.writeErr != nil {
 		// If the write side of the connection is broken, we cannot write responses
 		// for incoming calls, and cannot write requests for outgoing calls.
-		return fmt.Errorf("%w: %v", errClosing, s.writeErr)
+		return fmt.Errorf("%w: %w", errClosing, s.writeErr)
 	}
 	return nil
 }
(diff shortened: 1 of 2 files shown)


internal/jsonrpc2/conn.go
guglielmo-san: we should use the same approach for both cases

Tethys0: Applied the same %w wrapping to the read-shutdown path in aecb26d, with a focused errors.Is regression test. The pushed branch is ready for re-review.

guglielmo-san (dismissed):

nahapetyan-serob (approved):

PR #1187, opened 2026-08-20
auth: validate unadvertised RFC 9207 iss instead of rejecting it

### LINKED ISSUE #1152: OAuth: a correct but unadvertised RFC 9207 iss fails the whole sign-in (§2.4 asks for validation, not rejection)
**Describe the bug**

`AuthorizationCodeHandler` fails the entire sign-in when an authorization server returns the RFC 9207 `iss` parameter without advertising `authorization_response_iss_parameter_supported` in its metadata — even when the `iss` value is correct.

This happens *after* the user has completed the browser step, so the failure is maximally annoying: consent has been given, an authorization code has been issued, and the exchange never runs.

`auth/authorization_code.go` (v1.7.0), in `validateIssuerResponse`:

```go
} else {
	if iss != "" {
		return fmt.Errorf("authorization server does not advertise RFC 9207 iss parameter support but iss was received in the authorization response")
	}
}
```

**To Reproduce**

1. Connect to an MCP server whose authorization server returns `iss` in the authorization response but omits `authorization_response_iss_parameter_supported` from its `/.well-known/oauth-authorization-server` document. `https://mcp.sentry.dev/mcp` does this today.
2. Complete the browser sign-in.
3. The connection fails.

`validateIssuerResponse` is unexported, so the smallest reproduction is a unit test inside the `auth` package:

```go
func TestUnadvertisedIssuerIsValidatedNotRejected(t *testing.T) {
	const issuer = "https://mcp.example.com"

	// Fails today: the value is correct and is refused for not having been advertised.
	if err := validateIssuerResponse(issuer, issuer, false); err != nil {
		t.Errorf("correct issuer rejected because it was not advertised: %v", err)
	}
	// Should keep failing: this is the mix-up case the parameter exists for.
	if err := validateIssuerResponse("https://attacker.example", issuer, false); err == nil {
		t.Error("a mismatched issuer must be rejected whether or not it was advertised")
	}
}
```

**Expected behavior**

When `iss` is present, compare it with the expected issuer and reject on mismatch — regardless of the metadata flag. When the server advertised support and no `iss` arrives, keep the existing error.

RFC 9207 §2.4 conditions the client obligation on the parameter being *present*, not on it having been advertised: clients "MUST extract the value of the iss parameter from authorization responses they receive if the parameter is present", compare it with the issuer of the server the request was sent to, and reject the response if it does not match.

The duty to advertise sits on the server, in §2.3. A server that returns `iss` without advertising it is not meeting that duty, but the client-side consequence described in §2.4 is to validate the value that arrived, not to refuse the exchange. §2.4 also notes that RFC 6749 §4.1.2 already requires clients which do *not* support RFC 9207 to ignore an unrecognised `iss` — so failing closed here is stricter than both specifications, and strictly less useful than validating: the current code discards a value it could have checked.

**Additional context**

Observed against three hosted MCP servers on 2026-08-07. None advertises the parameter:

| server | `authorization_response_iss_parameter_supported` |
|---|---|
| `https://mcp.sentry.dev` | absent |
| `https://mcp.linear.app` | absent |
| `https://mcp.notion.com` | absent |

Sentry additionally returns `iss` in the authorization response, so signing in to it through this SDK is currently impossible. The other two are affected the moment they start sending one.

Our workaround, in case it is useful to anyone hitting this before it is fixed: we resolve the issuer ourselves from the protected-resource and authorization-server metadata, compare the `iss` from the redirect against it in our own callback handler (so the RFC 9207 comparison still happens), and then withhold the value from the `AuthorizationResult` we return from `AuthorizationCodeFetcher` when the server did not advertise support — so the library never sees a value it will refuse. That works, but it means every caller that wants to talk to these services has to re-implement discovery and the comparison.

Happy to send a PR for the `validateIssuerResponse` change if the direction looks right to you.

Found while adding MCP client support to Ollama; the SDK is otherwise doing all the heavy lifting of the authorization flow and this is the only interoperability problem we hit against real services.

asjdf:
## Summary

Fixes #1152.

Sentry, Linear, and Notion omit `authorization_response_iss_parameter_supported` from their authorization-server metadata. Sentry still sends a correct `iss` in the authorization response, so sign-in fails after the browser step: the user has already consented and an authorization code has been issued, but `validateIssuerResponse` rejects the exchange solely because the server did not advertise RFC 9207 support.

RFC 9207 §2.4 says clients MUST extract and compare `iss` when the parameter is present. §2.4 also notes that clients SHOULD discard an unadvertised `iss`, but that is a recommendation and local policy may accept servers that send `iss` without advertising. Validating a matching `iss` is interoperable with those hosted MCP servers and still catches mix-up attacks: a wrong `iss` is rejected whether or not support was advertised.

When `iss` is present, this change always compares it to the expected issuer (simple string equality). A match is accepted even if the server did not advertise support. A mismatch still returns `authorization response issuer %q does not match expected issuer %q`. Absence of `iss` remains an error only when the server advertised support.

## Test plan

- [x] `go test ./auth/ -run TestValidateIssuerResponse -count=1`
- [x] `go test ./auth/...`

New `TestValidateIssuerResponse` cases:

- `UnadvertisedIssCorrect`: `iss == expectedIssuer`, `issSupported == false` → no error
- `UnadvertisedIssWrong`: `iss == "https://attacker.example.com"`, `issSupported == false` → error containing `does not match expected issuer`

Existing advertised+match / advertised+mismatch / missing-iss cases are unchanged.

### DIFF
--- a/auth/authorization_code.go
+++ b/auth/authorization_code.go
@@ -124,6 +124,11 @@ type AuthorizationCodeHandlerConfig struct {
 	// See https://modelcontextprotocol.io/seps/2207-oidc-refresh-token-guidance.
 	RequestRefreshToken bool
 
+	// AcceptUnadvertisedIss accepts a matching RFC 9207 iss even when the
+	// authorization server metadata omits authorization_response_iss_parameter_supported.
+	// The zero value (false) keeps the historical reject-unadvertised-iss behavior.
+	AcceptUnadvertisedIss bool
+
 	// Client is an optional HTTP client to use for HTTP requests.
 	// It is used for the following requests:
 	//  - Fetching Protected Resource Metadata
@@ -367,7 +372,7 @@ func (h *AuthorizationCodeHandler) Authorize(ctx context.Context, req *http.Requ
 		// Purposefully leaving the error unwrappable so it can be handled by the caller.
 		return err
 	}
-	if err := validateIssuerResponse(authRes.Iss, asm.Issuer, asm.AuthorizationResponseIssParameterSupported); err != nil {
+	if err := validateIssuerResponse(authRes.Iss, asm.Issuer, asm.AuthorizationResponseIssParameterSupported, h.config.AcceptUnadvertisedIss); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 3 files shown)


auth/authorization_code.go
```diff
@@ -616,23 +616,23 @@ func (h *AuthorizationCodeHandler) getAuthorizationCode(ctx context.Context, cfg
 }
 
 // validateIssuerResponse validates the "iss" parameter in an authorization response
-// per [RFC 9207].
+// per [RFC 9207]. When iss is present it is always compared to expectedIssuer
+// (RFC 9207 §2.4), even if the server did not advertise
+// authorization_response_iss_parameter_supported. An unadvertised matching iss
+// is accepted; a mismatch is rejected. Absence of iss is an error only when
+// the server advertised support.
```
guglielmo-san: According to spec 

> However, there might be legitimate authorization servers that provide the iss parameter without indicating their support in their metadata. Local policy or configuration can determine whether to accept such responses, and specific guidance is out of scope for this specification.

What do you think about adding a configurable value in AuthorizationCodeHandlerConfig to let the client define the local policy if accepting or not

asjdf: Good call — added `AcceptUnadvertisedIss` on `AuthorizationCodeHandlerConfig` as the local policy knob.

- Zero value (`false`) keeps the released v1.7.0 behavior: if metadata does not advertise `authorization_response_iss_parameter_supported` and `iss` is present, we reject.
- Set `AcceptUnadvertisedIss: true` to accept a matching unadvertised `iss` (Sentry/Linear/Notion-style servers).
- A wrong `iss` is still rejected either way; we never accept a mismatch.

Pushed on this branch in 7fdf782.

guglielmo-san (approved):

guglielmo-san:

PR #1190, opened 2026-08-21
mcp: prune completed listen request IDs from the session

latent-9:
### Problem

`ServerSession` tracks the request IDs of in-flight `subscriptions/listen` streams so `Close` can cancel the parked handlers and avoid the jsonrpc2 drain deadlock. The ID is appended when the listen starts, but it is never removed when the listen handler returns.

Every completed listen stream (peer cancel, unsubscribe, stream break) therefore leaves one stale entry in the session for its entire lifetime. On a long-lived session where a client repeatedly subscribes and unsubscribes, the slice grows without bound: one leaked entry per cycle.

`Close` calls `jsonrpc2.Connection.Cancel` on each recorded ID, and a stale ID is a no-op there today, but the bookkeeping is meant to hold only in-flight listens and currently never shrinks.

### Change

Drop the request ID from the session's bookkeeping when the listen handler returns, matching the lifecycle of the in-flight handler that owns it.

### Tests

Three tests reproduce the leak, failing on `main` and passing here:

- `TestListenPrunedAfterSingleCompletion`
- `TestListenIDsDoNotAccumulate`
- `TestListenIDsLeakViaPublicSubscribeUnsubscribe`

The third exercises the leak through the public `Subscribe`/`Unsubscribe` API. `go test ./mcp/` passes.


mcp/listenleak_test.go
```diff
@@ -0,0 +1,110 @@
+// Copyright 2025 The Go MCP SDK Authors. All rights reserved.
+// Use of this source code is governed by an MIT-style
+// license that can be found in the LICENSE file.
+package mcp
+
+import (
+	"context"
+	"fmt"
```
guglielmo-san: can you add these tests to mcp_test.go? TestServerSessionCloseWithActiveListen is accessing the same listenIDs, and getSession can be deduplicated

latent-9:
Done. Moved the three tests into mcp_test.go next to TestServerSessionCloseWithActiveListen, reusing basicClientServerConnection for setup and a shared listenIDsCount accessor (the old test's inline lock block now uses it too), and removed listenleak_test.go.

Verified the tests still fail with the server.go fix reverted and pass with it.

guglielmo-san (approved):

guglielmo-san:
@latent-9 thank you for the contribution!

PR #1193, opened 2026-08-24
mcp: tolerate 404 on subscriptions/listen in stateless mode

### LINKED ISSUE #3129: Remote MCP calls to the Copilot-hosted GitHub MCP server fail (crush)
### Describe the bug

Remote MCP calls to the Copilot-hosted GitHub MCP server (`https://api.githubcopilot.com/mcp/`) fail during tool listing with a "session not found" error, specifically on the `subscriptions/listen` request, which is sent with an empty session ID. This started abruptly today - first occurence at `2026/08/20 14:29:20` UTC+2 - with no local config or token changes, and reproduces consistently across multiple client (Crush) versions (0.88.1, 0.89.0 and 0.90.0), so it does not appear to be a client-side regression. I have been using the remote MCP calls for the past 2-3 weeks successfully.

Notably, the `initialize` response's `capabilities.resources` object is empty (`{}`) — no `subscribe: true` is advertised — yet the client attempts a `subscriptions/listen` call regardless, which then fails with the empty-session error and tears down the connection.

I've isolated this to the server/session layer with a manual curl reproduction (see below, and double check it as I asked my local AI to give me the curl commands for this investigation): a raw `initialize` -> `notifications/initialized` ->`tools/list` sequence against the same endpoint, with the same PAT, completes successfully and returns a valid session ID. This suggests the issue is specific to how the session ID is being handled/propagated for the `subscriptions/listen` call path, not with authentication or the basic MCP handshake.

### Affected version

I don't know which version the remote is using. But I can see in this morning crush logs that it triggers no error, and then in the early afternoon and until now I have the errors. So whatever changed during that time period.

The curl initialise post returned a version, I hope this is relevant for you:

    "version":"github-mcp-server/remote-cea154657bcbd68eeea95ecedde676a3278e2c30"

### Steps to reproduce the behavior

1. Configure an MCP client (I used Crush, versions 0.90.0, 0.89.0, and 0.88.1, all reproduce identically) to connect to `https://api.githubcopilot.com/mcp/` via HTTP transport with a valid PAT in the `Authorization: Bearer` header. (see below)
2. Have the client perform the standard MCP handshake (`initialize` → `notifications/initialized` → `tools/list`).
3. Observe the client fail while listing tools, with the underlying transport attempting `subscriptions/listen` and receiving a "session not found" error with an empty session ID.

I'm using the following `.crushrc` configuration in my local project:

```
mcp add github --type http --url "https://api.githubcopilot.com/mcp/" \
  --timeout 10 --header Authorization "Bearer $GH_PAT"
```

Before calling `crush` I load my GitHub PAT in the GH_PAT env variable. I'm using a fine grained access token (which is not expired):

    Repository access: openporte/openporte
    Repository permissions:  Read access to Dependabot alerts, License compliance alerts, code, code quality, commit statuses, issues, metadata, pull requests, and security events

### Expected vs actual behavior

**Expected:** When starting crush, the MCP status for GitHub should be green.

**Actual:** When starting crush, the MCP status for GitHub is red with an error. The message is truncated in the TUI: `GitHub error: connection cl...` the rest isn't visible. But I have a log message which is in the next section.

### Logs

This is the relevant line from my crush log (the first time it broke):

```
2026/08/20 14:29:20 ERRO Error listing tools error="connection closed: calling "tools/list": client is closing: sending "subscriptions/listen": failed to connect (session ID: ): session not found" source=github.com/charmbracelet/crush/internal/agent/tools/mcp/init.go:552
```

And the previous startup was successful with:
```
2026/08/19 02:34:32 INFO Initializing MCP clients source=github.com/charmbracelet/crush/internal/agent/tools/mcp/init.go:294
```

Crush was running the whole time between this two dates. I restarted crush this afternoon because I added a new provider to its config. In my crush sessions this morning and yesterday, I referenced GitHub issues which were read from crush.

And from curl, `initialize` response capabilities (redacted of unrelated fields):
```json
{"capabilities":{"completions":{},"prompts":{},"resources":{},"tools":{}}, "protocolVersion":"2025-06-18"}
```
Note `resources` is empty; no `subscribe` flag advertised.

smlx:
When connecting to remote stateless servers (such as GitHub's MCP server), optional streams like subscriptions/listen may be rejected by middleware with a plain-text HTTP 404 before reaching a JSON-RPC handler.

Previously, checkResponse misclassified these 404s as lost stateful sessions (ErrSessionMissing) and tore down the transport connection because the error didn't wrap jsonrpc2.ErrRejected.

This change aligns with the MCP 2026-07-28 Streamable HTTP specification:

1. Conditionally check for an active session ID before returning ErrSessionMissing, since a 404 on a stateless connection cannot mean a session was lost.

2. Wrap subscriptions/listen failures with ErrRejected (matching server/discover) so the jsonrpc2 layer doesn't permanently break the transport when encountering non-compliant plain-text 404s.

---

I have manually tested this change against GitHub's hosted MCP server and confirmed that it fixes https://github.com/github/github-mcp-server/issues/3129

### DIFF
--- a/mcp/streamable.go
+++ b/mcp/streamable.go
@@ -2347,13 +2347,14 @@ func (c *streamableClientConn) Write(ctx context.Context, msg jsonrpc.Message) e
 	}
 
 	if err := c.checkResponse(ctx, requestSummary, resp); err != nil {
-		if requestMethod == methodDiscover && !errors.Is(err, jsonrpc2.ErrRejected) {
-			// Wrap the discover failure with ErrRejected so the jsonrpc2 layer
-			// doesn't set writeErr, which would prevent the legacy initialize
-			// fallback from succeeding on the same connection. This covers the
-			// case where a legacy server rejects server/discover with a
-			// non-JSON-RPC body (e.g. plain text 400), which checkResponse
-			// cannot classify as a per-call rejection on its own.
+		if (requestMethod == methodDiscover || requestMethod == methodSubscriptionsListen) && !errors.Is(err, jsonrpc2.ErrRejected) {
+			// Wrap the discover or subscriptions/listen failure with ErrRejected so
+			// the jsonrpc2 layer doesn't set writeErr, which would break the connection
+			// (preventing the legacy initialize fallback from succeeding, or breaking
+			// subsequent RPCs if subscriptions/listen fails). This covers the case
+			// where a server rejects these requests with a non-JSON-RPC body (e.g.
+			// plain text 404 or 400), which checkResponse cannot classify as a
+			// per-call rejection on its own.
 			err = fmt.Errorf("%w: %w", err, jsonrpc2.ErrRejected)
 		} else if !errors.Is(err, jsonrpc2.ErrRejected) {
 			// Only fail the connection for non-transient errors.
@@ -2592,10 +2593,11 @@ func (c *streamableClientConn) checkResponse(ctx context.Context, requestSummary
 	// §2.5.3: "The server MAY terminate the session at any time, after
 	// which it MUST respond to requests containing that session ID with HTTP
 	// 404 Not Found."
-	if resp.StatusCode == http.StatusNotFound {
+	// Only a stateful connection with a known session ID can experience ErrSessionMissing.
+	if sessionID := c.SessionID(); resp.StatusCode == http.StatusNotFound && sessionID != "" {
 		// Return an ErrSessionMissing to avoid sending a redundant DELETE when the
 		// session is already gone.
-		return fmt.Errorf("%s: failed to connect (session ID: %v): %w", requestSummary, c.sessionID, ErrSessionMissing)
+		return fmt.Errorf("%s: failed to connect (session ID: %v): %w", requestSummary, sessionID, ErrSessionMissing)
 	}
 	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
 		return fmt.Errorf("%s: %v", requestSummary, http.StatusText(resp.StatusCode))
--- a/mcp/streamable_client_test.go
+++ b/mcp/streamable_client_test.go
@@ -1965,3 +1965,67 @@ func TestStreamableClientHandlerErrorPropagation(t *testing.T) {
 		})
 	}
 }
+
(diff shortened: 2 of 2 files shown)


mcp/client_test.go
```diff
@@ -1092,3 +1093,55 @@ func TestClientConnectDiscover_UnsupportedVersionNegotiation(t *testing.T) {
...
+	server := NewServer(&Implementation{Name: "stateless-server", Version: "v1"}, nil)
+	server.AddReceivingMiddleware(func(next MethodHandler) MethodHandler {
+		return func(ctx context.Context, method string, req Request) (Result, error) {
+			if method == methodSubscriptionsListen {
+				select {
+				case <-listenReceived:
+				default:
+					close(listenReceived)
+				}
+				return nil, jsonrpc2.ErrMethodNotFound
```
guglielmo-san: this test was never broken. returning a JSON-RPC decodable error now would wrap the error with ErrRejecting preventing the connection to close. We should return a plain text error

mcp/client_test.go
```diff
@@ -1092,3 +1093,55 @@ func TestClientConnectDiscover_UnsupportedVersionNegotiation(t *testing.T) {
...
+			}
+			return next(ctx, method, req)
+		}
+	})
+
+	ct, st := NewInMemoryTransports()
```
guglielmo-san: you should use a Streamable connection to verify the actual fix

mcp/client_test.go
smlx: Ah yes, thanks. The plain text error is covered in `TestStreamableClient_StatelessSubscriptionsListen404`, so removing this test.

mcp/client_test.go
smlx: I removed this test since the case is covered in the other test.

smlx:
Thanks for the review, I pushed changes that:

1. Removed a redundant test.
2. Added a missing "call check" to the other test.
3. Use the SessionID() accessor instead of directly checking the field. This avoids a possible data race.

guglielmo-san (approved):

guglielmo-san:
@smlx thank you for the contribution!

PR #1198, opened 2026-08-26
mcp: reject Subscribe on a closed session and roll back failed registrations

### LINKED ISSUE #1171: ClientSession.Subscribe can silently succeed while the session is closing
Under the SEP-2575 path, ClientSession.Subscribe starts a subscriptions/listen request without awaiting the stream response. If the session has already started closing, or if Close races with Subscribe startup, Subscribe can return nil even though the listen stream was cancelled or rejected. In that case the local resourceSubs map can also keep an entry for a subscription that was never successfully started.

Reproduction:
1. Connect a ClientSession with protocolVersion20260728.
2. Close the session, or have Close start while Subscribe is entering the subscriptions/listen send path.
3. Call Subscribe for a resource URI.

Expected:
Subscribe reports ErrConnectionClosed and does not leave a local resource subscription registered.

Actual:
Subscribe can report success or context cancellation without consistently rolling back the resource subscription entry.

Environment:
go version go1.26.5 darwin/arm64

GerardGao:
## What

`ClientSession.Subscribe` (SEP-2575 path) starts a `subscriptions/listen` stream
without awaiting the send result, so on a session that is closing — or when the
listen send fails for any other reason — it can return `nil` even though the
stream never started. The local `resourceSubs` map also keeps an entry whose
cancel func is then never called, leaking the registration and misreporting the
URI as subscribed. Fixes #1171.

## How

- Add a `closed` flag to `ClientSession`, set at the top of `Close`.
- `Subscribe` checks the flag under `resourceSubsMu` before registering, so a
  call after `Close` has started returns `ErrConnectionClosed` and registers
  nothing. Because the check and the registration share the mutex with the
  `Close` cleanup sweep, a racing registration is either swept by `Close` or
  refused outright — no interleaving leaks.
- If `subscriptionsListen` reports an error, the registration is rolled back
  (entry removed and cancel func called), so a later `Subscribe` retries the
  listen instead of treating the URI as already subscribed.

`Unsubscribe` and idempotent double-`Subscribe` behavior are unchanged.

## Tests

- `TestResourceSubscriptions_SubscribeClosedSession`: Subscribe after Close
  returns `ErrConnectionClosed` and leaves no `resourceSubs` entry.
- `TestResourceSubscriptions_SubscribeListenFailureRollsBack`: a failed listen
  send (injected via `AddSendingMiddleware`) rolls back the registration and a
  retry re-attempts the listen.
- `TestResourceSubscriptions_SubscribeCloseRace`: 50 rounds of racing `Close`
  and `Subscribe` asserting that a failed `Subscribe` never leaves an entry.

`go test -race ./mcp/` and `go test ./...` pass; `gofmt`, `go vet`, and
`staticcheck` are clean.

### DIFF
--- a/mcp/client.go
+++ b/mcp/client.go
@@ -474,6 +474,11 @@ type ClientSession struct {
 	calledOnClose atomic.Bool
 	onClose       func()
 
+	// closed is set before Close tears the session down. Subscribe checks it
+	// under resourceSubsMu so that a subscription registered after Close has
+	// started cannot leak an entry that Close already cleaned up.
+	closed atomic.Bool
+
 	conn            *jsonrpc2.Connection
 	client          *Client
 	keepaliveCancel context.CancelFunc
@@ -563,6 +568,10 @@ func (cs *ClientSession) ID() string {
 //
 // Close is idempotent and concurrency safe.
 func (cs *ClientSession) Close() error {
+	// Mark the session closed before cancelling subscriptions so that a racing
+	// Subscribe either registers before the cleanup sweep (and is cancelled by
+	// it) or observes the closed flag and reports ErrConnectionClosed.
+	cs.closed.Store(true)
 	// Note: keepaliveCancel access is safe without a mutex because:
 	// 1. keepaliveCancel is only written once during Client.Connect (through startKeepalive),
 	//    which happens before any code that may call Close from another goroutine
@@ -1390,6 +1399,13 @@ func (cs *ClientSession) Subscribe(ctx context.Context, params *SubscribeParams)
 
 	var listenCtx context.Context
 	cs.resourceSubsMu.Lock()
+	if cs.closed.Load() {
+		// Close already started: the session can no longer open listen
+		// streams, and Close may have swept the subscription map already, so
+		// registering here would leak an entry that nothing cancels.
+		cs.resourceSubsMu.Unlock()
+		return ErrConnectionClosed
+	}
 	if _, exists := cs.resourceSubs[uri]; !exists {
 		var cancel context.CancelFunc
 		listenCtx, cancel = context.WithCancel(context.Background())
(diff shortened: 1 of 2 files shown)


GerardGao:
Rebased onto latest `main` (`3632967`) — this branch was 15 commits behind. No conflicts, and the diff is unchanged.

I also noticed no CI has run on this PR since it was opened. I think it needs a maintainer to approve the workflow run, since this is my first contribution to this repo — happy to be told if something else is blocking it instead.

On the overlap question: #1221 also touched `mcp/client.go`, but it changes the capability check in `Connect`, while this PR touches the `ClientSession` field, `Close`, and `Subscribe`. They don't overlap, and `Subscribe` on `main` still returns `subscriptionsListen`'s result directly with no rollback, so #1171 is still reproducible.

To confirm the change is actually load-bearing, I ran a deterministic oracle rather than relying on winning the race: `Close` sweeps `resourceSubs` to `nil`, so sequencing `Close` fully before `Subscribe` reliably hits the post-sweep window. On exact base `3632967` that fails 5/5 with `Subscribe-after-Close returned err=<nil>; resourceSubs entries=1 leaked=true` — the entry is registered into a freshly allocated map that nothing will ever cancel, since `Close` has already run. On this head the same oracle passes 10/10 with `err=connection closed; entries=0 leaked=false`.

Verification on the rebased head, mirroring this repo's CI: `gofmt`/`go vet`/`staticcheck v0.6.1` clean, `go test ./...` passes on both 1.25 and 1.26, `go test -race ./...` passes, docs are up-to-date, and both conformance suites pass against the baseline (client 460 passed, server 42/0 — including `resources-subscribe` and `resources-unsubscribe`).

Since #1171 isn't labelled `help wanted` and I didn't check in on the issue first, do say if you'd prefer a different approach here — happy to rework or close it.

PR #1199, opened 2026-08-26
mcp: allow restricting the protocol versions a server supports

### LINKED ISSUE #945: Set list of supported versions

guglielmo-san:
Fixes #945

### DIFF
--- a/mcp/server.go
+++ b/mcp/server.go
@@ -1557,7 +1557,7 @@ func (ss *ServerSession) updateState(mut func(*ServerSessionState)) {
 	copy := ss.state
 	ss.mu.Unlock()
 	if c, ok := ss.mcpConn.(serverConnection); ok {
-		c.sessionUpdated(copy)
+		c.sessionUpdated(copy, ss.server.protocolVersions)
 	}
 }
 
--- a/mcp/server.go
+++ b/mcp/server.go
@@ -176,13 +176,14 @@ type ServerOptions struct {
 
 	// SupportedProtocolVersions, if non-empty, restricts the protocol versions
 	// this server advertises in "server/discover" and is willing to negotiate,
-	// to those listed. If empty, every version supported by this version of the
-	// SDK is used.
+	// to those listed. If empty, every version returned by
+	// [SupportedProtocolVersions] is used.
 	//
-	// The value can only narrow support, never widen it: [NewServer] panics if
-	// it names a version the SDK does not support. The order of entries is
-	// irrelevant; the server always prefers the newest mutually supported
-	// version.
+	// Entries must come from [SupportedProtocolVersions], the set of valid
+	// values: the list can only narrow support, never widen it, and
+	// [NewServer] panics if it names a version that is not there. The order of
+	// entries is irrelevant; the server always prefers the newest mutually
+	// supported version.
 	//
 	// A request using the >= 2026-07-28 protocol at an excluded version is
 	// rejected with error code [CodeUnsupportedProtocolVersion].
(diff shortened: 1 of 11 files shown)


mcp/transport.go
```diff
@@ -114,7 +114,13 @@ type clientConnection interface {
...
 type serverConnection interface {
 	Connection
-	sessionUpdated(ServerSessionState)
+
+	// sessionUpdated is called whenever the server session state changes.
+	//
+	// supported is the set of protocol versions the server negotiates, newest
+	// first, so that a connection deriving the session's effective version
+	// reaches the same answer the initialize handshake did.
+	sessionUpdated(state ServerSessionState, supported []string)
```
yarolegovich: doesn't look clean. I'd rather we add a new field to session state: 
```
type ServerSessionState struct {
  ...
  NegotiatedProtocolVersion string `json:"negotiatedProtocolVersion,omitempty"`
}
```
compute it on initialize, and in sessionUpdated the logic becomes
```go
protocolVersion := state.NegotiatedProtocolVersion
if protocolVersion == "" { // the last spec version that did not protocol-version header
  protocolVersion = protocolVersion20250326
}
```

mcp/server.go
```diff
@@ -205,6 +232,21 @@ func NewServer(impl *Implementation, options *ServerOptions) *Server {
...
+		for _, v := range opts.SupportedProtocolVersions {
+			if !slices.Contains(supportedProtocolVersions, v) {
+				panic(fmt.Errorf("unsupported protocol version %q", v))
+			}
+		}
+		protocolVersions = slices.DeleteFunc(protocolVersions, func(v string) bool {
+			return !slices.Contains(opts.SupportedProtocolVersions, v)
+		})
+	}
+	if len(protocolVersions) == 0 {
```
yarolegovich: this is unreachable if `len(opts.SupportedProtocolVersions) > 0` and all elements are present in `supportedProtocolVersions`

mcp/server.go
```diff
@@ -169,6 +173,29 @@ type ServerOptions struct {
 	// GetSessionID is not consulted when [StreamableHTTPOptions.Stateless] is
 	// true, since stateless servers do not maintain sessions.
 	GetSessionID func() string
+
+	// SupportedProtocolVersions, if non-empty, restricts the protocol versions
+	// this server advertises in "server/discover" and is willing to negotiate,
+	// to those listed. If empty, every version supported by this version of the
+	// SDK is used.
+	//
+	// The value can only narrow support, never widen it: [NewServer] panics if
```
yarolegovich: Let's reference `[SupportedProtocolVersions]` as the source of valid values.

mcp/streamable.go
```diff
@@ -1541,7 +1541,11 @@ func (c *streamableServerConn) servePOST(w http.ResponseWriter, req *http.Reques
 					// Advertise only the legacy versions this stateful server accepts
 					// (i.e. exclude 2026-07-28, since that's what the request asked for
 					// and what we're rejecting).
-					legacyVersions := slices.DeleteFunc(slices.Clone(supportedProtocolVersions), func(v string) bool {
+					advertised := supportedProtocolVersions
```
yarolegovich, line 1552: line 371 in this file needs to be updated, currently checks supportedProtocolVersions ignoring server opts

yarolegovich (approved):

PR #1202, opened 2026-08-27
mcp: support per-resource cache policy for resources/read

### LINKED ISSUE #1185: mcp: support per-resource cache policy for resources/read

A server may expose both public static resources and tenant-specific resource templates. Today, assigning the correct `ttlMs` and `cacheScope` to each `resources/read` result requires receiving middleware—or the proposed `DefaultCacheable` callback in #1128—to inspect the request URI. That duplicates policy already known where the resource is registered.

Allow direct resources and resource templates to declare a cache hint at registration. The matched resource's hint should apply only to its `resources/read` result, take precedence over the server default, and not appear in `resources/list` or `resources/templates/list`.

For example, a server could register static documentation as `public` and a tenant-backed template as `private` without central URI matching.

The TypeScript SDK already supports this through `cacheHint` on `registerResource`.

- Receiving middleware works, but moves resource-specific policy away from registration.
- #1128's server-level callback is useful for broad or request-dependent defaults, but registration-level policy would complement it for stable per-resource semantics.

**Additional context**

Related: #1094 and #1128.

CocaineCong:
Fixes #1185

Desc: support per-resource cache policy for resources/read

Changes:
- Add exported CacheHint{ TTLMs int, CacheScope string } type.
- Add a CacheHint *CacheHint field to Resource and ResourceTemplate, tagged json:"-" so it never appears in resources/list or resources/templates/list.
- readResource now applies the matched resource's/template's hint to the resources/read result via a new (*Cacheable).applyCacheHint.
- Add UnitTest and passed.
- Update docs/server.md.

### DIFF
--- a/docs/server.md
+++ b/docs/server.md
@@ -996,6 +996,30 @@ server.AddSendingMiddleware(func(next mcp.MethodHandler) mcp.MethodHandler {
 })
 ```
 
+**Per-resource cache hints.** For `resources/read`, a resource or resource
+template can declare its caching policy at registration through
+`CacheHint`, without central URI matching in middleware. The matched
+resource's hint applies only to its `resources/read` result, takes precedence
+over the server default, and is never emitted in `resources/list` or
+`resources/templates/list`. A value the handler sets on the result still wins
+over the hint.
+
+```go
+// Public static documentation, cacheable by any intermediary.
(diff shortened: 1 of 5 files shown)


guglielmo-san:
Closed as the issue won't be addressed

PR #1203, opened 2026-08-27
mcp: allow per request Cacheable customization

### LINKED ISSUE #1094: No option to override forced Cacheable defaults (cacheScope public, ttlMs 0) on SDK-generated results

guglielmo-san:
Fixes #1094

### DIFF
--- a/mcp/protocol.go
+++ b/mcp/protocol.go
@@ -1191,13 +1191,13 @@ func (c Cacheable) GetTTLMs() int { return c.TTLMs }
 // GetCacheScope returns the cache scope.
 func (c Cacheable) GetCacheScope() string { return c.CacheScope }
 
-// normalizeCacheable fills in the protocol default for any cache field left
+// normalize fills in the protocol default for any cache field left
 // unset. An absent cacheScope means "public", but the field is required on the
 // wire, so the default is materialized here rather than sent empty.
 //
 // Values already present are preserved: this must not undo a decision made by
 // a resource handler or by [ServerOptions.SetCacheable].
-func (c *Cacheable) normalizeCacheable() {
+func (c *Cacheable) normalize() {
 	if c.CacheScope == "" {
 		c.CacheScope = "public"
 	}
--- a/mcp/protocol.go
+++ b/mcp/protocol.go
@@ -1191,9 +1191,16 @@ func (c Cacheable) GetTTLMs() int { return c.TTLMs }
 // GetCacheScope returns the cache scope.
 func (c Cacheable) GetCacheScope() string { return c.CacheScope }
 
-// setDefaultCacheableValues sets the default values for the cacheable fields.
-func (c *Cacheable) setDefaultCacheableValues() {
-	c.CacheScope = "public"
+// normalizeCacheable fills in the protocol default for any cache field left
+// unset. An absent cacheScope means "public", but the field is required on the
+// wire, so the default is materialized here rather than sent empty.
+//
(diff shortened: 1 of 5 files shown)


mcp/server.go
```diff
@@ -934,22 +942,22 @@ func filterSupportedVersions(t Transport) []string {
...
 }
 
-func (s *Server) listTools(_ context.Context, req *ListToolsRequest) (*ListToolsResult, error) {
-	s.mu.Lock()
-	defer s.mu.Unlock()
+func (s *Server) listTools(ctx context.Context, req *ListToolsRequest) (*ListToolsResult, error) {
 	if req.Params == nil {
 		req.Params = &ListToolsParams{}
 	}
+	s.mu.Lock()
```
yarolegovich: this is a scary change. if there's a recovery handler and paginateList panics we will be left with a locked mutex. I'd rather keep defer

mcp/protocol.go
```diff
@@ -1191,9 +1191,16 @@ func (c Cacheable) GetTTLMs() int { return c.TTLMs }
...
-// setDefaultCacheableValues sets the default values for the cacheable fields.
-func (c *Cacheable) setDefaultCacheableValues() {
-	c.CacheScope = "public"
+// normalizeCacheable fills in the protocol default for any cache field left
+// unset. An absent cacheScope means "public", but the field is required on the
+// wire, so the default is materialized here rather than sent empty.
+//
+// Values already present are preserved: this must not undo a decision made by
+// a resource handler or by [ServerOptions.SetCacheable].
+func (c *Cacheable) normalizeCacheable() {
```
yarolegovich: nit: `caceable.normalizeCacheable()` -> `cacheable.normalize()`

yarolegovich:
> Fixes https://github.com/modelcontextprotocol/go-sdk/pull/1049

wrong issue

yarolegovich (approved):

PR #1205, opened 2026-08-28
mcp: add an sse event size cap

yarolegovich:
update sse I/O code to limit the amount of memory a single non-terminated event can take

### DIFF
--- a/mcp/event.go
+++ b/mcp/event.go
@@ -65,23 +65,20 @@ func writeEvent(w http.ResponseWriter, evt Event) (int, error) {
 	return n, err
 }
 
-// defaultMaxEventSize bounds the number of bytes buffered for a single SSE
-// event before the input gets rejected.
-const defaultMaxEventSize = 16 << 20 // 16 MiB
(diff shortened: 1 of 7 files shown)


mcp/event.go
```diff
@@ -65,13 +65,24 @@ func writeEvent(w http.ResponseWriter, evt Event) (int, error) {
 	return n, err
 }
 
+// defaultMaxEventSize bounds the number of bytes buffered for a single SSE
+// event before the input gets rejected.
+const defaultMaxEventSize = 16 << 20 // 16 MiB
```
guglielmo-san: should not we allow the possibility to override this value?

guglielmo-san (dismissed):

mcp/event.go
yarolegovich: initially was thinking about hiding it under mcp_go_debug, together with json depth check, but thinking about it it's going to be more consistent if we expose this as a config and omit mcp go debug for the depth check, I can't imagine a legitimate need for very deep nesting

guglielmo-san (approved):

PR #1208, opened 2026-08-28
mcp: gate session lifecycle log demotion behind MCPGODEBUG

### LINKED ISSUE #1204: Session-lifecycle bookkeeping logs at info, spamming multiple log lines per request on stateless streamable HTTP
**Describe the bug**
Found this while working on adding stateless HTTP support to [the Prometheus MCP server](https://github.com/prometheus/prometheus-mcp)

TL;DR: the current implementation of the new 2026-07-28 spec, and the stateless HTTP support specifically, are spamming logs with in-actionable session bookkeeping.

Summary:
The server's internal `slog` records for session bookkeeping (`server connecting`, `server session connected`, `server session disconnected`, `session initialized`, `client log level set`) all log at `Info`. These logs were reasonable when they were added and HTTP was stateful since they were per-session events. Now though, they're _per-request_:

- Stateless streamable HTTP mints a session per POST: every single request replays the connect/disconnect lifecycle.
- **2026-07-28 clients carry `_meta` on every request**, so `setLevel` runs, logging `client log level set` once per request, on every transport (stdio included), even when the client set no level at all (the record fires with `level=""`).

The result: 
A server running at a normal info log level gets four lines of pure bookkeeping for every tool call, and this is before it logs anything of its own.

**To Reproduce**
go-sdk v1.7.0, stateless streamable HTTP, one client connect plus one `tools/call`:

```go
package main

import (
	"context"
	"fmt"
	"log/slog"
	"net/http"
	"net/http/httptest"
	"os"

)

func main() {
	logger := slog.New(slog.NewTextHandler(os.Stdout, &slog.HandlerOptions{
		Level: slog.LevelInfo,
		ReplaceAttr: func(groups []string, a slog.Attr) slog.Attr {
			if a.Key == slog.TimeKey && len(groups) == 0 {
				return slog.Attr{} // drop timestamps for a stable transcript
			}
			return a
		},
	}))

	server := mcp.NewServer(&mcp.Implementation{Name: "probe-server", Version: "0.0.1"}, &mcp.ServerOptions{
		Logger: logger,
	})
	mcp.AddTool(server, &mcp.Tool{Name: "echo", Description: "echo"}, func(ctx context.Context, req *mcp.CallToolRequest, input struct{}) (*mcp.CallToolResult, any, error) {
		return &mcp.CallToolResult{Content: []mcp.Content{&mcp.TextContent{Text: "ok"}}}, nil, nil
	})

	handler := mcp.NewStreamableHTTPHandler(func(*http.Request) *mcp.Server { return server }, &mcp.StreamableHTTPOptions{
		Stateless: true,
	})
	httpServer := httptest.NewServer(handler)
	defer httpServer.Close()

	ctx := context.Background()
	client := mcp.NewClient(&mcp.Implementation{Name: "probe-client", Version: "0.0.1"}, nil)
	session, err := client.Connect(ctx, &mcp.StreamableClientTransport{Endpoint: httpServer.URL}, nil)
	if err != nil {
		panic(err)
	}
	defer session.Close()

	fmt.Println("--- one tools/call ---")
	if _, err := session.CallTool(ctx, &mcp.CallToolParams{Name: "echo"}); err != nil {
		panic(err)
	}
	fmt.Println("--- end of tools/call ---")
}
```

**Expected behavior**
To see the debug logs in my debug stream 🙃 

**Logs**

Output from repro ^
```
level=INFO msg="server connecting"
level=INFO msg="server session connected" session_id=""
level=INFO msg="client log level set" level=""
level=INFO msg="server session disconnected" session_id=""
--- one tools/call ---
level=INFO msg="server connecting"
level=INFO msg="server session connected" session_id=""
level=INFO msg="client log level set" level=""
level=INFO msg="server session disconnected" session_id=""
--- end of tools/call ---
```

**Additional context**
That's four info lines per request, none of which carry information an admin can act on. At any real request rate, these events dominate the log stream. Note also `client log level set` firing with `level=""` — the client never asked for anything.

On stdio the amplification is smaller but still per-request -- the  `client log level set` event triggers on every request it sends.

These log events were added in #501 with the reasoning "to improve debuggability", so let's demote them to debug where they belong.

anneheartrecord:
Session bookkeeping records (`server connecting`, `server session connected`/`disconnected`, `session initialized`, `client log level set`) log at info. On stateless streamable HTTP a session is minted per POST, so every request replays the whole lifecycle — four info lines of bookkeeping per tool call before the server logs anything of its own. The `client log level set` record additionally fires on every request carrying `_meta` under the 2026-07-28 spec, on all transports, even when the client set no level.

This gates the demotion of the five lifecycle records to debug behind `MCPGODEBUG=demotesessionlifecyclelog=1`, so the default behavior is unchanged for existing users (per review). Warn/error records are unchanged. If the demotion should become the default in a future minor, flipping the gate is a one-line change. Adds a test covering both paths: lifecycle visible at info by default; silent at info and visible at debug with the variable set.

Fixes #1204


guglielmo-san:
I agree overall with the change, but this introduces a behavioral change for existing users. We should gate the change behind an MCPGODEBUG variable

anneheartrecord:
@guglielmo-san done — force-pushed as a single commit (`9eed2ee`). The five lifecycle records now stay at **info by default**; setting `MCPGODEBUG=demotesessionlifecyclelog=1` moves them to debug, so nothing changes for existing users unless they opt in. The knob follows the existing `MCPGODEBUG` compatibility-parameter pattern (cf. `allowsessionsinstateless`); if you'd rather have the demotion become the default in a future minor, flipping the gate is a one-line change.

Tests cover both paths (visible at info by default; silent at info and visible at debug with the variable set), `-race` clean. The only failures in `go test ./mcp/` on this machine are the two `*LocalhostProtection` tests, which fail identically on clean `main` here (local environment).

mcp/server.go
```diff
@@ -33,6 +33,26 @@ import (
...
 )
 
+// demotesessionlifecyclelog, when set to "1" via MCPGODEBUG, demotes routine
+// session bookkeeping records ("server connecting", "server session
+// connected", "server session disconnected", "session initialized", "client
+// log level set") from info to debug. On stateless streamable HTTP a session
+// is minted per request, so these fire constantly and drown out the server's
+// own logs (#1204). It is gated behind MCPGODEBUG because demoting them
+// changes log output that existing users may rely on.
+var demotesessionlifecyclelog = mcpgodebug.Value("demotesessionlifecyclelog")
```
guglielmo-san, line 43: a description should be added in mcpgodebug.rc.md files

anneheartrecord: Added in `a9b6d0d` — an entry under 1.8.0 next to `blockingcancelnotify`, in both `internal/docs/mcpgodebug.src.md` and the generated `docs/mcpgodebug.md`. Re-ran `go generate ./internal/docs/` so the two agree, and `go test ./internal/docs/` passes.

One thing worth flagging while you are looking at it: every other option on that page is opt-out (set to `1` to get the old behavior back), whereas this one is opt-in — the default keeps logging at info, which is what avoids the behavior change you raised. I said that explicitly in the entry, but it does sit a bit awkwardly next to the section's "will be removed in the 1.9.0 version" line, since dropping the knob would make the demotion unreachable rather than permanent.

If you would rather it follow the page convention, I can flip it to `keepsessionlifecycleloginfo`: demote by default, `1` restores info, and removal in 1.9.0 then means the demotion becomes permanent. Your call — happy either way.

anneheartrecord:
Rebased onto main (`5bc078a`) — the branch had drifted into a conflict. It was only in the `mcpgodebug` docs: 1.8.0 grew an "Options below were removed, according to plan" section, so the `demotesessionlifecyclelog` entry now sits with the other added options above it. The code diff is unchanged, `go generate ./internal/docs/` leaves no diff, and `TestSessionLifecycleLogLevel` passes on both the default and the opt-in path.

PR #1214, opened 2026-08-31
mcp: preserve single-request batch responses

### LINKED ISSUE #1211: mcp: preserve array response for single-request legacy batches
### Description

When `StreamableHTTPOptions.JSONResponse` is enabled, a legacy JSON-RPC batch containing one request is accepted for protocol revisions before `2025-06-18`, but the response is serialized as a bare JSON-RPC object rather than a one-element response array.

JSON-RPC 2.0 requires the response to a batch request to be an array, including when the batch contains a single call. This is observable for an initialization batch: the server creates a session and emits `Mcp-Session-Id`, but strict middleware or clients can reject the response shape.

### Reproduction

1. Start a stateful `StreamableHTTPHandler` with `JSONResponse: true`.
2. POST this request with the normal JSON and event-stream `Accept` values:

```json
[{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2025-03-26","capabilities":{},"clientInfo":{"name":"test","version":"1"}}}]
```

3. Observe a successful bare object response:

```json
{"jsonrpc":"2.0","id":1,"result":{...}}
```

Expected:

```json
[{"jsonrpc":"2.0","id":1,"result":{...}}]
```

### Cause

`streamableServerConn.servePOST` records the calls but not whether the incoming payload was a batch. `stream.deliverLocked` unwraps any response collection of length one.

### Proposed fix

Preserve the incoming `isBatch` value on the logical stream and only unwrap one response when the original request was not a batch. Add a regression for a one-request `2025-03-26` initialization batch with JSON responses enabled.

The behavior is present in `v1.7.0` and current `main` as of 2026-08-30.

wmh123wmh1:
## What does this PR do?

Preserve JSON-RPC batch response arrays when a batch contains a single request.

## Why is this needed?

A single-request legacy batch must be returned as a one-element JSON array, but the JSON response path currently unwraps it into a single JSON object.

## Changes

- Preserve the original batch state on the logical stream
- Keep single-request batch responses as arrays
- Add a regression test for legacy initialization batches

## Testing

- `go test ./mcp -count=1`
- `go test ./... -count=1`
- `git diff --check`

Fixes #1211

### DIFF
--- a/mcp/streamable.go
+++ b/mcp/streamable.go
@@ -988,6 +988,10 @@ type stream struct {
 	// Note: if we remove support for batching, this could just be a bool.
 	pendingJSONMessages []json.RawMessage
 
+	// isBatch reports whether the request that opened the stream was a JSON-RPC
+	// batch. Batch responses must remain arrays even when they contain one item.
+	isBatch bool
+
 	// w is the HTTP response writer for this stream. A non-nil w indicates
 	// that the stream is claimed by an HTTP request (the hanging POST or GET);
 	// it is set to nil when the request completes.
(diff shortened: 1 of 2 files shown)


guglielmo-san (approved):

guglielmo-san:
@wmh123wmh1 thank you for the contribution!

PR #1222, opened 2026-09-01
mcp: ignore Mcp-Session-Id on 2026-07-28 sessions

### LINKED ISSUE #1028: (possible) 2026-07-28 spec violation: mcp-session-id used in client requests with 2026-07-28
**Describe the bug**
Requests using 2026-07-28 should never set `mcp-session-id`.

However, the go SDK is sending this when the server _also_ violates the spec by setting it

TCP dump: First message is the server returning an invalid session id and then client sends it
```
2026-06-26T17:35:12.920442Z TCP 127.0.0.1:3000 --> 127.0.0.1:51940 (len:4243) (PSH,ACK)
HTTP/1.1 200 OK
content-type: text/event-stream
cache-control: no-cache
mcp-session-id: 8a392688-8526-4abb-bb21-65b140a9528b
transfer-encoding: chunked
date: Fri, 26 Jun 2026 17:35:12 GMT

F7F
data: {"jsonrpc":"2.0","id":1,"result":{"supportedVersions":["2026-07-28","2025-11-25","2025-06-18","2025-03-26","2024-11-05"],"capabilities":{"logging":{},"completions":{},"prompts":{"listChanged":true},"resources":{"listChanged":true},"tools":{"listChanged":true}},"serverInfo":{"name":"everything","version":"","icons":[{"src":"data:image/png;base64,iVBAAElFTkSuQmCC","mimeType":"image/png","sizes":["48x48"],"theme":"light"}],"websiteUrl":"https://example.com"},"instructions":"Use this server!","ttlMs":0,"cacheScope":"public"}}


0


2026-06-26T17:35:12.920451Z TCP 127.0.0.1:51940 --> 127.0.0.1:3000 (len:68) (ACK)

2026-06-26T17:35:12.920738Z TCP 127.0.0.1:51940 --> 127.0.0.1:3000 (len:650) (PSH,ACK)
POST / HTTP/1.1
Host: localhost:3000
User-Agent: Go-http-client/1.1
Content-Length: 276
Accept: application/json, text/event-stream
Content-Type: application/json
Mcp-Method: tools/list
Mcp-Protocol-Version: 2026-07-28
Mcp-Session-Id: 8a392688-8526-4abb-bb21-65b140a9528b
Accept-Encoding: gzip

{"jsonrpc":"2.0","id":2,"method":"tools/list","params":{"_meta":{"io.modelcontextprotocol/clientCapabilities":{"roots":{"listChanged":true}},"io.modelcontextprotocol/clientInfo":{"name":"mcp-client","version":"v1.0.0"},"io.modelcontextprotocol/protocolVersion":"2026-07-28"}}}


**To Reproduce**
Steps to reproduce the behavior:
1. Run the `listfeatures` client example against an invalid server

### PR DESCRIPTION
**Pull request by arpankernel**
2026-07-28 removed protocol-level sessions (SEP-2567), so the client must not echo an Mcp-Session-Id even if a non-compliant server sends one. Drop the session ID once the negotiated version is known to be modern, so it is not sent on later requests or used for a DELETE on close.

Added a regression test covering a modern session where the server sends a session ID; the client must not echo it on later requests.

Closes #1028

### DIFF
--- a/mcp/streamable.go
+++ b/mcp/streamable.go
@@ -2138,6 +2138,13 @@ var _ clientConnection = (*streamableClientConn)(nil)
 func (c *streamableClientConn) sessionUpdated(state clientSessionState) {
 	c.mu.Lock()
 	c.initializedResult = state.InitializeResult
+	// Protocol-level sessions were removed in 2026-07-28 (SEP-2567): a client
+	// must not echo an Mcp-Session-Id, so drop one that the initial response
+	// adopted before the negotiated version was known.
+	if state.InitializeResult != nil &&
+		state.InitializeResult.ProtocolVersion >= protocolVersion20260728 {
+		c.sessionID = ""
+	}
 	c.mu.Unlock()
 
 	// Under SEP-2575 (protocol version >= 2026-07-28) the standalone HTTP GET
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: mcp/streamable.go [hunk 1]
```diff
@@ -2150,6 +2150,13 @@ var _ clientConnection = (*streamableClientConn)(nil)
 func (c *streamableClientConn) sessionUpdated(state clientSessionState) {
 	c.mu.Lock()
 	c.initializedResult = state.InitializeResult
+	// Protocol-level sessions were removed in 2026-07-28 (SEP-2567): a client
+	// must not echo an Mcp-Session-Id, so drop one that the initial response
+	// adopted before the negotiated version was known.
+	if state.InitializeResult != nil &&
+		state.InitializeResult.ProtocolVersion >= protocolVersion20260728 {
+		c.sessionID = ""
+	}
```
guglielmo-san: this is a duplicate of the lines below

arpankernel: Thanks, good catch. Consolidated it so the version check is computed once and reused for both the session-id reset and the standalone-SSE branch.

arpankernel:
@guglielmo-san addressed your comment above (pulled it into one `modern` check). Anything else you'd like here, or is this good to go? Happy to rebase it current with main if that helps.

PR #1223, opened 2026-09-01
all: do not allow OPTIONS method in CORS config

jub0bs:
Contrary to popular belief, listing OPTIONS as an allowed method in a server's CORS configuration is not necessary for CORS preflight to succeed. It is only required if server developers wish to allow clients to make explicit use of that method, e.g. via the following client code:

```javascript
fetch('//example.com', {method: 'OPTIONS'})
```

This commit drops OPTIONS from the list of allowed methods throughout.


jub0bs:
cc @orius123

(Thanks for recommending [github.com/jub0bs/cors](github.com/jub0bs/cors), by the way!)

PR #1226, opened 2026-09-01
mcp: label resultType on results that bypass the dispatcher

### LINKED ISSUE #1225: mcp: tools/call omits required resultType when a receiving middleware short-circuits
**Describe the bug**
The SDK sets `resultType` on a `CallToolResult` only inside `Server.callTool`. A receiving middleware can reject a call by returning a result instead of calling `next`. That result never reaches `Server.callTool`, so the `tools/call` response carries no `resultType`. The 2026-07-28 schema requires the field, and strict clients reject the response as malformed.

This gap is the `tools/call` half of #1032. PR #1060 fixed the other half. It embedded `completeResultWithType` into `CompleteResult`, `DiscoverResult`, and the four list results, then called `setCompleteResultType` from `ServerSession.handle`.

PR #1060 left out the three multi-round-trip results: `CallToolResult`, `GetPromptResult`, and `ReadResourceResult`. Each one also carries `input_required`, so each holds its own unexported `resultType` field. They implement `setResultType` but not `isCompleteResult()`, so `setCompleteResultType` skips them.

A middleware that rejects a call is a normal use of a public extension point. Servers gate authentication, rate limits, and entitlements this way. But the SDK offers no way to build a spec-compliant `CallToolResult` outside the package:

- `resultType` is unexported.
- The SDK exposes no setter for it.
- `isResult()` seals `Result`, so no external type can implement it.

A server that uses `AddReceivingMiddleware` this way cannot return a compliant response.

Environment:

- `v1.7.0`, which contains #1060.
- Still present on `main` at `0d3036f785b6`. `CallToolResult` does not embed `completeResultWithType`, and `setCompleteResultType` is unchanged.
- Transport: Streamable HTTP, `Stateless: true`.

**To Reproduce**

The program below starts two servers. They differ only in whether a middleware short-circuits `tools/call`. Each one prints whether the response carries `resultType`.

```go
package main

import (
	"context"
	"encoding/json"
	"fmt"
	"io"
	"net/http"
	"net/http/httptest"
	"strings"

)

const meta = `"_meta":{"io.modelcontextprotocol/protocolVersion":"2026-07-28",` +
	`"io.modelcontextprotocol/clientInfo":{"name":"repro","version":"1.0"},` +
	`"io.modelcontextprotocol/clientCapabilities":{}}`

// gate rejects tools/call the way an auth, rate-limit, or entitlement
// middleware does: by returning a result instead of calling next.
func gate() mcp.Middleware {
	return func(next mcp.MethodHandler) mcp.MethodHandler {
		return func(ctx context.Context, method string, req mcp.Request) (mcp.Result, error) {
			if method != "tools/call" {
				return next(ctx, method, req)
			}
			return &mcp.CallToolResult{
				Content: []mcp.Content{&mcp.TextContent{Text: "rate limit exceeded"}},
				IsError: true,
			}, nil
		}
	}
}

func newServer(withGate bool) *mcp.Server {
	s := mcp.NewServer(&mcp.Implementation{Name: "repro", Version: "1.0.0"}, nil)
	mcp.AddTool(s, &mcp.Tool{Name: "greet"},
		func(context.Context, *mcp.CallToolRequest, struct{}) (*mcp.CallToolResult, any, error) {
			return &mcp.CallToolResult{Content: []mcp.Content{&mcp.TextContent{Text: "hi"}}}, nil, nil
		})
	if withGate {
		s.AddReceivingMiddleware(gate())
	}
	return s
}

func callTool(s *mcp.Server) map[string]json.RawMessage {
	ts := httptest.NewServer(mcp.NewStreamableHTTPHandler(
		func(*http.Request) *mcp.Server { return s },
		&mcp.StreamableHTTPOptions{Stateless: true}))
	defer ts.Close()

	body := fmt.Sprintf(`{"jsonrpc":"2.0","id":1,"method":"tools/call",`+
		`"params":{"name":"greet","arguments":{},%s}}`, meta)
	req, _ := http.NewRequest("POST", ts.URL, strings.NewReader(body))
	req.Header.Set("Accept", "application/json, text/event-stream")
	req.Header.Set("MCP-Protocol-Version", "2026-07-28")
	req.Header.Set("Mcp-Method", "tools/call")
	req.Header.Set("Mcp-Name", "greet")

	resp, err := http.DefaultClient.Do(req)
	if err != nil {
		panic(err)
	}
	defer resp.Body.Close()
	raw, _ := io.ReadAll(resp.Body)

	payload := raw
	if strings.HasPrefix(resp.Header.Get("Content-Type"), "text/event-stream") {
		for _, line := range strings.Split(string(raw), "\n") {
			if d, ok := strings.CutPrefix(strings.TrimRight(line, "\r"), "data: "); ok {
				payload = []byte(d)
				break
			}
		}
	}
	var msg struct {
		Result map[string]json.RawMessage `json:"result"`
	}
	if err := json.Unmarshal(payload, &msg); err != nil {
		panic(fmt.Sprintf("decode %v: %s", err, raw))
	}
	return msg.Result
}

func main() {
	for _, tc := range []struct {
		label    string
		withGate bool
	}{
		{"handler reaches dispatcher", false},
		{"middleware short-circuits tools/call", true},
	} {
		rt, ok := callTool(newServer(tc.withGate))["resultType"]
		fmt.Printf("%-38s resultType present=%-5v value=%s\n", tc.label, ok, rt)
	}
}
```

**Expected behavior**
- Both responses carry `resultType`, because both are `tools/call` results on a 2026-07-28 request.

**Observed behavior**
```
handler reaches dispatcher             resultType present=true  value="complete"
middleware short-circuits tools/call   resultType present=false value=
```

roncodingenthusiast:
Fixes #1225. See the [issue comment](https://github.com/modelcontextprotocol/go-sdk/issues/1225#issuecomment-5500818546).

## The bug

`2026-07-28` requires `resultType` on every result. `CallToolResult`, `GetPromptResult`, and `ReadResourceResult` keep `resultType` in an unexported field. Each of these results can be `complete` or `input_required`, so it cannot embed `completeResultWithType` the way the seven types in #1060 do.

Only `handleMultiRoundTripResult` sets that field. It runs inside `Server.callTool`, `getPrompt`, and `readResource`. Some middleware returns a result directly instead of calling `next` — for example, an auth, rate-limit, or entitlement gate. That middleware never reaches `handleMultiRoundTripResult`. The response then ships without the field, and strict clients reject it.

## The fix

The choice between `complete` and `input_required` now happens inside `setMultiRoundTripResultType`. Both call sites that need this choice call that function. `setCompleteResultType` now labels results it did not label before, so it is renamed `annotateResultType`. The new function picks a branch with a type switch.

Reading matters here. `setResultType` is unexported, so setting `InputRequests` is the only way for middleware to ask for input. If the code always labels results `complete`, the response ships as `{"resultType":"complete","inputRequests":{...}}`. A client that trusts the discriminator then drops the elicitation request.

Recomputing the label every time is safe. A result that did reach the dispatcher already carries the same label from that test. This also covers one edge case: a handler that returns `InputRequests` again after the shim's single reinvocation. Before this fix, that case shipped with no `resultType`.

Embedding `completeResultWithType` instead is not an option: it labels everything `complete` and overwrites `input_required`, breaking multi-round-trip flows.

**Skip nil results.** Middleware can return a typed nil value, which satisfies the `Result` interface. `handle` now checks for that case before it calls the labeling function. Before this change, that same case also panicked, one frame later in `annotateServerInfo` and then `GetMeta`.

## Checks

- `TestServerSessionHandle_SetsResultTypeWhenMiddlewareShortCircuits` fails on all three methods before the first change. Its fourth case fails if labeling overwrites a populated `InputRequests`.
- `TestServerSessionHandle_ResultTypeGate` records which requests carry `resultType`, including one known gap. See "Separate from this PR" below.
- `TestServerSessionHandle_NilResultFromMiddleware` panics without the nil check.
- The complete-path cases added to `TestServerSessionHandle_SetsResultTypeOnNewProtocol` also pass before this change. They close a gap in #1060, which added only three cases, and only for `input_required`.
- The exported API is unchanged. `go doc -all ./mcp`, with comments stripped, is identical before and after this change.
- With the reproduction from the issue, over the streamable transport: the dispatched `tools/call` and the short-circuited `tools/call` both now return `resultType: "complete"`.

I could not run `./scripts/conformance.sh`. CONTRIBUTING.md names that script, but the repo only has `client-conformance.sh` and `server-conformance.sh`. Both scripts need GNU `timeout`, and macOS does not include it.

## Also here

Ten comments named an exported `ResultType` field and `ResultTypeComplete` / `ResultTypeInputRequired` constants. None of those exist. Two of the comments were `[CallToolResult.ResultType]` doc links, and those links cannot resolve. They now point at `[CallToolResult.NeedsInput]` instead, because an unexported field is also not a valid link target.

### DIFF
--- a/mcp/mrtr.go
+++ b/mcp/mrtr.go
@@ -53,13 +53,19 @@ func handleMultiRoundTripResult(ss *ServerSession, logger *slog.Logger, res mult
 	if clientSupportsMultiRoundTrip(ss) {
 		// For older clients the resultType is left unset. Input requests will be handled
 		// by serverMultiRoundTripMiddleware client calls and handler reinvocation.
-		if hasInputRequests {
-			res.setResultType(resultTypeInputRequired)
-		} else {
-			res.setResultType(resultTypeComplete)
-		}
+		setMultiRoundTripResultType(res)
 	}
 	return nil
+}
+
+// setMultiRoundTripResultType labels res complete or input_required by
+// whether it carries input requests.
+func setMultiRoundTripResultType(res multiRoundTripResponse) {
+	if res.inputRequests() != nil {
+		res.setResultType(resultTypeInputRequired)
+	} else {
+		res.setResultType(resultTypeComplete)
+	}
 }
 
 func clientSupportsMultiRoundTrip(ss *ServerSession) bool {
--- a/mcp/protocol.go
+++ b/mcp/protocol.go
@@ -19,7 +19,7 @@ type resultType string
 
 const (
 	// resultTypeComplete indicates the result is final.
-	// This is the default when ResultType is empty.
+	// This is the default when resultType is empty.
 	resultTypeComplete resultType = "complete"
 
 	// resultTypeInputRequired indicates the server needs additional client
@@ -40,9 +40,16 @@ type completeResultResponse interface {
 	isCompleteResult()
 }
 
-func setCompleteResultType(res Result) {
-	if r, ok := res.(completeResultResponse); ok {
+// annotateResultType sets the resultType that protocol 2026-07-28 requires on
+// every result.
+func annotateResultType(res Result) {
+	switch r := res.(type) {
+	case completeResultResponse:
 		r.setResultType(resultTypeComplete)
+	case multiRoundTripResponse:
+		// A middleware can return one of these without reaching the
+		// dispatcher, leaving the field unset.
+		setMultiRoundTripResultType(r)
 	}
 }
 
@@ -226,9 +233,8 @@ type CallToolParams struct {
 	// marshaled to JSON.
 	Arguments any `json:"arguments,omitempty"`
 
-	// InputResponses maps input request IDs to responses, provided when
-	// retrying a call after receiving a result with ResultType
-	// ResultTypeInputRequired.
+	// InputResponses maps input request IDs to responses. Set it when retrying
+	// a call that asked for input. See [CallToolResult.NeedsInput].
 	InputResponses InputResponseMap `json:"inputResponses,omitempty"`
 	// RequestState is the opaque state from the previous input-required result.
 	// The client must echo this back when retrying.
@@ -249,9 +255,8 @@ type CallToolParamsRaw struct {
 	// Arguments (see [AddTool]).
 	Arguments json.RawMessage `json:"arguments,omitempty"`
 
-	// InputResponses maps input request IDs to responses, provided when
-	// retrying a call after receiving a result with ResultType
-	// ResultTypeInputRequired.
+	// InputResponses maps input request IDs to responses. Set it when retrying
+	// a call that asked for input. See [CallToolResult.NeedsInput].
 	InputResponses InputResponseMap `json:"inputResponses,omitempty"`
 	// RequestState is the opaque state from the previous input-required result.
 	// The client must echo this back when retrying.
@@ -305,7 +310,7 @@ type CallToolResult struct {
 	IsError bool `json:"isError,omitempty"`
 
 	// InputRequests is a map of server-assigned IDs to input requests.
-	// Populated only when ResultType is ResultTypeInputRequired.
+	// Set only when the result needs more client input.
 	// The client must fulfill these and echo the IDs back in InputResponses
 	// when retrying the call.
 	InputRequests InputRequestMap `json:"inputRequests,omitempty"`
@@ -317,10 +322,10 @@ type CallToolResult struct {
 	// Unauthenticated servers must encrypt, sign and verify this value.
 	RequestState string `json:"requestState,omitempty"`
 
-	// ResultType indicates whether this result is complete or requires further
-	// client input. Empty or ResultTypeComplete means the call succeeded
-	// normally. ResultTypeInputRequired means the client should fulfill the
-	// InputRequests and retry the call.
+	// resultType records whether the call finished or needs more client input.
+	// Empty or resultTypeComplete means it finished. resultTypeInputRequired
+	// means the client should fulfill InputRequests and retry.
+	// See [CallToolResult.NeedsInput].
 	resultType resultType
 	// The error passed to setError, if any.
 	// It is not marshaled, and therefore it is only visible on the server.
@@ -358,7 +363,8 @@ func (r *CallToolResult) GetError() error {
 	return r.err
 }
 
-func (*CallToolResult) isResult() {}
+func (*CallToolResult) isResult()     {}
+func (x *CallToolResult) isNil() bool { return x == nil }
 
 func (r *CallToolResult) setResultType(rt resultType) { r.resultType = rt }
 func (r *CallToolResult) requestState() string        { return r.RequestState }
@@ -373,7 +379,7 @@ func (r *CallToolResult) hasContent() bool {
 }
 
 // NeedsInput reports whether this result requires further client input.
-// This is true when the server returned ResultType "input_required".
+// This is true when the server returned a resultType of "input_required".
 // When NeedsInput returns true, check InputRequests for the set of
 // requests the server needs fulfilled before retrying the call.
 // An empty InputRequests with NeedsInput true indicates load-shedding.
@@ -665,7 +671,8 @@ type CompleteResult struct {
 	Completion CompletionResultDetails `json:"completion"`
 }
 
-func (*CompleteResult) isResult() {}
+func (*CompleteResult) isResult()     {}
+func (x *CompleteResult) isNil() bool { return x == nil }
 
 // CreateMessageParams holds parameters for a sampling/createMessage request.
 //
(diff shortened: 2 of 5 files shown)


mcp/server.go
```diff
@@ -2039,8 +2039,13 @@ func (ss *ServerSession) handle(ctx context.Context, req *jsonrpc.Request) (any,
 	if err != nil {
 		return nil, err
 	}
+	// A superset of the test handleMultiRoundTripResult uses, so a result that
+	// bypassed it -- one a middleware returned without calling next -- is
+	// never labeled less than one that did not.
+	if validatedMeta.usesNewProtocol || clientSupportsMultiRoundTrip(ss) {
+		annotateResultType(res)
+	}
```
guglielmo-san: i don't this is necessary. clientSupportsMultiRoundTrip depends on validatedMeta.usesNewProtocol

mcp/mrtr.go
```diff
@@ -53,15 +53,21 @@ func handleMultiRoundTripResult(ss *ServerSession, logger *slog.Logger, res mult
...
 }
 
+// setMultiRoundTripResultType labels res complete or input_required by
+// whether it carries input requests.
+func setMultiRoundTripResultType(res multiRoundTripResponse) {
+	if res.inputRequests() != nil {
+		res.setResultType(resultTypeInputRequired)
+	} else {
+		res.setResultType(resultTypeComplete)
+	}
```
guglielmo-san, line 68: probably a check on res != nil is needed in this case, in case a middleware returns a nil result

mcp/server.go
roncodingenthusiast: yeah after second thought I realize the thing I was trying to guard here would not happen in practice.

mcp/mrtr.go
roncodingenthusiast: good call, updating it

roncodingenthusiast:
@guglielmo-san i am assuming you will take care of merging this on your end? not sure if there is anything else i should do other than just wait for you to review, approve/merge

mcp/mrtr.go
```diff
@@ -53,15 +53,21 @@ func handleMultiRoundTripResult(ss *ServerSession, logger *slog.Logger, res mult
 	if clientSupportsMultiRoundTrip(ss) {
 		// For older clients the resultType is left unset. Input requests will be handled
 		// by serverMultiRoundTripMiddleware client calls and handler reinvocation.
-		if hasInputRequests {
-			res.setResultType(resultTypeInputRequired)
-		} else {
-			res.setResultType(resultTypeComplete)
-		}
+		setMultiRoundTripResultType(res)
 	}
```
guglielmo-san, line 57: why do we still need to set this inside mrtr ? Now it should be enough the single call from handle

roncodingenthusiast: We cannot remove this call from the code as it is now. The dispatcher reads `resultType` before `handle` runs.

- [L1018](https://github.com/modelcontextprotocol/go-sdk/blob/aa5b8ca9b0cc3dfec92517cac25c870104ceb2fa/mcp/server.go#L1018) keeps `"content"` in the JSON response as `null` and not `[]`.
- [L1087](https://github.com/modelcontextprotocol/go-sdk/blob/aa5b8ca9b0cc3dfec92517cac25c870104ceb2fa/mcp/server.go#L1087) returns early, before the Go check `res.Contents == nil` gives an error.

Removing the mrtr call breaks a bunch of pre-existing tests: (eg. `TestServerConformance/mrtr.txtar`, the three `TestMultiRoundTrip_ReadResource_*` tests...).

To remove it, I would also have to change the two dispatcher lines to read `InputRequests` and not `resultType`. I tried this approach and it seemed to work.  The result is cleaner, but it changes `input_required` detection on paths that are not related to #1225 or the current PR.

I am happy to do it in this PR if you want or keep this as is. What do you think?

guglielmo-san: I think it would look cleaner, and add it to this PR.
Thanks!

roncodingenthusiast: done ✅

roncodingenthusiast:
@guglielmo-san anything else needed for this? I have a small patch I made to my MCP so i am trying to figure when this will be release so i can remove the customization on my end

mcp/protocol.go
```diff
@@ -40,9 +40,20 @@ type completeResultResponse interface {
 	isCompleteResult()
 }
 
-func setCompleteResultType(res Result) {
-	if r, ok := res.(completeResultResponse); ok {
+// annotateResultType sets the resultType that protocol 2026-07-28 requires on
+// every result.
+func annotateResultType(res Result) {
```
guglielmo-san: I would rename it as "setResultType"

guglielmo-san (approved):

guglielmo-san:
@roncodingenthusiast thank you for the contribution! As we just released v1.8.0, this code change will be included in next release that will happen in ~2 months

PR #1228, opened 2026-09-02
mcp: notify sessions concurrently so one stalled peer cannot starve the rest

### LINKED ISSUE #1227: mcp: ResourceUpdated notifies subscribers serially, so one stalled stream write starves every other subscriber of the URI
### Describe the bug

When several sessions are subscribed to the same resource URI, `Server.ResourceUpdated` delivers the notification to them one at a time, and one slow session delays or fails delivery to all the others.

The delivery loop is `notifySessions` (`mcp/shared.go:452`) and `notifySubscribedSessions` (`mcp/server.go:768`):

```go
ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second) // one deadline for everyone
for _, s := range sessions {                                               // one session at a time
    handleNotify(ctx, method, req)
}
```

Two problems follow.

**1. Sessions are not isolated from each other.** Session B is only written to after session A's write returns. If A is slow, B waits. If A is slow enough that the shared 10s context expires, B's `handleNotify` fails immediately with the deadline error and B receives nothing for this update. Since `sessions` comes from a map, the order is random, so a different subscriber loses out on each call.

**2. The 10s deadline does not bound the slow write.** On the streamable HTTP transport the write ends in `stream.deliverLocked` (`mcp/streamable.go:1079`), which calls `s.w.Write(data)` on the request's `http.ResponseWriter` directly. That call takes no context. If the client has stopped reading its SSE response and the socket buffers are full, `Write` blocks until the client reads again, the connection dies, or the server's `WriteTimeout` fires — minutes, not 10 seconds. The whole loop, and every subscriber after A, is stuck for that long.

Concrete sequence:

1. Clients A and B subscribe to the same URI via `subscriptions/listen`.
2. A stops reading its stream but keeps the connection open. After enough notifications its socket buffers fill up.
3. The next `ResourceUpdated(uri)` writes to A first and blocks.
4. B gets nothing — for this notification, and for every later one, until A's connection finally dies.

When the write blocks: `Write` returns as soon as the bytes are in the kernel send buffer, so a client that stops reading only stalls the loop once its receive buffer and the server's send buffer are both full. With small notifications that takes many writes, so the trigger is a slow-but-connected consumer or a long-lived half-open connection rather than any brief hiccup. But once it does block, nothing bounds it.

The code already anticipates this: `shared.go:458` says `// TODO: make this timeout configurable, or call handleNotify asynchronously.`

Observed on v1.7.0; the code paths are unchanged on current `main`.

### Why it matters

A server keeping `subscriptions/listen` streams alive through an idle-timeout proxy has to write *something* to each stream periodically. The 2026-07-28 Streamable HTTP binding recommends a periodic SSE comment for exactly this, but the SDK has no way to emit one (#1229); `ping` is removed from the protocol in that revision, and stateless servers cannot make requests anyway (`streamable.go:1796`). So the only per-stream write a server author has today is `ResourceUpdated`. Under the behaviour above, one unresponsive client on a URI silences that keepalive for every other client on the same URI until a proxy closes their streams as well. #1229 would remove that particular need, but the serial delivery affects real change notifications just the same.

### Expected behavior

One session's slow write should affect only that session. For example, `notifySessions` / `notifySubscribedSessions` could start one goroutine per session, each with its own deadline, and wait for all of them (keeping today's "returns after attempting every session" behaviour). Separately, the stream write could honour the context, or `deliverLocked` could be arranged so a blocked write on one stream cannot hold up writes to other streams.

A per-session notification API (#745 / #1146) would also let server authors work around this by notifying each subscriber with its own context.

### To Reproduce

Two clients subscribed to one URI. Make the first stop reading its response (a raw `subscriptions/listen` POST whose body is never read), call `ResourceUpdated` on that URI until its socket buffers are full, and measure when the second, healthy client stops receiving notifications.

### Version

go-sdk v1.7.0 (`mcp/shared.go:452-470`, `mcp/server.go:764-782` and `:1114-1140`, `mcp/streamable.go:1079-1130`).

yhxlele:
`notifySessions` and `Server.notifySubscribedSessions` delivered a broadcast to its subscribers one session at a time, all under a single shared 10s context. A session whose write did not return promptly held up every session after it, and once the shared context expired the remaining sessions failed with the deadline error without ever being attempted. Because the streamable transport's stream write is a plain `http.ResponseWriter` write that does not observe the context, a peer that had stopped reading could hold the loop far longer than 10s. Details in #1227.

This sends to each session on its own goroutine with its own deadline (the same 10s, now a named constant), and waits for all attempts before returning, so `ResourceUpdated` and the list-changed broadcasts keep today's "returns after attempting every session" behaviour. A stalled peer now delays or fails only its own delivery.

Not changed here: the stalled write itself is still unbounded (the stream write does not honour the context). That is a transport-level change and can be taken separately; this PR removes the coupling between subscribers, which is the part that affected healthy peers.

## Test plan

- [x] `go test ./mcp/ -count=1 -race`
- [x] `TestNotifySessionsIsolatesStalledPeer` — one in-memory peer never reads its end of the pipe so the server's write to it blocks; a second, healthy session must still receive the notification while the first is blocked. Fails against the serial implementation (`healthy session was not notified while another session's write was stalled`).
- [x] `go vet ./mcp/`

Fixes #1227

### DIFF
--- a/mcp/server.go
+++ b/mcp/server.go
@@ -811,16 +811,24 @@ func (s *Server) notifySubscribedSessions(subscribers map[*ServerSession]jsonrpc
 	if len(subscribers) == 0 {
 		return
 	}
-	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
-	defer cancel()
+	// One goroutine per session, each with its own deadline, so a session
+	// whose write stalls neither delays nor fails the others (as in the
+	// notifySessions function in shared.go).
+	var wg sync.WaitGroup
 	for sess, reqID := range subscribers {
-		params := makeParams()
-		injectMetaSubscriptionID(params, reqID)
-		req := newRequest(sess, params)
-		if err := handleNotify(ctx, method, req); err != nil {
-			s.opts.Logger.Warn(fmt.Sprintf("calling %s: %v", method, err))
-		}
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			ctx, cancel := context.WithTimeout(context.Background(), notifyTimeout)
+			defer cancel()
+			params := makeParams()
+			injectMetaSubscriptionID(params, reqID)
+			if err := handleNotify(ctx, method, newRequest(sess, params)); err != nil {
+				s.opts.Logger.Warn(fmt.Sprintf("calling %s: %v", method, err))
+			}
+		}()
 	}
+	wg.Wait()
 }
 
 // injectMetaSubscriptionID stamps the listen request's JSON-RPC ID into the
--- a/mcp/server_test.go
+++ b/mcp/server_test.go
@@ -1932,3 +1932,76 @@ func TestServerSupportedProtocolVersions_NewProtocol(t *testing.T) {
 		t.Errorf("UnsupportedProtocolVersionData.Supported mismatch (-want +got):\n%s", diff)
 	}
 }
+
+// TestNotifySessionsIsolatesStalledPeer verifies that a session whose write
+// stalls — here a peer that never reads its end of the pipe — does not delay
+// or fail delivery to the other sessions in the same broadcast.
+func TestNotifySessionsIsolatesStalledPeer(t *testing.T) {
+	ctx := context.Background()
+	server := NewServer(testImpl, nil)
+
+	// The stalled session: nothing reads the client end until the end of the
+	// test, so the server's first write blocks (net.Pipe is synchronous).
+	stalledCT, stalledST := NewInMemoryTransports()
+	stalled, err := server.Connect(ctx, stalledST, nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// The healthy session: a real client that records the notification.
+	got := make(chan string, 1)
+	healthyCT, healthyST := NewInMemoryTransports()
+	healthy, err := server.Connect(ctx, healthyST, nil)
+	if err != nil {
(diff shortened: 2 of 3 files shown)


mcp/server.go
```diff
@@ -811,16 +811,24 @@ func (s *Server) notifySubscribedSessions(subscribers map[*ServerSession]jsonrpc
...
-		req := newRequest(sess, params)
-		if err := handleNotify(ctx, method, req); err != nil {
-			s.opts.Logger.Warn(fmt.Sprintf("calling %s: %v", method, err))
-		}
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			ctx, cancel := context.WithTimeout(context.Background(), notifyTimeout)
+			defer cancel()
+			params := makeParams()
```
guglielmo-san, line 824: executing this in parallel now introduces a race, as makeParams returns a shallow copy

yhxlele: Good catch. Fixed by cloning meta in injectMetaSubscriptionID.

yhxlele:
Conflict resolved.

yhxlele:
@guglielmo-san Whenever you get a chance, this one should be ready for another look. Thanks for the earlier catch.

guglielmo-san (approved):

guglielmo-san:

PR #1232, opened 2026-09-03
mcp: add StreamableHTTPOptions.StreamKeepAlive for periodic SSE comments on idle streams

### LINKED ISSUE #1229: mcp: add StreamableHTTPOptions.StreamKeepAlive — the SSE keep-alive comment the 2026-07-28 spec encourages for subscriptions/listen

The 2026-07-28 Streamable HTTP spec ("Receiving Messages") says:

> For long-lived streams — in particular the `subscriptions/listen` response stream — servers are encouraged to periodically emit an SSE comment line (a line beginning with a colon, e.g. `:\r\n`) as a keep-alive. This keeps the connection from being closed by intermediaries or client idle timeouts during quiet periods when no notifications are flowing. Per the SSE specification, any line beginning with a colon is a comment that carries no event data; clients must ignore such lines and must not treat them as malformed input.

go-sdk cannot do this. The stream's `http.ResponseWriter` is private to `stream`, and all writes go through `deliverLocked` under `stream.mu`, so a server author has no safe way to write a comment.

Without it, a quiet `subscriptions/listen` stream is closed by any idle-timeout proxy. The server just sees a cancelled request and unsubscribes. The client's `Subscribe` already returned, the stream ends without error, and `resourceSubs[uri]` stays set (`mcp/client.go:1393`), so even calling `Subscribe` again is a no-op. The application thinks it is subscribed and never hears anything again.

Nothing else in 2026-07-28 covers this: `ping` is removed from the protocol, `Last-Event-ID` resumption is removed, and stateless servers cannot send requests (`mcp/streamable.go:1798`). The only thing a server can write to the stream today is a notification, so servers end up sending fake `notifications/resources/updated` as a heartbeat. That makes every client re-read an unchanged resource and runs into #1227.

```go
// StreamKeepAlive, if non-zero, writes an SSE comment (":\n\n") and
// flushes on any SSE stream that has been idle for this duration, as
// the 2026-07-28 Streamable HTTP spec encourages for long-lived streams.
// Zero (the default) disables it.
StreamKeepAlive time.Duration
```

- Per stream, timer reset on every write.
- SSE streams only, after headers are committed (composes with #1197's delayed first flush).
- Written under `stream.mu` so it never interleaves with an event.
- A failed write is a disconnect: release the stream and cancel the request context.
- No client change needed: `scanEvents` already ignores comment lines, and the SDK already writes `: ok` on the GET stream (#410).

- Fake `resources/updated` heartbeat: works, wrong layer (see above).
- Client re-subscribe on stream end: worth having too (the spec says the client MAY reconnect on an abrupt close, and `resourceSubs` should be cleared when the stream ends), but it does not prevent the drop or the notifications lost in between.
- Exposing the `ResponseWriter`: every server behind a proxy needs the same thing, and the transport already owns the lock and the write path.

**Additional context**

The same spec section says servers SHOULD send `X-Accel-Buffering: no` on SSE responses; `acquireStream` does not. Easy to add at the same time.

Observed on v1.7.0; the code is the same on current `main`.

yhxlele:
The 2026-07-28 Streamable HTTP spec encourages servers to periodically emit an SSE comment line on long-lived streams, in particular the `subscriptions/listen` response, so that idle-timeout intermediaries do not sever them. The SDK had no way to do this: the stream's `ResponseWriter` is private and every write goes through `deliverLocked` under `stream.mu`. Servers behind a proxy ended up sending fake `notifications/resources/updated` as a heartbeat instead.

This adds `StreamableHTTPOptions.StreamKeepAlive time.Duration`. The response stream of a `subscriptions/listen` request gets a goroutine that writes the bare `:\n\n` comment from the spec example and flushes whenever the stream has carried no bytes for that duration. Zero selects `DefaultStreamKeepAlive` (30s), so servers follow the spec out of the box; a negative value disables the keep-alive. This is the same convention as `MaxRequestBodyBytes`. Other SSE responses are not kept alive; the standalone GET stream and `EventStore` paths are untouched.

Semantics:

- Idle-reset per stream. `stream.lastWrite` is stamped by every write in `deliverLocked`; the goroutine sleeps until `lastWrite + interval`, so a busy stream gets no comments and a quiet one gets exactly one per interval.
- Nothing is written before the listen acknowledgment. `deliverLocked` may still need to set a 400/404 status for a SEP-2575 error, which requires uncommitted headers, so the goroutine parks on a `committed` channel that the first write closes. The acknowledgment arrives within milliseconds, so this costs nothing.
- Writes happen under `stream.mu`, so a comment can never interleave with an event.
- A failed write closes the stream's `done` channel: `hangResponse` returns, the request context is cancelled, and the listen handler unwinds and unsubscribes. A dead peer is therefore noticed within one interval rather than at the next real notification.
- `X-Accel-Buffering: no` is set on SSE POST responses, which the same spec section recommends; the header is dropped again on the JSON error-override path.

Tests: a raw listen POST sees the ack and then only comments; a slow `tools/call` gets no comment at all; the option resolution (zero, negative, explicit, nil options) is checked; the goroutine semantics (park, idle-reset, failed write closes the stream) are tested directly; keep-alive goroutines end with their streams; and an end-to-end test puts an idle-timeout proxy in front of the server and checks that a quiet subscription dies with the keep-alive disabled and survives with it, with the SDK client ignoring the comments.

Fixes #1229

### DIFF
--- a/mcp/streamable.go
+++ b/mcp/streamable.go
@@ -218,6 +218,21 @@ type StreamableHTTPOptions struct {
 	// Requests using older protocol versions (including those routed through
 	// the allowsessionsinstateless compatibility path) are unaffected.
 	PropagateRequestCancellation bool
+
+	// StreamKeepAlive, if non-zero, writes an SSE comment to any SSE response
+	// stream that has carried no bytes for this duration, so that idle-timeout
+	// intermediaries do not sever long-lived streams such as the response to a
+	// subscriptions/listen request. The 2026-07-28 Streamable HTTP
+	// specification encourages this keep-alive; SSE clients ignore comment
+	// lines, so it has no protocol-level effect.
+	//
+	// On streams using protocol version 2026-07-28 or later, the keep-alive
+	// starts only after a first event has committed the response headers,
+	// since the HTTP status may still have to change (see #1229). A write
+	// failure ends the stream as a disconnect.
+	//
+	// If StreamKeepAlive is the zero value, no keep-alive is written.
+	StreamKeepAlive time.Duration
 }
 
 // DefaultMaxRequestBodyBytes is the default value used for
@@ -452,6 +467,7 @@ func (h *StreamableHTTPHandler) serveStateless(w http.ResponseWriter, req *http.
 		Stateless:                   true,
 		EventStore:                  h.opts.EventStore,
 		jsonResponse:                h.opts.JSONResponse,
+		streamKeepAlive:             h.opts.StreamKeepAlive,
 		logger:                      h.opts.Logger,
 		shouldPropagateCancellation: info.usesNewProtocol && (info.isSubscriptionsListen || h.opts.PropagateRequestCancellation),
 	}
@@ -678,11 +694,12 @@ func (h *StreamableHTTPHandler) serveStatefulPOST(w http.ResponseWriter, req *ht
 	sessionID = server.opts.GetSessionID()
 
 	transport := &StreamableServerTransport{
-		SessionID:    sessionID,
-		Stateless:    false,
-		EventStore:   h.opts.EventStore,
-		jsonResponse: h.opts.JSONResponse,
-		logger:       h.opts.Logger,
+		SessionID:       sessionID,
+		Stateless:       false,
+		EventStore:      h.opts.EventStore,
+		jsonResponse:    h.opts.JSONResponse,
+		streamKeepAlive: h.opts.StreamKeepAlive,
+		logger:          h.opts.Logger,
 	}
 
 	// Sessions without a session ID (GetSessionID returned "") are ephemeral:
@@ -847,6 +864,13 @@ type StreamableServerTransport struct {
 	// to write their own streamable HTTP handler.
 	jsonResponse bool
 
+	// streamKeepAlive is the idle interval after which an SSE comment is
+	// written to a stream; see [StreamableHTTPOptions.StreamKeepAlive].
+	//
+	// TODO: streamKeepAlive should be exported, like jsonResponse and logger,
+	// once users can write their own streamable HTTP handler.
+	streamKeepAlive time.Duration
+
 	// optional logger provided through the [StreamableHTTPOptions.Logger].
 	//
 	// TODO(rfindley): logger should be exported, since we want to allow users
@@ -871,6 +895,7 @@ func (t *StreamableServerTransport) Connect(ctx context.Context) (Connection, er
 		stateless:                   t.Stateless,
 		eventStore:                  t.EventStore,
 		jsonResponse:                t.jsonResponse,
+		streamKeepAlive:             t.streamKeepAlive,
 		logger:                      ensureLogger(t.logger), // see #556: must be non-nil
 		shouldPropagateCancellation: t.shouldPropagateCancellation,
 		incoming:                    make(chan jsonrpc.Message, 10),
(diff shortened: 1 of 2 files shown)


yhxlele:
Verified end to end against a real deployment rather than only `httptest`: a Go server on this branch with `StreamKeepAlive: 30 * time.Second`, its own periodic notification heartbeat disabled, one raw `subscriptions/listen` POST observing wire lines, and a go-sdk v1.7.0 client subscribed to the same resource. After a quiet period a database update fires the server's real change notification.

| Edge in front of the server | Quiet period | Wire | Delivery after the update |
| --- | --- | --- | --- |
| Traefik (no idle timeout) | 150s | ack, then `: keepalive` at +30.0s, +60.0s, +90.0s, +120.0s, +150.0s; no other bytes | 326ms, stream still open |
| nginx `proxy_read_timeout 45s` | 100s | ack, then `: keepalive` at +30s, +60s, +90s | 279ms, stream still open |
| nginx `proxy_read_timeout 20s` (control, shorter than the interval) | — | ack only; nginx closed the stream at +21.5s (`unexpected EOF`) | none: the subscription was gone |

So the comment resets a real proxy's read timer, the idle-reset spacing is exact (each comment lands `interval` after the previous byte, not on a fixed clock), the go-sdk client ignores the comments and keeps working, and without bytes the same proxy drops a quiet listen stream before the first interval. nginx also honoured and stripped `X-Accel-Buffering: no` (present on the direct connection, absent behind nginx), as expected.

guglielmo-san:
@yhxlele can you target for now only the subscriptions/listen case in the PR for now, discarding what is related to deprecated features (GET stream, EventStore..)

yhxlele:
Sure, deprecated feature related modification discarded.

mcp/streamable.go
```diff
@@ -1032,6 +1071,86 @@ func (s *stream) release() {
...
+		if s.done == nil {
+			s.mu.Unlock()
+			return
+		}
+		if wait := interval - time.Since(s.lastWrite); !s.lastWrite.IsZero() && wait > 0 {
+			s.mu.Unlock()
+			timer.Reset(wait)
+			continue
+		}
+		_, err := fmt.Fprint(s.w, ": keepalive\n\n")
```
guglielmo-san: let's remove the keepalive to keep it matching with the example in the spec

mcp/streamable.go
```diff
@@ -217,6 +217,22 @@ type StreamableHTTPOptions struct {
...
+	// lines, so it has no protocol-level effect. Other SSE responses are not
+	// kept alive.
+	//
+	// The keep-alive starts only after the listen acknowledgment has
+	// committed the response headers, since until then the HTTP status may
+	// still have to change (see #1229). A write failure ends the stream as a
+	// disconnect.
+	//
+	// If StreamKeepAlive is the zero value, no keep-alive is written.
+	StreamKeepAlive time.Duration
```
guglielmo-san, line 234: i would add a default value to use when this is set to zero, maybe 30 seconds, and -1 to disable it

mcp/streamable.go
yhxlele: Sure

mcp/streamable.go
yhxlele: Default to 30 seconds

guglielmo-san (approved):

guglielmo-san:

PR #1239, opened 2026-09-04
mcp: preserve structured content number precision

### LINKED ISSUE #1236: mcp: CallToolResult structuredContent loses large integer precision
**Describe the bug**

`mcp.ClientSession.CallTool` silently rounds large JSON integers in a tool result's `structuredContent` field.

This is the result-side equivalent of #1201. It also affects users of the low-level `Server.AddTool` API, so avoiding the typed tool argument path does not avoid this result decode.

**Environment**

- Go MCP SDK: `v1.7.0`
- Go: `go1.27.0`

**Observed behavior**

A server returns this valid JSON value in `structuredContent`:

```json
{"id":9007199254740993}
```

The official SDK client exposes the value as a `float64` with the value `9007199254740992`. No error is returned.

The loss occurs while the SDK decodes the JSON-RPC result into `CallToolResult.StructuredContent any`. Application code cannot recover the original integer after this decode.

**Expected behavior**

Generic JSON numbers in `structuredContent` retain their exact wire value. A decoder that uses `json.Number` or an exact raw JSON representation would avoid silent rounding.

**Why this needs an SDK fix**

A server bridge cannot repair the value after the official SDK has decoded the request or result envelope. Re-reading HTTP bodies would violate transport ownership and does not help in-memory transports. The SDK needs an exact-number decode path for generic JSON fields in parsed protocol envelopes.

CaliLuke:
CallToolResult currently decodes structuredContent through an any field, which converts JSON numbers to float64 and silently rounds integers outside the IEEE-754 safe range.

This change decodes structuredContent separately with UseNumber, preserving wire numbers as json.Number while retaining the existing decoded object, array, and primitive shapes. The field documentation now describes that representation. Users who depend on the previous float64 behavior can temporarily restore it with MCPGODEBUG=structuredcontentfloat64=1.

An end-to-end in-memory client/server regression covers 9007199254740993 from a low-level Server.AddTool result. Additional coverage verifies the compatibility option restores the legacy rounded float64 value.

Verification:
- gofmt -l .
- go test ./...
- go vet ./...

Fixes #1236

### DIFF
--- a/mcp/protocol.go
+++ b/mcp/protocol.go
@@ -280,6 +280,8 @@ type CallToolResult struct {
 	// result of the tool call. Per SEP-2106, it may marshal to any valid JSON
 	// value (object, array, or primitive) conforming to the tool's
 	// [Tool.OutputSchema].
+	// Numbers received from the wire are represented as [json.Number] to preserve
+	// their exact values.
 	//
 	// When using a [ToolHandlerFor] with structured output, you should not
 	// populate this field. It will be automatically populated with the typed Out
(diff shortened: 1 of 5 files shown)


mcp/protocol.go
```diff
@@ -401,12 +403,18 @@ func (x *CallToolResult) UnmarshalJSON(data []byte) error {
...
+		ResultType        resultType      `json:"resultType"`
 	}
 	if err := internaljson.Unmarshal(data, &wire); err != nil {
 		return err
 	}
+	if len(wire.StructuredContent) > 0 {
+		if err := internaljson.UnmarshalUseNumber(wire.StructuredContent, &wire.res.StructuredContent); err != nil {
+			return err
+		}
+	}
```
guglielmo-san, line 427: this is a behavioral change for existing users. Should be gated behind an opt-out MCPGODEBUG variable

CaliLuke: Addressed in 0769258. The exact-number behavior remains the default, and MCPGODEBUG=structuredcontentfloat64=1 restores the previous float64 decoding. I added coverage for the legacy rounded result and documented the temporary option for 1.9.0.

guglielmo-san (approved):

guglielmo-san:

PR #1242, opened 2026-09-05
mcp: decode a Base64-wrapped Mcp-Name before comparing it to the body

### LINKED ISSUE #1234: Streamable HTTP: a Base64-encoded Mcp-Name is rejected as a header mismatch instead of decoded
**Describe the bug**

`validateMcpHeaders` compares `Mcp-Name` with the body value as the raw header string. The 2026-07-28 spec lets a client Base64-wrap a name or resource URI that is not header-safe, and requires the server to decode it before comparing; the SDK decodes `Mcp-Param-*` that way but not `Mcp-Name`, so an encoded name is refused as a header mismatch.

`mcp/streamable_headers.go` (main): line 377 reads `nameInHeader := header.Get(nameHeader)` and compares it directly; `decodeHeaderValue` at line 496 exists and is applied only to `Mcp-Param-*` values at line 431.

Transport page, "Value encoding":

> Servers and intermediaries that need to inspect these values MUST decode them accordingly. In particular, servers MUST decode an encoded `Mcp-Name` or `Mcp-Param-{Name}` value before comparing it to the corresponding request body value during Server Validation.

Tool names are usually header-safe, so this mostly bites `resources/read`: a resource template with a path argument that contains non-ASCII, which the spec allows, forces a conformant client to encode the URI, and every such read is then rejected.

**To Reproduce**

Register a tool named `fast`, then:

```
curl -s -X POST http://127.0.0.1:8080/mcp \
  -H 'Mcp-Protocol-Version: 2026-07-28' \
  -H 'Mcp-Method: tools/call' \
  -H 'Mcp-Name: =?base64?ZmFzdA==?=' \
  -d '{"jsonrpc":"2.0","id":1,"method":"tools/call","params":{"name":"fast","arguments":{},"_meta":{"io.modelcontextprotocol/protocolVersion":"2026-07-28","io.modelcontextprotocol/clientInfo":{"name":"probe","version":"1"},"io.modelcontextprotocol/clientCapabilities":{}}}}'
```

Observed: `400` with `{"code":-32020,"message":"header mismatch: Mcp-Name header value '=?base64?ZmFzdA==?=' does not match body value 'fast'"}`.

**Expected behavior**

The header decodes to `fast`, matches the body, and the call runs. The same `decodeHeaderValue` already used for `Mcp-Param-*` would do it.

**Additional context**

go-sdk v1.7.0; unchanged on main. Working around it with a handler in front of the SDK's that decodes the header before delegating.

arpankernel:
The 2026-07-28 transport spec lets a client Base64-wrap a name or resource URI that is not header-safe (`=?base64?...?=`) and requires the server to decode it before comparing. `validateMcpHeaders` decoded `Mcp-Param-*` values that way but compared `Mcp-Name` raw, so a spec-legal encoded name was refused as a header mismatch.

This decodes `Mcp-Name` before comparing, mirroring the existing `Mcp-Param-*` path, and rejects an invalid Base64 wrapper with a clear error. Added table cases covering a valid encoded name and an invalid one (red before, green after).

Fixes #1234

### DIFF
--- a/mcp/streamable_headers.go
+++ b/mcp/streamable_headers.go
@@ -378,13 +378,18 @@ func validateMcpHeaders(header http.Header, msg jsonrpc.Message, toolLookup func
 			if nameInHeader == "" {
 				return fmt.Errorf("missing required Mcp-Name header for method %q", msg.Method)
 			}
-			var ok bool
+			// A name that is not header-safe may be Base64-wrapped by the client
+			// (=?base64?...?=), so decode before comparing, as with Mcp-Param-*.
+			decodedName, ok := decodeHeaderValue(nameInHeader)
+			if !ok {
+				return errors.New("header mismatch: Mcp-Name header contains invalid Base64 encoding")
+			}
 			nameInBody, ok = extractName(msg.Method, msg.Params)
 			if !ok {
 				return fmt.Errorf("failed to extract name from parameters for method %q", msg.Method)
 			}
-			if nameInHeader != nameInBody {
-				return fmt.Errorf("header mismatch: Mcp-Name header value '%s' does not match body value '%s'", nameInHeader, nameInBody)
+			if decodedName != nameInBody {
+				return fmt.Errorf("header mismatch: Mcp-Name header value '%s' does not match body value '%s'", decodedName, nameInBody)
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


guglielmo-san (approved):

guglielmo-san:

PR #1244, opened 2026-09-06
mcp: preserve JSON number precision in tool argument schemas

### LINKED ISSUE #1201: mcp: typed tool arguments lose precision for large int64 values
**Describe the bug**

When using the typed `mcp.AddTool` API, an integer larger than the IEEE-754
safe integer range can be silently rounded before it reaches the tool handler.
The value is accepted as a JSON integer, but the handler receives a different
value.

**To Reproduce**

Environment:

- Go MCP SDK: `v1.7.0`
- Go: `go1.25.7 linux/amd64`

Save the following as `main.go` in a module that requires
`github.com/modelcontextprotocol/go-sdk v1.7.0`:

```go
package main

import (
  "context"
  "fmt"
  "log"

)

type input struct {
  ID int64 `json:"id"`
}

func main() {
  const want int64 = 9007199254740993 // 2^53 + 1

  server := mcp.NewServer(
    &mcp.Implementation{Name: "server", Version: "repro"}, nil)
  mcp.AddTool(server, &mcp.Tool{Name: "echo_id"},
    func(_ context.Context, _ *mcp.CallToolRequest, in input) (
      *mcp.CallToolResult, input, error) {
      fmt.Printf("handler_received=%d\n", in.ID)
      return nil, in, nil
    })

  clientTransport, serverTransport := mcp.NewInMemoryTransports()
  ctx := context.Background()

  serverSession, err := server.Connect(ctx, serverTransport, nil)
  if err != nil {
    log.Fatal(err)
  }
  defer serverSession.Close()

  client := mcp.NewClient(
    &mcp.Implementation{Name: "client", Version: "repro"}, nil)
  clientSession, err := client.Connect(ctx, clientTransport, nil)
  if err != nil {
    log.Fatal(err)
  }
  defer clientSession.Close()

  _, err = clientSession.CallTool(ctx, &mcp.CallToolParams{
    Name: "echo_id",
    Arguments: map[string]any{"id": want},
  })
  if err != nil {
    log.Fatal(err)
  }
  fmt.Printf("want=%d\n", want)
}
```

Run:

```sh
go mod init example.com/mcp-int64-repro
go get github.com/modelcontextprotocol/go-sdk@v1.7.0
go run .
```

Observed output:

```text
handler_received=9007199254740992
want=9007199254740993
```

The issue is reproducible with the in-memory transport and does not depend on
an external server.

**Expected behavior**

The typed tool handler should receive exactly the integer sent by the client:

```text
handler_received=9007199254740993
want=9007199254740993
```

Values representable by Go `int64` should not be silently changed during
schema validation or default application.

**Logs**

No error is returned. The value is silently rounded.

**Additional context**

The behavior appears to come from the schema application path in
`mcp/tool.go`: tool arguments are decoded into `map[string]any`, and the
generic JSON number is then represented as `float64`. Applying defaults causes
the value to be marshaled again, making the rounding observable before the
typed handler decodes it.

A possible implementation direction is to preserve JSON numbers (for example,
using `Decoder.UseNumber()`) while applying the schema and defaults, while
retaining strict validation of trailing JSON data. The low-level
`Server.AddTool` handler path does not have this particular schema round-trip,
but it also does not provide the typed validation/default behavior.

This affects any tool that accepts IDs or counters outside the IEEE-754 safe
integer range, even though those values are valid JSON integers and valid Go
`int64` values.

jstar0:
Fixes #1201.

## The bug

Tool arguments and schema-validated outputs pass through `applySchema`. Decoding JSON numbers into `map[string]any` with the default `encoding/json` behavior converts them to `float64`; when defaults require a re-marshal, integers outside the exact `float64` range can be silently rounded. Using `json.Number` directly for schema validation is not compatible with the current `jsonschema-go` type classification, so the original number must be preserved separately from the validation representation.

## The fix

- Decode schema-processed values with `internal/json.UnmarshalUseNumber`, retaining the existing nesting-depth guard.
- Validate a `float64`-converted copy to preserve current JSON Schema behavior.
- When defaults require re-marshalling, serialize the preserved JSON number literals instead of rounded `float64` values.
- Keep exported API types unchanged.

## Verification

- `go test ./...` — passed on the narrowed head `fd4681d`.
- `go vet ./...` — passed.
- `gofmt -d internal/json/json.go mcp/server_test.go mcp/tool.go mcp/tool_test.go` — no changes.
- `git diff --check` — passed.
- `TestApplySchemaNumberPrecision` covers unsafe integers, nested data, arrays, and precise numbers when defaults force a re-marshal.
- `TestApplySchemaOutputNumberPrecision` covers the output-schema default path.
- `TestApplySchemaNumberErrors` preserves errors for unrepresentable values and invalid types.
- `TestToolArgumentIntegerPrecision` exercises the typed handler through an in-memory client/server call.

Hosted checks for exact head `fd4681dd320c08e0e3fcc86db77fbf2b9183777c` passed: CodeQL, Analyze (actions/go), client and server conformance, lint, race-test, and Test on Go 1.25 and 1.26.


jstar0:
Rebased onto `main` at `a5026ea` and reworked the change against the current `applySchema`, which has been restructured since I first wrote this (the `forOutput` parameter, unmarshalling output into `any`, and re-marshalling only when defaults were applied).

The bug survives that restructuring on both paths — confirmed on `a5026ea` before touching anything:

```
input  path: {"id":9007199254740992}
output path: {"id":9007199254740992}
```

for an input of `{"id":9007199254740993}`. Tool input always decodes to a `map[string]any`, so `appliedDefaults` is always true and the re-marshal always runs; an object-rooted output schema with a default re-marshals too.

Two things changed from the first version:

- `UnmarshalUseNumber` runs the same `checkMaxDepth` guard as `Unmarshal`, so it can't be used to slip past the nesting limit added in the meantime.
- Both decode sites now use it (arguments and output), and the float64-converted copy is what `Validate` sees.

Added `TestApplySchemaOutputNumberPrecision` for the output path alongside the existing coverage. `TestApplySchema` and `TestApplySchemaOutput` still pass unchanged.

`go test ./...` passes except `TestSSELocalhostProtection/0.0.0.0_via_localhost_rejects_evil.com` and `TestStreamableLocalhostProtection/0.0.0.0_via_localhost_rejects_evil.com`, which fail the same way on an unmodified checkout here because my resolver answers for `evil.com`.

jstar0:
Scope update: this PR is limited to issue #1201. I removed the separate StructuredContent precision follow-up so the diff stays focused on schema-processed tool arguments and outputs. On exact head fd4681dd320c08e0e3fcc86db77fbf2b9183777c, local go test ./..., go vet ./..., formatting/diff checks, and all 9 hosted checks pass. The PR description now reflects this scope; please review this head when convenient.

guglielmo-san:
@jstar0 thank you for the contribution, can you resolve the confilcts?

PR #1246, opened 2026-09-06
mcp: base64-encode a non-header-safe Mcp-Name

### LINKED ISSUE #1243: Streamable HTTP: client sets Mcp-Name raw and does not Base64-encode a non-header-safe name
**Describe the bug**

The Streamable HTTP client sets the `Mcp-Name` header from the raw name without Base64-wrapping a value that is not header-safe. The 2026-07-28 transport spec requires the client to encode such a value with the `=?base64?...?=` sentinel, the same rule it applies to `Mcp-Param-*`.

`mcp/streamable_headers.go`, `setStandardHeaders` (main line 210):

```go
if name, ok := extractName(msg.Method, msg.Params); ok {
    header.Set(nameHeader, name)
}
```

The value is set directly, unlike the `Mcp-Param-*` path, which runs each value through `encodeHeaderValue` (line ~250) and Base64-wraps it when `requiresBase64Encoding` is true.

Transport page, "Value encoding":

> The same encoding rule applies to the Mcp-Name header value. Tool and prompt names are only SHOULD-constrained to header-safe characters, so a name (or resource URI) outside the safe set is carried as: `Mcp-Name: =?base64?{Base64EncodedValue}?=`

**Impact**

A tool name or resource URI containing non-ASCII, control, or whitespace characters is sent as a raw header value. Depending on the HTTP stack it is either mangled or produces an invalid header, and a spec-compliant server that expects the sentinel encoding will not match it.

**Suggested fix**

Run the `Mcp-Name` value through `encodeHeaderValue` (or the same `requiresBase64Encoding` plus Base64 wrap) before setting the header, mirroring `Mcp-Param-*`.

This is the client-side counterpart to #1234 (server-side decode), fixed in #1242.

akashjainn:
Fixes: https://github.com/modelcontextprotocol/go-sdk/issues/1243

The 2026-07-28 transport spec only SHOULD-constrains names to header-safe characters, and lets a client Base64-wrap one that is not, using the `=?base64?...?=` sentinel. `setStandardHeaders` set `Mcp-Name` raw, while the sibling `Mcp-Param-*` path already ran its values through `encodeHeaderValue`, so a name outside the safe set went out unencoded.

This routes `Mcp-Name` through `encodeHeaderValue` as well.

#1242 landed the other direction — the server decoding the sentinel before comparing — so the two together close the round trip. Worth noting they interact: after #1242, a name that is literally shaped like the sentinel (`=?base64?abc?=`) is now rejected as invalid Base64, because the server tries to decode a value the client never encoded. The new test covers that case, and it fails on `main` today.

`extractName` also covers `resources/read`, so this applies to resource URIs, which is where non-ASCII is most likely to show up in practice.

## Testing

`TestMcpNameHeaderEncodingRoundTrip` drives `setStandardHeaders` and then feeds its output to `validateMcpHeaders`, so it pins the client and server halves against each other rather than asserting the encoding in isolation. Cases: an ASCII-safe name (must stay unencoded, no churn for the common path), non-ASCII, a leading space, a control character, and the sentinel-shaped literal above.

Verified failing before the change and passing after. `go test ./...` passes.

A note on process: [@remote-controlled-man](https://github.com/remote-controlled-man) asked about this issue earlier today and proposed much the same scope. I had the change written before I saw that comment and thought the #1242 interaction was worth surfacing either way, but I'm happy to close this if they'd rather carry it — please just say so.

https://claude.ai/code/session_011Wf643Lx36V5wdwuuf2EdT


guglielmo-san (approved):

guglielmo-san:

PR #1247, opened 2026-09-07
mcp: ignore Last-Event-ID on a 2026-07-28 POST instead of rejecting it

### LINKED ISSUE #1233: Streamable HTTP: a 2026-07-28 POST carrying Last-Event-ID is refused instead of ignored
**Describe the bug**

For protocol version 2026-07-28 the streamable HTTP server refuses any POST that carries a `Last-Event-ID` header, where the spec asks it to ignore the header.

`mcp/streamable.go`, `servePOST` (v1.7.0 line 1412, main line 1442):

```go
if len(req.Header.Values(lastEventIDHeader)) > 0 {
    http.Error(w, "can't send Last-Event-ID for POST request", http.StatusBadRequest)
    return
}
```

The 2026-07-28 transport page, "Backward compatibility with older revisions", says a server that receives such traffic from an older client SHOULD respond as follows, and lists: "A `Last-Event-ID` header: ignore it; streams are not resumable." Resumable streams left with this revision, so the header is a harmless leftover from a client that speaks the previous one, and as written it turns that client into one that cannot call anything: every POST is a text/plain 400.

**To Reproduce**

Against any `NewStreamableHTTPHandler` with `Stateless: true`:

```
curl -s -X POST http://127.0.0.1:8080/mcp \
  -H 'Mcp-Protocol-Version: 2026-07-28' \
  -H 'Mcp-Method: server/discover' \
  -H 'Last-Event-ID: stream-1:7' \
  -d '{"jsonrpc":"2.0","id":1,"method":"server/discover","params":{"_meta":{"io.modelcontextprotocol/protocolVersion":"2026-07-28","io.modelcontextprotocol/clientInfo":{"name":"probe","version":"1"},"io.modelcontextprotocol/clientCapabilities":{}}}}'
```

Observed: `400 can't send Last-Event-ID for POST request`.

**Expected behavior**

The header is ignored and the request is served. Keeping the rejection for sessions negotiated at an older version, where a POST with `Last-Event-ID` really is malformed, seems fine; the check just should not apply when the request is 2026-07-28.

**Additional context**

go-sdk v1.7.0; the code is unchanged on main as of today. I am working around it with a handler in front of the SDK's that deletes the header before delegating.

arpankernel:
The 2026-07-28 revision removed resumable streams, and its backward-compat guidance says a server should ignore a `Last-Event-ID` header left over from an older client rather than fail the request. `servePOST` rejected it with `400` for every protocol version.

This gates the rejection to pre-2026-07-28 and ignores the header otherwise. Added a test asserting the header is rejected before 2026-07-28 and ignored on it (red before, green after).

Fixes #1233

### DIFF
--- a/mcp/streamable.go
+++ b/mcp/streamable.go
@@ -1438,7 +1438,15 @@ func (c *streamableServerConn) acquireStream(ctx context.Context, w http.Respons
 //
 // It returns an HTTP status code and error message.
 func (c *streamableServerConn) servePOST(w http.ResponseWriter, req *http.Request) {
-	if len(req.Header.Values(lastEventIDHeader)) > 0 {
+	protocolVersion := protocolVersionFromContext(req.Context())
+	if protocolVersion == "" {
+		protocolVersion = protocolVersion20250326
+	}
+
+	// Last-Event-ID has no meaning on a POST. Before 2026-07-28 it is rejected;
+	// that revision removed resumable streams, so a leftover header from an
+	// older client is ignored rather than failing the request.
+	if protocolVersion < protocolVersion20260728 && len(req.Header.Values(lastEventIDHeader)) > 0 {
 		http.Error(w, "can't send Last-Event-ID for POST request", http.StatusBadRequest)
 		return
 	}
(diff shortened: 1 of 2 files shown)


guglielmo-san:
Closing, check [comment](https://github.com/modelcontextprotocol/go-sdk/issues/1233#issuecomment-5567707134)

PR #497, opened 2025-09-18
doc: document logging

jba:
For #442.

PR #1251, opened 2026-09-12
docs/server: fix logging defaults and the SetLoggingLevel method name

rishabhsai:
The Logging section of docs/server.md says "For stateful sessions" twice in a row with opposite meanings, so one of the two sentences has to be wrong. The default level of "info" is set in only one place in the SDK, `StreamableHTTPHandler.ephemeralConnectOpts` in mcp/streamable.go, which handles stateless requests; a session created through `Server.Connect` starts with an empty `ServerSessionState.LogLevel`, and `ServerSession.Log` returns early while the level is empty, which is exactly what the first sentence already says. The second sentence now reads "For stateless sessions". The same section also tells readers to call `SetLevel`, but there is no such exported method: the client-side call is `ClientSession.SetLoggingLevel`, and `setLevel` is only the unexported server handler, so the pkg.go.dev link on the page points at an anchor that does not exist. For context, the stateless default arrived in #411 and the paragraph describing it was written later in #497, which is where the duplicated "stateful" came from.

Checked with `go doc ./mcp ClientSession` (only `SetLoggingLevel` is exported) and `grep -rn 'LogLevel = "info"' mcp/` (a single hit, mcp/streamable.go:516); docs/server.md was regenerated with `go generate ./internal/docs`, and `go build ./... && go test ./...` pass.


docs/server.md
```diff
@@ -1066,8 +1066,8 @@ server produces logs that remain server-side, for use by server maintainers.)
 **Server-side**:
 The minimum log level is part of the server state.
 For stateful sessions, there is no default log level: no log messages will be sent
-until the client calls `SetLevel` (see below).
-For stateful sessions, the level defaults to "info".
+until the client calls `SetLoggingLevel` (see below).
+For stateless sessions, the level defaults to "info".
```
guglielmo-san: For legacy stateless sessions the level defaults to "info". For sessionless 2026-07-28 requests there is no server-side default: each request carries its level in _meta (io.modelcontextprotocol/logLevel), and nothing is logged until the client sends one. (Note: logging is deprecated as of 2026-07-28 / SEP-2577.)

rishabhsai: Applied your wording in 04140b2, with backticks around `_meta` and `io.modelcontextprotocol/logLevel` to match the rest of the docs; thanks.

guglielmo-san (approved):

PR #1268, opened 2026-09-13
mcp: classify the per-request protocol version by membership

### LINKED ISSUE #1260: mcp: an unknown protocol version in per-request `_meta` is not answered with UnsupportedProtocolVersionError
The [versioning page](https://modelcontextprotocol.io/specification/2026-07-28/basic/versioning) says that if a server does not implement the protocol version a request asks for, "whether the version is unknown to the server, or is a known version the server has chosen not to support", it MUST respond with an `UnsupportedProtocolVersionError` listing the versions it does support. The worked example on that page is a request asking for `1900-01-01`, answered with code `-32022` and a `data` object carrying `supported` and `requested`.

A server built on this SDK does not answer that way on stdio. It answers `method "tools/list" is invalid during session initialization`, a plain Go error that reaches the wire with code `0` and carries no `supported` list.

**Where it happens.** `validateRequestMeta` in `mcp/shared.go` (v1.7.0: function at line 543, the comparison at line 549; `main` at 5bc078a: function at line 570, comparison at line 577):

```go
protocolVersion, ok := meta[MetaKeyProtocolVersion].(string)
if !ok || protocolVersion < protocolVersion20260728 {
	return &validatedMeta{usesNewProtocol: false, initializeParams: nil}, nil
}
```

The classification is a string comparison, so a request counts as modern only when its version sorts at or above `2026-07-28`. Every published revision is in `supportedProtocolVersions`, so no published client is affected. But a version string the SDK does not know sorts wherever its characters put it, and one sorting below `2026-07-28` is classified as a legacy handshake: `usesNewProtocol` stays false, the request never reaches the `CodeUnsupportedProtocolVersion` check in `(*ServerSession).handle` (`mcp/server.go:1873` in v1.7.0, `:1963` on main), and it falls through to the initialization gate instead (`mcp/server.go:1901` in v1.7.0, `:1991` on main). The window holds values like `2026-01-01`, `2025-11-24`, `1900-01-01`, `1.0`, `0` and the empty string.

The same SDK already answers this question by membership one layer out. The header gate at `mcp/streamable.go:354` refuses an `Mcp-Protocol-Version` that is not in the supported list and sorts below `2026-07-28`, which is exactly the set `validateRequestMeta` accepts as a legacy handshake. One layer of the same server refuses those values and the other serves them.

That asymmetry is also why I only see this on stdio. Over HTTP either the header gate rejects the version first, or, when the version is carried only in `_meta`, the header-mismatch check refuses the request with `-32020` before the classification runs. stdio has no header, so nothing catches it there.

**Reproduction.** A stdio server with one tool and nothing else:

```go
package main

import (
	"context"

)

func main() {
	s := mcp.NewServer(&mcp.Implementation{Name: "repro", Version: "v0.0.1"}, nil)
	mcp.AddTool(s, &mcp.Tool{Name: "greet", Description: "greet"},
		func(ctx context.Context, req *mcp.CallToolRequest, args struct{}) (*mcp.CallToolResult, any, error) {
			return &mcp.CallToolResult{}, nil, nil
		})
	if err := s.Run(context.Background(), &mcp.StdioTransport{}); err != nil {
		panic(err)
	}
}
```

Driven with a single request. The `sleep` only holds stdin open long enough for the answer to be written; without it the process can exit on EOF first.

```sh
{ printf '%s\n' '{"jsonrpc":"2.0","id":1,"method":"tools/list","params":{"_meta":{"io.modelcontextprotocol/protocolVersion":"1900-01-01","io.modelcontextprotocol/clientInfo":{"name":"c","version":"1"},"io.modelcontextprotocol/clientCapabilities":{}}}}'; sleep 1; } | go run .
```

What I get:

```json
{"jsonrpc":"2.0","id":1,"error":{"code":0,"message":"method \"tools/list\" is invalid during session initialization"}}
```

What the versioning page requires: code `-32022` with `data.supported` and `data.requested`, which is the error this SDK already produces when a version is excluded by `ServerOptions.SupportedProtocolVersions`.

Observed on v1.7.0, reproduced on `main` at 5bc078a.

**What a server built on the SDK cannot do because of this.** It cannot answer an unknown version the way the specification requires, and the answer it gives instead carries the opposite meaning. The backward compatibility section of the same page says a recognized modern error such as `UnsupportedProtocolVersionError` identifies a modern server and makes the client retry with a supported version, while "anything else identifies a legacy server". A dual-era client probing a modern go-sdk server with a version the SDK does not know is therefore told, by an error carrying code `0`, that the server is legacy, and it falls back to `initialize`. It is given no `supported` list either, so it has nothing to retry with. The server cannot repair this from the outside: the gate returns at `mcp/server.go:1901` (v1.7.0), before `handleReceive` runs the receiving middleware chain at `:1925`, so no middleware ever sees the request. The only seam left is a wrapping `Transport` that parses `params._meta` before the session does, which means reimplementing this piece of the protocol outside the SDK.

I found this auditing a server against the lifecycle specification, not from a client failing in the field. No published revision is in the affected window, so what this breaks is the specified error contract and anything that exercises it, rather than a released client. The specification's own worked example landing inside that window is what convinced me it is worth reporting.

I am offering a pull request with a fix alongside this issue.

Part of #1257.

jmrplens:
`validateRequestMeta` decided whether a request follows the 2026-07-28 protocol by comparing `_meta.protocolVersion` against `2026-07-28` with `<`. This replaces that comparison with a membership test against `supportedProtocolVersions`, which is the rule `mcp/streamable.go:354` already applies to the `Mcp-Protocol-Version` header.

Nothing changes for a version the SDK supports: one predating 2026-07-28 still selects the legacy handshake, and one at 2026-07-28 or later is still validated as the full `_meta` triple. What changes is a version the SDK does not support at all. It now reports `usesNewProtocol`, so it reaches the `CodeUnsupportedProtocolVersion` check in `(*ServerSession).handle` and is answered with the error the versioning page requires.

Driving a stdio server with `_meta.protocolVersion` set to `1900-01-01`, the worked example on that page, before:

```json
{"jsonrpc":"2.0","id":1,"error":{"code":0,"message":"method \"tools/list\" is invalid during session initialization"}}
```

and after:

```json
{"jsonrpc":"2.0","id":1,"error":{"code":-32022,"message":"unsupported protocol version","data":{"supported":["2026-07-28","2025-11-25","2025-06-18","2025-03-26","2024-11-05"],"requested":"1900-01-01"}}}
```

There are two judgement calls in the diff, and I would rather state them than leave them to be found.

The first is that an unsupported version returns early with `initializeParams` carrying only that version, and the rest of the `_meta` triple is left unvalidated. The shape of `_meta` is defined by the revision the request names, and that is a revision this SDK does not know, so there is nothing to validate it against; answering "missing `io.modelcontextprotocol/clientCapabilities`" would send the client off to fix the wrong thing while its real problem is the version. The partly populated value cannot escape: `handle` runs the unsupported check before anything reads the capabilities, and `Server.protocolVersions` is always a subset of `supportedProtocolVersions`, so a version that fails the membership test in `validateRequestMeta` always fails it in `handle` too.

The second is that an empty `protocolVersion` still means "no version declared" rather than "unsupported version". `mcp/streamable.go` already reads the field that way when it decides whether a request carries a version at all (`metaVersion != ""`), and rejecting the empty string here would put the two layers back into disagreement, in the other direction. The `!ok` and `== ""` cases are therefore handled together.

**Behaviour an existing user would see change.** A request whose `_meta.protocolVersion` is a string not in `SupportedProtocolVersions()` is now refused with `-32022`, where before a value sorting below `2026-07-28` was treated as a legacy request: served if the session was already initialized, refused with the session-initialization error if it was not. No published protocol revision is in that set, so a client speaking any released version of MCP sees no difference. A deployment that put a sentinel or placeholder version in `_meta` and relied on it being ignored would see those requests refused, which I believe is the right answer but is a visible change.

This is not limited to stdio, and I want to be precise about where the header gate does and does not stand in front of it. The gate at `mcp/streamable.go:354` refuses an unsupported `Mcp-Protocol-Version` only when that version **also** sorts below `2026-07-28`, so an unknown version sorting above it (`2027-01-01`, say) passes the gate and reaches this classification; the deprecated HTTP+SSE transport in `mcp/sse.go` has no header check at all; and stdio, the in-memory transport and any custom transport have none either. The window the old comparison left open is therefore reachable on every transport, and a value sorting above `2026-07-28` with an incomplete `_meta` triple also changes answer, from `-32602` to `-32022`.

**The test that pins it** is `TestServerUnknownProtocolVersion_NewProtocol` in `mcp/server_test.go`. It drives `(*ServerSession).handle` with a `tools/list` request whose `_meta.protocolVersion` names a version the SDK does not know, and asserts that the returned error is a `*jsonrpc.Error` with code `CodeUnsupportedProtocolVersion`, `Data.Requested` equal to the version asked for, and `Data.Supported` equal to `SupportedProtocolVersions()`. Four cases: `1900-01-01`, `2025-11-24`, `1.0`, and `1900-01-01` again with the rest of the triple omitted, which is the case that pins the first judgement call above. Before the change all four fail with `handle returned method "tools/list" is invalid during session initialization, want a *jsonrpc.Error`. `TestValidateRequestMeta` in `mcp/shared_test.go` gains four cases at the unit level: two that fail before the change, and two that guard the classifications which must not move, a supported pre-2026-07-28 version and an empty one.

`go test ./...` and `gofmt -l .` are both clean.

Fixes #1260.


guglielmo-san (approved):

guglielmo-san:
@jmrplens thank you for the PR!

PR #1270, opened 2026-09-15
mcp: preserve SSE events buffered after the endpoint

yyydddkkk:
SSEClientTransport.Connect creates one event scanner to read the endpoint event
and a second scanner to read subsequent messages. Each scanner owns a buffered
reader, so bytes read ahead by the first scanner are discarded when it is
abandoned. A message coalesced with the endpoint can therefore disappear while
the connection remains open.

Reuse the event scanner across the endpoint handshake and the message-reading
goroutine. This preserves unread bytes without changing the public API.

The regression test exercises Connect and Read with a controlled HTTP response
body, covering both a message coalesced with the endpoint and a message arriving
after Connect returns. On unmodified main, the separate-message control passes
and the coalesced case times out; both pass with the fix.

Validation on linux/amd64 with Go 1.26.8:
- go test ./...
- go test -race ./...
- go vet ./...
- staticcheck v0.6.1 ./...
- gofmt and git diff --check

The HTTP response boundary is simulated; no third-party server or proxy was
used. This change concerns the legacy SSE transport. The Streamable HTTP
conformance suite was not run.


guglielmo-san (approved):

guglielmo-san:
@yyydddkkk thank you for the contribution!

PR #1273, opened 2026-09-15
mcp: refuse a call on a legacy session before initialize

### LINKED ISSUE #1271: mcp: three methods are served on a legacy session before the initialize handshake
Part of the audit indexed at #1257.

`Server.handle` refuses a call made before `initialize` on a legacy session, and the refusal lives in the `default` branch of its method switch. Any method with a `case` of its own therefore skips it, and three are served on a session that never handshook: `logging/setLevel`, `resources/subscribe` and `resources/unsubscribe`.

`resources/subscribe` is the one that matters, because it reaches state rather than merely answering. It starts a watcher and registers a subscriber, so a client that subscribes before `initialize` is delivered `notifications/resources/updated` for the lifetime of a session the server never agreed to. Its `case` today rejects new-protocol requests only.

This is the SDK disagreeing with itself rather than with the specification: the lifecycle page is what motivates the gate the `default` branch applies, and these three methods are outside it for no stated reason. The maintainer identified it during review of #1266: "In legacy sessions, no method can be server before the initialize handshake."

The discriminator is the request rather than the session: a call carrying no `_meta.protocolVersion` on a session with no recorded handshake, which is the condition the `default` branch already applies. A SEP-2575 session legitimately has no `initialize`, so the new-protocol path stays served through the same `usesNewProtocol` exemption it has now.

`initialize` and `ping` stay allowed, ping because the lifecycle page exempts it. `server/discover` is left alone: a legacy request there already receives method-not-found, which answers the client better than an initialization error. The two notification cases are outside the gate, since it covers calls and a notification carries no refusal.

Reproduced against current main with a test that drives each affected method on a fresh session over the in-memory transport. The pull request is linked below.

jmrplens:
Closes #1271.

`Server.handle` refuses a call made before `initialize` on a legacy session, and that refusal lives in the `default` branch of its method switch, so any method with a `case` of its own skips it. Three are served on a session that never handshook: `logging/setLevel`, `resources/subscribe` and `resources/unsubscribe`.

`resources/subscribe` is the one that reaches state rather than merely answering: it starts a watcher and registers a subscriber, so a client that subscribes before `initialize` is delivered `notifications/resources/updated` for the lifetime of a session the server never agreed to.

The check now runs after the switch, for every call carrying no new-protocol `_meta`, with `initialize` and `ping` exempt. That is the same condition the `default` branch already applied and the same error it produced; the change is that the three methods above stop skipping it.

`ping` stays allowed because the lifecycle page exempts it. `server/discover` is deliberately left alone: a legacy request there already receives method-not-found, which tells the client more than an initialization error would. The two notification cases stay outside, since the gate covers calls and a notification carries no refusal.

The SEP-2575 path is untouched. A session without an `initialize` handshake is exactly what that proposal is for, so the exemption is the request's own `_meta.protocolVersion` rather than anything about the session, and a new-protocol client is served as it is today.

Tests: `TestServerHandle_LegacyCallBeforeInitialize` drives each affected method on a fresh session and asserts the refusal, that `resources/subscribe` runs no handler and leaves no subscriber, and that `initialize` and `ping` are served. `TestServerHandle_NewProtocolCallWithoutInitialize` asserts a new-protocol session with no handshake still gets `tools/list`, and that the three methods keep their existing method-not-found rather than the initialization refusal. Without the change, the setLevel, subscribe and unsubscribe rows fail.

`go test ./mcp/ -count=1` is green, `gofmt -l` is empty and `go vet ./mcp/` is clean. This does not depend on #1266 and rebases onto it cleanly, verified with the suite green on the combination.

One decision worth your view: the refusal is a plain error, which on the wire carries code 0, exactly as the `default` branch produced before this change. The specification prescribes no code for it. If you would rather it were `-32600` or `-32603`, say so and I will change it here.


mcp/server.go
```diff
@@ -1989,17 +1989,28 @@ func (ss *ServerSession) handle(ctx context.Context, req *jsonrpc.Request) (any,
...
 
+	// The lifecycle spec lets a client send nothing but pings before the
+	// server has answered 'initialize'. This check used to sit in the default
+	// branch above, so every method with a case of its own escaped it:
+	// 'resources/subscribe' on a session that never handshook registered the
+	// session as a subscriber and delivered notifications to it. A
+	// new-protocol request is exempt because a SEP-2575 session has no
+	// 'initialize' to wait for; the request itself says which protocol it
+	// speaks, and at this point the session has recorded no version that
+	// could say otherwise.
```
guglielmo-san: let's simplify this comment with something like

// In legacy protocol versions, client cannot send requests other than pings 
// before the server has responded to the initialize request

mcp/server.go
```diff
@@ -1989,17 +1989,28 @@ func (ss *ServerSession) handle(ctx context.Context, req *jsonrpc.Request) (any,
...
+	if !initialized && !validatedMeta.usesNewProtocol && req.IsCall() &&
+		req.Method != methodInitialize && req.Method != methodPing {
+		ss.server.opts.Logger.Error("method invalid during initialization", "method", req.Method)
```
guglielmo-san, line 2005: let's replace the `req.IsCall()` with `req.Method != notificationInitialized`

mcp/server.go
jmrplens: Done in 8b01e5f. I kept your wording for the rule and one sentence for the exemption, since nothing on the line below says why a new-protocol request skips the gate. The archaeology moved to the commit message, which is where it belongs.

mcp/server.go
jmrplens: I would rather not, because it changes which messages the gate refuses rather than simplifying the expression.

`req.IsCall()` is false for every notification, so today the gate refuses calls only. Replacing it with `req.Method != notificationInitialized` extends it to notifications, and `notifications/cancelled` is the one that breaks: it has no case of its own in the switch above, so it reaches this check, and a client that gives up on a slow `initialize` sends exactly that message before the server has responded. The SDK's own client does it (client.go:353, where `ClientSession.Close` cancels the listen context to send `notifications/cancelled`) and the transport has a dedicated path for it (transport.go:261). With the replacement the server would refuse it and the in-flight `initialize` would never be cancelled. `notifications/progress` is in the same position.

The lifecycle language is about requests as well: "the client SHOULD NOT send requests other than pings before the server has responded to the initialize request".

There is also a scope point. `req.IsCall()` is what the check already did before this PR; all this PR changes is where the check lives, because sitting in the `default` branch let every method with a case of its own escape it. Widening it to notifications is a separate behaviour change and needs its own decision about `cancelled` and `progress`. Happy to open that as a follow-up issue if you think the gate should cover them.

guglielmo-san: You are right. The current implementation of the preempter for the cancelNofication is a bit weird and might require some work to make it more clean.
My request was based on the fact that v1.6.0 (before refactoring of handle() method) had this exact check https://github.com/modelcontextprotocol/go-sdk/blob/v1.6.0/mcp/server.go#L1432-L1439

guglielmo-san (approved):

PR #1274, opened 2026-09-15
mcp: record the negotiated protocol version on every path that records one

### LINKED ISSUE #1272: mcp: the negotiated protocol version is recorded on one path of four, so stdio accepts batches a SEP-2575 session forbids
Part of the audit indexed at #1257.

Four places record a protocol version on a `ServerSession` and only one of them records it in `NegotiatedProtocolVersion`:

- `ServerSession.initialize` records both, the negotiated one through `negotiatedVersion`.
- `Server.handle` records `InitializeParams` from the version a new-protocol client declares in `_meta`, once its support check has accepted it.
- `server/discover` records `InitializeParams` with the version it was asked about.
- The streamable handler synthesizes `InitializeParams` from the `MCP-Protocol-Version` header for an old-protocol request with no handshake.

Readers that ask what a session speaks therefore disagree depending on how the session began. The concrete failure is on stdio: `ioConn.sessionUpdated` reads only `NegotiatedProtocolVersion`, so a SEP-2575 session over stdio is treated as `2025-03-26` and accepts JSON-RPC batches. Batching was removed in `2025-06-18`, and the streamable handler already refuses them by reading the header, so the two transports disagree about the same session shape.

The support check each of the other three paths applies is all the negotiation SEP-2575 has: there is no handshake response in which a downgrade could be communicated, so a version those paths accept is one the server supports, and recording it as negotiated states what happened. `initialize` stays the one path that downgrades, because the lifecycle page requires it to answer with a version it supports and `InitializeResult.ProtocolVersion` is where the client learns the result.

The fallback in the unexported `protocolVersion()` accessor stays. `ServerSessionOptions.State` is exported, so a caller may supply `InitializeParams` with no negotiated version, and state persisted before #1199 carries none either.

Raised by the maintainer during review of #1266: "what do you think about setting the `ss.state.NegotiatedProtocolVersion` also in case of new protocol version?" The pull request is linked below.

jmrplens:
Closes #1272. Implements what you asked in #1266: "what do you think about setting the `ss.state.NegotiatedProtocolVersion` also in case of new protocol version?"

Four places record a protocol version and only `ServerSession.initialize` recorded it as negotiated. The other three record `InitializeParams` alone: `Server.handle` from a new-protocol client's `_meta`, `server/discover` from the version it was asked about, and the streamable handler from the `MCP-Protocol-Version` header. All three now record the negotiated field too.

The concrete failure this closes is on stdio. `ioConn.sessionUpdated` reads only `NegotiatedProtocolVersion`, so a SEP-2575 session over stdio was treated as `2025-03-26` and accepted JSON-RPC batches, which `2025-06-18` removed. The streamable handler already refuses them by reading the header, so the two transports disagreed about the same session shape. Recording the version on the `_meta` path settles it.

Why recording it as negotiated is accurate on those paths rather than a rename of a declared value: the support check each already applies is all the negotiation SEP-2575 has. There is no handshake response in which a downgrade could be told to the client, so a version those paths accept is one the server supports. `handle` refuses a `_meta` version outside `ss.server.protocolVersions` with `CodeUnsupportedProtocolVersion` before recording anything, `discover` records only a version its own answer lists, and the streamable header is what the transport page defines as the version negotiated earlier. `initialize` remains the one path that downgrades, as the lifecycle page requires, with the result visible in `InitializeResult.ProtocolVersion`.

One deliberate non-refusal: a request with no header at all, on a server that narrowed out `2025-03-26`, is still served as `2025-03-26`. The transport page says a server SHOULD assume that version when the header is absent, and nothing was declared to refuse.

`ServerSessionState.NegotiatedProtocolVersion` is exported, so its doc comment is rewritten to say what it now means on each path. The change is additive in JSON and is not breaking for any client that works today, with one exception worth stating plainly: a SEP-2575 client that sends JSON-RPC batches over stdio is now refused, which is the defect being fixed.

Tests: `TestServerHandle_RecordsNegotiatedVersionOnNewProtocolCall`, `TestServerDiscover_RecordsNegotiatedVersion`, and the old-protocol row of `TestEphemeralConnectOpts`. All three fail without the change.

`go test ./mcp/ -count=1` is green, `gofmt -l` empty, `go vet ./...` clean. Based on main, so it does not depend on #1266, and it rebases onto it cleanly with the suite green on the combination.

If you would rather have this inside #1266, since it answers a question raised there, say so and I will fold it in and close this. It carries one small follow-up either way: #1266's `protocolVersion()` doc says a SEP-2575 session carries its version in `InitializeParams` alone, which stops being true once this lands. The fallback itself stays, for a caller-supplied `ServerSessionOptions.State` and for state persisted before #1199.


mcp/server.go
guglielmo-san, line 1968: i realized here the check should be done against `ss.supportedVersions`

```
if validatedMeta.usesNewProtocol &&
		!slices.Contains(ss.supportedVersions, validatedMeta.initializeParams.ProtocolVersion)
```

jmrplens: I tried this and a test refuses it, so I have left the check where it was for now.

`StreamableServerTransport.SupportsProtocolVersion` (streamable.go:877) returns `t.Stateless && ...` for any version at or above 2026-07-28, so on a **stateful** session `ss.supportedVersions` does not contain 2026-07-28 at all. Validating an incoming new-protocol request against it refuses the one request whose whole job is to ask what the server supports. `TestStreamableStateful_AcceptsDiscover` goes from 200 to 400 with:

```
{"code":-32022,"message":"unsupported protocol version",
 "data":{"supported":["2025-11-25","2025-06-18","2025-03-26","2024-11-05"],"requested":"2026-07-28"}}
```

That test exists because a SEP-2575 client probes with `server/discover` carrying `_meta.protocolVersion: 2026-07-28` before it knows anything, and a stateful server has to answer it. A discover probe cannot be required to speak the version it is asking about.

Your underlying point stands, though, and I would rather not lose it: `server/discover` answers with `Session.supportedVersions` (server.go:928) while this check reads `Server.protocolVersions`, so a transport that filters versions out advertises one set and validates against another, and the error's `Supported` field then names versions the client was never offered.

The shape that does work, which I ran against the full package with `-race`: judge `methodDiscover` against the server's list and every other new-protocol method against the session's, and carry that same list into `Supported`. Happy to put it in this PR or in a follow-up, whichever you prefer.

guglielmo-san: I was wrong, it is actually more correct to use server.ProtocolVersion here

mcp/server.go
```diff
@@ -948,6 +948,14 @@ func (s *Server) discover(ctx context.Context, req *ServerRequest[*DiscoverParam
...
 		req.Session.updateState(func(state *ServerSessionState) {
 			state.InitializeParams = init
+			// A client keeps speaking the version it asked about when the
+			// answer lists it, so that is the version this session negotiated.
+			// One whose version is not listed picks another from the list on
+			// its next call, which nothing here can predict, so only the
+			// declared version is recorded for it.
+			if slices.Contains(versions, init.ProtocolVersion) {
+				state.NegotiatedProtocolVersion = init.ProtocolVersion
+			}
```
guglielmo-san, line 958: what do you think about 


```suggestion
if slices.Contains(versions, init.ProtocolVersion) {
req.Session.updateState(func(state *ServerSessionState) {
state.InitializeParams = init
			// A client keeps speaking the version it asked about when the
			// answer lists it, so that is the version this session negotiated.
			// One whose version is not listed picks another from the list on
			// its next call, which nothing here can predict, so only the
			// declared version is recorded for it.
				state.NegotiatedProtocolVersion = init.ProtocolVersion
			}
			}
```

If a request reaches this point, for sure the protocol version requested is >=20260728 due to the check in the handle() func

```
case methodDiscover:
		// In case of methodDiscover call the state.initializeParams is populated
		// within the discover handle function to make sure the method is supported
		// when the user is probing a pre-2026-07-28 server.
		if !validatedMeta.usesNewProtocol {
			return nil, &jsonrpc.Error{
				Code:    jsonrpc.CodeMethodNotFound,
				Message: fmt.Sprintf("%q is only supported in protocol version >= %s", req.Method, protocolVersion20260728),
			}
		}
```

jmrplens: I went to check this properly and came back agreeing with you more than when I started, so let me separate what holds from what does not.

The `handle()` check does guarantee the client asked for a version >= 20260728. But that is a different quantity from the one the guard tests: the guard asks `ContainsFunc(versions, v >= protocolVersion20260728)`, which is whether **this transport serves any** new version, not whether it serves **the client's**. So the premise is true and does not by itself license replacing the guard.

What does license it is that your form implies the guard. If the client's version is in `versions`, then `versions` contains a version >= 20260728, so the leak the guard exists for stays prevented: the stateful `StreamableHTTPHandler` case where a `discover` request creates a session never surfaced through `Mcp-Session-Id`.

The behaviour it changes is narrower than I first thought, and it may well be an improvement. A client declaring a version >= 20260728 that this transport does not serve keeps its `InitializeParams` today and would keep nothing under your form. Following that through:

- The MRTR predicate is unaffected. `protocolVersion()` falls back to `InitializeParams.ProtocolVersion` and returns `""` when there is none, and the predicate already answers `""` as "not legacy", so such a session is served the same pattern either way.
- The elicitation capability check reads `iparams.Capabilities.Elicitation` and would refuse for a nil, which is the right answer for a session this transport cannot serve.
- `serveStatefulPOST`'s safety net keys on `InitializeParams() == nil`, so under your form that session gets closed rather than kept. For a client whose version we do not serve, closing looks more correct than holding a session it can never use.

So the honest summary is that your version is simpler, keeps the property the guard was written for, and turns an unservable session into a closed one instead of a half-recorded one. I am happy to take it.

One request, and it is only about shape: I would rather it came as its own commit with a test naming the case, a client declaring a supported-by-spec but unserved version, than folded into this one, because this pull request is about recording the negotiated version on the paths that record one and that is a change to what gets recorded at all. If you would rather see it here, say so and I will add it here instead.

guglielmo-san: yes add it here with a comment explaining the logic behind

jmrplens: Done in 7930405, folded into the single commit and rebased onto `main` (the conflict was only the tests #1273 added beside these).

`discover` now records both fields together, and only when its answer lists the version the client declared, with the comment saying why: a client whose version is not listed picks another on its next call, and a nil `InitializeParams` is what lets `serveStatefulPOST` close a stateful session the client was never told about. `TestServerDiscover_UnservedVersionRecordsNothing` names that case with a transport that filters 2026-07-28 out, the way a stateful `StreamableHTTPHandler` does: the answer omits the version and nothing is recorded. The listed case stays covered by `TestServerDiscover_RecordsNegotiatedVersion`, and the commit body carries the reasoning.

guglielmo-san (approved):

PR #1278, opened 2026-09-17
oauthex: build defaultDiscoveryTransport lazily instead of at import

### LINKED ISSUE #1276: Regression: v1.8.0 import initializes Go HTTP proxy env cache before app startup
Importing `github.com/modelcontextprotocol/go-sdk/mcp` at `v1.8.0` appears to prime/cache Go’s `net/http.ProxyFromEnvironment` before application startup code can set proxy env vars, causing later `os.Setenv("HTTPS_PROXY", ...)` to be ignored by default HTTP clients. This did not happen in `v1.7.0`.

**To Reproduce**

1. Create `main.go`:

```go
package main

import (
	"log"
	"net/http"
	"os"

	_ "github.com/modelcontextprotocol/go-sdk/mcp"
)

func main() {
	os.Setenv("HTTPS_PROXY", "http://127.0.0.1:9")

	req, _ := http.NewRequest(http.MethodGet, "https://example.com/", nil)
	proxy, err := http.ProxyFromEnvironment(req)

	log.Printf("proxy=%v err=%v", proxy, err)
}
```

2. Use `go-sdk v1.8.0`:

```go
module repro

go 1.25.0

require github.com/modelcontextprotocol/go-sdk v1.8.0
```

3. Run:

```sh
go mod tidy
go run .
```

4. Observe actual output:

```text
proxy=<nil> err=<nil>
```

5. Change only the SDK version to `v1.7.0`:

```sh
go get github.com/modelcontextprotocol/go-sdk@v1.7.0
go run .
```

6. Observe expected/previous output:

```text
proxy=http://127.0.0.1:9 err=<nil>
```

Control case: remove the blank `mcp` import, and the proxy is also detected correctly.

kilarusravankumar:
Fixes #1276

## What's happening

`oauthex` builds `defaultDiscoveryTransport` as a package-level var, so it
runs at import time. That path calls `proxyConfigured`, which invokes
`http.DefaultTransport`'s proxy func (`http.ProxyFromEnvironment`). net/http
reads the proxy env vars only once and caches the result, so this import-time
call freezes the proxy config before `main()` can set `HTTPS_PROXY`.

New in v1.8.0, which is why it's a regression — just importing `.../mcp`
(which pulls in `oauthex`) triggers it, even with no use of the SDK.

## Fix

Wrap the setup in `sync.OnceValue` so it runs on first use instead of at
import. Still built once and reused.

## Testing (removed)

- Ran the issue's repro against this branch via a local `replace`: prints
  `proxy=http://127.0.0.1:9` now, `proxy=<nil>` before.
- Added a regression test in `oauthex`.
- `go test ./...` passes.


oauthex/oauth2_test.go
```diff
@@ -0,0 +1,15 @@
...
+)
+
+func TestImportDoesNotCacheProxyEnv(t *testing.T) {
+	t.Setenv("HTTPS_PROXY", "http://127.0.0.1:9")
+	req, _ := http.NewRequest(http.MethodGet, "https://example.com/", nil)
+	got, err := http.ProxyFromEnvironment(req)
+	if err != nil || got == nil {
+		t.Fatalf("proxy not detected: got=%v err=%v", got, err)
+	}
+}
```
guglielmo-san: this test fails when is not run in isolation

kilarusravankumar: Yeah, that's it — another [discovery test ](https://github.com/modelcontextprotocol/go-sdk/blob/826e653c270433bbaee6d5d46a771a66f9e15d98/oauthex/discovery_test.go#L113)calls newDiscoveryClient which calls defaultDiscoveryTransport() before my test, so ProxyFromEnvironment is already cached by the time my t.Setenv runs. It only passed because I ran it alone.

I'll drop the test. Since that proxy env is read once per process, any test I write here is at the mercy of whatever runs first, so it's not reliable. The change itself is small and the fix is clear.

is that okay ?

yarolegovich: let's drop the test, but leave a comment explaining why OnceValue is used, thanks for the fix!

kilarusravankumar: Sure @yarolegovich , done

guglielmo-san (approved):

PR #1285, opened 2026-09-21
mcp: preserve explicit null structured content across transport

Tangerg:
A tool result containing `structuredContent: null` currently loses the field when decoded and re-encoded because decoding into `any` turns explicit null into a nil interface, which `omitempty` then removes. This also prevents clients from distinguishing a present null result from an absent structured result.

Keep present null as `json.RawMessage("null")` after decoding. Other values retain the existing decoder behavior. The regression test performs actual client/server calls over in-memory transports and checks omitted content, null, exact large integers, nested content, arrays, booleans, and strings.

Validation: `go build ./...`, `go vet ./...`, `go test ./...`, and `go test -race ./mcp` passed.


guglielmo-san (approved):

guglielmo-san:

